PR #509, opened 2026-06-12
Personas workbench: destination-native rebuild with Console-style UX parity

rmusser01:
## Summary
- Rebuilds the Characters/Personas surface as a destination-native workbench on the `personas` route per the approved spec (`Docs/superpowers/specs/2026-06-09-personas-workbench-design.md`) and ADR-007, replacing both the legacy CCP screen and the thin snapshot shell (legacy routes `ccp`/`characters`/`prompts` resolve here).
- Full workflows: character + persona CRUD with validation and optimistic-lock recovery, card import (PNG/JSON) and JSON/PNG export, search with safe FTS fallback, saved-conversations panel with read-only transcript + Continue-in-Console, draft-aware streaming preview conversations, Attach to Console / Start Chat handoffs.
- Console-"ASCII" visual parity and keyboard-first interaction: design-system tokens (no hardcoded hex), shared flat-button vocabulary (contract-documented), ListView navigation, context-sensitive Esc, real Ctrl+S, mode keys, managed focus, change-based dirty tracking with truthful footer hints and live header state.
- Retires ~7,300 lines of legacy code (ccp_screen, shim, sidebar chrome, orphaned card/editor/conversation widgets); relocates the shared message contracts into `Persona_Widgets/personas_pane_messages.py`. `Persona_Widgets/personas_state.py`/`personas_messages.py` remain byte-identical to PR #506.
- Also fixes systemic CSS bugs found during QA: a dead global `VerticalScroll { height: 100%% }` rule breaking scroll layouts app-wide, `.ds-field-row` forced-horizontal layout, and dead `.--highlight` selectors (personas occurrences; repo-wide sweep filed as task-102).

## Test Plan
- [x] Personas + contract suites green: 328 passed (workbench, widgets, preview, panes, shell routes, non-obscuring-focus + design-system contracts)
- [x] Notes/Library regression gates green except 15 verified pre-existing failures (14 reproduce pre-merge, 1 came in with origin/dev)
- [x] `import tldw_chatbook.app` clean post-retirement; legacy route resolution covered by tests
- [x] Visual QA evidence committed under `Docs/superpowers/qa/personas-workbench/ux-polish/` (real stylesheet + agentic_terminal theme)
- [ ] Reviewer: spot-check the workbench live (`ccp`/`personas` routes) and the conflict-resolution areas (notes_screen, library_screen, _agentic_terminal.tcss)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
[...]

PR #517, opened 2026-06-12
fix: Personas workbench UAT findings (two P0 crashes + UX)

rmusser01:
Found by live UAT (running the real app via \`tldw-serve\` + Playwright, driving the actual rendered terminal) of the merged Personas workbench. Two of these are **P0 crashes currently live on \`dev\`** that the green test suite never caught because they live in the dialog/seam layer.

## Fixes (each its own commit)

1. **P0 — Personas screen was unopenable.** Commit \`2d0d1c89\` (a GitHub-web-applied review suggestion, the last commit before #509 merged) duplicated the \`def __init__\` line in \`EditCharacterRequested\` instead of keeping \`super().__init__()\`, producing an \`IndentationError\` that crashed the app the moment you click Personas. Restored the body.

2. **P0 — Import / Export JSON / Export PNG all hard-crashed the app.** The file picker \`Filters(...)\` was built with glob **strings** (\`\"*.json\"\`) where the API requires \`(name, Callable[[Path], bool])\` tuples; \`Filter.__call__\` then did \`\"*.json\"(path)\` → \`TypeError: 'str' object is not callable\`, tearing down the whole session. Converted to callables matching the house idiom (e.g. \`STTS_Window.py\`). Added a regression test that asserts every personas filter is callable (the existing tests bypass the dialog, which is why this shipped).

3. **Honest footer hint.** \`Ctrl+1\`–\`Ctrl+5\` mode chords aren't transmissible by standard terminals/xterm.js, so they're silently dead while \`[\`/\`]\` work. Footer now advertises \`[ ] mode\` instead of \`ctrl+1-5 mode\`.

4. **Preview transcript ordering.** The provider-error status line rendered *above* the greeting/user lines (out of chronological order). Moved the status region below the transcript, adjacent to the input.

5. **Narrow-width safety.** Added a \`min-width\` to the work area so the center column doesn't collapse at ~80 cols; full responsive pass filed as TASK-108.

## Verification (live, not just tests)
- Re-ran UAT on this branch: Personas **opens**, Export JSON and Import both render their file dialogs with **no session crash**, preview transcript is chronological. Screenshots captured.
- 216 personas/preview/widget/footer tests + 143 contract/handler tests green.

## Follow-up flagged (not in this PR)
The same glob-string \`Filters\` crash is latent in other screens (\`file_picker_dialog.get_eval_file_filters\`, \`ccp_*_handler.py\`, eval dialogs, transcription viewer). Worth a sweep — listed in the PR thread.

---
## Summary by cubic
Fixes two P0 crashes in the Personas workbench and cleans up preview and layout behavior. Personas opens, import/export no longer crash, and the preview ordering and footer hint are corrected.

- **Bug Fixes**
  - Ensures proper initialization in EditCharacterRequested so Personas opens.
  - Replaced glob-string file filters with callables for Import/Export JSON/PNG; added regression tests to ensure filters are callable.
  - Moved preview status below the transcript; added a test asserting DOM order.

- **UI Improvements**
  - Footer now advertises “[ ] mode” instead of “ctrl+1-5 mode” for terminals that don’t send ctrl+digits.
  - Added a min-width to the work area to prevent collapse at narrow widths (partial step toward TASK-108).

<sup>Written for commit 38add3c0c0bf2b5ba5c8bfe08c3eadc9ec2d6169. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/517?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_personas_preview.py
+++ b/Tests/UI/test_personas_preview.py
@@ -123,6 +123,42 @@ async def test_seed_empty_greeting_clears_transcript():
         assert pane.transcript_text() == ""
 
 
+async def test_status_region_renders_below_transcript():
+    """The status line is its own region beneath the transcript, never above.
+
+    Regression guard: a provider/error status must not interleave above the
+    chronological greeting -> you -> character transcript. We assert DOM
+    ordering (transcript precedes status precedes input) and that, with both
+    a transcript and a status set, the greeting/user lines still read in
+    order while the status sits in its separate region.
+    """
+    app = PreviewApp()
+    async with app.run_test() as pilot:
+        pane = pilot.app.query_one(PersonasPreviewPane)
+        await pane.seed_greeting("Greetings, detective.")
+        pane.append_user("Who are you?")
+        pane.set_status("anthropic is not ready - configure in Settings")
+        await pilot.pause()
+
+        body = pilot.app.query_one("#personas-preview-body")
+        child_ids = [c.id for c in body.children]
+        transcript_pos = child_ids.index("personas-preview-transcript")
+        status_pos = child_ids.index("personas-preview-status")
+        input_pos = child_ids.index("personas-preview-input")
+        # Status must sit between the transcript and the input, not above it.
+        assert transcript_pos < status_pos < input_pos
+
+        # The transcript itself stays in chronological order.
+        assert _line_texts(pilot.app) == [
+            "character: Greetings, detective.",
+            "you: Who are you?",
+        ]
+        # The status lives in its own region, separate from the transcript.
+        assert "anthropic is not ready" in str(
+            pilot.app.query_one("#personas-preview-status", Static).renderable
+        )
+
+
 async def test_double_seed_same_tick_does_not_crash():
     app = PreviewApp()
     async with app.run_test() as pilot:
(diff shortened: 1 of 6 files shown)


rmusser01:
**Follow-up: same `Filters` glob-string crash is latent elsewhere** (out of scope here, worth a sweep — each will `TypeError` and crash the session the same way when its file dialog opens):
- `tldw_chatbook/Widgets/file_picker_dialog.py` (`get_eval_file_filters` / first Filters block)
- `tldw_chatbook/Widgets/Evals/eval_dialogs.py`
- `tldw_chatbook/Widgets/transcription_history_viewer.py`
- `tldw_chatbook/UI/CCP_Modules/ccp_dictionary_handler.py`, `ccp_character_handler.py`, `ccp_prompt_handler.py`

(`file_picker_dialog.py`'s `create_filter(...)` helper and `chat_attachment_handler.py` are already correct — they return callables.)

PR #648, opened 2026-07-17
feat(agents): native provider tool-calls with fence-protocol fallback (task-243)

rmusser01:
## Native provider tool-calls with fence-protocol fallback (task-243)

Closes out the vertical-slice spec's deferred half: agent runs on tool-capable providers now use **native `tools=` function-calling** — multi-call batches, no per-turn fence-protocol prompt overhead (~1.3k tokens/turn at a 14-skill catalog per the task-231 review) — while every other provider, llama_cpp included, keeps the fence protocol **byte-identically**. Executed via SDD (5 reviewed tasks + live gate); final whole-branch review: APPROVE, 0 findings.

### How it works
- **Capability ground truth** (`Agents/native_tools.py`): `NATIVE_TOOLS_PROVIDERS` contains only handler-verified raw-OpenAI-dict providers (openai, groq, openrouter, mistral, deepseek, moonshot, custom-openai ×2). Anthropic/Google/Cohere are excluded — their handlers normalize responses and silently drop tool-use blocks (follow-up filed as **task-246**). groq/deepseek had a map gap (handlers accept `tools` but `PROVIDER_PARAM_MAP` dropped it) — fixed.
- **Service** (`agent_service._make_call_model`): native = `config.native_tools ∧ provider_supports_native_tools(api_endpoint)` → sends `tools=` (ToolSchema→OpenAI conversion, progressive disclosure recomputed per turn), suppresses `render_tool_protocol`, parses `message.tool_calls` into `ModelTurn.tool_calls`. Native-parsed calls flow through the **same** permission gate (`allowed_tools ∩ disclosed_names`) — nothing bypasses it.
- **Engine**: `ModelTurn.assistant_message` (the provider's own message, tool_calls array included) is echoed into history verbatim; each call's result is answered as `role="tool"` keyed on `call_id`. Fence turns (`assistant_message=None`, `call_id=""`) produce exactly the pre-existing dicts — pinned by a key-set regression test. The multi-call dispatch loop needed **zero changes** (AC #3).
- **Gateway** (`console_provider_gateway`): `stream_chat` gains a `tools=` passthrough and accumulates streamed `delta.tool_calls` fragments (id/name once, arguments concatenated) into a single `ProviderToolCalls` sentinel yielded last — **only when tools were requested**. Plain Console sends are provably unchanged; a review Critical (mapping-level guards that broke `tools=None` byte-identity on raw-dict chunks) was caught and fixed with a pinning regression test.
[...]

PR #655, opened 2026-07-17
feat(agents): model-turn budget tier + load_tools bare-name fallback (task-244)

rmusser01:
Closes task-244 — the second of the three task-231 tool-call-flow follow-ups (after #648's native tool-calls).

### 1. Model-turn budget tier (ACs #1–2)
`RunBudget` gains **`max_model_turns`** — provider round-trips become the run's primary limiter instead of raw step entries. The measured pain (Skills Phase-2 gate): a successful 4-model-turn discovery run burned 10 of 16 steps because tool_call/tool_result bookkeeping counts the same as a provider call.

- Engine defaults are **byte-identical by construction**: the default `max_model_turns=8 == max_steps=8`, each model turn appends ≥1 step, and the step check runs first — so the new check is provably unreachable at `RunBudget()` (stated in the field comment, pinned by tests; the pinned `"step budget exhausted"` copy is untouched, the new distinct copy is `"model-turn budget exhausted"`).
- `CONSOLE_RUN_BUDGET` → `RunBudget(max_steps=32, max_wall_seconds=480.0, max_model_turns=8)`: 8 provider calls primary, steps a pure backstop. The documented 4-turn/10-step discovery floor plus **two** additional real tool rounds (6 turns / 16 steps) now completes with 2 turns of headroom — proven by a test that drives the real loop under the real imported budget.
- `clamp_child_budget` carries the field through to sub-agents.

### 2. `load_tools` bare-name fallback (ACs #3–4)
Models routinely echo the tool **name** from a `find_tools` result line instead of the catalog id, previously burning a whole round on a generic "No valid tools found to load". `load_schemas` now falls back to `registry.resolve_name()` only after the direct catalog-id lookup KeyErrors; the allow-list, disclosed-dedup, and room-slice gates below are byte-identical and verified to apply to name-resolved schemas (including a resolvable-but-disallowed refusal test and an unresolvable-junk pin).

### Verification
SDD with an independent review per task (both clean — tie-break unreachability proof and the collision/stale-cache/security traces verified against real code) + final whole-branch review: **APPROVE** (composition checked: nothing reconstructs `RunBudget` from persisted run records; rail UI unaffected; strictly more permissive than the old budget for everything task-243's native path shipped). Cross-suite 274 passed; DB suite 212 passed; controller suite 54 passed after the one docstring minor.

---
## Summary by cubic
Make model turns the primary run limiter and add a `load_tools` bare-name fallback, plus in-batch dedupe to avoid duplicate loads. This satisfies task-244 ACs while keeping engine defaults identical and improving console runs.

- **New Features**
  - Add `RunBudget.max_model_turns` (default 8 equals `max_steps`); the loop enforces it with a distinct "model-turn budget exhausted" error, and `clamp_child_budget` carries the field.
  - Update console budget to `RunBudget(max_steps=32, max_wall_seconds=480.0, max_model_turns=8)` so the 4-turn/10-step floor plus two extra tool rounds (6 turns / 16 steps) completes with headroom.
  - `load_tools` now falls back to `registry.resolve_name()` when a catalog-id lookup fails; allow-list, dedup, and room-slice gates remain unchanged.
  - Tests cover the model-turn cap, console floor-plus-two scenario, successful bare-name loading, unchanged errors for disallowed/unresolvable names, and dedupe behavior.

- **Bug Fixes**
  - Deduplicate same-batch tool loads by name in both the runtime and service layers (handles name/id aliases) to prevent duplicate activation.
  - Use an O(1) model-turn counter in the loop.
  - Console failure docstring lists the new "model-turn budget exhausted" reason.

<sup>Written for commit 08173216ebfd0cd148610a5ff5b56ae2b46f5c6e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/655?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## Review findings — disposition (293997b4)

**All three fixed:**
- **Gemini (agent_service.py): same-batch duplicate via name/id aliases** — valid, and it also exposed a pre-existing vector (a repeated catalog id in one batch had the same effect). Fixed at BOTH layers: `load_schemas` dedupes by name within the batch (disclosed_names only guards prior rounds), and the loop's own F1-b last-line-of-defense now dedupes within `new_loaded` too. Two new tests: service-level `["calculator", "builtin:calculator"]` → `"loaded: calculator"` exactly once + a full round-trip; engine-level duplicate schemas in one batch admit one into `active`.
- **Qodo: quadratic model-turn counting** — replaced the per-iteration `sum(...)` with an O(1) counter incremented beside the `STEP_MODEL` append (the plan explicitly allowed either; counter is strictly better).
- **Qodo: missing test docstring** — added (pins the unreachability invariant + child clamp passthrough).

Merged latest dev (clean, no backlog-ID collisions this time); post-merge sweep **290 passed / 0 failed** (Agents + bridge + swap + gateway + rail + agent-runs DB).

PR #659, opened 2026-07-17
feat(llm): Anthropic native tool-calls — full block conversion, live-gated (task-263)

rmusser01:
Closes task-263 (Anthropic-only scope per user decision; Google/Cohere filed as tasks 266/267). Anthropic joins `NATIVE_TOOLS_PROVIDERS` — agent runs on Anthropic now use native function-calling instead of the fence protocol, including **parallel multi-tool batches**.

### How: the handler becomes bilingual
All conversion is contained in `chat_with_anthropic` (+ one pure helper) — everything upstream keeps speaking OpenAI shapes, and the gateway/service/engine needed **zero changes**:
- **Request side**: OpenAI `tools=` → Anthropic `{name, description, input_schema}` (entries already Anthropic-shaped pass through — historical contract preserved); `assistant.tool_calls` echo → `tool_use` content blocks (text block first; junk entries guarded after a review Important caught empty-name/empty-content live-400 vectors); `role="tool"` results → `tool_result` blocks **coalesced into one user turn** (role alternation).
- **Non-streaming**: `tool_use` content blocks → OpenAI `message.tool_calls` (`arguments` JSON-encoded; key absent on text-only responses — byte-compat pinned).
- **Streaming**: the previously-dead `tool_calls_delta` placeholder goes live — `content_block_start` + `input_json_delta` events emit OpenAI-style incremental `delta.tool_calls` fragments keyed by 0-based tool position, exactly the shape the Console gateway's task-243 accumulator merges (pinned by a cross-layer test that feeds the handler's real SSE output through the real gateway pieces).

### Live gate (AC #3 — flip strictly after a real round-trip)
`Docs/superpowers/qa/anthropic-native-2026-07/` — real api.anthropic.com (claude-haiku-4-5, streaming ON), run through the real Console reply engine **before** the flip via an in-harness set override:
- **Case A** (1.5s): single calculator round-trip — tools accepted (no 400), no fence protocol, `tool_result` follow-up turn accepted, correct answer.
- **Case B** (2.0s): a **genuine parallel two-tool batch in one reply** — the first live parallel batch in the program (the OpenAI-compatible gate only ever saw serialized calls) — dispatched as one engine batch, both results paired by id and accepted; also live-covers interleaved two-block streaming reassembly.

API key handling: git-excluded local file, read at runtime, never logged/echoed/committed (final review grepped diff + evidence: clean).

### Verification
[...]

PR #662, opened 2026-07-17
feat(llm): Google/Gemini native tool-calls — full conversion incl. Gemini 3 thought signatures, live-gated (task-266)

rmusser01:
Closes task-266 (Cohere remains the last fence-only normalizer — task-267). Google/Gemini joins `NATIVE_TOOLS_PROVIDERS` — agent runs on Gemini now use native function-calling, including **parallel multi-tool batches**, on current Gemini 3-family models.

### The handler becomes bilingual (task-263 containment pattern)
- **Request side**: OpenAI `tools=` → one `{"functionDeclarations": [...]}` entry (Gemini-shaped entries pass through; blank-name entries dropped locally); `assistant.tool_calls` echoes → `functionCall` parts (list-form multimodal content keeps its text parts — the same bug class review-caught on the Anthropic sibling, fixed identically); `role="tool"` results → `functionResponse` parts **coalesced into one user turn** with the function name resolved via an id→name map built from preceding echoes (Gemini has no call ids) plus a positional fallback — Gemini's own pairing convention for same-name parallel calls. `response` is always a JSON object (dict-parseable content used directly, else `{"result": <str>}`).
- **Response side**: the non-streaming `functionCall` → OpenAI `tool_calls` parsing already existed — now pinned by tests incl. the `parse_native_tool_calls` round-trip. **Streaming** emits whole `functionCall` parts as complete OpenAI fragments (running cross-stream index; malformed parts skipped without aborting the stream — a reviewer-reproduced abort vector, fixed + pinned).

### The live gate earned its keep — two real failures before PASS
1. `gemini-2.5-flash` 404s ("no longer available to new users") despite being listed by `/v1beta/models` → switched to `gemini-flash-latest`.
2. **Gemini 3 models 400 unless the response part's `thoughtSignature` is echoed back verbatim** on follow-up `functionCall` parts. Fixed by carrying it opaquely through the OpenAI shapes (`google_thought_signature` on the entry: non-streaming parser + streaming fragments + request-converter re-attach + a gateway accumulator extra-key preservation), three tests. The final review traced the signature surviving the REAL streaming service path (accumulator → bridge → `ensure_tool_call_ids` → echo → converter), and per-call signatures pair with their own parts by construction (proven live in the parallel case).

**Gate: PASS** (`Docs/superpowers/qa/google-native-2026-07/`, streaming ON): case A single calculator round-trip (1.8s, correct); case B genuine parallel two-tool batch in one reply (2.4s, both results + signatures accepted). Key handling: git-excluded local file, never logged/echoed/committed (final review grepped diff + evidence clean).

### Verification
SDD: 2 tasks + coordinator gate; independent review per task — two Importants caught, fixed, re-verified (list-content text loss; malformed-functionCall stream abort). Final whole-branch review (opus): **APPROVE, 0 blocking** (one informational Minor deferred: the accumulator extra-key passthrough is unscoped — allow-list later if ever needed). Suites: Agents 137, google-native file 19, targeted cross-suite 119.

Out-of-scope observation for a future nit: the google `PROVIDER_PARAM_MAP` entry keys `'temperature': 'temp'` — a nonexistent generic name, so temperature is silently dropped for google today (pre-existing).

---
## Summary by cubic
Adds native tool-calls for Google/Gemini using its function-calling API, including streaming, and enables Google in `NATIVE_TOOLS_PROVIDERS` after a live gate. Satisfies task-266 by converting OpenAI shapes to/from Gemini and supporting Gemini 3 thought signatures.

- **New Features**
  - Convert OpenAI `tools` to Gemini `functionDeclarations` with passthrough for Gemini-shaped entries.
  - Convert assistant `tool_calls` and `role="tool"` history to Gemini `functionCall`/`functionResponse` parts; results coalesce into a single user turn with id→name + positional fallback pairing.
  - Stream Gemini `functionCall` parts as complete OpenAI `delta.tool_calls` fragments with stable indexes; gateway accumulator preserves extra fragment keys.
  - Round-trip Gemini 3 `thoughtSignature` via `google_thought_signature` on tool-calls; re-attached on requests.
  - Added tests for request/response/streaming and a service-level check; live gate on `gemini-flash-latest` passed (incl. a parallel two-tool batch), then Google was enabled.

- **Bug Fixes**
  - Keep assistant list-content text when tool-calls are present.
  - Guard malformed `functionCall` parts in both streaming and non-streaming paths; streams continue and parsers skip bad parts.
  - Skip unpairable tool results (id miss + positional fallback exhausted) instead of emitting empty-name `functionResponse`; blank-name calls are ignored and do not consume indexes.
  - Accumulator preserves provider extra keys across fragments, keeping falsy-but-present values and dropping only `None`.

<sup>Written for commit 3296ef00e007e4d547384d849ed4827735b9bfe8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/662?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## Review findings — disposition (fa79b9c1 + dev merge)

**Fixed (all three code findings):**
- **Gemini (High): non-streaming parser lacked the `isinstance(functionCall, dict)` guard** the streaming path already has → adopted (with the args dict-guard and signature carry preserved); a malformed part now skips instead of crashing the parser (test: `test_non_streaming_malformed_function_call_part_is_skipped`).
- **Qodo 3: empty-name `functionResponse` on unpairable results** (id-map miss + positional fallback exhausted) → the orphan result part is skipped with a warning instead of 400ing the whole request (test: `test_unpairable_tool_result_is_skipped_not_empty_named`).
- **Qodo 4: falsy extra fragment keys dropped** contradicting the verbatim-echo doc → `if value is not None:` — present-but-falsy extras (e.g. `""`) now survive verbatim, `None` still drops (accumulator test extended to pin both).

**Declined with reasoning:**
- **Qodo 1 (`GATE_DBDIR` via path_validation.py)** — same disposition as PR #659's identical finding: the QA harness under `Docs/superpowers/qa/` is a developer-run evidence script whose env var is set by the coordinator in the same shell; `path_validation.py` guards the application's input boundaries, and importing it into a point-in-time QA artifact would misstate the trust model.
- **Qodo 2 (import grouping in the harness)** — the `sys.path.insert` between stdlib and project imports is load-bearing: the standalone script cannot resolve `tldw_chatbook.*` before the path setup runs. Identical accepted structure as the anthropic gate harness on dev.

Merged latest dev (RAG PRs #653/#656/#657 — clean, no conflicts); post-merge sweep **281 passed / 0 failed**.

PR #678, opened 2026-07-17
feat(ui): footer shortcut hints render per-screen (task-264)

rmusser01:
## Summary

Live QA (MCP Hub Phase 4) found the footer shortcut-hint system was dead app-wide: `AppFooterStatus` was mounted once on the App's DEFAULT screen, which every pushed `BaseAppScreen` occludes permanently — `App.query_one` resolves against the default screen **by design** (`App._get_dom_base`), so every `set_workbench_shortcuts()` registration (Console F6 set, MCP mode set) "succeeded" against an invisible widget.

**Decision (AC#3, recorded in `Docs/superpowers/plans/2026-07-17-footer-shortcut-hints.md`):** per-screen mounting of the existing `AppFooterStatus`, fed by the unchanged registration API. Repairing the single global strip is architecturally impossible under `push_screen` navigation (screens are full-viewport), and the app-owned-chrome migration is explicitly deferred by the design contract.

## Changes

- **`BaseAppScreen.compose()`** yields its own `AppFooterStatus(id="screen-footer-status")`, replacing a `Footer` that rendered nothing on every screen except Console. The App-level default-screen instance stays (covers the pre-push startup frame; different id).
- **Registrations persist across screen recompose** (`register_footer_shortcuts`/`clear_footer_shortcuts` on `BaseAppScreen`, re-seeded during `compose()`): Settings' `recompose=True` reactives and Library/Chat `refresh(recompose=True)` replace the footer widget — without persistence, hints silently reset (reproduced; pinned by `test_registration_survives_screen_recompose`). Clear mirrors the widget's source-ownership race guard.
- **Callers are screen-aware:** chat/mcp/library/settings screens use the persisting API; personas + `chat_token_events` resolve the active screen; `app.py` gains `_active_footer_status` (active-screen lookup, default-screen cache fallback, `ScreenStackError` guard for shutdown timer ticks).
- **Library and Settings register the hint sets their retired Footer rendered** (`u` / `s·r·t`) — caught in review; the old Footer showed their `show=True` bindings. Console's `Ctrl+K`/`Ctrl+T` fold into `CONSOLE_WORKBENCH_SHORTCUTS` (Alt-hints stay functional but unrendered — unverifiable over xterm.js anyway).
[...]

PR #687, opened 2026-07-17
test(ui): pin footer CSS-subset and personas-no-recompose invariants (task-289)

rmusser01:
## Summary

task-264 left two safety invariants as comments; this turns both into red tests (test-only change, +190 lines in `Tests/UI/test_screen_footer_hints.py`).

**Guard 1 — CSS drift.** `AppFooterStatus.DEFAULT_CSS` (the widget-baked layout rules that keep stylesheet-less harnesses matching production geometry) must stay a faithful *subset* of the live bundle source (`css/components/_widgets.tcss` footer section) **and** of the built `tldw_cli_modular.tcss` — the second check catches an edited source that was never rebuilt. Implemented as a parsed-declaration comparison (comments stripped, whitespace-normalized), not string matching; the section markers use the full comment text because the bare marker sits *inside* a comment and splitting there leaves a dangling `*/` that corrupts the first selector.

**Guard 2 — personas recompose.** `PersonasScreen` drives its footer through the non-persisting `set_shortcut_context` path (its hint set is dynamic), which is only safe while the screen never recomposes — a screen-level recompose replaces the footer widget and silently resets hints (the task-264 fix-wave bug). AST-based guard fails if the file gains `recompose=True` inside a `BaseAppScreen` subclass **or** a module-level `.refresh(recompose=True)` helper (the exact pattern living in `library_screen.py`), unless an actual `.register_footer_shortcuts()` call (the persisting API) is present — matched as an AST call, not a substring, so a comment mentioning the API cannot disarm it. Child-widget classes are excluded: their recompose never touches the screen's footer.

## Verification

- **Every rule mutation-verified in both directions:** `height: 1 → 2` in DEFAULT_CSS fails both CSS guards; a recompose reactive injected into `PersonasScreen` fails; a module-level `screen.refresh(recompose=True)` helper fails; that helper plus a `# TODO: migrate to register_footer_shortcuts` comment **still fails**; an actual `register_footer_shortcuts(...)` call passes. Clean tree: 9/9 (whole file) + `test_app_footer_shortcut_context.py` 3/3.
- Independent review (probing parser false-pass/false-fail modes, escape-hatch bypass, scope gaps): verdict "with fixes (optional)" — both Important findings applied (substring escape hatch → AST call-match; class-body-only scope → module-level refresh coverage) and the parser's selector-list limitation documented (it produces a *loud* false-fail, never a silent pass).

Closes task-289.

---
## Summary by cubic
Adds two test-only drift guards for the footer: keep `AppFooterStatus.DEFAULT_CSS` a subset of the live and built bundles, and block `PersonasScreen` recomposes while using non-persisting hints unless the persisting API is used. Hardened the CSS parser to merge duplicate selector blocks and added a self-check to prevent blind AST detection. Addresses task-289.

- **New Features**
  - CSS drift guard: parsed-declaration subset check against `_widgets.tcss` footer and built `tldw_cli_modular.tcss`; parser now merges duplicate selector blocks in source order.
  - Personas recompose guard: fails on any `recompose=True` in a `BaseAppScreen` subclass or module-level `screen.refresh(recompose=True)`; escape hatch requires an AST-detected `.register_footer_shortcuts(...)`; adds a runtime-backed self-check to ensure direct `BaseAppScreen` detection.

<sup>Written for commit 0ac37f833a3bf1ab96ac23811d3e11b2067c9c0e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/687?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_screen_footer_hints.py
+++ b/Tests/UI/test_screen_footer_hints.py
@@ -14,6 +14,10 @@ app. These tests pin that contract directly against the real screens/
 registration methods, not a hand-rolled fake.
 """
 
+import ast
+import re
+from pathlib import Path
+
 import pytest
 from textual.app import App, ComposeResult
 from textual.widgets import Static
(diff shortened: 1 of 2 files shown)


rmusser01:
**Qodo findings adjudication** (3 bugs / 1 rule violation reported):

- **#2 CSS parser loses duplicate selector blocks — FIXED (00bd44ed).** Correct catch: CSS merges duplicate blocks per-property in source order; the parser now does the same (`setdefault(...).update(...)`), so a legitimate split of one selector across two blocks can no longer false-fail the guard. Height-drift mutation re-verified still failing; clean tree 9/9.
- **#3a alias/indirect inheritance blinds the screen-class detection — FIXED (00bd44ed)** with a loud self-check rather than heavier analysis: the test now asserts at runtime that `PersonasScreen` is a `BaseAppScreen` subclass AND that the AST walk found ≥1 direct-`BaseAppScreen` class — if inheritance ever goes indirect/aliased, the guard fails with instructions instead of silently scanning nothing.
- **#3b any `.register_footer_shortcuts()` call anywhere disarms the guard — DECLINED.** The guard is scoped to one file whose only screen is PersonasScreen; any call of that API in `personas_screen.py` is, realistically, the personas migration the guard exists to await. Distinguishing "the right receiver" would need dataflow analysis — disproportionate for a lightweight tripwire whose failure mode is only ever guard-doesn't-fire (never a false failure), a trade-off the independent pre-PR review also assessed and accepted.
- **#1 test docstrings not Google-style — DECLINED.** File-local precedent: the existing tests in this file (`test_console_registration_updates_the_screens_own_footer` et al., merged in #678) use the same multi-line prose docstring shape without a blank-line-after-summary; test functions here aren't documented as Google-style public API. Consistency with the file wins.

Also merged latest dev (clean; brings the task-292 media-search fix). Post-merge: 12/12 across the footer suites.

PR #690, opened 2026-07-17
feat(llm): Cohere native tool-calls via v2 /chat migration (task-267)

rmusser01:
## Summary

`chat_with_cohere` was the last named fence-only provider — it normalized responses and dropped tool-call data. This PR completes the native-tools program (task-263 Anthropic, task-266 Google, task-286 dead-key audit): **all providers with tool support now use native tool-calls**, with the fence protocol remaining the fallback for local/unlisted providers.

**Design decision (user-approved):** migrate the handler from Cohere's legacy **v1 `/chat` to v2 `/chat` first**, then map tools. v1's shapes are structurally hostile — flat `parameter_definitions` cannot express nested JSON Schema (MCP tools would be inexpressible), tool results travel as a separate request field outside the history model, and calls have no ids. v2 is OpenAI-shaped end-to-end, making the tool mapping near-1:1. The migration changes ALL Cohere traffic; text-only behavior was byte-compat pinned **before** any tool code landed.

## Changes

- **Task 1 — v2 migration, text paths:** `messages` array (system folded in), v2 param names, both response paths rewritten for text with the OpenAI-normalized shape pinned (content, `COMPLETE→stop`/`MAX_TOKENS→length`, no `tool_calls` key when absent); `num_generations` (v1-only) dropped with a debug log; inline default model aligned to config's `command-a-03-2025`.
- **Tasks 2-4 — native tools:** OpenAI `tools` passthrough with junk-entry guards; assistant `tool_calls` + `role:"tool"` history → v2 echo shapes (document-content, positional `tool_call_id` fallback); non-streaming `message.tool_calls` → OpenAI shape (attached only when present); streaming `tool-call-start`/`tool-call-delta` → `delta.tool_calls` fragments honoring the gateway `_ToolCallAccumulator` contract (cross-layer test imports the **real** accumulator); `tool_plan` round-trips as the `cohere_tool_plan` extra via `_PRESERVED_FRAGMENT_EXTRAS` — the shipped `google_thought_signature` mechanism, additive-only to shared code.
- **Live gate (pre-flip, real `api.cohere.com/v2/chat`, streaming, real Console bridge→gateway→handler stack; evidence in `Docs/superpowers/qa/cohere-native-2026-07/`):**
  - Case A — single calculator round-trip: PASS (native tools sent, no fence, tool-result turn accepted, correct answer).
  - Case B — parallel two-tool reply: **found a real bug.** No-arg streamed calls accumulate `arguments=""` (`tool-call-start` seeds the empty string; no deltas follow) and Cohere 400s the echo (*"tool arguments must be a stringified JSON object"*). Fixed at the request boundary only (`""→"{}"` — streaming emission untouched, so the accumulator contract is preserved), regression-pinned, rerun PASS with two `tool_call` steps in one stream.
- **Flip (only after PASS):** `NATIVE_TOOLS_PROVIDERS` gains `cohere`; registry docstring no longer names any fence-only provider; registry test flipped; service-level pin asserts the real registry (override removed). Committed gate outputs independently verified to contain zero key material.

## Verification

- 34 handler tests (`Tests/Chat/test_cohere_native_tools.py`, TDD per task) + real-accumulator cross-layer pin + service fence-suppression pin; PROVIDER_PARAM_MAP invariants green (signature unchanged).
- Implementer sweep pre-flip: Tests/Chat + Tests/Agents = **1183 passed**. Flip-affected suites post-flip: 162 passed. Post-dev-merge: 138 passed.
- SDD: task review (opus) **Approved / spec PASS**; final whole-branch review (opus) **ready to merge, no Critical/Important** — its one follow-up (the summarization helper still calls v1 `/chat`; pre-existing, out of scope) is filed as **task-297**.

Closes task-267.

---
## Summary by cubic
Migrated Cohere from v1 `/chat` to v2 `/chat` and enabled native tool-calls end-to-end. This completes task-267 and adds `cohere` to `NATIVE_TOOLS_PROVIDERS`, with fence kept as fallback for non-native providers.

- **New Features**
  - Moved `chat_with_cohere` to v2 `/chat` with OpenAI-shaped `messages` and `tools`.
  - Passthrough OpenAI `tools`; accept `role:"tool"` history; emit `tool_calls` for non-streaming and streaming deltas.
  - Round-trip v2 `tool_plan` as `cohere_tool_plan`.
  - Added `cohere` to `NATIVE_TOOLS_PROVIDERS`; service sends native tools and suppresses the fence path.
  - Kept text-only behavior byte-compatible; default model aligned to `command-a-03-2025`.

- **Bug Fixes**
  - Normalize empty streamed tool-call `arguments` from `""` to `{}` at echo to avoid Cohere 400s.
  - Stream terminator now emits a single `[DONE]` on success; `finally` only handles errors.
  - Streamed `cohere_tool_plan` is echoed correctly (message-level with fallback) and preserved without leaking on wire.
  - Hardened tool converters: `_cohere_tools_payload` drops non-function junk; response parser skips invalid/blank-name function entries.

<sup>Written for commit 5d803ef2514613ec1f36724bb4735595200eb0e9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/690?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
**Qodo findings adjudication** (6 findings; 4 fixed in the latest commit, 2 declined):

- **#3 duplicate `[DONE]` terminator — FIXED.** Real: `message-end` yielded `[DONE]` and the generator's `finally` unconditionally yielded a second. The `finally` now only covers abnormal termination.
- **#4 streamed `tool_plan` not echoed — FIXED.** Correct catch on a real mechanism gap: the non-streaming path attaches `cohere_tool_plan` at message level (round-trips fine), but streamed turns carry it inside the accumulated `tool_calls` entry, which the echo builder never read. Message-level miss now falls back to the entry, and `_cohere_request_tool_calls`'s reconstruction guarantees the extra never leaks to the wire.
- **#6 malformed tools entries forwarded — FIXED.** v2's native tools shape IS the OpenAI shape, so the "native passthrough" else-branch had nothing legitimate to pass through — non-function dict entries are now dropped with a warning.
- **#7 response converter crash/undispatchable entries — FIXED.** Non-dict `function` (AttributeError risk) and blank names are now skipped; an all-junk list falls through to plain content handling.
- **Verification:** 4 regression tests added (38/38 in the cohere suite, 142 across the flip-affected suites), and — since #4/#6 change wire payloads — the **live gate was re-run post-fix: cases A+B PASS** (`gate-A-B-postreview-2026-07-17.txt`, scanned clean of key material).
- **#1 logging full SSE event / raw decode-failure payload — DECLINED.** File-local precedent: the shipped Anthropic streaming path (PR #659) logs undecodable payloads identically (`Anthropic Stream: Could not decode JSON: {event_data_str}`), and the full-event debug dump fires only for UNKNOWN event types — a diagnostic necessity for new provider event vocabularies. These are provider *response* fragments at debug/warning level; API keys never appear in response bodies.
- **#2 `GATE_DBDIR` path validation — DECLINED.** Established precedent (same class declined on the native-tools QA harnesses in #659/#662): the gate script is a developer-run QA harness under `Docs/superpowers/qa/`, its env var is developer-controlled, and `path_validation.py` guards application boundaries, not local dev tooling.

Gemini consumer bot is sunset; CodeRabbit skipped (non-default base). Latest dev merged.

PR #695, opened 2026-07-17
fix(settings,library): coalesce the Settings mount recompose storm (task-290); skills-import outcome names the skill (task-291)

rmusser01:
## Summary

Two small follow-ups from this week's reviews, one branch, separated commits.

### task-290 — Settings mount-time recompose storm (c4481e3e + fixes)

Opening Settings used to trigger **three** full-screen composes. Timeline-tracing the real compose/refresh/focus events found three compounding causes:

1. **A phantom startup recompose:** `__init__` assigned the two `recompose=True` reactives (loading rows) via plain assignment, firing `refresh(recompose=True)` on the unmounted screen — the flag survived into mount and forced a recompose of the *just-composed* screen. Now seeded via `set_reactive` (no watchers/validators existed on these reactives — verified).
2. **Two independent thread workers**, each recomposing at its own nondeterministic landing. Now one combined `_refresh_sync_rows` worker computing both row sets and applying them in a single `call_from_thread` hop — Textual's flag-based `_check_recompose` coalesces the same-slot pair into **one** recompose. The solo manual-sync worker keeps its own group for targeted refreshes; the combined worker deliberately gets its own group so a solo refresh can't cancel an in-flight combined pass.
3. **Focus loss on storm landings:** `Widget.focus()` defers its `set_focus` via `call_later`, so a storm recompose landing mid navigation-dance destroyed the focus target before the deferred focus processed (the route-to-API-key test caught this the moment timings shifted). Fixed with a one-shot pending-focus intent: recorded when navigation focuses a provider field, cleared **only** when the intended widget actually lands focus (clearing on any focus was too eager — the stale category-chip focus the intent supersedes lands first), consumed by the post-recompose restore which re-issues it against the fresh children; genuinely-post-recompose focus (user action) always wins.

Pinned by: a compose-count ceiling test (mount composes ≤ 2), a **deterministic interleave test** (drives `_apply_sync_rows` mid navigation-dance, asserts the field ends focused), and the formerly-racy navigation test made settle-first (8/8 stable). Full settings hub suite: **239 passed**.

### task-291 — skills-import success line names the skill (a816c65e + fixes)

The success copy was byte-identical for every import ("1 imported · …"), so a stale outcome from the previous skill read exactly like a fresh one — the diagnosed false-pass gap for repeat imports, and unattributable for users. Now `Imported "<name>" · re-review it in the trust panel`; all test assertions are name-specific, so stale text can never satisfy the next import's expectation. Per review: the loose-file branch reports the **service-derived** (kebab-normalized) name from the returned record, not the raw file stem — `My Notes.md` imports as `my-notes` and now says so (new test). Skills suites: **10 + canvas green (168 total)**.

## Review

Independent pre-merge review (opus): "with fixes" — the Important (loose-file name divergence) fixed with the reviewer's prescribed approach + test; all three Minors applied (dead method removed, pending-intent reset on navigation-away, stale helper docstring). It verified the risks directly: no lingering-intent path survives unbounded, the `changed` guard exactly matches "an assignment will recompose", no callers remain of the removed methods, and `set_reactive` drops no load-bearing side effects.

Closes task-290, task-291.

---
## Summary by cubic
Coalesces Settings mount-time recomposes into a single recompose and preserves focus during refresh (task-290). Names the imported skill in the skills-import success line and uses the service-derived name for loose files to prevent stale-text false passes (task-291).

- **Bug Fixes**
  - Settings (task-290): Combined workers into `_refresh_sync_rows` that applies both row sets in one hop; seeded reactives via `set_reactive` to avoid a phantom recompose; added a pending-focus intent with post-recompose restore and reset on navigation-away; removed the unused `_apply_server_sync_workspace_handoff_rows`; corrected the inline comment to match behavior; tests cap compose calls at ≤2 and cover interleaved recomposes.
  - Library (task-291): Success message now reads `Imported "<name>" · re-review it in the trust panel`; loose-file imports show the stored kebab-case name (service-derived); tests updated to assert name-specific outcomes.

<sup>Written for commit b4b83c0097e4044585ad311023883ab23ea459a2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/695?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
**Qodo adjudication** (1 bug, 0 rule violations): **FIXED.** The `_pending_navigation_focus_selector` attribute comment still described the first-iteration "cleared on ANY descendant focus" behavior, while the shipped logic clears only on a satisfied-match landing (plus restore-consumption and navigation-away reset) — exactly the kind of drift that misleads future edits to subtle focus logic. Comment now states the real lifecycle. Focus/storm/interleave tests re-run: 8/8.

Gemini consumer bot is sunset; CodeRabbit skipped (non-default base).

PR #698, opened 2026-07-17
fix(llm): migrate summarize_with_cohere to the v2 /chat API (task-297)

rmusser01:
## Summary

`summarize_with_cohere` was the repo's **last Cohere v1 `/chat` caller** (the chat handler migrated in task-267 / PR #690, live-gated) — the two Cohere code paths had diverged on API version, flagged by #690's final review. This closes the divergence: grep-verified zero `cohere.ai`/v1 references remain in `tldw_chatbook/`.

## Changes

- v2 request shape: `messages` array (system inline, blank omitted), `stream` flag, `api.cohere.com/v2/chat` on both paths. Error string formats preserved exactly for the dispatcher caller.
- Non-streaming: v2 parts-array `message.content` parsing. Streaming: `content-delta` extraction.
- Inline default model aligned to the config section's own `command-a-03-2025` (the old `command-r-plus` fallback disagreed with it); dead second `system_message` default removed; streaming 4xx now reports the same pinned `Cohere: API request failed: <body>` format as non-streaming (previously a bodyless `HTTPError` through the outer except).

## The interesting finding

The independent review prescribed mirroring `chat_with_cohere`'s strict SSE `data:`-line filter — and the **live smoke caught that fix breaking real streaming to zero chunks**: this helper sends `accept: application/json`, so the real v2 stream is **raw JSON lines with no SSE framing** (unlike the chat handler, which negotiates SSE on the same endpoint). The parser now handles both framings — strips a `data:` prefix when present, skips standalone `event:` lines, parses bare JSON — with framing-specific regression tests for each shape. Another data point for the live-verification rule: same endpoint, different `Accept`, different wire format.

## Verification

- 10 tests at the `requests` boundary (`Tests/Chat/test_cohere_summarize_v2.py`): request shape, both stream framings, interleaved event lines, parts-array concat, error formats, no-key short-circuit (config-fallback patched so a machine's real key can't skew it).
- **Live smoke against the real API, re-run after every fix wave:** both paths return genuine summaries; streaming reassembles identically from 18 chunks.
- Independent review (sonnet): "with fixes" — the Important + two adjacent Minors applied; the AC (no v1 caller remains) independently grep-verified by the reviewer.

Closes task-297.

---
## Summary by cubic
Migrated `summarize_with_cohere` to Cohere v2 `/chat`, unifying request/stream behavior and hardening streaming/error handling. Removes the last v1 path, preserves error strings, and sets the default model to `command-a-03-2025`.

- **Migration**
  - Uses `https://api.cohere.com/v2/chat` with a `messages` array (blank system omitted) and `stream` flag.
  - Non-streaming: concatenates text from v2 `message.content` parts.
  - Default model set to `command-a-03-2025`.
  - Satisfies task-297; 12 tests pin request shape, both streaming framings, and error formats.

- **Bug Fixes**
  - Streaming parser handles SSE `data:` lines and raw JSON lines; skips `event:` and non-JSON metadata; guards non-dict JSON events.
  - Streaming 4xx returns the same "Cohere: API request failed: <body>" format as non-streaming; `response.json()` only on 200 to keep that format on non-JSON bodies.
  - Streaming response closes deterministically even if the consumer drops early.

<sup>Written for commit f0c9e989f0c030cbbc9d49418a4cbfe19b6154ce. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/698?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Chat/test_cohere_summarize_v2.py
+++ b/Tests/Chat/test_cohere_summarize_v2.py
@@ -202,3 +202,40 @@ def test_streaming_parses_raw_json_lines_without_sse_framing(mock_post):
     ))
 
     assert chunks == ["Raw ", "lines."]
+
+
+@patch("requests.Session.post")
+def test_non_200_with_non_json_body_keeps_pinned_error_format(mock_post):
+    """Gemini/Qodo #698-2: a gateway/CDN HTML error body must not break the
+    pinned failure format via a JSONDecodeError into the outer except."""
+    mock_response = Mock()
+    mock_response.status_code = 502
+    mock_response.text = "<html>Bad gateway</html>"
+    mock_response.json.side_effect = json.JSONDecodeError("x", "<html>", 0)
+    mock_post.return_value = mock_response
+
+    result = summarize_with_cohere("test-key", "Text.", "Prompt.")
+
+    assert result == "Cohere: API request failed: <html>Bad gateway</html>"
+
+
+@patch("requests.Session.post")
+def test_streaming_skips_non_object_json_and_closes_response(mock_post):
+    """Gemini #698-1 + Qodo #698-1: a JSON list/primitive line must not crash
+    the generator, and the response is closed when the stream ends."""
+    mock_response = Mock()
+    mock_response.status_code = 200
+    mock_response.raise_for_status = Mock()
+    mock_response.iter_lines.return_value = [
+        b'[1, 2, 3]',
+        b'"just a string"',
+        b'{"type": "content-delta", "delta": {"message": {"content": {"text": "Still works."}}}}',
+    ]
+    mock_post.return_value = mock_response
+
+    chunks = list(summarize_with_cohere(
+        "test-key", "Text.", "Prompt.", streaming=True,
+    ))
+
+    assert chunks == ["Still works."]
+    mock_response.close.assert_called_once()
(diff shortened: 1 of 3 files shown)


rmusser01:
**Bot findings adjudication** (Gemini 2 medium, Qodo 2 bugs + 1 rule violation — all four distinct findings FIXED in the latest commit, none declined):

- **Gemini/Qodo (same finding): `response.json()` before the status check — FIXED.** A non-JSON error body (gateway/CDN HTML) raised `JSONDecodeError` into the outer except, returning the generic processing string instead of the pinned `Cohere: API request failed: <body>` format. `json()` now runs only on 200; pinned by a 502-HTML regression test.
- **Gemini: non-dict parsed stream event crashes the generator — FIXED.** `json.loads('[1,2]')` → `.get()` `AttributeError` propagating to the consumer. `isinstance` guard skips non-object events; regression test covers list and primitive lines.
- **Qodo [rule]: streaming response not deterministically closed — FIXED.** The generator now wraps its event loop in `try/finally response.close()` (also matches the `chat_with_cohere` sibling), covering consumer-abandoned streams; the test asserts `close()` fires.
- **Qodo: warning spam on harmless SSE metadata/comment lines — FIXED.** Non-JSON, non-`data:`/`event:` lines (`id:`, `: keep-alive`) log at debug.

Suite: 12/12. **Live smoke re-run post-fix: both paths return real summaries (streaming 18 chunks)** — the same live check that earlier caught the strict-SSE filter regression.

PR #737, opened 2026-07-21
fix(console): classify provider failures with response-body detail + image-poisoning recovery hint (task 335)

rmusser01:
## Summary

Fifth batch from the Console UX review backlog (#720): the last P1 regression — `j3-provider-error-discards-detail-poisons-conversation`.

**Before:** an image send to a non-vision llama.cpp rendered `Agent run failed: Server error '500 …' For more information check: https://developer.mozilla.org/…` — the server's actual message (`image input is not supported - hint: you may need to provide the mmproj`, verified by curl during the review) was discarded, and because history re-sends the image every turn, every later plain-text send failed with the same opaque 500.

**After:** `Agent run failed: provider returned HTTP 500 (image input is not supported - hint: you may need to provide the mmproj). This conversation includes an image attachment; if the model can't accept images, remove that message (select it and use Delete) or switch to a vision-capable model.`

### The two leak paths
1. `describe_stream_failure`'s `(detail)` parenthetical appended the full `str(exc)` — for httpx status errors that IS the status line + MDN boilerplate. For status errors the detail is now the **response body** (best-effort JSON message extraction, truncated), and MDN boilerplate is stripped from any detail.
2. **The review's actual live path**: the agent service's `run_turn` catch-all stamped raw `str(exc)` into the STEP_ERROR summary that becomes user-facing `Agent run failed: {reason}` copy. It now classifies at capture, where the exception and its response body are still live. (Found by the integration test flowing through the outcome path, not the exception path my first fix covered.)

### Structure
- Classifier extracted to a shared `Chat/provider_failures.py` — the Agents layer must not import the Console controller; the controller re-exports `describe_stream_failure` for existing importers (top-of-file, E402-clean).
- The recovery hint fires on **both** failure sites via a session-level image probe.

## Verification
- 4 RED-first tests, including a controller-level integration with a real `httpx` 500 carrying the review's curl-proven body plus an image message in history → classified row with body detail + recovery hint, zero MDN.
- `Tests/Chat`: 1120 passed (one known pre-existing dev-tip failure). Console sweep: **1129 passed**, all 14 failures in the known dev-tip set.
- Honest caveat (in the task notes): live E2E of the *attach* step was blocked by headless-clipboard→xterm paste friction; the provider body behavior rests on the review's live curl evidence plus the real-httpx integration test.

Task file marked Done with implementation notes.

---
## Summary by cubic
Fix provider error UX in Console by surfacing the server’s response message and adding an image-rejection recovery hint; prevents opaque `httpx` 500s and reduces image-poisoned conversations. Addresses TASK-335.

- **Bug Fixes**
  - Classify failures via shared `Chat/provider_failures.py` (controller re-exports `describe_stream_failure`).
  - For HTTP errors, show response-body message (JSON-aware), collapse/trim text, and strip MDN boilerplate.
  - Agent runtime now uses the classifier for `STEP_ERROR` summaries so `Agent run failed: …` includes the body detail.
  - If session history includes images and the error is HTTP, append a hint to remove the image or switch to a vision model; tests cover JSON/plain bodies, MDN stripping, and controller/agent paths.

<sup>Written for commit f735351d239f194b6a33b52b52a366b0d3a769bf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/737?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Chat/test_console_provider_failure_copy.py
+++ b/Tests/Chat/test_console_provider_failure_copy.py
@@ -1,0 +1,113 @@
+"""TASK-335: provider failures must surface the server's message, never MDN links.
+
+The agent-runtime send path rendered raw httpx text — `Server error '500
+...' For more information check: https://developer.mozilla.org/...` — while
+the response body's actionable hint ("you may need to provide the mmproj")
+was discarded, and the re-sent image poisoned every later send with the
+same undiagnosable 500 (UX review finding
+j3-provider-error-discards-detail-poisons-conversation, REGRESSION).
+"""
+
+from types import SimpleNamespace
+
+import httpx
+import pytest
+
+from tldw_chatbook.Chat.console_agent_bridge import ConsoleAgentBridge
+from tldw_chatbook.Chat.console_chat_controller import (
+    ConsoleChatController,
+    describe_stream_failure,
+)
+from tldw_chatbook.Chat.console_chat_models import ConsoleMessageRole
+from tldw_chatbook.Chat.console_chat_store import ConsoleChatStore
+from tldw_chatbook.DB.AgentRuns_DB import AgentRunsDB
+
+MMPROJ_BODY = (
+    '{"error": {"message": "image input is not supported - '
+    'hint: you may need to provide the mmproj", "code": 500}}'
+)
+
+
+def _http_500(body: str) -> httpx.HTTPStatusError:
+    request = httpx.Request("POST", "http://127.0.0.1:9099/v1/chat/completions")
+    response = httpx.Response(500, request=request, text=body)
+    return httpx.HTTPStatusError(
+        "Server error '500 Internal Server Error' for url "
+        "'http://127.0.0.1:9099/v1/chat/completions'\n"
+        "For more information check: "
+        "https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500",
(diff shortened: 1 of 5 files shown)


rmusser01:
Rebased onto latest dev (dc196563f, clean). Review adjudication: Gemini reviewed with no findings. Qodo's two suggestions are both worthwhile follow-up features rather than defects in this fix, and both exceed this task's AC: (1) a preflight capability check before sending when history contains images — the pre-send gate already blocks NEW image attachments to non-vision models (the review praised that gate); extending it to inherited history is a design change that overlaps task-350's sub-agent/history rendering cluster; (2) persisting structured error objects instead of formatted strings touches the message schema and belongs with the persistence program. Both noted in the task file as declined-for-scope. Pre-merge verification on the rebased tree: 107 across the touched suites, full Console selection **1129 passed with all 14 failures in the known dev-tip set**. Merging.

PR #742, opened 2026-07-21
fix(settings): stop Theme/Splash Screen categories crashing the app

rmusser01:
## Summary

Selecting **Theme** or **Splash Screen** in the Settings sidebar crashed the entire app (reported by an end user as "accessing the themes page causes the application to crash").

**Root cause:** PR #713 added `THEME` and `SPLASH_SCREEN` to the Settings sidebar without adding matching entries to the hardcoded guidance table in `SettingsScreen._inspector_guidance()`. The method ended in a bare `guidance[category]` lookup, so selecting either category raised `KeyError` — and because the lookup runs inside `compose_content` → `_render_impact_pane`, the uncaught exception inside compose took down the whole app. This hit every user on dev with completely default state.

Existing tests never caught it because they mounted `SettingsThemeEditor` / `SettingsSplashScreenViewer` directly, bypassing `SettingsScreen`'s category navigation.

## Fix

- Add proper `_inspector_guidance` entries for `THEME` and `SPLASH_SCREEN` (affected config / recovery / boundary rows, matching the existing entries' style).
- Replace the bare `guidance[category]` with `guidance.get(category, <generic fallback>)` so a future missing entry degrades to generic guidance instead of crashing the app from inside compose.

## Verification

TDD (RED confirmed with the exact `KeyError` before the fix, then GREEN):

- `test_inspector_guidance_covers_every_settings_category` — sweeps every `SettingsCategoryId` through `_inspector_guidance`, so any future sidebar addition without guidance fails a unit test instead of crashing users.
- `test_theme_category_opens_without_crashing` / `test_splash_screen_category_opens_without_crashing` — real navigation through `SettingsScreen` via `DestinationHarness`, asserting the editor/viewer actually mount.

Local results: `Tests/UI/test_settings_configuration_hub.py` → 245 passed; theme editor, splash viewer, and appearance test files green. (CI checks are intentionally cancelled in this repo — verified locally.)

---
## Summary by cubic
Fixes a crash when opening the Theme or Splash Screen categories in Settings by adding missing inspector guidance and a safe fallback to prevent KeyError inside compose. Adds regression tests to cover all categories and verify both views mount.

- **Bug Fixes**
  - Add guidance entries for THEME and SPLASH_SCREEN.
  - Hoist the guidance table to module-level `_INSPECTOR_GUIDANCE`, and fall back to generic guidance with a once-per-category warning instead of raising.
  - Tests: sweep asserts every non-domain category is an explicit key in `_INSPECTOR_GUIDANCE`; navigation tests confirm the Theme editor and Splash Screen viewer mount.

<sup>Written for commit ae08c2ab2fdb71495871fd2694a4d1783eb0d073. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/742?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_settings_configuration_hub.py
+++ b/Tests/UI/test_settings_configuration_hub.py
@@ -131,6 +131,62 @@ def test_theme_and_splash_not_in_guided_mutation_categories():
     assert SettingsCategoryId.SPLASH_SCREEN not in settings_screen_module.GUIDED_SETTINGS_MUTATION_CATEGORIES
 
 
+def test_inspector_guidance_covers_every_settings_category():
+    """Every non-domain sidebar category must have an explicit guidance entry.
+
+    Regression: THEME and SPLASH_SCREEN were added to the sidebar without
+    guidance entries, so selecting them raised KeyError inside compose and
+    crashed the whole app. The runtime now returns a generic fallback so a
+    missing entry can never crash, which means asserting a truthy return would
+    always pass. This test guards the stronger property instead -- that every
+    category outside DOMAIN_SETTINGS_CATEGORY_IDS is an explicit key in
+    _INSPECTOR_GUIDANCE -- so a new category without real guidance fails CI
+    rather than silently shipping the generic fallback to users.
+    """
+    screen = SettingsScreen(_build_test_app())
+    for category in SettingsCategoryId:
+        guidance = screen._inspector_guidance(category)
+        assert guidance, f"no inspector guidance for {category!r}"
+        if category in settings_screen_module.DOMAIN_SETTINGS_CATEGORY_IDS:
+            continue
+        assert category in settings_screen_module._INSPECTOR_GUIDANCE, (
+            f"{category!r} has no explicit guidance entry and would fall back "
+            f"to generic text; add it to _INSPECTOR_GUIDANCE"
+        )
+
+
+@pytest.mark.asyncio
+async def test_theme_category_opens_without_crashing():
+    """Selecting the Theme category mounts its editor without crashing compose.
+
+    The editor widget mounting is the regression signal: the original bug
+    raised KeyError inside compose, so a crash would leave the editor unmounted
+    and time this wait out. Editor content is verified in
+    test_settings_theme_editor.py.
+    """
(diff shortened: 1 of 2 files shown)


rmusser01:
Rebased onto latest `dev` (`015bd6f3d`) and addressed the Qodo review:

1. **Guidance coverage test ineffective (Bug, High)** — fixed. Hoisted the guidance table to a module-level `_INSPECTOR_GUIDANCE` and changed the sweep test to assert every *non-domain* `SettingsCategoryId` is an **explicit key** in it, not merely that the lookup returns something truthy. Verified this is a real guard: temporarily removing the `THEME` entry now fails the test (previously it passed because the `.get()` fallback is always truthy).

2. **Fallback guidance is silent (Bug, Medium)** — addressed. The fallback now emits a `logger.warning` once per category (deduped via a module-level `set`) so a missing entry is visible in logs without spamming every compose pass. It still returns generic guidance rather than raising, since this runs inside compose.

3. **Async tests missing docstrings (Rule)** — fixed. Added Google-style docstrings to both async tests.

4. **Duplicated `size=(190, 55)` (Rule)** — respectfully declining. This literal (and `size=(180, 50)`) appears ~40 times across this test file with no named constant; introducing one only for these two tests would make them inconsistent with the established file convention. Qodo flagged this Low and noted similar refactors were declined previously.

Also made the two navigation tests robust against pre-existing mount-storm timing flakiness by polling for the mounted editor/viewer (`_wait_for_selector`) instead of a fixed settle, and dropped the incidental "Theme Library" content assertion (editor content is already covered by `test_settings_theme_editor.py`; widget mount is the actual crash-regression signal).

Local verification: `Tests/UI/test_settings_configuration_hub.py` → **245 passed**.

PR #759, opened 2026-07-22
fix(console): dock feedback toasts to the top so they don't cover/swallow composer clicks (task 352)

rmusser01:
## Summary

Console UX review task **352** (P2, high confidence). Textual docks notification toasts **bottom-right** by default — directly over the Console composer's **Send/Attach/Save** cluster and the staged-chip strip. Because toasts capture clicks (click-to-dismiss), a click aimed at those controls during a ~5s toast dismisses the toast instead of pressing the button. The finding measured this across three journeys: every attach/clear/limit toast (J3), the boot "Model catalog" toast (J4), and the small-terminal boot toast (J6). The app had no toast override.

## Change

A one-rule `ChatScreen.DEFAULT_CSS` docks the Console screen's toast rack to the **top-right**:

```css
ChatScreen ToastRack {
    dock: top;
    align: right top;
    margin-top: 1;
    margin-bottom: 0;
}
```

A top-docked rack can never overlap the bottom composer cluster, so feedback no longer obscures — nor swallows clicks aimed at — the composer controls. The only thing beneath a top-right toast is the header's read-only status chips.

**Why `DEFAULT_CSS`, not the CSS bundle:** the bundle isn't loaded by the test harnesses (they use widget `DEFAULT_CSS` only). Keeping the rule in `ChatScreen.DEFAULT_CSS` means it applies in **both** the real app and the harness — which is what makes the fix regression-testable. Scoped to `ChatScreen` (2-type selector beats Textual's 1-type `ToastRack` default) and additive to `BaseAppScreen.DEFAULT_CSS` via the MRO.

Toast duration is left at Textual's short default — with the occlusion removed the ~5s lifetime is no longer harmful, and changing the global notify timeout across ~95 `notify()` sites is out of scope.

## Verification

- **RED→GREEN** test `test_console_toast_rack_docks_top_not_over_composer`: mounts a `ToastRack` under the Console screen and asserts the CSS docks it top (the headless notification system doesn't render toasts, so the computed `dock`/`align` style is asserted directly). Was `bottom` before, `top` after.
- 44 layout tests confirm the added `DEFAULT_CSS` is additive (BaseAppScreen styling intact).
- Served-app capture: the composer region is clean at boot (previously the "Model catalog" toast sat over it). The boot catalog-refresh toast can't be re-triggered live because the capture harness aborts outbound https, but the top-dock mechanism is deterministic and proven by the test.
- Full Console sweep: **1218 passed / 15 failed — all 15 within the known baseline, 0 outside**.

Files: `tldw_chatbook/UI/Screens/chat_screen.py`, `Tests/UI/test_console_toast_placement.py`.

---
## Summary by cubic
Move Console feedback toasts to the top-right so they no longer cover the composer or intercept clicks. Satisfies TASK-352 acceptance criteria.

- **Bug Fixes**
  - Override `ToastRack` placement via `ChatScreen.DEFAULT_CSS` (`dock: top; align: right top`), scoped to `ChatScreen`.
  - Keep the rule in widget `DEFAULT_CSS` (not the CSS bundle) so it applies in both the app and test harness.
  - Add regression test `test_console_toast_rack_docks_top_not_over_composer` that mounts a `ToastRack` and asserts the top dock; uses a private `textual.widgets._toast` import since `ToastRack` isn’t publicly exported in headless mode.

<sup>Written for commit 6414fe1f6ba24a74605b23125328476ae73d798b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/759?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_console_toast_placement.py
+++ b/Tests/UI/test_console_toast_placement.py
@@ -1,6 +1,12 @@
 """Console feedback toasts must not sit over the composer action cluster (task-352)."""
 
 import pytest
+
+# Textual does not export ToastRack publicly (it is not in `textual.widgets`),
+# and the headless notification system does not mount one, so the test reaches
+# into the private module to construct one directly and assert the Console
+# screen's CSS docks it. Mirrors the existing `textual.widgets._markdown`
+# private import already used under Tests/.
 from textual.widgets._toast import ToastRack
 
 from Tests.UI.test_destination_shells import _build_test_app, _wait_for_selector
@@ -10,7 +16,7 @@ from Tests.UI.test_product_maturity_gate1_core_loop_screen_adaptation import (
 
 
 @pytest.mark.asyncio
-async def test_console_toast_rack_docks_top_not_over_composer():
+async def test_console_toast_rack_docks_top_not_over_composer() -> None:
     """Textual docks notification toasts bottom-right by default — directly over
     the Console composer's Send/Attach/Save cluster and staged-chip strip — and
     toasts capture clicks, so a click aimed at those controls during a toast
--- a/Tests/UI/test_console_toast_placement.py
+++ b/Tests/UI/test_console_toast_placement.py
@@ -1,0 +1,40 @@
+"""Console feedback toasts must not sit over the composer action cluster (task-352)."""
+
+import pytest
+from textual.widgets._toast import ToastRack
+
+from Tests.UI.test_destination_shells import _build_test_app, _wait_for_selector
+from Tests.UI.test_product_maturity_gate1_core_loop_screen_adaptation import (
+    ConsoleHarness,
+)
+
+
+@pytest.mark.asyncio
+async def test_console_toast_rack_docks_top_not_over_composer():
+    """Textual docks notification toasts bottom-right by default — directly over
+    the Console composer's Send/Attach/Save cluster and staged-chip strip — and
+    toasts capture clicks, so a click aimed at those controls during a toast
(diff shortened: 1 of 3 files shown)


rmusser01:
Addressed Qodo's review:

- **#2 missing return type** — fixed: added `-> None` to the test (6414fe1f).
- **#1 Private ToastRack import** — kept, with a clarifying comment. Textual does not export `ToastRack` publicly (`from textual.widgets import ToastRack` raises), and the headless notification system doesn't mount one, so constructing it directly from the private module is the only way to assert the screen's CSS docks it. There's existing precedent for this under `Tests/` (`textual.widgets._markdown`).
- **#3 Non-Google docstring in test** — not changing: the test takes no arguments and returns nothing, so a Google `Args:`/`Returns:` block would be empty; the sibling `test_console_native_chat_flow.py` uses descriptive prose docstrings for all 142 of its tests.

Rebased onto latest dev. Local: toast-placement test green, 44 layout tests confirm the DEFAULT_CSS is additive, full Console sweep 1218 passed / 15 baseline.

PR #771, opened 2026-07-22
RAG SP1: per-embedding-config index isolation (foundation)

rmusser01:
## SP1 — Per-embedding-config index isolation (foundation)

First of a **3-part program** (RAG settings screen + swappable profiles). This PR delivers **only SP1**; the SP2 (profile system) and SP3 (settings screen) specs are included as **design docs for context** — no code for them here. Overview: `Docs/superpowers/specs/2026-07-21-rag-settings-profiles-overview.md`.

### Why
Profiles will let users change the embedding model / chunking / distance metric. Today all vectors land in one `default` collection separated only by `source_type`, so such a change would silently mix incompatible vectors (dimension-mismatch crash, or — for chunk/metric changes — silent wrong rankings). SP1 makes collections **keyed by the config that built them** so that swap is safe. It's the prerequisite for SP2/SP3.

### What
- **`collection_fingerprint.py`** (pure): a versioned fingerprint over index-determining config only — `embedding.model` + `max_length`, all 10 `ChunkingConfig` fields, `vector_store.distance_metric`. Query-time settings (top_k, hybrid_alpha, reranking, citations) share one index. Normalized (TOML `"400"` == int `400`), ASCII-only Chroma-safe names, no `..`.
- **Single seam:** `RAGService.__init__` resolves the collection via `fingerprinted_collection_name(self.config)` and stamps provenance metadata — inherited by `EnhancedRAGService(V2)`, so ingestion, backfill, and search all resolve the same fingerprinted collection (embed/query parity).
- **`collection_indexes.py`:** provenance-aware, **idempotent, race-safe** legacy-`default` adoption migration (wired into `create_rag_service`, never blocks service creation) so existing users keep their index on upgrade — plus a `list_indexes` / `delete_index` / `index_status` admin API over the existing store CRUD.
- **Honest empty state:** a fresh fingerprinted collection reuses the existing `semantic_index_is_empty` probe (empty until backfilled), distinct from zero-results.

### Notable finds during implementation
- **Latent production bug fixed** (surfaced by the empty-honesty test): `ChromaVectorStore.get_collection_stats` raised numpy "truth value ambiguous" on `peek()["embeddings"]` (numpy arrays in chromadb 1.5.8), swallowed → masked the real count with `{count:0,error:…}` on essentially every call, silently degrading health-check → DEGRADED and `get_chunk_count` → 0. Fixed with explicit `len()` checks; a trustworthy positive-int count now flows to all consumers.
- **Two chromadb 1.5.8 migration landmines** handled: `modify()` rejects the `hnsw:space` key even to restate it (stripping it preserves the metric, held in `configuration_json`); and the per-`persist_directory` `SharedSystemClient` cache errors on `Settings` mismatch (the migration client now matches the store's `Settings` exactly). An e2e regression test locks the second (verified by break-and-revert).

### Known limitation (documented)
Only the canonical `default` legacy collection is auto-adopted; a user with a custom `[rag.vector_store].collection_name` gets a fresh empty fingerprinted collection and must re-index (out of scope per SP1 spec §4).

### Testing
`Tests/RAG/` — **468 passed, 8 skipped** (pre-existing optional-dep skips), 0 failures. New: `test_collection_fingerprint.py`, `test_collection_indexes.py`, `test_index_isolation_integration.py` (28 tests, real chromadb 1.5.8). Cross-SP invariant (adopted fingerprint == SP2 imported-profile fingerprint; embed-config == query-config) is documented and lands as an end-to-end test in SP2.

### Process
SDD-executed: fresh subagent + spec/quality review per task, plus a final opus whole-branch review = **ready to merge, 0 Critical / 0 Important**. backlog `task-476`.

---
## Summary by cubic
Isolates RAG vector indexes per embedding/chunking/distance config by fingerprinting collection names, preventing mixed or incompatible vectors and enabling safe future profile swaps. Adds an idempotent migration that adopts the legacy `default` collection, small index admin helpers, and fixes `chromadb` quirks while enforcing distance-metric precedence.

- **New Features**
  - Versioned collection fingerprint over embedding model, `max_length`, all chunking fields, and distance metric; query-time settings share an index. ASCII- and Chroma-safe names.
  - RAG services resolve collections via the fingerprint and stamp provenance metadata, keeping ingest, backfill, and search aligned.
  - Index admin helpers: list, delete, and status for fingerprinted collections.
  - Honest empty state: new collections report empty until backfilled.
  - Migration: automatically adopts the canonical legacy `default` collection (also on the no-profiles fallback path); custom `collection_name` users get a fresh empty index and must re-index.

- **Bug Fixes**
  - Fixed `chromadb` 1.5.8 numpy truthiness issue in collection stats, restoring accurate counts and health checks.
  - Handled `chromadb` 1.5.8 quirks: ignore `hnsw:space` on modify and match `Settings` to avoid client cache collisions; covered by an e2e regression test.
  - Enforced `hnsw:space` precedence on collection creation so provenance metadata can’t override the configured distance metric.

<sup>Written for commit c2c21f4514516a84f84b9b9decba0e9fe6d5e40a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/771?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Thanks for the reviews. Resolution:

- **hnsw:space override risk (Qodo #3)** — fixed. `ChromaVectorStore.collection` now merges `collection_metadata` **before** `hnsw:space`, so the real distance metric always wins regardless of caller-supplied metadata; locked by `test_collection_metadata_cannot_override_hnsw_space`.
- **No-profiles fallback path** — `create_rag_service`'s ultimate fallback now also runs `maybe_adopt_legacy_collection`, so that (unreachable-in-practice) path preserves an existing index rather than opening an empty fingerprinted one.
- **persist_directory not path-validated (Qodo #1)** — declined for SP1 and filed as follow-up **task-482**. This is a pre-existing, store-wide pattern (`ChromaVectorStore` at vector_store.py:199/273 already passes the config path straight to `PersistentClient`). Validating only the new module would risk a normalized-vs-raw path-string divergence from the store → the `SharedSystemClient` per-persist_directory client-cache collision this PR specifically avoids. task-482 hardens it uniformly across both Chroma client sites.

`Tests/RAG/` green. Rebasing onto latest dev and merging.

PR #777, opened 2026-07-22
feat(console): task-457(a) cold-send optimistic user echo

rmusser01:
## What

Implements **task-457(a)**: the Console first-send no longer waits on the provider readiness probe before the user's message appears. The USER row is now echoed into the transcript **before** `resolve_for_send`, so a slow/cold provider no longer leaves the transcript blank while the composer clears — the message reads as *sent*, not lost.

(This PR supersedes its earlier backlog-only scoping content. The core-model blocker recorded then is resolved here.)

## Behaviour

- **Cold/slow provider:** the echoed row appears immediately; the composer keeps the draft.
- **Not-ready provider (block):** the echoed row persists next to an honest SYSTEM block-row (previously the message was **silently dropped**), and the draft is kept so the user re-attempts. Staged attachments stay staged for the retry.
- **Success:** unchanged flow after the echo.

## Core-model piece (the recorded blocker, now solved)

A blocked-but-echoed USER row must not pollute the **next** send's provider context. New `ConsoleChatStore.mark_message_send_blocked(id)` flips a never-streamed row to `status="failed"` **without** the assistant-stream terminal guard (`mark_message_failed` raises for non-assistants; `append_message` returns a snapshot so direct mutation is a no-op). `_provider_messages_for_session(skip_failed=True)` then already excludes it.

Two rendering gates keep a `failed` USER row honest:
- the transcript stream-state suffix (`[failed]`) is now **assistant-only**;
- the `retry` message action is now **assistant-only** (a USER row has no assistant response to regenerate — the user re-sends from the composer).

## Auto-title ordering

Because the persisting append creates the durable conversation from `session.title` and sets `persisted_conversation_id` (after which `_maybe_auto_title_session` early-returns), the auto-title now runs **before** the append. The conversation is created with the derived title (e.g. `"hello"`) instead of the default `"Chat 1"`, keeping the workspace conversation rail correct after persistence. Regression covered by `test_console_send_refreshes_workspace_conversation_rail_after_persistence`.

## Tests

RED→GREEN:
- `test_console_chat_store.py::test_mark_message_send_blocked_fails_a_user_row_for_context_exclusion`
- `test_console_chat_controller.py::test_not_ready_provider_still_echoes_the_user_message` (roles == [user, system], row failed, draft kept) + updated `test_blocked_provider_wip_copy_is_normalized_once_in_controller`
- `test_console_message_actions.py::test_failed_user_row_offers_no_retry_action`
- `test_console_native_transcript.py` blocked-user body renders `hello` (no `[failed]` suffix)

Full runs: 274 Chat + transcript tests green; `test_console_native_chat_flow.py` = 186 passed. The 2 `continue`/`regenerate` failures are **pre-existing on the current dev tip** (verified by reverting the 4 production files to origin/dev — they fail identically without this change); `test_console_workspace_conversation_search_blank_query_clears_error_cache` is a load-contention flake (passes 3/3 in isolation) and is unrelated to this change (it does not touch the send path).

## Value note

Resolve is ~24ms warm, so this mainly helps a genuinely cold/slow provider startup — but it removes a real "did my message send?" ambiguity and stops the silent-drop on a blocked send.

### DIFF
--- a/Tests/Chat/test_console_chat_controller.py
+++ b/Tests/Chat/test_console_chat_controller.py
@@ -43,6 +43,11 @@ class BlockedGateway:
         )()
 
 
+class RaisingProbeGateway:
+    async def resolve_for_send(self, selection):
+        raise RuntimeError("probe boom")
+
+
 class StreamingGateway:
     async def resolve_for_send(self, selection):
         return type(
@@ -343,6 +348,26 @@ async def test_not_ready_provider_still_echoes_the_user_message():
     # context, and the draft is preserved for a re-attempt.
     assert messages[0].status == "failed"
     assert result.should_clear_draft is False
+
+
+@pytest.mark.asyncio
+async def test_probe_exception_after_optimistic_echo_marks_row_blocked():
+    """TASK-457(a) (Qodo #777 review): if the readiness probe raises (or is
+    cancelled) after the optimistic USER echo, the echoed row must still be
+    failed so a never-sent message cannot leak into the next send's provider
+    context (skip_failed only drops failed rows). The error still propagates."""
+    store = ConsoleChatStore()
+    controller = ConsoleChatController(
+        store=store, provider_gateway=RaisingProbeGateway()
+    )
+
+    with pytest.raises(RuntimeError):
+        await controller.submit_draft("hello")
+
+    messages = store.messages_for_session(store.active_session_id)
+    assert [message.role.value for message in messages] == ["user"]
+    assert messages[0].content == "hello"
+    assert messages[0].status == "failed"
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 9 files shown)


rmusser01:
Addressed both focus areas from the Qodo review in `dd63c4d4b`:

**Exception Handling** — real bug, now fixed. Moving the echo before the probe meant a `resolve_for_send` that *raises or is cancelled* left the USER row `"complete"`, so it could leak into the next send's context (`skip_failed` only drops `"failed"`). `submit_draft` now wraps the probe: any unsuccessful exit (not-ready **or** exception/cancel) marks the echoed row send-blocked; the exception path re-raises so the caller still sees the failure. Regression test: `test_probe_exception_after_optimistic_echo_marks_row_blocked` (probe raises → row `"failed"`, error propagates).

**Missing Validation** — fixed. `mark_message_send_blocked` now rejects non-USER rows and mid-stream (`pending`/`streaming`) rows. The optimistic echo is always a USER row, so this stops a mistaken caller from flipping an assistant/system or in-flight row to `"failed"` and bypassing the assistant terminal-state guards (`mark_message_failed`). Regression test: `test_mark_message_send_blocked_rejects_non_user_rows`.

276 Chat-suite green + send-path native-flow green.

PR #778, opened 2026-07-22
Bound Console conversation history by tokens before dispatch (TASK-322)

rmusser01:
## What

Bounds the native Console conversation history by real tokens against the model window before dispatch (closes **TASK-322**). Previously the send path sent the entire conversation every turn with only an *image* budget — long conversations hit provider 400s (un-continuable) or silent local front-truncation (dropping the system prompt first), with unbounded token cost.

Spec: `Docs/superpowers/specs/2026-07-22-console-history-token-budget-design.md`
Plan: `Docs/superpowers/plans/2026-07-22-console-history-token-budget.md`

## How

- **New pure module `console_history_budget.py`** — `bound_messages_to_window(...)` drops oldest **whole turns** (a user + its following rows) until the payload fits, **always preserving** the leading `system` prefix and the current turn (last user → end). `budget = window − response_reservation − max(512, window//50)`; `response_reservation = self.max_tokens or 1024`.
- **Multimodal-aware counting adapter** — the existing `count_tokens_messages` crashes on the Console's vision payloads (`content` is a list of parts; it does `content.split()` on a list). The adapter flattens text parts + adds a flat per-image estimate, delegating text to the existing counter so tasks 320/321 flow through the same seam.
- **Wired at the single dispatch choke point** — trim at the top of `_stream_assistant_response`, *before* the agent-vs-direct branch, so both the direct provider path and the agent/tool path (`agent_messages = list(provider_messages)`) get the bounded list.
- **Display-only trim note** — when history was dropped, a `SYSTEM` transcript row "Earlier messages were trimmed to fit the model's context window (N dropped)." `_provider_message_payloads` filters `SYSTEM` rows, so it renders but is never resent. The session lookup is guarded against a session-close race (mirrors `_append_failure_system_row`).

## Scope boundaries (documented in the spec)

- Built on the existing `token_counter` seam — **tasks 320/321** sharpen the window table + non-OpenAI tokenizer transparently. Unknown local models get a conservative 4096 default (over-trims until 320/325); the `bound_messages_to_window(window=...)` override is ready for **task-325** to feed a configured limit.
- **Intra-agent-loop** growth (tool-call/result accumulation across iterations) is **task-326 (RunBudget)**'s scope, not this.

## Testing

- **8 unit tests** (boundary, system/current-turn preservation, degenerate, whole-turn drop, multimodal without crash + per-image cost, window override) — trim-logic tests inject a deterministic `count_fn` since **tiktoken is absent** in the venv; the multimodal test uses the real counter.
- **3 controller tests** — a trimming send trims `messages_seen` + appends the note; a fitting send does neither; a session-close race is swallowed.
- **114 passed** across the feature + controller suites; full `Tests/Chat` sweep = **zero novel regressions** vs `origin/dev` (the pre-existing failures — 6 missing-`pytest-mock` errors + 1 anthropic-tools — are in files this branch doesn't touch).
- Whole-branch review (opus): Ready to merge, 0 Critical/Important.

---
## Summary by cubic
Bounds Console conversation history by real tokens before dispatch to keep chats continuable and control cost. Hardened multimodal counting, optimized trim selection, and budget now uses the dispatch `resolution` to prevent race-driven under-trims (TASK-322).

- **New Features**
  - Added `tldw_chatbook/Chat/console_history_budget.py` with `bound_messages_to_window(...)` that drops oldest whole turns, always keeps the leading `system` prompt and the current turn, and respects a response reservation (`max_tokens` or `DEFAULT_RESPONSE_RESERVATION`).
  - Added multimodal-aware `count_console_messages_tokens(...)` to flatten text parts and add a per-image estimate, avoiding crashes on list `content`.
  - Trim is applied at the top of `_stream_assistant_response` in `console_chat_controller.py`, so both direct provider and agent/tool paths use the bounded messages, and a display-only `SYSTEM` note is appended when trimming occurs (filtered from the next provider payload).

- **Bug Fixes**
  - `count_console_messages_tokens` now skips non-string multimodal text parts to prevent send-path crashes on malformed payloads.
  - `bound_messages_to_window` now uses a binary search to choose the drop count (O(n log n)) with identical results.
  - Trim budgets against `resolution.model`/`resolution.provider`/`resolution.max_tokens` instead of mutable controller state, avoiding under-trim races; added a regression test in `Tests/Chat/test_console_chat_controller.py`.

<sup>Written for commit da0b5779e2cebbc9fad33cdeb6de7d651497cd53. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/778?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Thanks — addressed in `b99d849`:

**1. Non-string text join crash (`console_history_budget.py`) — fixed.** `count_console_messages_tokens` now filters multimodal text parts to `isinstance(t, str) and t` before joining, so a malformed/legacy non-string `text` value is skipped rather than raising `TypeError` on the send path. This matches the controller's existing `isinstance(..., str)` guards on the same data. Added a regression test (`test_non_string_text_part_does_not_crash_and_is_skipped`).

**3. Quadratic trimming loop (`console_history_budget.py`) — fixed.** The token count is monotonically non-increasing as more oldest turns drop (each turn contributes ≥ 0 tokens), so `bound_messages_to_window` now binary-searches the minimal drop count instead of re-counting the whole payload once per dropped turn. That takes the long-history path from O(n²) to O(n log n); the chosen drop count and returned messages are identical to dropping one turn at a time (all existing exact-count tests pass unchanged). Added `test_long_history_drops_exact_turns_via_binary_search` (50 turns).

**2. Zero `max_tokens` treated as unset (`console_chat_controller.py`) — declining, with evidence.** `max_tokens=0` is not a valid configured value here: `console_session_settings.py:478-479` validates `max_tokens` as `>= 1` when present (else blank/None), so the only falsy values that reach the controller are `None`/blank. `self.max_tokens or DEFAULT_RESPONSE_RESERVATION` therefore correctly maps "unset" to the 1024-token reservation. The suggested `if self.max_tokens is not None` change would instead reserve **0** response tokens for a stray `0`, letting history fill the entire window and immediately overflowing the reply — strictly less safe than the current behavior for a value the settings layer already forbids.

rmusser01:
Thanks for the second pass — addressed in `8219dbc`:

**Selection Race — fixed.** Good catch. The trim was reading `self.model`/`self.provider`/`self.max_tokens`, but the dispatch immediately below sends `resolution` (`stream_chat(resolution, provider_messages)`) — the model/provider/max_tokens `resolve_for_send` captured at the send site. A provider/model switch racing the awaits between that capture and `_stream_assistant_response` (skill substitution, chat dictionaries, world-info — the same awaits task-427's owner-lookup guard was added for) could budget against a different window than the model actually dispatched, under-trimming into a provider 400. Now sourced from `resolution` so the budget and the dispatch always agree. Added `test_trim_budgets_against_resolution_model_not_controller_state` (tiny-window `resolution` + huge-window `self.model` → still trims).

**Incomplete Counting — accepted by design (documented).** The counter is an intentionally conservative *estimate*, not an exact accounting: (1) the native Console send path's payload is USER/ASSISTANT content only — `_provider_message_payloads` filters `tool`/`system` rows out, so `tool_calls`/function-arg metadata aren't present in the dispatched list to count; (2) the budget already subtracts a safety margin (`max(512, window//50)`) on top of the response reservation precisely to absorb estimation error; and (3) the degenerate over-budget case keeps system+current anyway and lets the provider/local server handle any residual — it never hard-fails. Tightening the tokenizer (including any provider-specific metadata) is the documented seam of **task-321** (real non-OpenAI tokenizer), which flows through this same counter with no logic change here.

PR #780, opened 2026-07-22
RAG SP2a: profile system storage foundation

rmusser01:
## SP2a — RAG profile system: storage foundation

Second sub-project of the RAG settings screen + profiles program; the **safe half of SP2**. Makes the existing `ConfigProfileManager` a *correct, file-backed, user-facing* profile store — **without** touching the risky config-resolution wiring (that's **SP2b**, its spec already written, a separate user-gated PR). Depends on SP1 (index isolation, #771 merged). Backlog `task-483`.

All production changes are in `tldw_chatbook/RAG_Search/config_profiles.py`.

### What & why
- **Fixed the broken serialization round-trip.** `ProfileConfig.from_dict` used `RAGConfig(**dict)`, leaving sub-configs as raw dicts (`.rag_config.embedding.model` → AttributeError). Now routes through `RAGConfig.from_dict` — the prerequisite for file storage.
- **Fixed the builtins, which silently ignored their own settings.** Most builtins set *dead attributes* (`.chunking.size`/`.overlap`/`.search.top_k`/`.default_type`, `type="in_memory"`) that aren't the real dataclass fields — so "fast" (chunk 256) and "accuracy" (chunk 512) actually ran identical defaults. Corrected to the real fields + valid enum values (keyword→`plain`, store type `memory`), removed a stray `.chunking.method` dead attr, and fixed `validate_profile` to read real fields. Now fast/balanced/accuracy genuinely differ.
- **Stable, rename-safe `id` + `read_only` marker.** All ~13 builtins stamped read-only; user profiles keyed by a `_slugify` id that survives rename.
- **File-per-profile storage** under `rag_profiles/` (one JSON per user profile), with one-time legacy `custom_profiles.json` → per-file migration (idempotent, no-clobber). `default=str` handles chroma `persist_directory` Paths on write; `from_dict` restores them on read.
- **CRUD with read-only guards:** `save_profile` / `delete_profile` / `rename_profile` / `clone_profile` — none can mutate a builtin; clone deep-copies any profile (incl. a builtin) into a fresh writable one.

### Notable review catches (all fixed)
The read-only invariant — "a builtin can never be mutated/shadowed/deleted" — turned out to have **three** distinct holes, each reproduced and closed:
1. `save_profile` id-collision: a new profile named "High Accuracy" (→ id `high_accuracy`) could overwrite the builtin. Guarded.
2. **Load/migration path**: a *legacy* custom profile named "High Accuracy" migrated to `high_accuracy.json` and shadowed the builtin on every boot. Fixed with uniquify-on-collision in both the loader and the migration.
3. The uniquify fix itself introduced an **order-dependent file-clobber** (checked in-memory but not disk, so it could overwrite an unrelated on-disk profile). Fixed with a disk-aware uniquify + write-before-delete ordering.
Also fixed: `clone_profile` shared the `tags` list by reference (corrupted the source); `create_custom_profile` returned a half-built object (same `RAGConfig(**dict)` anti-pattern); and a `.search.top_k` ripple in `enhanced_rag_service_v2.py` from the builtin rename.

### Testing
`Tests/RAG/` — **492 passed, 8 skipped, 0 failed**. New: `Tests/RAG/test_config_profiles.py` (23 tests: round-trip incl. chroma Path, builtin settings applied + valid enums, id/read_only + backfill, file storage + legacy migration, CRUD guards, all three read-only-invariant holes as regression locks).

### Out of scope (SP2b, next)
Config-resolution unification (`resolve_active_rag_config`), routing `config.py`'s loader + the active pointer through the profile, `reset_shared_rag_service()`, first-run "Imported settings" import, embed==query parity. This PR does not touch `config.py`'s loader, `ingestion_indexing.py`, `rag_factory.py`'s resolution, or `[rag.service].profile`.

### Follow-ups filed
- `task-484` — three builtins (`hybrid_full`/`research_papers`/`technical_docs`) use the correct field `chunking_method` with invalid values ("hierarchical"/"structural") that would raise on the enhanced-chunking path (pre-existing, different bug class).

### Process
SDD-executed: fresh subagent + spec/quality review per task, plus a final opus whole-branch review; every finding traced to file:line and fixed or filed.

---
## Summary by cubic
Foundation for the RAG profile system (SP2a): makes `ConfigProfileManager` a correct, file-backed, user-facing store with stable ids and read-only built-ins; adds path-safe ids and deterministic load/migration to prevent clobbers. Addresses TASK-483.

- **New Features**
  - File-per-profile storage under `rag_profiles/` with one-time auto-migration from `custom_profiles.json`.
  - `ProfileConfig` gains stable `id` and `read_only`; built-ins are read-only; safe CRUD (`save_profile`, `delete_profile`, `rename_profile`, `clone_profile`) with disk-aware id uniquify; rename keeps id/file; `create_custom_profile` uniquifies ids upfront.
  - Reliable round-trip: `ProfileConfig.from_dict` uses `RAGConfig.from_dict`; `to_dict` copies `tags`; JSON write handles `Path` via `default=str`.

- **Bug Fixes**
  - Path-safe ids and load hardening: `_profile_path` rejects non-slug ids; loader treats filename stem as the source of truth and self-heals divergence; glob is materialized/sorted and reassignment widens to avoid clobber when stems collide (e.g., `foo-bar.json` vs `foo_bar.json`) using disk+memory-aware uniquify.
  - Migration isolation: per-entry try/except so one bad entry can’t abort others; legacy blob is always renamed via `os.replace`; stray files are never treated as the legacy blob; names canonicalize onto reserved filenames.
  - Immutable read-only guard: capture builtin ids as a frozenset so save/delete/rename cannot be bypassed by flipping `read_only`; saving cannot collide with a read-only profile; clone uses independent `tags`.
  - Built-ins use real fields and valid enums; `validate_profile` checks real fields; closes ways to mutate or shadow built-ins (save collisions, legacy migration, hand-placed files) with guards and self-heal; updated `simplified/enhanced_rag_service_v2.py` to read `default_top_k`.

<sup>Written for commit f0639b9fd3dd0c280dea300d7fd4472a4cc3ba1d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/780?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/plans/2026-07-22-rag-profile-system-sp2a-storage.md
+++ b/Docs/superpowers/plans/2026-07-22-rag-profile-system-sp2a-storage.md
@@ -1,0 +1,599 @@
+# SP2a — Profile System: Storage Foundation — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Turn the existing `ConfigProfileManager` into a *correct*, file-backed, user-facing profile store: fix the broken serialization round-trip, fix the builtins that silently ignore their own settings, mark builtins read-only, give profiles a rename-safe stable id, store one file per user profile, and add real CRUD.
+
+**Architecture:** All changes live in `tldw_chatbook/RAG_Search/config_profiles.py`. `ProfileConfig` gains a stable `id` and a `read_only` flag. `ProfileConfig.from_dict` routes `rag_config` through `RAGConfig.from_dict` (proper nested-dataclass reconstruction). The ~13 builtins are corrected to the real dataclass field names + valid enum values and marked `read_only`. The single `custom_profiles.json` blob is replaced by one JSON file per user profile under the existing `rag_profiles/` dir (legacy blob auto-migrated). CRUD (`save_profile`/`delete_profile`/`rename_profile`/`clone_profile`) guards against mutating builtins.
+
+**Tech Stack:** Python 3.11+, existing `ProfileConfig`/`RAGConfig` dataclasses, `json`, pytest. Tests pass an explicit temp `profiles_dir` to `ConfigProfileManager(...)` (note: `get_profile_manager()` re-instantiates per call — it is NOT a singleton — so tests construct the manager directly with a temp dir).
+
+## Global Constraints
+
+- **Spec:** `Docs/superpowers/specs/2026-07-21-rag-profile-system-design.md`; overview `Docs/superpowers/specs/2026-07-21-rag-settings-profiles-overview.md`. This plan is **SP2a only** — the storage foundation.
+- **OUT OF SCOPE (this is SP2b, a later PR):** the config-resolution unification (`resolve_active_rag_config`), routing `config.py`'s loader through the active profile, deprecating `AppRAGSearchConfig.rag.*` value keys, the active-profile pointer wiring, `reset_shared_rag_service()`, the first-run "Imported settings" import, and the embed==query parity test. **Do NOT touch `config.py`'s loader, `ingestion_indexing.py`, `rag_factory.py`'s resolution, or `[rag.service].profile` in SP2a.**
+- **Owner decision:** builtins are fixed to use the REAL fields (not left as-is), so fast/balanced/accuracy actually differ.
+- **Real dataclass fields (verbatim):** `ChunkingConfig.chunk_size`, `.chunk_overlap`, `.chunking_method` (NOT `.size`/`.overlap`). `SearchConfig.default_top_k`, `.default_search_mode` (values `"plain"|"semantic"|"hybrid"`), `.score_threshold` (NOT `.top_k`/`.default_type`). `VectorStoreConfig.type` ∈ `{"chroma","memory","auto"}` (NOT `"in_memory"`).
+- **Value mapping for `default_type` → `default_search_mode`:** `"keyword"→"plain"`, `"semantic"→"semantic"`, `"hybrid"→"hybrid"`. This is a value map, not a blind rename.
+- **No DB migration** (profiles are files). **Builtins are never written to disk.**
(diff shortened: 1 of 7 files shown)


rmusser01:
Addressed the bot + independent review findings across three fix waves (pushed):

**Security / correctness (Important):**
- **Unvalidated `profile_id` paths (Qodo #1):** `_profile_path` now raises on any non-slug id, and the loader treats the **filename stem as authoritative** (re-slugified), so a hand-edited profile JSON with a traversal `id` (`../…`) can never drive `open`/`unlink`/`rename` outside `rag_profiles/`. This also closes the id/filename-mismatch orphaning (Qodo #5).
- **Read-only guard bypass (Qodo #4):** enforcement is now anchored to an immutable `_builtin_ids` frozenset captured at load, not the mutable `read_only` flag — flipping `get_profile(...).read_only = False` no longer lets you delete/rename/save a builtin.
- **Migration not isolated/idempotent (Qodo #6 + review):** each legacy-blob entry now has its own try/except (one bad profile can't strand the rest → no silent loss on upgrade); the blob rename uses `os.replace` (overwrite-safe); and `_migrate_legacy_blob` only acts on a genuine `{"profiles":[...]}` blob, so a stray file is never destroyed.

**Data-loss holes found by adversarial review (all reproduced + fixed + regression-locked):**
- The filename-authority self-heal could overwrite a not-yet-loaded sibling file whose name slugified to the same id — fixed with a disk-occupancy-aware reassign + materialized/sorted glob (no writes during a live directory scan). Opus adversarially proved the clobber class is now closed by induction.
- A profile canonicalizing onto the reserved `custom_profiles.json` name could be mis-migrated away — fixed by excluding reserved names from id assignment.

**Style (Qodo #2/#3):** Google-style docstrings on the CRUD methods; test imports regrouped stdlib→third-party→local.

`Tests/RAG/` = **501 passed, 8 skipped, 0 failed** (32 in `test_config_profiles.py`, incl. regression locks for every hole above). SP2b (config-resolution unification) remains a separate follow-up.

PR #781, opened 2026-07-22
fix(settings): provider Test evaluates & honestly reports the draft config (TASK-432)

rmusser01:
## Summary

Fixes the provider-Test "trust bug" from the RP/character-card UX review: with an unsaved draft endpoint, the Test correctly *exercised* the draft (the live probe hit it) but the evidence line printed the **saved** value — the proof contradicted what was tested. This makes the **Test evaluate and honestly report the unsaved draft config end-to-end** (endpoint, API key, key env-var, provider, model), tagging each unsaved value.

**Discovery:** dev moved since the review — the review's AC#2 (clickable button) and AC#3 (non-hotkey path) were already delivered by task-189's `#settings-test-provider` button. The real remaining defect was AC#1, extended to a full draft-honesty pass.

## Changes (all in `settings_screen.py` + one new test file)

- **`overlay_provider_draft_config(...)`** — pure helper: deep-copies `app_config` and overlays only the *dirty* draft provider fields onto `api_settings[<provider>]`. Never mutates the input.
- **`_provider_test_staged_config(provider)`** — overlays only fields in `SettingsDraft.dirty_keys` (the app's own unsaved signal) and feeds the merged config to `get_provider_readiness` (unchanged), so `ready`/`api_key_source`/`env_var` reflect the draft. Returns the config unchanged when nothing is dirty.
- **`_build_provider_readiness_findings(...)`** — resolved-input evidence builder (no widget reads → unit-testable on a bare screen) with per-value `(draft)` tags: endpoint `(draft)`, `api_key_source=draft api_key (unsaved)`, env-var `(draft env var)`, model `(draft)`.

### Security
The draft API-key **value** is never placed in a finding — only a source label — and every finding stays `redact_secret_text`-wrapped. (Redaction was switched to per-finding, skipping only the one constant marker literal, because whole-string redaction mangled `api_key_source=draft api_key (unsaved)` — its regex reads `api_key…=value` as a secret. Verified: no secret leaks; the draft key value never even enters the findings list.)

### AC#2 / AC#3
No production change (already wired). Pilot tests pin the real `Button.press → handle_test_provider → action_settings_test_category(allow_text_entry_focus=True)` path, plus an **honest** AC#3 companion (`test_t_hotkey_does_not_run_test_while_input_focused`) proving the `t` hotkey no-ops while an input is focused — which is *why* the clickable button is needed.

## Testing

`Tests/UI/test_settings_provider_test_draft.py` — **13 tests**: 5 pure-overlay, 4 bare-instance findings (incl. draft-key-value-never-printed), 4 pilot (AC#2, AC#3-from-focus, AC#3-hotkey-no-op, draft-endpoint wiring end-to-end). All RED→GREEN; each pilot mutation-checked (fails when its code path is broken). No-draft path byte-identical to before (20 existing hub provider-test/readiness tests stay green).

SDD-executed (per-task implement + spec/quality review, whole-branch opus review = "ready to merge, no Critical/Important").

## Follow-up

Filed **TASK-483** (low, pre-existing): `redact_secret_text` redacts env-var *values* only when the var *name* matches `API_KEY|TOKEN|SECRET|PASSWORD`, so a custom-named credential env var prints raw — harden to `present`/`missing` or positional redaction. Out of scope for 432's ACs.

Closes TASK-432.

---
## Summary by cubic
Fixes the Provider Test “trust bug” so it evaluates the unsaved draft config and reports exactly what it tested, tagging unsaved values, and tightens secret handling so no raw secrets are echoed. Satisfies TASK-432 AC#1–#3.

- **Bug Fixes**
  - Readiness now runs against the draft by overlaying only dirty fields (endpoint, env var, API key) into a deep-copied `app_config` before `get_provider_readiness`.
  - Evidence shows the exact draft values used, tagged as unsaved: endpoint “(draft)”, model “(draft)”, env var “(draft env var)”, and `api_key_source=draft api key (unsaved)` when applicable. No draft → unchanged output.
  - Secrets are never printed: env-var findings show `<redacted>` when set; endpoint userinfo passwords are masked; API-key values are never shown; per-finding redaction is preserved.
  - Added pure helpers for testability and kept pilot coverage: the clickable `#settings-test-provider` runs the check, and the “t” hotkey no-ops while an input is focused.

<sup>Written for commit 0be00eb4b5449bf43ca0509e1a0705586dc3b33f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/781?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Rebased onto latest `dev` and addressed the Qodo review:

- **Security (secrets echoed in evidence):** the env-var finding now emits the established `<redacted>` marker for any set value instead of the raw `os.environ` value (`redact_secret_text` is assignment-name based and missed custom-named credential vars like `MY_CUSTOM_CRED`). Standard-name output (`OPENAI_API_KEY=<redacted>`) is byte-identical — existing hub tests still pass. Endpoint userinfo passwords are masked via a new `_mask_url_userinfo` helper (`scheme://user:pass@host` → `user:***@host`). This folds in the env-var value-redaction I'd initially deferred.
- **Rule (hardcoded key-like test literals):** renamed the test fixtures to clearly-fake placeholders (`fake-*-not-real`).

An internal adversarial review of the fix then caught a real robustness bug I'd introduced: `_mask_url_userinfo` accessed `urlsplit(...).port` (a lazy property that raises `ValueError` on a malformed/out-of-range port) outside the try/except, so a user-typed endpoint with basic auth + a typo'd port (e.g. `http://admin:pass@localhost:99999`) would have crashed the Test action. Rewrote it to reconstruct host:port from the raw `netloc` substring — which also fixes IPv6-with-userinfo bracket corruption — and added tests for the malformed-port, IPv6, and username-only cases. (Deliberately does not fall back to returning the raw string, which would re-leak the password.)

Residual **query-param** credential gap (custom-named `?mycred=...`) — the same name-based-redaction class, for query strings — filed as follow-up **TASK-483**.

Tests: `Tests/UI/test_settings_provider_test_draft.py` 15 pass; hub redaction/provider-test/readiness slice unaffected.

PR #788, opened 2026-07-23
fix(llm): llama.cpp endpoint default is the server root + legacy caller tolerates /completion (TASK-433)

rmusser01:
## Summary

Fixes the RP-review bug where a user who keeps the prefilled llama.cpp endpoint gets failures after a green-looking setup. The config default was `http://localhost:8080/completion` (llama.cpp's *native* endpoint), but the legacy chat caller appends `/v1/chat/completions` unless the URL already ends with it → `.../completion/v1/chat/completions`. The caller was also fragile for *any* partial path (the `/v1` placeholder form would double to `.../v1/v1/...`).

The native-Console send path already tolerated this via `console_provider_gateway.normalize_llamacpp_base_url`; the legacy `chat_with_llama` path did not.

## Changes

- **Server-root default (AC#1/AC#2):** `config.py` `[api_settings.llama_cpp].api_url` → `http://localhost:8080` + an honest comment (only that line; `llama_api_IP` and `[providers] llama_cpp` untouched). Settings placeholders (`llama_cpp`/`local_llamacpp`) aligned to `http://127.0.0.1:9099` (dropped `/v1`).
- **Tolerant legacy caller (AC#1):** `chat_with_llama` now normalizes its base URL via the **existing public** `normalize_llamacpp_base_url` (deferred local import, applied only to a non-empty value after the empty-guard) before the shared caller appends the path. So `/completion`, `/v1`, bare root, and the full path all resolve to `http://localhost:8080/v1/chat/completions`. Retroactively fixes users who saved `/completion` (no migration). Scoped to `chat_with_llama` (covers `llama_cpp`/`local_llamacpp`/`local_llamafile`); the shared caller and other providers are unchanged.

## Testing

- Config-default test (server root, no `/completion`).
- Normalizer-contract test (strip-to-root incl. scheme-less input; reverse-proxy prefix left unchanged).
- Parametrized `chat_with_llama` URL-capture (`/completion`, `/v1`, bare root → exactly `…/v1/chat/completions`) — **mutation-verified to fail without the fix**.
- `Tests/Chat/test_chat_functions.py` + `test_console_provider_gateway.py`: **120 passed**; placeholder tests 3 passed; deferred import doesn't break module load.

SDD-executed (per-task implement + spec/quality review; whole-branch composition review = "no defects, ready to merge").

## Follow-up

`normalize_llamacpp_base_url` is duplicated in `console_session_settings.py`; deduping is a noted follow-up (out of scope here).

Closes TASK-433.

---
## Summary by cubic
Fixes the llama.cpp endpoint default and makes the legacy caller tolerant, so users who keep the prefilled value (or saved `/completion`/`/v1`) get working chat. Satisfies TASK-433 AC#1 and AC#2.

- **Bug Fixes**
  - Default `[api_settings.llama_cpp].api_url` is now `http://localhost:8080` (server root); Settings placeholders for `llama_cpp`/`local_llamacpp` updated to `http://127.0.0.1:9099`.
  - Legacy `chat_with_llama` normalizes the base URL via `normalize_llamacpp_base_url`, avoiding `.../completion/v1/...` and `.../v1/v1/...`; supports `/completion`, `/v1`, root, or full path; reverse-proxy prefixes are preserved.
  - Tests added/updated for config default, normalizer contract, URL capture, and UI placeholder. No migration needed; existing saved values continue to work.

<sup>Written for commit d5547f9077cc3b2ef61a5c20779af83d026d0804. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/788?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Rebased onto latest `dev` (clean, no conflicts; key tests re-verified green). Bot review: Gemini is sunset, CodeRabbit skips non-default-branch PRs, and Qodo posted **no actionable findings** — it endorses the approach ("set the default and UI hints to server root, and add llama-specific normalization inside `chat_with_llama` using the already-public console normalizer via deferred import … fixes new and existing users including saved `/completion` or `/v1` while keeping the shared caller and other providers unchanged"). Nothing to address.

The SDD whole-branch review returned no defects (`chat_with_llama` empty-guard ordering, scope containment with zero diff to the gateway/session-settings, config correctness, placeholder consistency all confirmed; URL-capture test mutation-verified). The two `test_console_native_chat_flow` continue/regenerate failures are pre-existing on `dev` (verified identical at the base commit), unrelated to this change.

PR #789, opened 2026-07-23
Token budget for the agent RunBudget (TASK-326)

rmusser01:
Adds a **token-spend ceiling** to the agent RunBudget (**TASK-326**) — the documented complement to task-322. **322 bounds conversation history *into* the agent loop at dispatch; 326 bounds the loop *itself*** — the tool-call/result messages that accumulate across iterations within a single turn, which previously had no spend ceiling.

## What changed (3 units, mirroring the existing step/model-turn/wall-clock caps)

- **`agent_models.py`** — `RunBudget.max_total_tokens: int = 0` (**0 = unlimited sentinel** → default runs byte-identical); `ModelTurn.tokens` (prompt+completion for one provider call); `RunOutcome.total_tokens` (measured run spend, for accounting); `clamp_child_budget` passes the field through (each run has its own budget).
- **`agent_runtime.py` `run_agent_loop`** — accumulates `total_tokens += turn.tokens` per model turn and adds a top-of-loop check (after wall-clock) → `add(STEP_ERROR, "token budget exhausted")` + `RUN_STUCK` (a clear terminal step, not silent truncation). An `_outcome()` helper closure routes all 8 terminal returns so spend is always reported. Because the check is at the *top*, a run that produces its final answer on the crossing turn still completes `RUN_DONE` (stop-the-loop, not fail-the-answer). The `and` guard is deliberately fail-closed.
- **`agent_service.py` `_make_call_model`** — **hybrid** token source: real `resp["usage"]` (via `_usage_total_tokens`, crash-proof on malformed input) when the provider reports it, else the task-321 estimate (`count_tokens_messages(payload) + estimate_tokens(text)`). Set on both `ModelTurn` return paths.

**Semantics:** `max_total_tokens` is cumulative *spend* (Σ prompt+completion across turns, since the growing prompt is re-sent each call), not a context-window size.

## Testing
Full agent suite **275 passed**. Loop tests (exhaustion→STUCK, 0-sentinel never trips, DONE-on-crossing-turn boundary, `total_tokens` accounting); `_usage_total_tokens` (total / prompt+completion / None-on-malformed, never crashes); call_model hybrid on **both** native and non-native return paths. Executed via subagent-driven development (implementer + reviewer per task, whole-branch opus review **Ready-to-merge**, 0 Critical/Important).

## Deferred follow-ups (non-blocking, from reviews)
- `_usage_total_tokens` doesn't reject `bool`/negative token counts (no real provider emits them).
- Error-path `RunOutcome` (`RUN_ERROR`) reports `total_tokens=0` (partial pre-crash spend lost).
- `total_tokens` is return-only, not persisted to AgentRuns_DB (surfacing in a consumer UI/DB was out of scope).

---
## Summary by cubic
Adds a token-spend ceiling to the agent loop via `RunBudget.max_total_tokens`, accumulating per-turn tokens and stopping cleanly when the budget is exceeded. Defaults to unlimited (`0`), so behavior is unchanged unless configured; aligns with TASK-326 and complements task-322.

- **New Features**
  - Data model: `RunBudget.max_total_tokens` (0 = unlimited), `ModelTurn.tokens`, `RunOutcome.total_tokens`; `clamp_child_budget` preserves the cap.
  - Runtime: accumulates `total_tokens` and, after the wall-clock check, stops with `RUN_STUCK` + "token budget exhausted" when over budget; a final answer on the crossing turn still returns `RUN_DONE`.
  - Service: `_usage_total_tokens` reads provider `resp["usage"]`; else falls back to `count_tokens_messages(payload)` + `estimate_tokens(text)`; `tokens` set on both native and non-native `ModelTurn` paths.

- **Bug Fixes**
  - When provider usage is absent, normalize model ids by stripping provider prefixes (e.g., `openai/gpt-4o-mini` → `gpt-4o-mini`) and pass the provider to the token counter to avoid undercount.
  - `_usage_total_tokens` now rejects bool/negative/zero totals and invalid prompt/completion values, falling back to the estimator to keep spend accounting correct.

<sup>Written for commit 0e20dbf6a6bc515ed07f640d2411a5eba23715cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/789?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/plans/2026-07-23-agent-token-budget.md
+++ b/Docs/superpowers/plans/2026-07-23-agent-token-budget.md
@@ -1,0 +1,474 @@
+# Agent RunBudget Token Budget Implementation Plan (TASK-326)
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Give the agent runtime a token-spend ceiling — `RunBudget.max_total_tokens` — that accumulates prompt+completion tokens per run and stops the loop cleanly (like the existing caps) when exceeded, with real provider usage when reported and a token_counter estimate otherwise.
+
+**Architecture:** Three files. `agent_models.py` gains the data fields (`ModelTurn.tokens`, `RunBudget.max_total_tokens`, `RunOutcome.total_tokens`). `agent_service.py` populates `ModelTurn.tokens` from `resp["usage"]` or the task-321 estimator. `agent_runtime.py`'s pure loop accumulates the spend and adds a top-of-loop budget check mirroring the step/model-turn/wall-clock caps, reporting spend via an `_outcome` helper closure.
+
+**Tech Stack:** Python, pytest. The runtime calls providers with `streaming=False`, so usage is on the response dict.
+
+## Global Constraints
+
+- **Worktree:** all work in `/Users/macbook-dev/Documents/GitHub/tldw_chatbook-agent-token-budget` (branch `feat/agent-token-budget`, off `origin/dev @ c10b5eb9e`). Never touch the main checkout `/Users/macbook-dev/Documents/GitHub/tldw_chatbook`.
+- **Test command (venv is in the main checkout):**
+  `cd /Users/macbook-dev/Documents/GitHub/tldw_chatbook-agent-token-budget && /Users/macbook-dev/Documents/GitHub/tldw_chatbook/.venv/bin/python -m pytest <path> -v`
+- **Sentinel default `max_total_tokens = 0` = unlimited** — default runs must be byte-identical (the check is guarded `if budget.max_total_tokens and …`).
+- **`max_total_tokens` is cumulative SPEND** (`Σ(prompt_i + completion_i)` across turns, since the growing prompt is re-sent each call), not a context-window size.
+- New fields `ModelTurn.tokens` and `RunOutcome.total_tokens` are **trailing and defaulted (0)** — backward compatible with every existing construction.
+- **Terminal on exhaustion:** `add(STEP_ERROR, summary="token budget exhausted")` then return `RUN_STUCK` — a clear step, not silent truncation. Placed **after** the existing wall-clock check.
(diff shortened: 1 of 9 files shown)


rmusser01:
Thanks — fixed in `9aabb8d`.

**Prefixed models undercount tokens — fixed.** In the estimate fallback (`_make_call_model`, taken only when the provider reports no usage), a provider-qualified id like `openai/gpt-4o-mini` is now normalized by stripping the prefix (`-> gpt-4o-mini`) so `count_tokens_messages`'s GPT-family framing detection matches and the correct per-message overhead is applied; the endpoint is also passed as the `provider` hint so the chars-fallback ratio is right. This mirrors the OpenRouter `upstream/model` handling already in `get_model_token_limit` (task-320). Added a discriminating regression test asserting the `openai/gpt-4o-mini` estimate equals the bare `gpt-4o-mini` estimate (it was lower before the fix).

Note the trigger is narrow (a provider-prefixed model AND a provider that reports no `usage` — OpenRouter, the common prefixed case, does report usage, taking the exact-count path), but the fix makes the fallback consistent and correct. Full agent service suite: 38 passed.

rmusser01:
Good catch — fixed in `0e20dbf6`.

**Malformed usage tokens accepted — fixed.** `_usage_total_tokens` now uses `type(x) is int` (which rejects `bool`, an int subclass) and requires a **positive** `total_tokens`, or a `prompt_tokens`+`completion_tokens` pair that are both non-negative real ints summing to `> 0`; anything else returns `None` and falls to the estimate. This prevents a malformed `usage` block (`True`/`False`, negative, or zero) from corrupting or shrinking the accumulated `ModelTurn.tokens` the runtime enforces `max_total_tokens` on. Extended the malformed-usage test to cover bool / zero / negative / mixed cases (and that a valid `0 + 5` still yields `5`). 38 passed.

PR #793, opened 2026-07-23
fix(personas): workbench preserves selection + preview across navigation (TASK-434)

rmusser01:
## Summary

Fixes the RP-review bug where the Personas workbench resets to "Selected: none" (blank center, "Console blocked", collapsed preview) after every Personas → Console → back round-trip — dropping the working context the workbench↔Console loop encourages.

**Root cause:** the app never caches screens; it stores `screen.save_state()` on navigate-away and calls `restore_state()` on a *fresh* `PersonasScreen` on return. `PersonasScreen` overrode neither, so the workbench `self.state` (a `PersonasWorkbenchState` dataclass) and the preview were lost.

## Changes

- **Preview primitives:** `PersonasPreviewPane.greeting_text` (the greeting lives in the pane, **not** in `controller.history`) + `PersonasPreviewController.restore_conversation(greeting, history, seeded_for)`, which sets `seeded_for` **before its first `await`** so the still-pending character-load worker's `handle_character_loaded` guard *preserves* the restored turns instead of erasing them.
- **Screen persistence:** `PersonasScreen.save_state`/`restore_state` capture the workbench dataclass + preview `{greeting, history, seeded_for}`; a deferred `_apply_pending_restore` at the end of `on_mount` re-applies the selection via the existing `_select_*` flow (guarded so a stale/deleted entity degrades to a blank center); `_select_character` gains a `restore_preview` branch (mutually exclusive with `reset_for_character`, whose `invalidate()` would wipe the restored turns).

So a character round-trip restores the selected item, mode, center pane **and** the preview conversation (greeting + turns) — AC#1 + AC#2.

## Scope

Restoration is **gated to `characters` mode** (the task's driver and the review scenario). A whole-branch review found that restoring non-character modes (dictionaries/lore/personas) naively would leave an empty list + mis-toggled Preview/Try-It widgets, because their row-render + widget-visibility live only in `_apply_mode` (which restore doesn't call). Non-character modes therefore fall back to today's clean default view — no regression, no broken view. **Full non-character-mode restore is filed as TASK-488.**

## Testing

- Preview primitives: `greeting_text` + `restore_conversation` (seeded-for-first ordering **mutation-verified**).
- Full-screen round-trip: AC#1 selection/center restore, AC#2 greeting+turns survive, the async-worker guard (drives `handle_character_loaded` directly), stale-entity degrade, and the non-character-mode gate — all RED→GREEN, mutation-verified non-vacuous.
- 320 Personas regression tests pass.

SDD-executed (per-task implement + spec/quality review; whole-branch opus review → one Medium finding fixed + follow-up filed).

Closes TASK-434.

---
## Summary by cubic
Fixes the Personas workbench losing its selection and preview when navigating to Console and back by persisting the workbench state and preview transcript. Implements TASK-434 for Characters mode so your context survives within a session.

- **Bug Fixes**
  - `PersonasScreen.save_state`/`restore_state` now snapshot `PersonasWorkbenchState` and preview `{greeting, history, seeded_for}`; the selection is re-applied after mount.
  - Added `PersonasPreviewPane.greeting_text` and `PersonasPreviewController.restore_conversation()` to rebuild greeting + turns; `seeded_for` is set before the first await to prevent late character-load from erasing turns.
  - `_select_character` accepts a `restore_preview` path to rebuild the transcript instead of resetting it.
  - On failed restore (stale/deleted entity), the selection is fully cleared and the center is blanked to avoid stale console actions.
  - Restore is gated to Characters mode; other modes return to the default view (follow-up in TASK-488).
  - Tests cover preview restore, round-trip, late worker race, stale-entity handling, and the non-character-mode gate.

<sup>Written for commit 25355ac18da4648754ede6c297620906776fa4bf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/793?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/plans/2026-07-23-personas-workbench-selection-persistence.md
+++ b/Docs/superpowers/plans/2026-07-23-personas-workbench-selection-persistence.md
@@ -1,0 +1,380 @@
+# TASK-434 Personas workbench selection+preview persistence — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Returning to the Personas screen within a session restores the previously selected item, mode, and center view (AC#1), and the preview conversation (greeting + turns) survives the round-trip (AC#2).
+
+**Architecture:** The app stores `screen.save_state()` on navigate-away and calls `restore_state()` on a fresh screen when returning. `PersonasScreen` will override both to persist its `PersonasWorkbenchState` dataclass and the preview conversation, and re-apply the selection at the end of `on_mount`.
+
+**Tech Stack:** Python 3.11+, Textual, pytest (+ pytest-asyncio).
+
+## Global Constraints
+
+- The preview is split: `PersonasPreviewController.history` holds only sent turns; the **greeting** is in `PersonasPreviewPane._greeting` (NOT in history). Capture and restore **both**.
+- In `restore_conversation`, set `self.seeded_for` **before the first `await`** — `invalidate()` does NOT cancel the character-load worker, whose `handle_character_loaded` must hit the `:133` guard (`refresh_greeting_seed`, no erase).
+- `restore_state` runs **before** mount; rebuild `PersonasWorkbenchState` filtered to known dataclass fields (version-drift tolerant), and only stash a pending restore when there is a real selection.
+- Non-goals: restoring an unsaved editor, cross-restart persistence, changing the screen-caching model.
+- Files: `tldw_chatbook/Widgets/Persona_Widgets/personas_preview_pane.py`, `tldw_chatbook/UI/Persona_Modules/personas_preview_controller.py`, `tldw_chatbook/UI/Screens/personas_screen.py`, and tests.
+
+---
+
+### Task 1: Preview capture + restore primitives
+
+**Files:**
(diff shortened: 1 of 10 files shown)


rmusser01:
Rebased onto latest `dev` and addressed the Qodo review (Qodo endorsed the approach, incl. gating restore to Characters mode + the TASK-488 follow-up for other modes):

- **Bug (stale selection after restore fail):** the `_apply_pending_restore` except block now performs a full deselect degrade — `self.state.clear_selection()` + `await PersonasInspectorPane.clear_selection()` + `_show_center(None)` + `_sync_title_and_console_actions()` — mirroring the file's existing deselect pattern. So a stale/deleted-entity restore consistently reports **no** selection (Console attach/Start-Chat correctly disabled, inspector cleared), not just a blank center over a lingering selection. Added a test asserting `state.selected_entity_id is None` after a stale restore (RED-verified: fails when the except only hides the center).
- **Docstring:** Google-style `Args:`/`Returns:` on `restore_conversation`.
- **Type hint:** `seeded_for: str | None` annotated.
- **Import ordering:** `import dataclasses` was already alphabetically placed in the stdlib group — no change.

Tests: 29 pass (state + preview + preview-restore). The SDD whole-branch review had already found + fixed the non-character-mode gap (gated to Characters; TASK-488 filed).

PR #795, opened 2026-07-23
RAG SP2b: config-resolution unification onto the active profile

rmusser01:
## SP2b — RAG profile system: config-resolution unification

The **correctness-critical half of SP2** — makes the active RAG profile the *single* config source the engine reads, so ingestion and search can never diverge. Depends on SP1 (index isolation, #771) and SP2a (profile storage, #780), both merged. Backlog `task-487`. SP3 (settings screen UI) is next and last.

### The core problem it fixes
Two independent resolution paths existed: the **search** path (`RAGConfig.from_settings`) applied a full env overlay on top of scattered `[AppRAGSearchConfig.rag.*]` values; the **ingestion** path used `profile.rag_config` with **no env**. So with any `RAG_*` env var set — or once a profile was actually selected — the config used to *embed* and the config used to *query* diverged (a latent dimension-mismatch / silent-wrong-results bug).

### What landed (new `simplified/active_config.py`)
- **`resolve_active_rag_config()`** — the single resolver: the active profile's `rag_config` (deep copy) + the env-override layer. `RAGConfig.from_settings` delegates to it; the shared service (`get_shared_rag_service`) builds from it; `fusion.resolve_hybrid_alpha` reads from it. The old ~350-line `from_settings` re-deriver is deleted; the scattered `AppRAGSearchConfig.rag.*` *value* keys are deprecated (only the `rag.indexing` toggle and the vector-store deployment knob remain).
- **Parity** — the shared service routes `profile_name ∈ {None, active pointer}` through the resolver (the real consumer forwards the active pointer *as* `profile_name`, so the naive "is None" check would have missed it). Locked by a wiring test + a fingerprint-equality test.
- **`set_active_profile()`** — writes the single pointer `[rag.service].profile` (correct nested-section form, round-trip-verified against a real temp config) and resets the shared service.
- **First-run import** — captures the resolved active config into an "Imported settings" active profile, wired into `get_shared_rag_service` **before** the lock (its `reset` re-enters that non-reentrant lock → in-lock wiring would self-deadlock; proven by a lock-acquisition test) and guarded by a once-flag. Its SP1 fingerprint equals SP1's adopted legacy-collection fingerprint, so upgraders keep their index.

### Notable review catches (all fixed, each reproduced)
- The parity heuristic missed the live chat consumer (forwards the active pointer as `profile_name`) → routed both through the resolver.
- The cross-SP fingerprint-invariant test was **tautological** (compared the config to a re-resolution of itself) → rewritten to capture `pre_fp` before the import mutates the pointer, and extended with a **fingerprint-affecting-env-var** case (RED-proven).
- First-run import had a partial-failure non-idempotency (created-but-never-activated) → heals on next call; and its no-deadlock test was writing to the real `~/.local/…` → isolated.
- Spec §6 said to snapshot the *legacy* config; the implementation correctly snapshots the *active profile* (pre-SP2b ingestion built the collection from the built-in, so SP1 adopted under the built-in fingerprint — following §6 literally would have *broken* the invariant for default users). Reconciled the spec + softened the profile description + filed `task-495` for query-knob capture.

### Testing
`Tests/RAG/` — **519 passed, 8 skipped, 0 failed**. New: `test_active_config_resolution.py`, `test_config_unification_parity.py`, `test_first_run_import.py`.

### Out of scope (SP3)
The settings-screen UI + off-thread reload orchestration. NOTE: between this merge and SP3, the *old* narrow Library-RAG settings controls still write `[AppRAGSearchConfig.rag.*]` keys the engine now ignores — those edits no-op on the engine until SP3 rebuilds that screen.

### Process
SDD-executed: fresh subagent + spec/quality review per task, plus a final opus whole-branch review; every finding reproduced and fixed or filed.

---
## Summary by cubic
Unifies RAG config resolution onto the active profile so ingestion and search always use the same settings. Adds a single resolver and a first‑run import that auto‑captures current settings (including env overrides) to keep existing indexes—no manual steps.

- **New Features**
  - Added `simplified/active_config.py` with `resolve_active_rag_config()`; `RAGConfig.from_settings`, `get_shared_rag_service` (including when the active pointer is passed as `profile_name`), and `fusion.resolve_hybrid_alpha` now read from it.
  - Added `set_active_profile()` to write `[rag.service].profile` and reset the shared service.
  - First run: creates and activates an “Imported settings” profile from the resolved active config; its fingerprint matches SP1’s adopted legacy-collection fingerprint.
  - Deprecated scattered `[AppRAGSearchConfig.rag.*]` value reads (only the `rag.indexing` toggle remains). Docs updated in `README_enhanced_services.md`.
  - Satisfies Linear task‑487; follow‑up task‑495 tracks merging non‑fingerprint query‑time legacy keys.

- **Bug Fixes**
  - Env overrides now include `RAG_VECTOR_STORE` when resolving from a deep‑copied profile; normalized like `default_vector_store_type()`.
  - Cached default `get_profile_manager()` as a singleton for hot paths; added `reset_profile_manager_cache()` for tests.
  - `set_active_profile()` validates IDs, checks `save_setting_to_cli_config` success, and only resets the shared service on a successful pointer write.
  - First‑run import is once‑per‑process and runs before the shared‑service lock to avoid deadlocks; heals partial runs and skips under pytest.

<sup>Written for commit 95d6cf34e7c6bbceada72ad3fedf75ca9d0464b1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/795?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Addressed all 7 Qodo findings (pushed; rebased on latest dev):

- **#4 RAG_VECTOR_STORE override lost (Bug/correctness):** fixed — you're right that `copy.deepcopy` doesn't rerun `VectorStoreConfig.__post_init__`, so the env was silently dropped. `_apply_env_overrides` now applies `RAG_VECTOR_STORE` explicitly via the same `_normalized_type_setting` helper `default_vector_store_type()` uses; regression test covers both directions (chroma-profile + env=memory → memory, and vice-versa).
- **#5 hybrid-alpha resolver too heavy (Bug/perf):** fixed at the root — `get_profile_manager()` now returns a cached singleton for the default dir (its docstring already promised 'get or create the global'), so the query hot path no longer re-instantiates `ConfigProfileManager` (mkdir + 13 builtins + disk glob) per call; explicit-`profiles_dir` calls stay fresh, and `reset_profile_manager_cache()` + an autouse fixture keep tests isolated.
- **#6 pointer write result ignored (Bug/reliability):** `set_active_profile` now checks `save_setting_to_cli_config`'s bool and skips the service reset + logs a warning on a failed write.
- **#7 first-run import race (Bug/reliability):** the once-flag check-and-set is now guarded by a dedicated `_first_run_lock` (kept separate from `_shared_service_lock` to preserve the before-the-lock ordering that avoids the reset-deadlock).
- **#3 input validation:** `set_active_profile` validates `profile_id` (rejects empty / traversal / non-slug) before any write.
- **#2 torch import:** kept the guarded `try/except ImportError` — `optional_deps.py` has no torch accessor to route through (and it mirrors the original `from_settings`).
- **#4b docstrings:** Google-style Args/Returns/Raises on the three new public functions.

`Tests/RAG/` 535 passed / 8 skipped; independently RED-proofed the #4 and #5 fixes.

PR #798, opened 2026-07-23
Anthropic prompt caching + cache-hit usage metrics (TASK-323/324)

rmusser01:
Enables Anthropic prompt caching and logs cache-hit usage metrics (**TASK-323 + 324**) — a "pure missed win" from the LLM-harness audit: the request prefix is already byte-stable, so `cache_control` breakpoints earn cost/latency savings on ordinary chats today.

## What changed (both in `LLM_API_Calls.py`)

**TASK-323 — Anthropic `cache_control` breakpoints**
- `_anthropic_supports_caching(model)` gate — modern Claude (`claude-3/3.5/3.7/4+`); excludes legacy `claude-2*`/`instant`.
- System prompt → structured content block with one `cache_control: {"type":"ephemeral"}` breakpoint (the largest stable prefix; caches tools+system per Anthropic's hierarchy).
- Optional second breakpoint on the **last converted** tool (a fresh dict — never mutates the caller's input), gated on the converted list being non-empty. ≤4 breakpoints (real max = 2).
- Non-caching models keep the plain-string system + untouched tools; non-Anthropic providers unaffected. Applies to streaming + non-streaming (payload built before the streaming branch).

**TASK-324 — cache-hit usage metrics**
- Appended `openai_api_cached_tokens` (from `prompt_tokens_details.cached_tokens`) to `chat_with_openai`'s existing usage block, and `anthropic_api_cache_read_input_tokens` / `anthropic_api_cache_creation_input_tokens` to `chat_with_anthropic`'s existing block — **no duplication** of the pre-existing input/output/total histograms. All reads graceful on absent/`null` fields.

## Testing
**51 passed** across the affected suites. New tests: gate; caching-model system/tools carry `cache_control`; non-caching model unchanged; ≤4 breakpoints; OpenAI + Anthropic cache metrics present→exact / absent→0. Executed via subagent-driven development (implementer + reviewer per task, whole-branch opus review **Ready-to-merge**, 0 Critical/Important).

**AC#4 (cache read on 2nd request)** is verified structurally (payload carries `cache_control`; the metric reads the response's `cache_read_input_tokens`); the literal cross-call read is a documented live check (two identical requests with a >2048-token system prompt).

## Known baseline (not from this PR)
`Tests/Chat/test_anthropic_native_tools.py::test_anthropic_shaped_tools_pass_through_untouched` fails identically on `origin/dev` — a separate `_anthropic_tools_payload` anthropic-shape drop bug (task-263 territory), deliberately out of scope here.

## Deferred follow-ups (non-blocking)
- Streaming-path cache-usage metrics (final SSE event).
- Test-helper de-dup; explicit `prompt_tokens_details: None` case (no code-path gap).

---
## Summary by cubic
Adds Anthropic prompt caching via `cache_control` and logs cache-hit usage for OpenAI and Anthropic to cut latency/cost and improve visibility (TASK-323, TASK-324). Also coerces null usage fields to zero to keep metrics clean.

- **New Features**
  - Anthropic caching: gate via `_anthropic_supports_caching(model)`; send system as a structured block with `cache_control: {"type":"ephemeral"}` (when non-empty) and add `cache_control` to the last converted tool; non-caching models unchanged; works for streaming and non-streaming.
  - Metrics: on non-streaming, log `openai_api_cached_tokens` from `usage.prompt_tokens_details.cached_tokens`, and `anthropic_api_cache_read_input_tokens` + `anthropic_api_cache_creation_input_tokens`; appended to existing token histograms; missing fields log 0.

- **Bug Fixes**
  - Present-but-null usage fields are coerced to 0 for all three cache metrics to avoid non-numeric histogram values; tests cover explicit `null` cases.

<sup>Written for commit c8b19a2c0a858cc20671604d6304284a3384c0e6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/798?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/plans/2026-07-23-anthropic-prompt-caching.md
+++ b/Docs/superpowers/plans/2026-07-23-anthropic-prompt-caching.md
@@ -1,0 +1,376 @@
+# Anthropic Prompt Caching + Cache Metrics Implementation Plan (TASK-323/324)
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Activate Anthropic prompt caching by emitting `cache_control` breakpoints on the stable system/tools prefix (gated to cache-capable Claude), and log the provider-reported cache-hit usage fields for OpenAI and Anthropic.
+
+**Architecture:** Two changes in `tldw_chatbook/LLM_Calls/LLM_API_Calls.py`. Task 1 adds a `_anthropic_supports_caching` gate and puts `cache_control` on the system content block + last converted tool in `chat_with_anthropic`'s payload. Task 2 appends cache-token `log_histogram` metrics to the existing OpenAI and Anthropic non-streaming usage blocks.
+
+**Tech Stack:** Python, pytest. Tests mock `requests.Session.post` and inspect the sent JSON / spy `log_histogram`.
+
+## Global Constraints
+
+- **Worktree:** all work in `/Users/macbook-dev/Documents/GitHub/tldw_chatbook-anthropic-caching` (branch `feat/anthropic-prompt-caching`, off `origin/dev @ e293b3313`). Never touch the main checkout `/Users/macbook-dev/Documents/GitHub/tldw_chatbook`.
+- **Test command (venv in main checkout):** `cd /Users/macbook-dev/Documents/GitHub/tldw_chatbook-anthropic-caching && /Users/macbook-dev/Documents/GitHub/tldw_chatbook/.venv/bin/python -m pytest <path> -v`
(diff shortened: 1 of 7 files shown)


rmusser01:
Good catch — fixed in `c8b19a2c`.

**Anthropic/OpenAI cache-token nulls — fixed.** `.get(key, 0)` returns `None` for a present-but-`null` field, so `log_histogram` could record a non-numeric value. All three new cache metrics now null-coalesce with `or 0`: `openai_api_cached_tokens` = `(usage.get("prompt_tokens_details") or {}).get("cached_tokens") or 0`, and the two Anthropic metrics = `usage.get("cache_*_input_tokens") or 0`. This covers absent, present-but-null, and 0 → always a numeric 0. Added regression tests with explicit `null` usage fields for both providers.

PR #801, opened 2026-07-23
Skills: Codex-style $-mention invocation (breaking: /skill-name removed)

rmusser01:
## Summary

**Codex-style `$`-mention skill invocation** — migrates Console user skill invocation to the ecosystem convention ([Codex](https://developers.openai.com/codex/skills): `/skills` browses, `$skill-name` invokes — anywhere in the prompt).

Breaking by design (user-approved full-Codex model): `/skill-name` invocation is **hard-removed**.

Design + plan: `Docs/superpowers/specs/2026-07-23-skills-dollar-mention-invocation-design.md`, `Docs/superpowers/plans/2026-07-23-skills-dollar-mention-invocation-plan.md`.

## What this does

- **`$skill-name` is THE invocation**, in two position-dependent forms:
  - **Leading** (`$pdf-processing extract the tables`): the arg-bearing whole-message form — trailing text → `{{args}}`, inline replaces the message, fork takes over the turn, untrusted refuses. Today's semantics, re-sigiled.
  - **Embedded** (`summarize the draft, $style-guide it, then list questions`): **argless** expansion — the rendered inline body is spliced at the mention's position, all surrounding prose preserved. Multiple mentions each expand (one `execute_skill` per unique name).
- **Safety rails on embedded mentions:** exact case-**sensitive** match against canonical lowercase skill names (`$PATH`/`$HOME`/`$5` stay literal, even if a same-named skill exists); markdown code spans skipped (fences + inline backticks; an odd backtick count fail-safes by masking the whole line); no recursion; fork skills can't be embedded (left literal); an untrusted embedded mention stays literal with a non-aborting system note (your prose is never lost to a refusal).
- **Fast path:** a message without `$` never touches the skills service.
- **Hard removal of `/skill-name`** at both layers (composer fallback resolver deleted; controller branch re-sigiled). A `/former-skill` draft gets the standard unknown-command hint, then an armed second Enter sends it literally. `/skills` stays as the browse list (rows teach `$name`); its run form now responds `Run skills by typing $<name> — /skills only lists them.`; the picker submits `$name`.
- **Dead code swept with the migration:** the orphaned skill-run marker machinery and a write-only candidates snapshot (plus its two mount/resume refresh workers) are deleted; the draft preview annotator flags leading `$` (and no longer mislabels `/` drafts).
- Ephemeral-only: transcripts keep your raw text; the skills → dictionaries → world-info payload ordering is unchanged; model-invoked skills (`console_agent_bridge`) are untouched.

## Verification

- Built subagent-driven: 5 TDD tasks, each spec+quality reviewed. The review gate caught real bugs the plan had missed: a code-span masking hole (a stray backtick could un-mask a guarded `` `$name` `` and splice into it — now fail-safe), an unconditional skills-service fetch on every plain-text send (now short-circuited), and a coverage-vs-dead-code inconsistency (resolved by deleting the genuinely orphaned machinery, premise verified against git history).
- Final **whole-branch review returned MERGE READY** (no Critical/Important). Bounded-execution verified: 1000 repeated `$name` mentions → 1 `execute_skill` call.
- Full gate: `Tests/Chat/` + skill-command + composer + `Tests/Skills/` suites → **1520 passed**; only pre-existing baselines (`test_anthropic_native_tools`, one known flaky) remain.
- Rebased onto latest `dev` (+69 commits; `chat_screen.py` overlap verified free of semantic collisions with the deletions).

## Breaking change / release note

Skills are now invoked with **`$skill-name`** (leading for args, embedded anywhere for argless expansion) — matching Codex, opencode, and the emerging Agent Skills ecosystem convention. `/skill-name` no longer invokes; `/skills` still lists (and now teaches the `$` form). Stored transcripts are unaffected; historical raw `/name` turns retried send as literal text.

## Out of scope (follow-ups)

Interactive `$`-autocomplete popup (none existed for `/` either); embedded args (`$name{...}`); fork reference-file reachability (next layer of the skills program); retained-dead `_console_command_run_skill`/picker-modal cleanup; the indented-leading-mention edge (an indented `$name args` is treated as embedded/argless — pre-existing whitespace-sensitivity, faithfully re-sigiled).

---
## Summary by cubic
Switches Console skill invocation to Codex-style $-mentions with embedded expansion. Removes /skill-name; hardens parsing (CommonMark backticks, indented leading “$”, blocked-skill notes) and keeps a fast path when “$” is absent.

- **New Features**
  - $skill-name is the only invocation:
    - Leading: “$name args” runs the skill with args (inline or fork), unchanged semantics.
    - Embedded: “... $name ...” splices the argless rendered body in place; multiple mentions splice; one execute per unique skill.
  - Safety: exact case-sensitive matching (lowercase names); code spans are skipped with CommonMark-style backtick pairing (odd or unclosed spans mask safely); leading whitespace before “$name args” is tolerated as leading form; no recursion; fork skills cannot be embedded; blocked or untrusted mentions are detected and left literal with a non-blocking system note (leading untrusted still refuses).
  - Performance: messages without “$” bypass the skills service.

- **Migration**
  - /skill-name invocation removed. /skills still lists skills and now tells users to run them by typing $name; the picker submits “$name [args]”.
  - Deleted dead code: skill-run marker machinery and a write-only candidates snapshot. Preview annotator now flags leading $ only and skips multimodal drafts.
  - Transcripts keep raw text; payload ordering unchanged; model-invoked skills paths unaffected.

<sup>Written for commit 96149fd381b030c16320ac59edc4216c9043775b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/801?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/plans/2026-07-23-skills-dollar-mention-invocation-plan.md
+++ b/Docs/superpowers/plans/2026-07-23-skills-dollar-mention-invocation-plan.md
@@ -1,0 +1,527 @@
+# Codex-Style `$`-Mention Skill Invocation Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Migrate Console user skill invocation to the Codex convention — `$skill-name` becomes THE invocation (leading form keeps `{{args}}`; embedded mentions splice argless at position), and `/skill-name` invocation is hard-removed.
+
+**Architecture:** A pure mention scanner (code-span-aware, exact case-sensitive matching) lives in `console_skill_resolver`; `_apply_skill_substitution` in the controller gains a leading-`$` branch (today's semantics re-sigiled) plus an embedded splice pass with a new non-aborting notes channel; the composer's `/`-fallback claiming and the `/skills` run form are deleted.
+
+**Tech Stack:** Python 3.11+, pytest. No new dependencies.
+
+**Design doc:** `Docs/superpowers/specs/2026-07-23-skills-dollar-mention-invocation-design.md` (read it if a requirement here seems ambiguous — it governs).
+
+## Global Constraints
+
+- **Sigil rules (verbatim from spec):** leading `$skill-name args…` = arg-bearing whole-message form (args → `{{args}}`, inline-replace / fork-takeover, untrusted → refuse); embedded `$skill-name` anywhere else = **argless** splice at the mention's position preserving ALL surrounding prose; multiple embedded mentions each expand.
+- **Embedded matching is exact + case-SENSITIVE** against canonical (lowercase) skill names: `$PATH`/`$HOME`/`$5` stay literal. Leading form keeps `resolve_skill_command`'s case-insensitive/unique-prefix behavior.
+- **Code spans skipped:** mentions inside ``` fenced blocks or inline `backtick` spans stay literal.
+- **No recursion:** single pass over the user's original text; a `$mention` inside a spliced body is inert.
+- **Fork cannot be embedded:** embedded mention whose `execute_skill` result has `execution_mode != "inline"` stays literal (no note). Embedded untrusted (SkillTrustBlockedError) stays literal + a system note. A leading resolved mention does NOT get its args embedded-scanned.
+- **Hard remove `/skill-name` at BOTH layers:** the composer fallback resolver (chat_screen registration + `make_skill_fallback_resolver` itself) and the controller's leading-`/` branch. `/skills` bare list stays; its `<name> [args]` run form is removed.
+- **Payload ordering preserved:** skills → dictionaries → world-info at all 4 send sites; the deliberate skip-skills site (search "may execute skills with side effects") stays skill-free.
+- **Ephemeral only:** splicing happens on the provider payload; transcripts keep raw text.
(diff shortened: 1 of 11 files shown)


rmusser01:
Addressed the Qodo review in `96149fd38`:

**Fixed**
- **Blocked skills never noted** — the detection population now includes user-invocable `blocked_skills`: a leading `$blocked-name` refuses (via the existing execute-time `SkillTrustBlockedError` path, preserving the render-fresh discipline) and an embedded one stays literal **with** the skipped-skill note; non-user-invocable skills remain undetectable either way. New tests for all three cases.
- **Whitespace breaks leading args** — leading detection now runs on `content.lstrip()`, matching the preview annotator; an indented `  $code-review fix it` is the args-bearing leading form again. Pinned by test.
- **Multi-backtick code bypass** — `_code_span_mask` now pairs backtick **runs** of equal length (CommonMark inline-code rule) and fail-safes by masking the whole line whenever any run is unmatched (generalizes the previous odd-count rule). ``` `` $path `` ``` inside inline code no longer expands; all prior masking behavior pinned green. Live-verified.
- **Multimodal preview honesty** (half of finding 3) — `_annotate_skill_commands` no longer annotates list-content drafts, since send-time substitution deliberately skips them (replacing list content would drop attachments). Preview and send now agree.

**Deferred with rationale**
- **Multimodal mentions never run** — pre-existing behavior carried over verbatim from the `/` implementation (the early return predates this PR). Making skills work on attachment sends needs a design decision about attachment preservation under inline-replace/fork (a rendered string body can't carry image parts). Filed as a follow-up rather than a drive-by.

**Declined**
- **Magic `execution_mode` strings** — the `"inline"`/`"fork"`/`"local"` literals predate this PR, mirror the schema's `Literal["inline", "fork"]` contract, and match the file's existing convention; a constants refactor would touch schema + bridge + controller well beyond this migration's scope.

Targeted suites 165 passed; full `Tests/Chat/` + skill-commands 1324 passed with only the known pre-existing `test_anthropic_native_tools` baseline.

PR #808, opened 2026-07-23
feat(roleplay): preview real speaker names + italic *action* text (task-437)

rmusser01:
## Summary

Makes the Roleplay preview transcript feel in-genre (TASK-437, from the RP UX review): replies are labelled with the character's real name instead of "character:", and single-asterisk `*action*` spans render italic instead of as literal asterisks.

## Changes

**AC#1 — real speaker names**
- `PersonasPreviewPane` gains `_character_label`/`_user_label` + `set_speakers(*, character=None, user=None)` (empty/None keeps the current label); the five label sites read those fields.
- Wired at `_select_character` (covers normal + TASK-434 restore) and the controller's `handle_character_loaded` (async refine). The user label stays `"you"` — the active persona/user name is **TASK-442**'s remit.

**AC#2 — styled `*action*`**
- A `_styled_line(line) -> rich.text.Text` helper escapes Rich markup then italicizes single-asterisk spans, used at the four render sites. `self._lines`/`transcript_text()` stay **plain** (raw for copy/handoff/restore); only the mounted `Static` content is the styled `Text`. Passing a `Text` (not a markup string) keeps the repo's `Static.renderable` shim plain and makes `MarkupError` impossible. The Markdown widget was rejected (it over-interprets chat text into lists/headings).

**Review fix (regression):** `_character_label` was set on selection but never reset — character → persona → Test Reply showed the stale name. Added `reset_speakers()`, called from `_apply_mode` (a mode switch clears the character context; a later selection re-sets).

## Testing

New preview tests (named labels; default fallback; `*action*` italic + literal-bracket regression preserved; reset-on-mode-switch) + an end-to-end wiring test. The end-to-end workbench suite selecting the "Detective Sam" stub now labels lines `"Detective Sam: …"` (23 assertions updated — pure prefix swaps). Task review: spec ✅, code quality Approved; mutation-checked non-vacuous.

## Notes

Display-only change — the `"character:"/"you:"` prefix is not parsed anywhere (the LLM payload uses `history` role-dicts; `transcript_text()` is only a handoff text-blob + a non-empty guard).

---
## Summary by cubic
Roleplay preview now shows the character’s real name and renders `*action*` text in italics to make transcripts feel in-genre. Implements TASK-437 AC#1 and AC#2.

- New Features
  - Replies and greeting lines use the selected character’s name via `set_speakers(character=...)` at selection; user label stays "you" until TASK-442.
  - Single-asterisk `*action*` spans render italic using Rich `Text`; transcript storage stays plain.

- Bug Fixes
  - Relabel already-rendered character lines when the name changes; `_after_character_save` calls `set_speakers(character=...)` to keep prefixes in sync.
  - Reset speaker labels on mode switches and after deleting the selected character via `reset_speakers()` to prevent stale names.
  - Avoid relabeling on same-character reloads that preserve the transcript to prevent mixed prefixes.

<sup>Written for commit 642af4160c5923cf54f1e8b2702a43195e544021. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/808?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Thanks @qodo-code-review — fixed in the latest commit.

**🐞 Stale prefixes after refine (correctness):** confirmed real for same-character reload/edit flows where the loaded record name differs from the selection name. `handle_character_loaded` was calling `set_speakers()` even in the branch that preserves an in-progress transcript, but `set_speakers` only relabels future lines → the already-rendered lines kept the old prefix (mixed/stale, including in `transcript_text()`/Open-in-Console). The speaker label is already set once at selection (`_select_character`, from the display name), so I dropped the redundant re-set in `handle_character_loaded`. Added a controller unit test pinning that the preserve branch refreshes the greeting seed but does not relabel. Preview + restore + preview-integration suites green.

PR #762, opened 2026-07-22
Skills Foundation: trust isolation, recovery & discoverability (Layer 0, Spec 1)

rmusser01:
## Summary

**Skills Foundation — Trust Isolation, Recovery & Discoverability (Layer 0, Spec 1).**

This is the first layer of a program whose north star is *"a user asks an agent to install a skill/pack from a GitHub link."* Before any of that, the trust foundation had to be solid — a UAT of installing the `obra/superpowers` skillset found a new user could get **permanently stuck**. This PR fixes that.

Full design and plan: `Docs/superpowers/specs/2026-07-21-skills-foundation-trust-design.md`, `Docs/superpowers/plans/2026-07-21-skills-foundation-trust-plan.md`.

## Problems fixed (from the UAT)

- **#1 Cross-profile trust dead-end (P0).** The rollback-protection marker was stored in the OS keychain under a *fixed global* account while the manifest is per-profile, so a fresh profile inherited a foreign marker and reported *"manifest cannot be verified"* — a state with **no working action** (Unlock/Review/Approve all disabled, no "Set up" button). The only escape was manual on-disk/keychain deletion.
- **#4 Trust setup was unguided** and order-dependent — nothing told a new user trust must be set up before imported skills work.
- **#5 Trust controls were unreachable** — buried at the bottom of a long editor whose Body field ate the scroll.
- **#6** Minor palette entry ambiguity.

## What this does

- **Profile-scopes both OS-keyring entries** (generation marker *and* derived-key cache) by a hash of the resolved store dir. This kills the cross-profile dead-end **and** closes a real security leak — the key cache was globally shared, exposing one profile's derived trust key material to another.
- **`reset_trust()` + a structured `trust_posture()`** with dedicated **orphaned-manifest** detection and a **None-vs-raise safety invariant**: a genuinely-absent marker (`None`) is recoverable and offers re-setup; a *raised* `SkillTrustMarkerUnavailable` (e.g. a transiently-locked keychain) is treated as transient — **never** a destructive path. No trust state is a UI dead-end anymore.
- **Adaptive trust header on the Skills list** — *"Skill trust isn't set up — Set up"* / *"needs to be set up again after an update"* / *"N need review"* / quiet *"ready"* — so trust is discoverable and one-click, with **no scrolling** and no need to open a skill. Reset is confirm-gated ("your skills are not deleted") and doubles as the forgot-passphrase escape from `locked`.
[...]

PR #814, opened 2026-07-24
Skills: reference-file reachability — the skill_file runtime tool

rmusser01:
## Summary

**Reference-file reachability — the `skill_file` runtime tool.** Makes the skills program's faithfully-stored bundles (PR #784) actually *usable*: agents can now read a skill's bundled `references/*` on demand, with progressive disclosure instead of context-stuffing.

Follows the trust foundation (PR #762), bundle fidelity (PR #784), and `$`-mention invocation (PR #801). Design + plan: `Docs/superpowers/specs/2026-07-23-skills-reference-file-reachability-design.md`, `Docs/superpowers/plans/2026-07-23-skills-reference-file-reachability-plan.md`.

## What this does

- **One trust-gated read seam** — `read_skill_file(skill_name, relative_path)`: policy gate (new `skills.read_file.launch.local` registry entry — the engine fails closed on unknown ids) → **per-read trust re-verification** (a skill revoked mid-run stops being readable immediately) → path validation → contained read (resolve/symlink/containment — an intermediate-symlinked-directory escape was live-probed and blocked in review) → binaries return a clean refusal string → 100 KB cap with truncation marker. Scope-service passthrough is local-only with a clean server-mode rejection.
- **`skill_file` is the fourth runtime tool** (joining `spawn_subagent`/`find_tools`/`load_tools`): schema pinned per run whenever bindings exist — always active from step 1, never hidden behind tool-disclosure budgets — with authorization via a per-run `SkillFileBindings` object, never `allowed_tools` membership. Unauthorized or hallucinated names get a named per-call refusal; a misbehaving reader fails the call, never the run.
- **Who gets it:** model-invoked skill forks always read their *own* bundle (granted before spawn, independent of frontmatter `allowed-tools` — reading your own references is the progressive-disclosure baseline), and native-tools Console turns get the tool for the skills the user's `$mentions` actually rendered that turn (leading + spliced embedded; never blocked/fork-literal mentions). Plain sends get nothing — no tool, no dangling promise.
- **Discovery travels with capability:** a `Bundled files (readable via skill_file): references/api.md (2 KB), …` block is pre-rendered at substitution (pure string work from metadata already in hand) but inserted only where the tool's existence is certain — bridge-side onto the run's own message copy (copy-safe; the non-mutation contract is pinned by a test proven non-vacuous) and into spawned fork bodies. Never the stored transcript.
- `execute_skill` results carry additive `reference_files` metadata (derived from the payload already computed — no extra bundle walk).

## Verification

- Built subagent-driven: 6 TDD tasks, each spec+quality reviewed. Review-gate catches this run: a live symlink-escape probe validating containment; a double unbounded bundle-walk on every skill invocation (fixed, RED-proven); the Library shadow-name drift guard firing exactly as designed when `skill_file` joined the runtime tool names; a run-crash hardening for contract-violating readers; and direct coverage forced for the copy-safe block-append (an in-place-mutation variant fails the test).
- Final **whole-branch review returned MERGE READY** (no Critical/Important): end-to-end security (single gated seam, no scope-service bypass, revocation backstop verified), bindings lifecycle (fresh per run, no cross-run leaks), byte-identical block formats, clean layering (`agent_runtime` has no Skills_Interop import).
- Full gate: `Tests/Skills/` + `Tests/Agents/` + `Tests/Chat/` + skill-command suite → **1807 passed**, only the known pre-existing baseline.
- Rebased onto latest `dev` (+49): the rebase surfaced a **fifth** substitution call site in dev's new edit-and-resend path still unpacking the old 3-tuple (guaranteed runtime `ValueError`) — integrated with the 5-tuple contract and kwargs threaded, mirroring the sibling sites; targeted suites green.

## Out of scope (follow-ups)

Script **execution** (`scripts/*` running — the trust-gated next layer); per-conversation binding accumulation (per-turn by design; seam named); binary reads; multimodal-turn bindings; exact-100 KB-boundary unit pin; per-row skill attribution in multi-skill blocks.

---
## Summary by cubic
Adds a trust-gated `skill_file` runtime tool so agents can read a skill’s bundled `references/*` on demand, with progressive disclosure. Hardens the read seam based on review (supports exact `SKILL.md`, stronger containment, and an undisclosed‑tool leak fix).

- **New Features**
  - `skill_file(skill_name, path)` runtime tool with per-run `SkillFileBindings`; schema pinned and name-dispatched in the runtime; `skill_file` added to `RUNTIME_TOOL_NAMES`.
  - Availability: skill forks can always read their own bundle; console turns grant reads for skills invoked via `$mentions`; plain sends grant nothing.
  - Secure read seam (local-only): policy `skills.read_file.launch.local`, per-read trust re-check, path containment/symlink escape blocking done before any file ops, text-only with 100 KB cap and truncation marker; exact `SKILL.md` is readable; server mode cleanly rejected.
  - `execute_skill` returns `reference_files`; bridge appends one “Bundled files (readable via `skill_file`)” block to the run’s message copy and spawned fork body only when the tool exists (never stored transcripts).

- **Bug Fixes**
  - Edit-and-resend path updated to the new 5‑tuple `_apply_skill_substitution` contract, avoiding a runtime `ValueError`.
  - Tool call hardening: a bad `skill_file` reader fails only the call, not the run.
  - Avoided redundant bundle walks by deriving `reference_files` from the `get_skill` payload.
  - Review hardening: allow reading exact `SKILL.md`; align runtime dispatch with schema pinning to avoid undisclosed‑tool leakage; check containment before any stat/is_file to block symlink oracles; add a local‑only docstring to `SkillsScopeService.read_skill_file`.

<sup>Written for commit e13ba32953c42bb5401f341efdc2e0a12dde673e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/814?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/plans/2026-07-23-skills-reference-file-reachability-plan.md
+++ b/Docs/superpowers/plans/2026-07-23-skills-reference-file-reachability-plan.md
@@ -1,0 +1,534 @@
+# Reference-File Reachability (`skill_file` runtime tool) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Let agents read a skill's bundled reference files on demand: a `skill_file(skill_name, path)` runtime tool available to model-invoked skill forks and to native-tools Console turns driven by `$skill` invocations.
+
+**Architecture:** One trust-gated, contained read seam in Skills_Interop; `skill_file` becomes the fourth RUNTIME tool (schema pinned via `runtime_schemas`, name-branch dispatch in `run_agent_loop`, authorization via a per-run mutable `SkillFileBindings` seeded from the turn's `$skill` names and extended when `SkillRunner` spawns a skill child). A "Bundled files" block is rendered only where the tool's existence is certain (bridge `run_reply` + `SkillRunner.run`).
+
+**Tech Stack:** Python 3.11+, pytest. No new dependencies.
+
+**Design doc:** `Docs/superpowers/specs/2026-07-23-skills-reference-file-reachability-design.md` (governs on any ambiguity).
+
+## Global Constraints
+
+- **Read seam order (verbatim from spec §1):** `_enforce("skills.read_file.launch.local")` → `_require_trusted_skill(skill_name)` (per-READ trust re-verification; raises `SkillTrustBlockedError`) → `validate_supporting_file_path(relative_path)` → contained read via the `_read_text_preserving_newlines` discipline → binary = clean refusal string, never raises → cap **100,000 chars** with `truncated: True` + trailing marker.
+- **Policy registry entry is REQUIRED** (`_resource("skills.read_file", actions=(LAUNCH,))` in the `server_skills` capability block) — the engine FAILS CLOSED on unknown action ids.
+- **Runtime-tool pattern:** schema pinned into `runtime_schemas` when bindings non-empty (active from step 1, never disclosure-gated); dispatch via a name-branch in `run_agent_loop` BEFORE `deps.invoke_tool`; authorization = membership in `SkillFileBindings.authorized`, NOT `config.allowed_tools`. `agent_runtime.py` must NOT import Skills_Interop.
+- **Bindings:** per-run, mutable; seeded from `turn_skill_bindings` (leading-resolved + embedded-SPLICED names only — never blocked/fork-literal mentions); `_BridgeSkillRunner.run` adds the spawned skill's name BEFORE `spawn`. Own bundle always readable — independent of frontmatter `allowed-tools`.
+- **"Bundled files" block** renders ONLY bridge-side (`run_reply`, trailing user message of its OWN `agent_messages` copy) and in `SkillRunner.run` (spawned body) — never at substitution time, never in the stored transcript, never on plain sends.
+- **Collision:** `SKILL_FILE_TOOL_NAME` joins `RUNTIME_TOOL_NAMES` (agent_models.py:33) — the existing consumers (`_non_colliding_skill_entries`, bridge composition) then exclude a skill literally named `skill_file` automatically.
+- **Sync everywhere:** the agent runtime runs on a worker thread with no event loop; the reader closure wraps the async scope-service call with `asyncio.run(...)` exactly like `BuiltinToolProvider.invoke` / `_BridgeSkillRunner.run` do.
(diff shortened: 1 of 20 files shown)


rmusser01:
Addressed the Qodo review in `e13ba3295`:

**Fixed**
- **SKILL.md read rejected** — spec/impl mismatch confirmed (the spec says the body is readable through the seam). Exact-match special case before the validator; `skill.md`/nested `references/SKILL.md` still rejected; containment unchanged. Live-verified.
- **Undisclosed tool still dispatches** — dispatch wiring now uses the SAME predicate as schema pinning (`bindings is not None and bindings.authorized`), evaluated per run — fork children still get both (the grant happens before spawn). An empty-authorized hallucinated call now falls through to the generic unknown-tool path, matching the tool-absent expectation.
- **Stat before containment** — reordered as defense-in-depth: containment (central `get_safe_relative_path` resolve check) now runs before `is_file`/`stat`, and an intermediate-symlink escape yields an error **indistinguishable** from a missing file (live-verified: `link/secret.md` and `link/nonexistent.md` produce identical `local_skill_file_not_found` errors — no existence/size oracle, and the trust-rescan-to-stat TOCTOU window is closed). Worth noting the practical exposure was already near-zero: the per-read trust re-scan quarantines any bundle containing a symlink before the read path runs.
- **Google-style docstring** on `SkillsScopeService.read_skill_file` (Args/Returns/Raises).

**Declined with evidence**
- **"read_skill_file skips path_validation.py"** — factually inaccurate: the read goes through `_read_text_preserving_newlines`, which is built on the central `path_validation` helpers (`validate_path_simple` ×2 + `get_safe_relative_path`) — the identical discipline the pre-existing skill-body read uses; the flagged pre-read checks were additional guards, not replacements (and the ordering nuance was fixed above).
- **Pydantic for `skill_file` tool args** — the loop's sibling runtime-tool branches (`spawn_subagent`/`find_tools`/`load_tools`) all read `call.args` with the same raw-dict coercion (file convention); boundary validation lives in the read seam; `agent_runtime` deliberately keeps a minimal import surface. Wrapping one branch in a BaseModel would be locally inconsistent for no security gain.

Tests: `Tests/Skills/` + `Tests/Agents/` → 429 passed, 0 failed (4 new).

PR #817, opened 2026-07-24
feat(roleplay): alternate greeting selector in the preview (task-438)

rmusser01:
## Summary

Lets the Roleplay preview start from an **alternate greeting** (TASK-438, from the RP UX review). Previously the preview always seeded the primary `first_message` and alternate greetings were unreachable anywhere in the app.

## Changes

**AC#1 — greeting selector**
- A `Select` (`#personas-preview-greeting-select`) in a new preview row, shown **only when the character has ≥1 alternate**. The controller owns the placeholder-processed greetings list `[first_message, *alternate_greetings]` + a `_current_greeting_index`; picking → `PreviewGreetingSelected(index)` → `handle_greeting_selected` → `pane.seed_greeting(chosen)`.

**AC#2 — Reset returns to the chosen greeting**
- `seed_greeting` sets `pane._greeting`, and Reset already re-renders from `_greeting` — no new Reset logic.

## Review-driven fixes (the `Select` widget had two subtle traps)
- **Same-character reload** (edit+save the card after picking an alternate) must **keep** the chosen index, else Reset silently reverted to the primary — fixed with `keep_index`/clamp.
- **Critical:** `Select.set_options()` internally snaps `value→0`, firing a spurious `Select.Changed(0)` that was misread as a user reselection and `invalidate()`d the in-progress transcript on reload — fixed by wrapping the programmatic population in `with self.prevent(Select.Changed)`. A test that drives the **real Select widget** guards this (the earlier controller-direct test missed it).

## Testing

Pane tests (selector shown/hidden, picking posts the message); integration tests (pick re-seeds; **Reset returns to the chosen greeting**; same-character reload preserves the choice + the transcript; no-alternates hides the selector). Task review + two adversarial re-reviews (both found real bugs, both fixed + mutation-verified). Suites green.

## Scope

Preview only (not the card view / Console). A `Select.set_options()`-resets-value footgun noted at a couple of other call sites (which already carry hazard-comments) — out of scope.

---
## Summary by cubic
Adds a greeting selector to the Roleplay preview so you can start a conversation from any alternate greeting. Reset returns to the chosen greeting, and the choice now persists across navigation restores. (TASK-438)

- **New Features**
  - Greeting dropdown in the preview, shown only when a character has alternates.
  - Choosing an option re-seeds the preview from that greeting.
  - Selector hides when no alternates or the preview is cleared.
  - All greetings get the same placeholder processing as the primary.

- **Bug Fixes**
  - Keeps the chosen greeting after same-character reloads (edit/save), so Reset still uses it.
  - Prevents a spurious `Select.Changed` from programmatic repopulation that could wipe the transcript.
  - Persists the selected greeting across save/restore navigation, preventing silent reversion to the primary.

<sup>Written for commit 868acf399074cfbbbcbd57587aa065a68048ab69. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/817?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/plans/2026-07-23-alternate-greeting-selector.md
+++ b/Docs/superpowers/plans/2026-07-23-alternate-greeting-selector.md
@@ -1,0 +1,300 @@
+# TASK-438 Alternate greeting selector — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** When a character has alternate greetings, the Roleplay preview shows a dropdown to pick which greeting seeds the conversation, and Reset returns to the chosen greeting (not the primary).
+
+**Architecture:** A `Select` in the preview pane (shown only when alternates exist). The controller owns the placeholder-processed greetings list `[first_message, *alternate_greetings]` and a `_current_greeting_index`; picking an option re-seeds via `pane.seed_greeting(chosen)`, which sets `pane._greeting` so the existing Reset (`_render_seed_lines` from `_greeting`) returns to the chosen greeting with no new Reset logic.
+
+**Tech Stack:** Python 3.11+, Textual, pytest.
+
+## Global Constraints
+
+- The selector lives in the **preview pane** only (not the card view / Console). It is shown **only when `len(greetings) > 1`** (≥1 alternate); a character with no alternates shows no selector (unchanged behavior).
+- `alternate_greetings` is a top-level `list[str]` on the record, available reliably only from the **async `handle_character_loaded`** path (and synchronously on a cached re-selection). Do not attempt to populate synchronously on a cold first selection.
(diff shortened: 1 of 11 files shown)


rmusser01:
Thanks @qodo-code-review — all three addressed.

- **🐞 Restore resets greeting (correctness):** confirmed real on the TASK-434 navigation-restore path — `restore_conversation` reseeded the alternate but left `_current_greeting_index=0`, so the async `handle_character_loaded(keep_index=True)` kept index 0 and `refresh_greeting_seed` overwrote the reset seed back to the primary (Reset reverted). Fixed by persisting `_current_greeting_index` in `save_state` and restoring it via `restore_conversation`'s new `greeting_index` arg. Covered by a controller overwrite test + a save/restore round-trip index assertion.
- **📘 Docstrings:** added the missing Google-style `Args` for `set_greetings(selected_index)` and `handle_greeting_selected(index)`.

PR #845, opened 2026-07-24
fix(config): call-time user-data-dir resolution — restores test HOME isolation (task-519)

rmusser01:
Root-cause fix for a test-isolation hazard that bit **three** programs (SP2b twice, SP3 once): `get_user_data_dir()`'s fallback was the module constant `BASE_DATA_DIR_CLI = Path.home()/...`, frozen at import — before any test fixture patches `HOME` — so unmocked default-dir consumers (profile manager, first-run import, validators) silently read/wrote the developer's real `~/.local/share/tldw_cli`.

**Fix:**
- `_default_base_data_dir()` resolves `HOME` at **call time** — semantics otherwise identical to before for every real user. (`XDG_DATA_HOME` is **deliberately not honored**: the pre-existing default never consulted it, and adding it would silently relocate an XDG user's entire data dir on upgrade with no migration — caught in review, locked by inverse-XDG tests.)
- `Tests/conftest.py`: removed the dead `get_data_dir` hasattr-patch (that function never existed — the fixture's data-dir patch had been silently no-opping) and the autouse isolation now pre-arms SP2b's first-run once-flag.
- **Retired the `PYTEST_CURRENT_TEST` production guard** in `_maybe_run_first_run_import` — the organic suite now exercises the real first-run wiring; the two wiring tests re-arm the flag directly.
- Settings' "User data directory" display now uses the same call-time resolution (it renders this path in production; the frozen constant could diverge).

**Proof:** RED/GREEN on the frozen-home behavior; scratch-HOME run of the previously-leaking suites → writes landed under scratch, and `find -newer <marker>` on the real data dir showed no new files. `Tests/RAG/` 536 passed; settings suite at its 7-failure pre-existing baseline. Backlog `task-519` → Done. Reviewed (one Important round: the XDG migration hazard, fixed + RED-proven).

---
## Summary by cubic
Resolves the user data directory at call time to restore test HOME isolation and prevent writes to the real `~/.local/share/tldw_cli`. Settings now use the same path logic. Addresses task-519.

- **Bug Fixes**
  - `get_user_data_dir()` now falls back to `_default_base_data_dir()` which resolves `HOME` at call time; deliberately ignores `XDG_DATA_HOME`.
  - Removed the `PYTEST_CURRENT_TEST` guard in `RAG_Search/ingestion_indexing._maybe_run_first_run_import`; `Tests/conftest.py` pre-arms `_first_run_import_attempted = True` only if the module is already loaded (no implicit import).
  - Settings screen uses `_default_base_data_dir()` for its read-only path display to match runtime resolution.
  - Added regression tests for call-time `HOME` resolution and XDG ignore; updated RAG tests to drop the old guard bypass and removed the dead `config.get_data_dir` patch.

<sup>Written for commit bfe7236285b6b14870bd319af90abb09a395dde2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/845?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/RAG/test_first_run_import.py
+++ b/Tests/RAG/test_first_run_import.py
@@ -162,8 +162,9 @@ def test_get_shared_rag_service_calls_first_run_import_at_most_once(
 
     calls = []
     monkeypatch.setattr(ac, "ensure_imported_profile", lambda: calls.append(1) or None)
-    # Bypass the test-mode safety skip so the wiring path under test actually runs.
-    monkeypatch.setattr(ii, "_running_under_pytest", lambda: False)
+    # No test-mode skip to bypass anymore (task-519 removed the
+    # PYTEST_CURRENT_TEST guard): the _reset_first_run_wiring fixture already
+    # resets _first_run_import_attempted so this exercises the real path.
 
     fake_service = object()
     ii.set_shared_rag_service(fake_service)
@@ -203,7 +204,9 @@ def test_first_run_import_runs_before_shared_service_lock_is_held(
         acquired.append(got)
 
     monkeypatch.setattr(ac, "ensure_imported_profile", _fake_ensure_imported_profile)
-    monkeypatch.setattr(ii, "_running_under_pytest", lambda: False)
+    # No test-mode skip to bypass anymore (task-519 removed the
+    # PYTEST_CURRENT_TEST guard): the _reset_first_run_wiring fixture already
+    # resets _first_run_import_attempted so this exercises the real path.
     # Avoid a real (possibly slow/dependency-gated) RAGService build -- this
     # test is only exercising the lock-acquisition ordering, not construction.
     monkeypatch.setattr(simplified_pkg, "create_rag_service", lambda **kwargs: object())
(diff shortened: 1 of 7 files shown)


rmusser01:
Addressed (pushed): the autouse fixture no longer imports the RAG stack. It now arms `_first_run_import_attempted` only when `tldw_chatbook.RAG_Search.ingestion_indexing` is ALREADY in `sys.modules` — the first-run wiring lives inside that module, so if it isn't imported it cannot fire; and if a test imports it later, task-519's call-time HOME resolution already isolates any first-run write under the patched HOME. Zero cost for non-RAG suites; the two wiring tests (which import + re-arm the flag themselves) still pass.

PR #851, opened 2026-07-24
Security hardening: git-clone RCE + tool-sandbox + pickle→json + eval RLIMIT (TASK-330/331/332)

rmusser01:
## Summary

Four independent defensive-security fixes — the final security cluster of the LLM-harness review 2026-07. Closes **TASK-330**, **TASK-332**, and **TASK-331 #1/#2** (331 #3 deferred to a follow-up).

### TASK-330 — git-clone transport/argument injection (MEDIUM)
`_clone_git_repository` built a `git clone` argv from an ingestion-source `repo_url`/`ref` with no scheme validation, no `--` separator, and no protocol restriction — so `repo_url = "ext::sh -c '<cmd>'"` was arbitrary-shell **RCE**, and a leading-dash `repo_url`/`ref` was git-option injection. Fix:
- New `validate_git_repo_url` / `validate_git_ref` in `Utils/input_validation.py`: allowlist **explicit `https://` and `ssh://` only** (scp-shorthand rejected as ambiguous); reject `ext`/`file`/`fd`/`git`/`http` schemes, leading-dash **and dashed ssh host/user** (CVE-2017-1000117 class), whitespace/backslash/**zero-width unicode**, and ref control chars/`..`.
- `_clone_git_repository` validates both inputs, inserts a literal `--` before the URL, and runs with `env` `GIT_ALLOW_PROTOCOL=https:ssh` + `GIT_PROTOCOL_FROM_USER=0` (git blocks non-{https,ssh} even across redirects/submodules). Defense-in-depth: even a validator gap can't make a positional parse as an option.

### TASK-331 #1 — real file-tool sandbox root
`ReadFileTool`/`ListDirectoryTool`/`WriteFileTool` passed the literal strings `"file"`/`"directory"` as `validate_path`'s sandbox-root argument (confining to a bogus `<cwd>/file`, fails-closed only by accident). Fix: a real `[tools] file_sandbox_root` config (default `get_user_data_dir()/tool_sandbox`) at all three sites — plus a fix to shared `validate_path` so its hidden-file check inspects only the **user-supplied portion relative to the sandbox root** (the default root lives under `~/.local/share`, whose `.local` component was tripping the whole-path check and dead-lettering the tools; this also unbreaks the same latent bug in other `get_user_data_dir()`-based callers). Confinement verified: absolute/traversal/**symlink** escape all rejected; dotted user-path components still rejected.

### TASK-331 #2 — pickle→json + a crash fix
`ToolResultCache` deserialized its on-disk cache with `pickle` (code-execution anti-pattern). Swapped to `json` (all cached values are JSON-serializable; no migration — see below). Also fixed a **second bug**: `get_tool_executor()` did `from ..config import USER_DATA_DIR` — a name that doesn't exist — so it raised `ImportError` whenever caching was enabled (this is why no legacy pickle files exist). Now uses `get_user_data_dir()`.

### TASK-332 — surface non-enforced eval-sandbox memory limit
The eval code-runner set `RLIMIT_AS` (256MB) in `except (ValueError, OSError): pass` — on macOS/BSD that `ValueError` is silently swallowed, so the memory cap doesn't apply (only `RLIMIT_AS`; `RLIMIT_NPROC` works). Fix: parent-side `_memory_limit_enforced()` → a one-time WARNING + `results["sandbox_warnings"]` (no silent gap), and a corrected docstring. Time is still bounded cross-platform by the subprocess timeout.

## Testing

169 tests across the touched areas (git validators incl. adversarial bypass regressions, `path_validation` + all its callers, git-clone hardening, media suite, all `Tools/` tests, eval security + RLIMIT surfacing, file-operations integration).

Built via SDD (fresh implementer + reviewer per task; opus whole-branch review **READY TO MERGE**). Adversarial per-task review caught and fixed **2 Criticals**: the git validator's SSH host-leading-dash option-injection + ref trailing-newline bypasses, and the sandbox being non-functional under the default config.

## Follow-ups filed (TASK-545/546/547)

- **545** — wire the built-in tool executor into the existing MCP Allow/Ask/Deny permission gate (TASK-331 #3, deferred as a cross-system + UI build ~as large as this whole cluster; fs tools remain default-OFF meanwhile).
- **546** — restrict the `file://`/local-path read vector in the git ingestion source (secondary to the clone-time transport allowlist).
- **547** — `get_tool_executor()` reads `get_cli_setting("tools", {})` where `{}` is treated as the default slot, so the entire `[tools]` config (all tool flags + cache) is currently unreachable — a pre-existing bug (which also means the fs tools can't even be enabled today = extra-safe posture).

## Scope note

TASK-331 #3 (the confirmation/permission gate) is **not** in this PR — decomposed into TASK-545. The Task-331#1 sandbox fix makes the fs tools functional-within-a-sandbox while that gate is deferred; the confined, default-off, narrow-empty-sandbox posture is the intended, on-the-record tradeoff.

---
## Summary by cubic
Hardened repo ingestion, file tools, and the eval runner to close RCE and sandbox gaps. Blocks `git clone` transport/arg injection, confines file tools to a real sandbox, switches the tool cache to `json` with safer startup, and warns when macOS memory limits aren’t enforced. Addresses TASK-330, TASK-331 (#1/#2), and TASK-332.

- **Bug Fixes**
  - Git clone (TASK-330): Validate repo URL/ref (allow only explicit `https://`/`ssh://`), reject scp shorthand, leading dashes, whitespace/zero‑width/control chars; add `--` separator; run with `GIT_ALLOW_PROTOCOL=https:ssh` and `GIT_PROTOCOL_FROM_USER=0`. Tests ensure malicious inputs fail before spawning `git` (incl. SSH host/user leading‑dash and ref trailing‑newline).
  - File tools sandbox (TASK-331 #1): Use `[tools] file_sandbox_root` (default `<user data dir>/tool_sandbox`). Update `validate_path` so the hidden‑file check inspects only user‑supplied segments; dotted ancestors (e.g. `~/.local/...`) no longer block valid paths. Traversal/symlink escapes still rejected.
  - Tool cache (TASK-331 #2): Replace `pickle` with `json` in `ToolResultCache`; corrupt files degrade gracefully. Fix `get_tool_executor()` to use `get_user_data_dir()` and defer disk‑load when constructed without a running loop (lazy start on first async get/set). Tests cover JSON round‑trip, corrupt files, and no “no running event loop” crash.
  - Eval runner (TASK-332): Add `_memory_limit_enforced()` and `_sandbox_warnings`; surface non‑enforced `RLIMIT_AS` on macOS via one‑time WARNING and per‑result `sandbox_warnings`. Docstring corrected.

- **Migration**
  - No data migration; cache now writes JSON. A sandbox directory is created at `<user data dir>/tool_sandbox`. Tools remain default‑off.
  - Follow‑ups filed: TASK-545 (permission gate), TASK-546 (restrict `file://` local read), TASK-547 (fix tools config reachability).

<sup>Written for commit 476ce0d3c1d4ba9f51e413c272fca0685a369f38. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/851?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/plans/2026-07-24-security-hardening-cluster.md
+++ b/Docs/superpowers/plans/2026-07-24-security-hardening-cluster.md
@@ -1,0 +1,725 @@
+# Security Hardening Cluster Implementation Plan (TASK-330 + 331#1/#2 + 332)
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Four independent defensive-security fixes: git-clone transport/argument-injection hardening, a real file-tool sandbox root, pickle→json (+ a crash) in the tool-result cache, and surfacing the silently-unenforced eval-sandbox memory limit.
+
+**Architecture:** Each fix is confined to one subsystem. New git-URL validators live in `Utils/input_validation.py`; the other fixes are in-place edits to `Media/local_media_reading_service.py`, `Tools/file_operation_tools.py`, `Tools/tool_executor.py`, and `Evals/specialized_runners.py`.
+
+**Tech Stack:** Python 3.11, pytest, unittest.mock. No new dependencies.
+
+**Spec:** `Docs/superpowers/specs/2026-07-24-security-hardening-cluster-design.md` (committed `b0cc04ab3`). TASK-331 **#3** (permission/confirmation gate) is DEFERRED to its own spec — do NOT implement it here.
+
+## Global Constraints
+
+1. **Git URL allowlist:** explicit scheme ∈ `{"https","ssh"}` ONLY. Reject `ext`, `file`, `fd`, `git`, `http`, scp-shorthand (`user@host:path`), absent scheme, whitespace/backslash/control chars, and a leading `-`. Same env on the clone: `GIT_ALLOW_PROTOCOL="https:ssh"`, `GIT_PROTOCOL_FROM_USER="0"`, plus a literal `"--"` argv separator before the URL.
+2. **Sandbox root** default = `get_user_data_dir() / "tool_sandbox"` (the real API; `USER_DATA_DIR` does NOT exist), config-overridable via `[tools] file_sandbox_root`, `mkdir(parents=True, exist_ok=True)`.
+3. **Cache serializer** = `json` (NOT pickle); no migration needed (no legacy pickle files exist — the path crashed before ever writing one). The existing try/except around load/save already degrades gracefully.
+4. **RLIMIT surfacing** is PARENT-side (`platform.system()=="Darwin"` → not enforced); one-time WARNING + `results["sandbox_warnings"]`. Only `RLIMIT_AS` is the gap; do NOT touch `RLIMIT_NPROC`.
+5. **DEFERRED (do not touch):** TASK-331 #3 permission gate; the `file://`-local-read vector (follow-up).
+6. Worktree `/Users/macbook-dev/Documents/GitHub/tldw_chatbook-sec-hardening` (branch `feat/security-hardening-cluster`); tests via `/Users/macbook-dev/Documents/GitHub/tldw_chatbook/.venv/bin/python -m pytest` FROM the worktree. Never touch the main checkout. `git add` only each task's listed files, never `-A`.
+7. **Line numbers below are as-of origin/dev `5a402cbb7` — re-verify with `grep -n` before editing; the target text is authoritative.**
+
+**Baseline:** `Tools/` and `Media/` and the git-clone path have essentially no existing tests for the touched code; `Tests/Evals/test_code_execution_security.py` exists. Run `/Users/macbook-dev/Documents/GitHub/tldw_chatbook/.venv/bin/python -m pytest Tests/Evals/test_code_execution_security.py -q` before Task 5 and note any pre-existing failures — report, don't fix.
+
+---
+
+### Task 1: Git URL + ref validators (`Utils/input_validation.py`)
+
+**Files:**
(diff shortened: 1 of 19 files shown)


rmusser01:
Thanks — one fixed, one declined with evidence.

**Fixed — 🐞 Cache load needs running loop** (`476ce0d3c`). Correct: `ToolResultCache.__init__` did an unguarded `asyncio.create_task(self._load_from_disk())`, and my TASK-331 fix (repointing the persist path from the non-existent `USER_DATA_DIR` to `get_user_data_dir()`) made that path reachable — so constructing the cache from a sync context (`get_tool_executor()`) with no running loop would raise `RuntimeError`. Now the load is guarded (`_maybe_start_load()` swallows the no-loop `RuntimeError`) and started **lazily** on the first async `get()`/`set()`, where a loop exists. Added regression tests: sync construction no longer raises, and a pre-written on-disk cache still loads lazily on first `get()`. (The existing cache-enabled test never caught this because it ran inside `asyncio.run(...)`.)

**Declined — 🐞 Tools config never read.** This is a real bug, but **pre-existing and deliberately out of scope for this PR**: `get_tool_executor()` reads `tools_config = get_cli_setting("tools", {})`, where `{}` is treated as the default-value slot, so the whole `[tools]` section is unreachable and every tool flag defaults False. My diff only touched the adjacent lines to fix the `USER_DATA_DIR` `ImportError` — it did not introduce this. I already filed it as **task-547** and am deferring it intentionally: fixing the config read makes the fs/mutating tools (`write_file`/`create_note`/…) *enable-able*, and those must land **with** their confirmation/permission gate (**task-545**, TASK-331 #3, deferred), not before it. Enabling mutating tools while their gate is still deferred would be the wrong ordering; the current "config unreachable" state is in fact an extra-safe posture until 545+547 land together.

PR #852, opened 2026-07-24
Console branching//rewind follow-ups, batch 1 (tasks 500, 549, 550)

rmusser01:
## What

First batch of the Console branching//rewind follow-up backlog — the small hardening trio, each marked Done with Implementation Notes in its backlog file:

- **task-500 — Phase A cleanups**: deleted the production-dead `_iter_console_tree_messages` helper + its orphaned unit test (resume loads the full tree since Phase A Task 8); added visited-set cycle guards to both active-path ancestry walks (defensive-only, mirrors `_nearest_persisted_ancestor_id`) with manufactured-cycle termination tests; added a real-DB multi-root resume test (genuine all-USER root fork via first-message edit-&-resend: the second root loads off-path, isn't shown, stays swipe-reachable).
- **task-549 — rewind session guard**: `_apply_console_rewind_choice` no-ops with a notify if the active session changed since the modal opened (covers restore/summarize/None; theoretical under today's modal, robust for future surfaces), with session-switch tests for both actions.
- **task-550 — stale summary clear**: resume now best-effort clears the persisted `context_summary`/`summary_boundary_message_id` pair when the boundary is dangling (guarded + non-fatal, same pattern as the write-throughs); valid boundaries restore unchanged; real-DB tests for both paths.

## Testing

151 tests green across the covering suites (store/tree/summary/rewind-restore/resume + both e2es); `Tests/UI/test_console_native_chat_flow.py` fully green post-deletion (exit 0).

## Note for the maintainer

**PR #803 (tasks 498–501) is still open** — it files task-500 as To-Do. This PR lands task-500 as Done, so #803 should drop its `task-500` copy on its next rebase (or I can rework #803 to only carry 498/499/501 — say the word).

---
## Summary by cubic
Hardens Console branching and rewind flows by adding ancestry cycle guards, guarding rewind against session switches, and clearing stale persisted summaries on resume. Removes a dead tree-walk helper and adds targeted real-DB tests, including multi-root resume and stale-summary handling.

- **Bug Fixes**
  - Add visited-set guards to `active_path_message_ids` and `_recompute_active_path` to prevent hangs on malformed cycles (Task-500).
  - Rewind: `_apply_console_rewind_choice` no-ops with a warning if the active session changed while the modal was open; covers restore/summarize/None (Task-549).
  - Resume: clear persisted `context_summary` when its boundary message is missing; keep valid boundaries unchanged (Task-550).
  - Remove dead `ChatScreen._iter_console_tree_messages` and its orphaned test; add tests for multi-root resume, cycle termination, session-switch guard (restore/summarize), and summary boundary clear/restore.

<sup>Written for commit 8425716cc051a557398cff21a98b5aac0190a766. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/852?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Chat/test_console_chat_store_tree.py
+++ b/Tests/Chat/test_console_chat_store_tree.py
@@ -325,3 +325,38 @@ def test_deferred_active_leaf_pointer_written_on_append_after_swipe():
         assert db.get_conversation_active_leaf(conversation_id) == a2_pid
     finally:
         db.close_connection()
+
+
+def test_active_path_message_ids_terminates_on_manufactured_cycle():
+    """A visited-set guard keeps the ancestry walk from hanging on a cycle.
+
+    Real DBs can't produce a cyclic parent chain (unique PKs), so this is
+    defensive-only hardening (mirrors ``_nearest_persisted_ancestor_id``'s
+    guard) against a malformed test double or future in-memory bug -- hand
+    -wire a 2-cycle into ``_native_parent_by_message`` directly since the
+    public API refuses to create one.
+    """
+    store, sid = _store_with_session()
+    u = store.append_message(sid, role=ConsoleMessageRole.USER, content="hi")
+    a = store.append_message(sid, role=ConsoleMessageRole.ASSISTANT, content="yo")
+    # hand-wire a cycle: a -> u -> a (bypassing the normal append plumbing,
+    # which can never produce this)
+    store._native_parent_by_message[u.id] = a.id
+
+    ids = store.active_path_message_ids(sid)
+    assert set(ids) <= {u.id, a.id}
+    assert len(ids) <= 2
+
+
+def test_recompute_active_path_terminates_on_manufactured_cycle():
+    """Same cycle guard for ``_recompute_active_path``'s node-materializing walk."""
+    store, sid = _store_with_session()
+    u = store.append_message(sid, role=ConsoleMessageRole.USER, content="hi")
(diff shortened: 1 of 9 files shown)


rmusser01:
Addressed Qodo's two findings:

1. **Docstring (fixed)** — `active_path_message_ids` now carries Google-style `Args:`/`Returns:`/`Raises:` sections (commit on this branch).

2. **"Summary cleared on truncation" (declined with evidence)** — the only production caller feeding `_resolve_context_summary_on_resume` is the Console resume path, which loads the tree with `depth_cap=10_000, root_limit=10_000` (raised from the service's 50-defaults in Phase A specifically so the full tree loads intact — see the comment at the `get_conversation_tree` call in `chat_screen.py`). A truncated load therefore requires a conversation exceeding 10,000 nodes/roots; and even in that pathological case the boundary is unusable in-memory for that session regardless (the mapping fails open before the clear), and would dangle identically on every future same-cap resume — so the clear removes a pair that could never be applied again anyway. No change made.

Also rebased on latest dev; touched suites green (149 passed) and the full `test_console_native_chat_flow.py` is 199/199 on the rebase.

PR #860, opened 2026-07-25
Console branching//rewind follow-ups, batch 2 (tasks 548, 501)

rmusser01:
Second batch of Console branching//rewind follow-ups.

## task-548 — inspector next-send preview reflects boundary compaction
`build_context_snapshot` now annotates ids, applies the same `_apply_context_summary_compaction` as the dispatch choke point (after the dictionary transform), strips the private key, and derives the duplicated `system` field from the payload's own leading system row — so the inspector preview matches the actually-dispatched payload when a /rewind summary is active.

## task-501 — keep transcript selection across sibling swipe
A `<`/`>` sibling swipe now hands the landed sibling's id off as a PENDING selection: the screen holds it (remount-proof across recomposes) and transfers it onto whichever transcript instance receives the post-swipe push; `set_messages` applies it at ingest, before the stale-selection clear. The action row therefore stays available and repeated `<`/`>` presses compare branches back-to-back with no intervening row re-click — the primary use case branching was built for. "continue" and every other selection-clearing action keep the PR #359 clear-on-swap rule.

## Verification
- `test_console_sibling_swipe_buttons_navigate_between_regenerated_siblings` extended: selection follows both swipe directions, repeated swipe with no re-click, boundary disabled states re-pinned under the new resting state.
- 177 passed across the swipe test + console store (base/sibling/tree) + transcript selection-contract suites; full `test_console_native_chat_flow.py` at 196 passed with only the 4 pre-existing baseline failures (identical on clean HEAD).
- task-501 live-verified in the real TUI (tmux, local llama_cpp): regenerate to 3 siblings, `<`/`<`/`>` back-to-back with no row re-clicks — counter walked (3/3)→(2/3)→(1/3)→(2/3) with the action row and guide persistent throughout.

---
## Summary by cubic
Aligns the inspector next-send preview with `/rewind` boundary compaction and keeps transcript selection across sibling swipes. Improves preview accuracy and removes re-clicks when comparing branches.

- **Bug Fixes**
  - task-548: Preview mirrors dispatch compaction. Annotates ids, compacts after dictionary transform, strips `NATIVE_MESSAGE_ID_KEY` (also in degraded snapshots), and derives `system` from the payload’s leading system row. Covers no-summary and dangling-boundary cases. Tests updated.
  - task-501: Selection follows `<`/`>` swipes. Landed sibling id is handed off as a pending selection, moved to the active transcript, and applied in `set_messages` before the stale-selection clear. Repeated swipes need no re-click; "continue" still clears.

<sup>Written for commit c841802baf5a7aa41baebb413c742af1489f3765. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/860?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Chat/test_console_chat_controller.py
+++ b/Tests/Chat/test_console_chat_controller.py
@@ -2418,6 +2418,13 @@ async def test_build_context_snapshot_isolates_assembly_errors():
     assert payload["messages"][0]["content"] == "Hello"
     assert payload["messages"][1]["content"].startswith("Follow up")
     assert payload["system"] == []
+    # Qodo (PR #860): the failure here fires inside the annotate->strip
+    # window, so the degraded payload must strip the private id-threading
+    # key too -- it must never surface in the inspector snapshot.
+    assert all(
+        controller_module.NATIVE_MESSAGE_ID_KEY not in row
+        for row in payload["messages"]
+    )
 
 
 def test_annotate_skill_commands_multimodal_text_part():
--- a/Tests/Chat/test_console_rewind_summarize.py
+++ b/Tests/Chat/test_console_rewind_summarize.py
@@ -548,3 +548,78 @@ async def test_native_message_id_key_stripped_before_provider():
         and "[Summary of earlier conversation]" in row.get("content", "")
         for row in gateway.captured_messages
     )
+
+
+# ---------------------------------------------------------------------------
+# task-548: the inspector next-send preview mirrors boundary compaction
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.asyncio
+async def test_snapshot_reflects_boundary_compaction():
+    """With an active summary, build_context_snapshot compacts like a real send."""
+    store = ConsoleChatStore()
+    controller = ConsoleChatController(store=store, provider_gateway=SummaryGateway())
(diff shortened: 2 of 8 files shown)


rmusser01:
Addressed Qodo's finding:

**Snapshot leaks native ids (confirmed + fixed)** — the except handler in `build_context_snapshot` built the degraded payload from the annotated `provider_messages` without stripping `NATIVE_MESSAGE_ID_KEY`, so a failure inside the annotate→strip window leaked the private key into the inspector snapshot. The degraded path now strips it exactly like the success path, and the existing `test_build_context_snapshot_isolates_assembly_errors` (whose mocked failure fires precisely in that window) now pins the absence — verified red before the fix, green after.

Also rebased on latest dev (including #852); 515 tests green across the flow/controller/store/transcript/rewind suites (one unrelated conversation-browser test flaked under full-battery load and passes in isolation).

PR #832, opened 2026-07-24
feat(console): image-generation card with variants — /generate-image (Phase 2a)

rmusser01:
## Image Generation — Console Card with Variants (Phase 2a)

Phase 2a of the in-chat image-generation program (Phase 1 foundation merged in #800). This brings generation **into the Console conversation**: `/generate-image` produces a persistent assistant message rendered as a bordered "Image Generation" card with working variant navigation — mirroring tldw_server's webui pattern and the program's reference design.

### What's included

**Command & flow**
- `/generate-image <prompt>` — leading `:backend` token overrides the backend (`/generate-image :swarmui a red dragon`). Registered as a real Console command (not the skill fallback); dispatch runs before the send-readiness gate.
- Configurable initial batch (`[image_generation].default_batch`, default 1); sequential generation on a thread worker (`asyncio.to_thread`) — never on the UI loop; partial batches keep the successes ("k/N generated"); zero successes → error status, no orphan message, **draft restored to the composer**.
- In-flight guards (per-session for the command, per-message for regenerate), always cleared in `finally`.

**Storage (schema v24 → v25)**
- New `message_generation_metadata` sidecar table keyed `(message_id, position)` (prompt/negative/backend/model/seed/style/params per variant), CASCADE-delete, deliberately **not** sync-integrated and prompts **not** FTS-indexed (v19/v24 precedent). Fresh DBs reach v25 via the migration chain.
- Variant image bytes ride the existing attachments contract (position 0 = messages-row scalar mirror, ≥1 = `message_attachments`). **The kept/selected variant IS position 0.**
- **Narrow persistence ops** — `append_message_attachment_with_metadata` (single INSERT + sidecar, one transaction) and `swap_message_attachment_with_scalar` (two-variant swap + sidecar re-key, one transaction). The full-list `update_message_content(attachments=...)` rewrite is deliberately never used for these: in-memory attachment bytes may be `None` (unrehydrated) and that contract overwrites — a data-loss footgun caught at spec review and regression-tested.

**Card & actions**
- New transcript row kind mirroring the image-row pattern: bordered "Image Generation" panel — browsed variant image (graphics/pixels modes, existing render cache) + details (Style/Source/Seed/Prompt/Negative) + `n/N`. Plain image row suppressed for card messages (no double render, no 16-slot window burn). Byte-less variants render a placeholder, never crash.
[...]

PR #822, opened 2026-07-24
Web-fetch hardening: SSRF protection + size caps/timeouts (TASK-328/329)

rmusser01:
## Summary

First app-wide SSRF protection + response size caps/timeouts for outbound URL fetching. Closes **TASK-328** (SSRF protection) and **TASK-329** (size caps/timeouts) from the LLM-harness-review stream.

Before this branch there was **no SSRF protection anywhere**: `validate_url` was syntactic-only and intentionally permitted localhost/metadata/internal hosts, every fetcher followed redirects with no post-redirect re-validation (a validated public URL returning `302 → http://169.254.169.254/` was followed), and many fetchers read unbounded bodies / had no timeout.

## What's in it

**New `tldw_chatbook/Utils/egress.py`** — the single SSRF policy + guarded transport helpers:
- **One policy rule:** a URL is allowed iff every resolved IP is public and not a cloud-metadata endpoint, OR its host is user-trusted (`trusted_origins`), OR it's in `[web_security] allowed_hosts`. Metadata endpoints (169.254.169.254 / AWS-IPv6 / Alibaba / `metadata.*.internal`) are blocked **even for trusted origins** — only the allowlist overrides them.
- Sync **and** async DNS evaluation (async uses the event loop — never blocks it); classification via `ipaddress.is_global` (covers RFC1918/loopback/link-local/ULA/CGNAT/reserved/multicast + IPv4-mapped IPv6). Fail-closed: DNS failure blocks; any bad IP among multiple records blocks; the kill switch short-circuits before DNS.
- Four **guarded fetch helpers** (httpx sync/async, `requests` with session-auth-safe hops + `sink=` streaming, aiohttp): pre-check → redirects disabled → manual hop loop re-validating every `Location` → decompressed-byte cap that aborts mid-stream. Cross-origin hops strip `Authorization`/`Cookie`/`Proxy-Authorization` (and suppress `requests.Session.auth`).
- **Playwright** paths: pre-`goto` check + post-navigation redirect-chain validation (route handlers don't see server redirect hops — empirically confirmed on Playwright 1.58).

**Posture (local-first friendly):** URLs the user explicitly configured (feed sources, Confluence `base_url`, ingest URLs) may resolve to private IPs; **content-derived** URLs (redirect targets, sitemap/crawl-discovered links, feed items) must resolve public. Trust is threaded **fail-closed** — shared pipeline functions default to no trust and never auto-trust their own input URL.

[...]

PR #862, opened 2026-07-25
Image-gen follow-ups: P1 polish (task-497) + adopt Utils/egress SSRF policy (task-498)

rmusser01:
## Summary

Engine-cluster follow-ups from the image-gen program (P1 #800 / P2a #832 / P2b #850 / P3 #859). Two backlog tasks, each SDD-reviewed per-task plus a whole-branch final review (merge-ready, zero findings).

### task-497 — Image-gen Phase-1 polish (`50368cecc`)
- `http_client`: explicit `timeout=0`/falsy values no longer silently replaced by the default (`_DEFAULT_TIMEOUT if timeout is None else timeout`), pinned by test.
- Registry: adapter-load-failure coverage + autouse registry reset fixture; lazy-import `AttributeError` now names the missing attribute; cold-start test restores `sys.modules` via try/finally.
- New opt-in live-backend suite `Tests/Image_Generation/test_live_backends.py` (`integration, optional, slow`; env-gated per backend, matching the config loader's secrets map).
- AC6 (demo-screen worker refactor) recorded as omitted per the AC's own alternative, with rationale in the task file.

### task-498 — Adopt `Utils/egress.py` SSRF policy (`6c08ec6a0` + `c2c6527b2`)
- `Image_Generation/http_client.py`'s placeholder guard now delegates to the app's central egress policy (#822): private/link-local/metadata blocking, per-hop redirect revalidation. `Utils/egress.py` itself untouched.
- Trust distinction: adapters pass `trusted_origins=origin_set(<self-built URL from configured base_url>)` so operator-configured local backends (SwarmUI etc.) keep working; API-response-extracted URLs are fully enforced. Cloud-metadata IPs blocked regardless of trust (pinned).
- ModelStudio's previously-dead image-host allowlist now enforced through the adopted policy.
- Review round: cross-origin redirect hops now strip `Authorization`/`Cookie`/`Proxy-Authorization` (reusing `egress._hop_headers`) in both `fetch_json` and `fetch_image_bytes`; same-origin redirects keep credentials (regression-guarded) so local backends are unaffected.
- SSRF payload matrix: private ranges, loopback, link-local, metadata, scheme abuse, redirect-to-private — all exercised against production code paths.

Also files backlog task-563 (P3 polish minors) and task-564 (three pre-existing personas/library test failures), worked next in this campaign.

## Testing
- `Tests/Image_Generation/` + `Tests/Utils/test_egress.py`: 124 passed, 6 skipped (opt-in live-backend tests).
- Cross-tree image-gen test sweep over `Tests/`: 0 failures. `ruff check` clean; `python -c "import tldw_chatbook.app"` clean.
- Final whole-branch review confirmed: no unguarded fetch path remains in the package (all adapter HTTP goes through the validated helpers); consumer sweep clean (no imports of the changed helpers outside the package).

---
## Summary by cubic
Adopts the app-wide egress/SSRF policy for image generation and completes P1 polish, with per-hop checks, safe redirects, and trusted-origin threading across adapters so local backends keep working. Implements task-497 and task-498 ACs.

- **New Features**
  - All adapter HTTP and image fetches use `Utils/egress`: private/link‑local/metadata IPs blocked, each redirect hop revalidated. `fetch_json` and image byte fetches strip `Authorization`/cookies on cross‑origin redirects; same‑origin redirects keep credentials.
  - Threaded `trusted_origins` from adapter `base_url`s across `swarmui`, `novita`, `openrouter`, `together`, and `modelstudio`. ModelStudio’s image-host allowlist now runs through the same policy, using the same `trusted_origins` for the allowlist gate and the fetch. `evaluate_url_policy` now delegates to the policy and blocks private/link‑local/metadata before host checks.
  - Added `Tests/Image_Generation/test_live_backends.py`: opt‑in live smoke tests (one per backend). Expanded unit tests cover redirects, scheme abuse, and API‑returned private-IP image URLs.

- **Bug Fixes**
  - `create_client(timeout=0)` now respects an explicit zero timeout.
  - `__getattr__` in `Image_Generation` raises a descriptive `AttributeError`.
  - Cold‑start test restores `sys.modules` after purging to avoid process‑wide side effects.
  - `run_generation` raises a clear error when an adapter fails to load.

<sup>Written for commit f9e7110129d30cf7d6d692fe55e9879898c0bbe9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/862?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:

**Bug 1 (redirect origin check host-only, ignores scheme/port): DECLINED here — filed as task-568 for a central fix.**
The hostname-only comparison is a deliberate reuse of `Utils/egress.py`'s own primitive: `guarded_fetch_httpx` makes its credential-stripping decision via the same `host_of()` (hostname-only) comparison, shipped app-wide in #822. This PR's task was *adopt the central policy*; forking a stricter scheme/port-aware comparison only inside `Image_Generation` would create exactly the policy divergence the task exists to remove, while every other egress consumer (Web_Scraping, Subscriptions) kept the weaker rule. The substantive point is fair — and the app is indeed internally inconsistent (SwarmUI's image-URL gate already compares scheme+host+port) — so the correct fix is upgrading the comparison **centrally** with all consumers and the trusted-origins flows re-verified together: filed as **task-568** (on this branch, `f15eda06e`) with default-port normalization and the downgrade/port-change test matrix spelled out.

**Rule 497145 (path_validation for env-derived paths in the live test): DECLINED.**
`TLDW_LIVE_SD_CPP_BINARY` / `TLDW_LIVE_SD_CPP_MODEL_PATH` are the test *operator's own opt-in configuration* pointing at their own local binary/model, read solely to decide skip-vs-run for an `integration/optional/slow` suite that never runs in CI. `validate_path` exists to confine untrusted user-supplied paths to a trusted base directory; there is no base directory to confine to here (the operator may legitimately keep models anywhere), and treating the operator's own env as untrusted input would just force a bypass parameter. Same rationale as this rule's prior declines on earlier image-gen PRs.

PR #864, opened 2026-07-25
Image-gen follow-ups: P2a Console polish (task-558)

rmusser01:
## Summary

Console-cluster follow-up from the image-gen program: the non-blocking findings deferred from the P2a whole-branch review (#832). One backlog task, SDD-reviewed with one fix round.

### task-558 — Image-gen P2a polish (`3a0fbeb53` + `5a5b7747e`)
- **In-memory restore parity**: `ChatScreen.restore_state`'s tab-switch path now rehydrates `generation_metadata` the same way it rehydrates attachments (same probe/graceful-degradation contract), so a tab-switch-restored generation message keeps its card. The DB-resume path is unchanged.
- **Resolved model on the card**: `ImageGenResult` gained optional resolved-model/seed fields; SwarmUI and sd.cpp adapters capture the model where the backend reports it, threaded into per-variant metadata, and the card renders a conditional "Model" row (fix round). `resolved_seed` is plumbed but deliberately unpopulated — reviewer independently verified against sd.cpp source and SwarmUI API docs that neither backend reliably reports a resolved seed; nothing is fabricated.
- **Draft restore on raise**: a generation batch that raises now restores the composer draft exactly like the zero-success path (same guards; `CancelledError` not swallowed).
- **Stale test narrative** fixed in `test_console_generation_store.py` (comments claimed `restore_persisted_session` doesn't hydrate — false since the resume fix; redundant manual hydrate calls dropped) and the hydration is documented in the docstring.
- **Coverage nits**: generation-vs-sibling action precedence pinned with `sibling_count=3`; exact-80-char content-marker boundary; empty-negative-prompt card branch.
- **Cosmetics**: inert `console-generation-card*` CSS classes resolved with real TCSS rules (bundle regenerated via build_css, `check_bundle_sync` clean); `chat_persistence_service.py` docstring no longer promises `CharactersRAGDBError` for a raw `sqlite3.IntegrityError`, matching sibling convention.

## Testing
- Targeted: `Tests/Chat/test_console_generation_*` + `Tests/Image_Generation/` → 80 passed; `Tests/UI/test_console_native_chat_flow.py` → 200 passed. All foreground.
- `ruff check` clean on touched files; `python -c "import tldw_chatbook.app"` clean; CSS bundle reproducible.
- Every behavioral AC was TDD'd red→green; reviewer independently re-ran suites and verified the tests assert real rendered output.

Sibling PR: #862 (engine cluster, tasks 497+498) — different files except `Image_Generation/adapters/`, where this PR touches only response-parsing regions (no overlap with #862's egress/redirect changes).

---
## Summary by cubic
Polishes the Console image generation flow (task-558): tab‑switch restore now rehydrates generation metadata so cards persist, cards show the resolved model when known, and the composer draft is restored on batch errors. Adds CSS and doc tweaks with new tests that pin behavior and fallback paths.

- **New Features**
  - In‑memory restore hydrates generation metadata so generation cards survive tab switches.
  - `ImageGenResult` adds `resolved_model`/`resolved_seed`; `run_generation_batch` prefers these when present and cleanly falls back when missing.
  - Backfills `resolved_model` in `stable_diffusion_cpp` (filename) and `swarmui` (explicit or configured default); seed remains unset by design.
  - Cards conditionally render a “Model” row when available.

- **Bug Fixes**
  - Restore the typed draft and show a system error if `run_generation_batch` raises; in‑flight guard always releases.
  - Card details omit “Model” when absent and “Negative” when empty; action precedence pins variant navigation over stale sibling fields; exact 80‑char content‑marker boundary asserted.
  - Tab‑switch restore skips stale `messages_by_session` keys during metadata rehydrate; test added to prove no crash.
  - CSS adds real `console-generation-card*` rules and removes an inert class; `chat_persistence_service` docstring stops over‑promising `CharactersRAGDBError` on sidecar writes.

<sup>Written for commit d06e60e9020a45a54a89d7aea8e8c8d4c20b214b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/864?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:

**Bug (restore hydration KeyError): DECLINED — structurally unreachable; pinned with a regression test instead (`6c9b8ecd9`).**
The premise is that `_rehydrate_console_message_generation_metadata` iterates raw payload keys. It doesn't: `_restore_native_console_state` **rebuilds** `restored_messages_by_session` keyed only by the ids of sessions it actually constructs from `sessions` (messages are pulled per-session via `messages_by_session.get(session.id)` — a stale/orphaned `messages_by_session` key is never copied in), and every one of those sessions is registered by the `store.restore_state(...)` call that precedes hydration. So `hydrate_generation_metadata`'s unknown-session lookup cannot be reached from this path. Rather than guard dead code, the commit adds `test_restore_generation_metadata_survives_stale_session_key_in_payload`, which feeds exactly the claimed poison payload (an orphaned `messages_by_session` key) and asserts the restore completes with the known session hydrated — if the construction invariant ever changes, that test goes red.

**Rule 497152 (missing docstrings on two swarmui tests): FIXED** in `6c9b8ecd9` — the two `resolved_model` tests now carry one-line docstrings, matching the other tests added in this PR.

**Rule (underscore-prefixed class names): DECLINED.**
`_ResWithResolved` / `_FakeComposer` are private module-level test doubles; underscore-prefixed helper classes are the established idiom across this test suite (158 files under `Tests/` define `class _...`). PascalCase applies to public classes; these are deliberately non-public.

PR #865, opened 2026-07-25
Image-gen follow-ups: P3 expression-generation polish (task-563)

rmusser01:
## Summary

Personas-cluster follow-up from the image-gen program: the deferred polish minors from the P3 whole-branch review (#859, ✨Generate for avatar + expression slots). One backlog task, SDD-reviewed, approved with zero findings.

### task-563 — P3 expression polish (`ef2c8d9bf`)
- **Narrow-width Expressions row** (pre-existing since the row shipped): the action row is now horizontally scrollable (same idiom as `MainNavigationBar`), so Import/Export/Generate-all stay reachable at 80 and 120 cols; Tab-focus auto-scrolls buttons into view. The old 200-col-only test claim is replaced by an 80/120/200 width-parameterized reachability test — reviewer reverted the fix locally and confirmed the 80-col case genuinely fails without it.
- **In-slot "Generating…" affordance** (spec §1 had promised it): set at all three dispatch sites, cleared in a `finally` that re-checks editor identity — a stale clear from character A cannot clobber character B's in-flight indicator after a mid-generation switch (dedicated leak test).
- **Generate-all overwrite confirmation**: when a staged avatar or populated expression slot exists, Generate-all asks first via the screen's existing `ConfirmationDialog`/`push_screen_wait` idiom; no dialog when everything is empty. The modal is awaited from an async worker, not a thread.
- **Honest k/N summary**: `_apply_expression_upload` now returns whether the DB write actually persisted; the sweep counts only genuine writes, so a slot that errors no longer inflates the "k/4 generated" summary (per-slot error notify preserved).
- **Cosmetics**: comment at `_after_character_save`'s record-reread fallback documenting the style-reset invariant; the generate-all narrow race is now guarded (an `(character_id, "all")` in-flight key blocks duplicate single-slot dispatches during a sweep, symmetric both directions, no starvation).

One disclosed micro-nuance: the sweep worker's pre-loop editor lookup silently aborts on an editor-unmount race instead of the old misleading "0/4 generated" notify — reviewer assessed the window as vanishingly narrow and the new behavior no worse.

## Testing
- `Tests/UI/test_personas_expression_generate.py` + `test_personas_expression_slots.py`: 69 passed (baseline 52). All TDD'd red→green; all foreground.
- Broader personas UI sweep: only the three known pre-existing failures (task-564's scope), reproduced identically on the base commit.
- `ruff` byte-identical to base on touched files; `python -c "import tldw_chatbook.app"` clean; no TCSS bundle impact (grep-verified).

Also carries the task-563/564 backlog filings (duplicated from #862 so merge order doesn't matter; rebase-before-merge collapses the duplicate add).

Sibling PRs: #862 (engine: 497+498), #864 (Console: 558). All three are independent against dev.

---
## Summary by cubic
Polishes Personas image generation UX: the Expressions row works at narrow widths, shows per-slot “Generating…” hints, confirms overwrite for Generate all, fixes the k/N summary, and blocks redundant clicks. Also forces confirmation when overwrite status can’t be read, suppresses the k/N summary on user cancel, and notifies if the dialog can’t be shown. Satisfies Linear task-563 (AC1–AC5).

- **New Features**
  - Expressions action row is horizontally scrollable at 80/120/200 cols; Tab focus auto-scrolls buttons into view.
  - In-slot “Generating…” indicator for avatar and each expression; set on dispatch and cleared on success/failure without leaking across character switches.
  - Generate all asks before overwriting a staged avatar or populated expression slots; no dialog when everything is empty.

- **Bug Fixes**
  - Accurate k/N summary: `_apply_expression_upload` returns `bool`; the sweep counts only persisted writes.
  - Guard narrow race: refuse single-slot or avatar generate while `(character_id, "all")` is in-flight to prevent duplicates.
  - Consent gate hardens: unreadable overwrite state forces confirmation; user-declined confirmation aborts with no writes and no summary; dialog failure cancels the sweep and notifies the user.

<sup>Written for commit 81879296faea010cfc54998950678f462697fc47. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/865?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_personas_expression_generate.py
+++ b/Tests/UI/test_personas_expression_generate.py
@@ -1129,6 +1129,77 @@ async def test_generate_all_worker_declining_confirmation_aborts_with_no_writes(
     assert not any("generated" in msg.lower() for msg, _sev in notifications)
 
 
+async def test_generate_all_worker_unreadable_expression_state_fails_closed_into_confirmation(
+    personas_editor_with_saved_character, monkeypatch
+):
+    """A DB read failure while checking for existing images must fail CLOSED:
+    when overwrite status is unknown, the confirmation dialog is forced
+    rather than silently skipped (Qodo PR #865 — consent gate must not fail
+    open on the exception path)."""
+    app, screen, db, char_id = personas_editor_with_saved_character
+    _configure_backend(monkeypatch)
+    _set_description(screen, "A cheerful adventurer.")
+
+    def _raise(*args, **kwargs):
+        raise RuntimeError("db unavailable")
+
+    monkeypatch.setattr(db, "get_character_expression_image", _raise)
+
+    push_calls: list = []
+
+    async def _fake_push_screen_wait(dialog):
+        push_calls.append(dialog)
+        return False  # Cancel — sweep must abort without writes
+
+    monkeypatch.setattr(app, "push_screen_wait", _fake_push_screen_wait)
+    run_calls: list = []
+    monkeypatch.setattr(
+        personas_screen_module,
+        "run_generation",
+        lambda request: run_calls.append(request),
+    )
+    screen._expression_generate_inflight.add((char_id, "all"))
+
+    await screen._generate_all_expression_images_worker(char_id)
+
+    assert len(push_calls) == 1  # unknown overwrite state ⇒ dialog forced
+    assert run_calls == []
+
+
+async def test_generate_all_worker_confirm_dialog_failure_notifies_user(
+    personas_editor_with_saved_character, monkeypatch
+):
+    """When the overwrite-confirmation dialog itself fails to show, the
+    aborted sweep must tell the user instead of silently doing nothing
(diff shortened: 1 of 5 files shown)


rmusser01:
Both Qodo findings confirmed real and fixed in `a141499f3` (TDD, both red first): the overwrite check now **fails closed** — an unreadable expression-image state forces the confirmation dialog instead of silently counting as no-image — and a confirmation dialog that cannot be shown now notifies the user that Generate-all was cancelled instead of aborting with only a log line. Expression suites 56 + 15 green.

PR #866, opened 2026-07-25
Fix three pre-existing personas/library test failures (task-564)

rmusser01:
## Summary

Kills the three known-red tests that every branch since the image-gen P3 sweeps has had to re-attribute as baseline. Each was root-caused to the commit that drifted it; two turned out to be genuine product bugs, one a stale test stub. SDD-reviewed, approved with zero findings.

### task-564 (`8fd1d02c9`, `92a633752`, `8fbc4bd23`, `6221d35fa`)

1. **`test_import_offpage_name_conflict_message`** (library-scale) — test-side drift: task-445 (`5eb878cfda`) deliberately added `timeout=6.0` to `_notify` call sites and updated the workbench tests but missed this file's stub. Fix: stub signature now matches the real `_notify` exactly.
2. **`test_export_json_rejects_hidden_directory_destination`** — genuine product bug since inception (`ac995f72b`): `validate_path(target, base_directory=target.parent)` never checked whether the immediate parent itself is a hidden directory. Fixed at root in `Utils/path_validation.py` (also closes the identical hole in `Home/active_work_adapter.py`). The check is deliberately narrow — it rejects only when `base_directory`'s **own basename** is dotted, never walking ancestors, so legitimate app paths under `~/.config`/`~/.local` are unaffected; reviewer swept every caller in the repo to confirm none passes a dotted basename.
3. **`test_character_book_errors_render_in_editor_footer`** — product regression from `414183488`, which loosened `validate_character_book` to return `ok=False` only for non-dict input (everything else became `(True, [warnings])`) but left the editor footer gated on `if not ok:`, silencing the warnings it still reports. Fix: gate on `if book_errors:` — renders exactly what the lenient validator reports, resurrecting nothing it deliberately silenced. The fix hunk (lines ~6364-6371) has a ~125-line gap to the nearest hunk of sibling PR #865 in the same file — verified conflict-free in either merge order.

## Testing
- `test_personas_library_scale.py` + `test_personas_workbench.py`: 211 passed, 0 failed (previously 3 known-red).
- `-k path_validation`: 25 passed; caller-adjacent suites (Tools/Home/Character_Chat integration): 94 passed.
- `ruff` clean on touched files (2 pre-existing unrelated F821s in personas_screen.py confirmed identical on base); `python -c "import tldw_chatbook.app"` clean.

Sibling PRs: #862 (engine), #864 (Console), #865 (personas) — all independent against dev.

---
## Summary by cubic
Fixes three pre-existing personas/library test failures by aligning a test stub, tightening `validate_path` to reject dotted base directories, and restoring character book error rendering (Linear: task-564). Adds regression tests to pin the new path rule.

- **Bug Fixes**
  - Path validation: `validate_path` now rejects destinations whose `base_directory` name is dotted, preventing exports into hidden directories without affecting `~/.local` and similar paths.
  - Personas editor: `_validate_character` now gates on `book_errors` instead of the `ok` flag so character book issues show in the footer and block Save.

- **Tests**
  - Updated `_notify` stub in `test_personas_library_scale.py` to accept the `timeout` kwarg, matching the real signature.
  - Added `test_hidden_base_directory_blocked` and `test_base_directory_under_hidden_ancestor_allowed` in `Tests/Utils/test_path_validation.py` to pin the hidden base-directory rule.

<sup>Written for commit dd2ede0ee3542511303ce2496d93b82378e6ccd5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/866?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Qodo's finding (no direct unit regression test for the new hidden-`base_directory` branch) confirmed and fixed in `e64f52fd7`: `Tests/Utils/test_path_validation.py` now pins both directions — a dotted base name raises, and a base directory merely living **under** a dotted ancestor (app-owned roots like `~/.local/share`) stays valid. 14 passed.

PR #867, opened 2026-07-25
Image-gen follow-ups: P2b features (task-559) — user styles, TTS stop, LLM context

rmusser01:
## Summary

Final cluster of the image-gen follow-ups campaign: the feature-sized P2b deferred items (task-559), built in three per-unit-reviewed slices plus a whole-branch final review (merge-ready; zero Critical/Important).

### Unit 1 — user-defined style templates + picker previews (`3923afe28`)
- User templates load from `<user_data_dir>/image_generation_styles/*.toml` (filename stem = id) and `[image_generation.styles.<id>]` config sections, merged over the 13 built-ins (directory > config section > built-in on id collision). Malformed templates skip with a logged warning — the picker and `@style` resolver never crash on untrusted input.
- Every consumer now reads the merged registry (`get_all_templates()` single cache source); this also fixed a latent bug where the SwarmUI widget resolved against the raw built-in dict its own dropdown no longer matched.
- The style picker gained a detail pane showing the base-prompt/negative snippet for the highlighted template (90-char truncation, `markup=False` — a template containing `[red]` renders literally, test-pinned).

### Unit 2 — Console TTS stop (`c943d8dbe` + `90fcae316`)
- Per-message 🔊→⏹ toggle posting the exact `TTSPlaybackEvent(action="stop")` legacy widgets use — no parallel audio plumbing. Pause/save deliberately skipped (pause is a no-op stub app-side; save races the 5s auto-cleanup) and documented.
- Two real pre-existing bugs fixed in the shared handler: stop never actually silenced the system audio player (only deleted the cached file), and — caught by review with a reproduction — the play path's unconditional 5-second cache cleanup made stop blind for any clip longer than 5s while the UI claimed "Stopped speaking." Stop identity now lives in a `_last_played_audio_files` map immune to the cleanup, guarded by the player's actual current file so stopping message A can never kill message B's audio. Legacy chat's stop button is genuinely fixed by the same change.
- Honest limitation (no playback-end event exists in the stack): the ⏹ reverts on stop/new-speak, not on natural end; a stale ⏹ click is a safe no-op and idle stop no longer shows a misleading notify.

### Unit 3 — LLM-composed generation context (`323f30081` + `4bb2cd823`)
- `/generate-image` with no prompt now composes the prompt from the last N conversation turns via the session's active Console provider (`chat_api_call`, single non-streaming call in a dedicated executor with a real `future.result(timeout=…)` bound, inside the existing off-loop generation chain).
- Fallback is the load-bearing design: kill-switch off, provider missing, call raising, timeout, or empty/garbage response all land silently on the existing keyword extractor — generation always proceeds, at most a debug log (no conversation content logged). Reviewer traced every failure edge and confirmed nothing escapes.
- Config: `[image_generation]` `context_llm_enabled` (default true), `context_llm_turns` (10), `context_llm_timeout_seconds` (15), documented in the shipped default TOML. Composed prompts remain visible in the generation card; `@style` templates apply in a single pass (no double composition).

## Testing
- Final-review totals (all foreground): style/generate/picker suites 232; `Tests/TTS/` + transcript 814 passed/14 skipped; Console native chat flow 199; broader `Tests/Chat/` 2141 passed/69 skipped. `ruff` clean (2 pre-existing F841s at merge-base); `import tldw_chatbook.app` clean; CSS bundle sync clean.
- Review rounds: unit 2's Critical (stop-blind-after-5s) was found by adversarial review, reproduced, fixed with the repro as a red test; units 1 and 3 approved with only cosmetic/doc minors, all addressed.

## Merge note
Overlaps sibling PR #864 (`console_generate_image.py`, `chat_screen.py`, TCSS, 4 test files) — whichever merges second should rebase and rerun. #862/#865/#866 share no files with this branch.

---
## Summary by cubic
Adds user-defined image styles, a speak→stop TTS toggle, and LLM-built context for prompt-less `/generate-image`. Completes task-559 P2b features and improves Console image generation and TTS control.

- **New Features**
  - User style templates: load from `[image_generation.styles.<id>]` and `<user_data_dir>/image_generation_styles/*.toml`, merged over built-ins; malformed entries are skipped.
  - Style picker: shows a brief prompt/negative preview for the highlighted template; all consumers now use a single merged registry.
  - LLM context: when no prompt is provided, the app composes one from recent turns via the active provider; controlled by `[image_generation]` `context_llm_enabled`, `context_llm_turns`, `context_llm_timeout_seconds` with safe fallback; composed prompt stays visible and `@style` applies once.
  - Console TTS: per-message 🔊 swaps to a ⏹ action that posts `TTSPlaybackEvent(action="stop")`; pause/save intentionally out of scope.

- **Bug Fixes**
  - Stop now actually interrupts audio, survives the 5s cache cleanup, and won’t silence unrelated playback. Uses a single-slot last-played tracker plus a player-file identity check.
  - LLM context executor is now a bounded, shared single-worker pool to prevent thread leaks; if already occupied, requests fail fast and fall back to the keyword extractor.
  - SwarmUI sidebar now resolves templates from the merged registry, fixing mismatches with the dropdown.

<sup>Written for commit eca493bd1ea48229824cde18b985eb61343f4846. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/867?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:

**Bug 1 (orphaned LLM timeout threads): FIXED** in `5cc73f184` per the suggested direction — the per-call `ThreadPoolExecutor` is replaced by one shared module-level single-worker executor with a non-blocking in-flight guard: while a timed-out call is still hanging, a new context attempt fails fast into the keyword-extractor fallback instead of queueing behind it. The abandoned call still can't be cancelled (inherent without provider-level network timeouts — noted in the docstring), but accumulation is now bounded at exactly one thread, with a red→green saturation test plus a test-only executor reset for cross-file isolation.

**Bug 2 (`_last_played_audio_files` unbounded growth): FIXED** in `c68aa0c2f` — replaced the dict with a single `(message_id, path)` slot overwritten on every play, matching `SimpleAudioPlayer`'s single-slot reality. Stop semantics unchanged (all four pre-existing stop tests pass untouched); a no-growth test pins the bound.

**Rule 497144 (key-like test literals): FIXED** in `5cc73f184` — `sk-test`/`sk-x` replaced with clearly-fake non-`sk-` strings in both test files.

**Rule 497152 (missing `Args:` on `stop_audio_playback_if_current`): FIXED** in `c68aa0c2f`.

**Rule 497145 (path_validation for template-dir files): DECLINED.** The loader never consumes a user-supplied path *string* — it enumerates `*.toml` inside the app-owned user-data directory it just resolved, the same trust class as the operator's own `config.toml`. This is the established repo convention for exactly this pattern: `Chat_Dictionary_Lib` globs `*.md`/`*.yaml` from the chat_dicts folder, `RAG_Search/config_profiles.py` globs `*.json` from the profiles dir, and `Utils/Splash_Screens` globs `*.py` — all opened directly. A symlink planted there is the local operator linking to a file they can already read; there is no privilege boundary for `validate_path` to defend in a local TUI, and template *ids* (the only part that flows onward) are already regex-validated.

**Rule 1011859 (Pydantic for template validation): DECLINED.** `_coerce_generation_template`'s contract is *never raises* — malformed entries skip with a warning so a bad user file can't take down the picker or the `@style` resolver. That skip-with-warning coercion guard is the repo's standing idiom for untrusted-import data (the Roleplay lore/world-book importers use the same `_coerce` pattern); a Pydantic model raising `ValidationError` would need the identical try/except wrapper around every entry, adding a dependency surface without changing behavior.

**Rule 497158 (streaming=False): DECLINED.** The streaming requirement targets user-visible chat responses. This call is an internal composition step whose short single-paragraph output is consumed whole before generation dispatches — there is no UI surface to stream to; the user-visible artifact is the composed prompt on the generation card, which appears exactly as before.

PR #868, opened 2026-07-25
fix(agents): agent-runtime durability hardening (TASK-327)

rmusser01:
Closes **TASK-327** — the final task of the LLM-harness review 2026-07 (320-334). Five bounded durability fixes to the agent runtime, one per AC.

## What changed

| AC | Fix |
|----|-----|
| **#1** | **Cyclic loop detection.** The one-slot `last_key`/`repeat_count` guard only caught *consecutive* identical calls, so `A→B→A→B` never tripped. Replaced with a loop-local `deque` of call-keys plus a pure `_detect_cycle()`: periods 1..`MAX_LOOP_PERIOD`(=4), smallest-period-first, trip threshold 3 for period 1 (backward-compatible) and 2 full repeats for periods ≥2. |
| **#2** | **Orphaned-run reconcile.** A crash between `create_run` (`running`) and the run-end `set_status` left rows stuck `running` forever. `AgentRunsDB.reconcile_orphaned_runs()` flips them to `error` with `COALESCE(result, 'Interrupted by app restart')`, once per DB file per process, file-backed only. |
| **#3** | **WAL + busy_timeout** on `AgentRunsDB` connections (`base_db.py` untouched — it's shared by every other DB). |
| **#4** | **Per-tool-call timeout.** `RunBudget.max_tool_call_seconds` (default `300.0`, `0` = unlimited), enforced by a module-level `_call_with_timeout` in the impure `agent_service` seam so the pure runtime stays pure. |
| **#5** | **Doc-only:** recorded `ConsoleChatController` as the sole run-serialization point; the bridge's `_live`/`_historical_cache` are display state, not a guard. |

## Two things review caught that are worth calling out

**The design spec was wrong about MCP.** It claimed MCP tools were exempt from the new timeout because they self-time-out. They are not — `MCPToolProvider` is registered into the *same* per-run `ToolCatalogRegistry` as builtins, so MCP calls flow through the wrapper like any other tool. Worse, the originally-specified 120s default collided with MCP's own ~186s worst case (121s approval wait + 65s execution): an "Ask"-gated tool the user approved after ~2 minutes would be reported to the agent as timed-out while the abandoned worker still executed it for real — and an agent retry could execute it twice on one approval. The default is now `300.0`, deliberately above MCP's worst case at default config. The spec and plan carry a `CORRECTION` block rather than a silent edit.

**The new loop-detection message dropped the tool name.** `f"loop detected: {period}-cycle repeated {repeats}x"` is not internal — `ConsoleChatController._agent_failure_visible_copy` renders it verbatim as `Agent run stuck: …`. Users would have gone from *"loop detected: calculator repeated 3x with identical args"* to *"loop detected: 1-cycle repeated 3x"*. The summary now names the offending tools.

Also folded in from review: the timeout's `join` now polls cancellation every 0.5s (a single blocking join left **Stop** dead for up to 300s, defeating AC#4's own purpose); `_call_with_timeout` catches `BaseException` (an `asyncio.CancelledError` from `BuiltinToolProvider`'s `asyncio.run` previously escaped as a `KeyError` into the pure loop); `_swept_paths` registers only *after* a successful commit, so a transient `database is locked` retries instead of silently disabling reconciliation for the whole process; `busy_timeout` is set before the WAL conversion (the one PRAGMA that can contend).

## Deliberate non-changes

- `max_tool_call_seconds` is **not** clamped to the parent's remaining wall-clock in `clamp_child_budget` (unlike `max_wall_seconds`) — clamping it down makes premature timeouts more frequent, which is exactly the trigger for the abandoned-thread double-execution hazard. The parent's `max_wall_seconds` already bounds the run.
- No `force=` on `reconcile_orphaned_runs`; no `base_db.py` change.

## Testing

`Tests/Agents/ Tests/DB/test_agent_runs_db.py` → **278 passed**, 3 warnings (all pre-existing: a `requests`/`urllib3` version mismatch, and a "coroutine was never awaited" from `test_mcp_tool_provider.py`, a file this branch doesn't touch). No leaked-thread warnings despite the new daemon threads.

Built with subagent-driven development: 5 TDD tasks, a per-task spec+quality review each, and an adversarial whole-branch review plus scoped re-review at the end.

---
## Summary by cubic
Hardens the agent runtime for durability by catching cyclic tool-call loops, bounding each tool call with a timeout, and making DB operations more resilient. Aligns with TASK-327.

- **Bug Fixes**
  - Loop detection: detects cycles up to period 4 (e.g., A→B→A→B) and stops runs with an error that names the looping tool(s).
  - Per-tool timeout: adds `RunBudget.max_tool_call_seconds` (default 300s, 0 = unlimited) and enforces it in `agent_service` with a per-call daemon thread that polls cancellation; MCP tools go through this wrapper; skill calls bypass it.
  - DB resilience: `AgentRunsDB` reconciles crash-orphaned `running` rows to `error` on open (once per file per process; keeps existing `result`; skips `:memory:`), and enables WAL with `busy_timeout=5000` (timeout set before WAL).
  - Tool catalog: guards both caches to avoid a partial-reset race under per-call threads.
  - Docs: clarify `ConsoleChatController` is the sole run-serialization point; the bridge’s `_live`/`_historical_cache` are display state only.

<sup>Written for commit c320114feea915ad6697580a06a216eed32ffe95. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/868?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Thanks — both findings reviewed. **Declining both, with evidence.**

---

### 1. "PRAGMA outside `transaction()`" (Rule violation) — declined, the suggested fix is impossible *and* would break the feature

Three independent reasons:

**(a) It would be infinite recursion.** `transaction()` is *built on* `_get_connection()` — `AgentRuns_DB.py:94` is `with closing(self._get_connection()) as conn:`. Wrapping `_get_connection`'s own setup PRAGMAs in `transaction()` would have `_get_connection` call `transaction()` call `_get_connection`, forever.

**(b) SQLite refuses `journal_mode=WAL` inside a transaction.** `transaction()` opens with `BEGIN IMMEDIATE` (`AgentRuns_DB.py:95`). Verified empirically:

```python
conn.execute("BEGIN IMMEDIATE")
conn.execute("PRAGMA journal_mode = WAL")
# sqlite3.OperationalError: cannot change into wal mode from within a transaction
```

So applying this suggestion would not just be stylistically different — it would raise on every connection and take the class down.

**(c) It is the established pattern here, and the line it flags as new is adjacent to an identical pre-existing one.** `conn.execute("PRAGMA foreign_keys = ON")` sits one line above at `AgentRuns_DB.py:48`, untouched by this PR, doing exactly the same thing. Every sibling DB class in the codebase (ChaChaNotes, Client_Media, Evals, Prompts, Library_Ingest_Jobs, Kanban) likewise sets WAL during connection setup.

The rule ("wrap all database operations in `transaction()`") is about *data operations*. Connection-level PRAGMAs are connection construction — they are what `transaction()` runs on top of, not something it can contain.

---

### 2. "Wall budget bypassed" (Bug) — declined; known, documented, and deliberately not clamped

The observation is accurate: `run_agent_loop` checks `max_wall_seconds` between loop iterations, so a slow tool can overrun the wall budget by up to one tool timeout. But:

**This is pre-existing and strictly improved by this PR, not introduced by it.** Before this change there was *no* per-tool bound at all — `BuiltinToolProvider.invoke` did a bare `asyncio.run(tool.execute(**args))`, so a hung provider blocked the run **forever**. The PR replaces "unbounded" with "bounded at 300s, with cancellation polled every 0.5s". The wall-check-only-between-iterations behavior is unchanged and is documented in the design spec's AC#4 gap analysis.

**The 300 > 240 comparison uses a default that the real caller overrides.** The Console path — the only shipped agent caller — uses `CONSOLE_RUN_BUDGET = RunBudget(max_steps=32, max_wall_seconds=480.0, max_model_turns=8)` (`console_agent_bridge.py:104`). In the deployed configuration the tool timeout (300s) is *below* the wall budget (480s), so it fires first.

**Clamping to remaining wall-clock was explicitly considered and rejected — it makes things worse.** Our own final review raised this and it was adjudicated with a recorded ruling: clamping the per-call timeout *down* makes premature timeouts **more frequent**, and a premature timeout is exactly the trigger for the abandoned-thread double-execution hazard this PR's default is designed to avoid. Because Python cannot kill a thread, a timed-out call keeps running; for a side-effecting "Ask"-gated MCP tool the user approves late, that means the agent is told "timed out" while the tool really executes — and an agent retry executes it twice on one approval. A sub-agent with 1s of parent budget left would, under the suggested clamp, get a 1s tool timeout and hit that window routinely.

The `max_wall_seconds` bound already terminates the run at the next iteration; user-initiated **Stop** is now honored within ~0.5s during a hung call (this PR added the cancellation polling). The ruling and its reasoning are recorded in the design spec's CORRECTION block and in `clamp_child_budget`'s neighbourhood, so a future reader doesn't re-derive it.

PR #869, opened 2026-07-25
feat(agents): raise per-message tool-calling round cap 8 -> 20

rmusser01:
Direct user request: raise the per-message tool-calling limit from 8 to 20. No backlog task filed.

## What "8" actually was

There is no per-turn tool-call cap in the codebase. Three things default to 8, and only one of them produces the observed behavior:

- **`RunBudget.max_model_turns = 8`** — a Console run alternates model → tool → model → tool, so this cap is what yields ~8 tool calls per user message. `console_agent_bridge`'s own comment calls it "this run's **PRIMARY** limiter". **This is the one raised.**
- `RunBudget.max_steps = 8` — the engine default, but Console overrides it to 32, so changing it alone would not affect the app.
- `RunBudget.max_active_tools = 8` — how many tool *schemas* are disclosed to the model (`schemas[:budget.max_active_tools]`), not a call limit.

## Why three numbers changed, not one

Raising the turn cap alone just moves the wall to whichever of the other Console caps binds first, so they are sized together:

| Cap | Was | Now | Why |
|---|---|---|---|
| `max_model_turns` | 8 | **20** | the requested change — ~20 tool-calling rounds per message |
| `max_steps` | 32 | **64** | a fence round costs 3 steps (`STEP_MODEL` + `STEP_TOOL_CALL` + `STEP_TOOL_RESULT`), so 19 rounds + wrap-up = **58**. At 32 the step check fired around round 10 and the 20-turn cap would never have been reachable. 64 clears 58 while staying a real backstop — a native multi-call batch costs `1 + 2N` steps per turn, so heavy parallel batches can still hit it, which is the backstop doing its job. |
| `max_wall_seconds` | 480 | **1200** | 480 was derived as "25-50s/turn × up to 8 model turns" at the slow local-model pace this gate exercises; 20 turns at that pace needs ~500-1000s, so 480 would have become the new binding limit around turn 10. A backstop, not a target — fast cloud models finish well inside it, and Stop is honored within ~0.5s (task-327's cancellation polling). |

## Engine defaults are unchanged

`RunBudget()` keeps `max_steps=8`, so `max_model_turns=20` stays **provably unreachable** at engine defaults (every model turn appends ≥1 step, so with `max_model_turns >= max_steps` the step check always fires first). The raised cap bites only for callers that raise `max_steps` to match — today, only the Console bridge.

## Tests

`test_run_budget_default_model_turns_equal_steps_and_child_clamp_carries` pinned `max_model_turns == max_steps == 8` as "the unreachability invariant". Equality was never the property that mattered — `>=` is — so it is renamed and re-asserts the real invariant plus the new concrete values.

Two guards added, because the failure mode here is a future edit raising one cap without the others:
- `test_console_budget_step_cap_admits_a_full_model_turn_run` — asserts `max_steps >= 3*(max_model_turns-1)+1`. Verified discriminating: the half-change (turns→20, steps left at 32) fails it.
- `test_console_budget_reaches_its_model_turn_cap_before_step_cap` — drives a real run to the cap and asserts it stops on `"model-turn budget exhausted"`, not the step budget.

`Tests/Agents/` → **253 passed**, 3 pre-existing warnings.

---
## Summary by cubic
Raised the per-message tool-calling round cap in the Console from ~8 to ~20 by making the model-turn budget the primary limiter and sizing step/time backstops to match. Engine-default behavior is unchanged.

- **New Features**
  - `CONSOLE_RUN_BUDGET`: `max_model_turns` 8→20, `max_steps` 32→64, `max_wall_seconds` 480→1200.
  - `RunBudget.max_model_turns` default 8→20; still unreachable at engine defaults (`max_steps` stays 8). Tests ensure `max_model_turns >= max_steps`, the step cap admits a full 20-turn run, and real runs stop on the model-turn budget first.

- **Refactors**
  - Replaced literals with named Console constants: `CONSOLE_MAX_MODEL_TURNS`, `CONSOLE_MAX_STEPS`, `CONSOLE_MAX_WALL_SECONDS`; added `DEFAULT_MAX_MODEL_TURNS` in `agent_models`.
  - Documented sub-agent policy: children inherit `max_model_turns` and `max_steps` (no clamping); lifetime is bounded by the parent’s wall-clock clamp.

<sup>Written for commit 8ded47a0bb31a22255997d2e67b6b18d44d3a410. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/869?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Agents/test_agent_models.py
+++ b/Tests/Agents/test_agent_models.py
@@ -72,18 +72,21 @@ def test_budget_defaults():
     ) == (8, 240.0, 2, 8, 4000)
 
 
-def test_run_budget_default_model_turns_equal_steps_and_child_clamp_carries():
+def test_run_budget_default_model_turns_unreachable_and_child_clamp_carries():
     """Pin the engine-default unreachability invariant and child passthrough.
 
-    At ``RunBudget()`` the model-turn cap must equal ``max_steps`` (so the
-    new check can never fire before the step check), and
+    At ``RunBudget()`` the model-turn cap must be at least ``max_steps`` (so
+    the model-turn check can never fire before the step check), and
     ``clamp_child_budget`` must carry ``max_model_turns`` through unclamped.
     """
     b = RunBudget()
     # Unreachability invariant: each model turn appends >=1 step, so with
-    # max_model_turns == max_steps the step check always fires first (or
-    # ties) at defaults -> engine-default behavior byte-identical.
-    assert b.max_model_turns == b.max_steps == 8
+    # max_model_turns >= max_steps the step check always fires first (or
+    # ties) at defaults -> engine-default behavior unchanged. The cap was
+    # raised to 20 for callers that raise max_steps to match (the Console
+    # bridge); it must never drop below max_steps here.
+    assert b.max_model_turns >= b.max_steps
+    assert (b.max_model_turns, b.max_steps) == (20, 8)
     child = clamp_child_budget(
         RunBudget(max_model_turns=3), parent_remaining_seconds=30.0
     )
--- a/Tests/Agents/test_agent_runtime.py
+++ b/Tests/Agents/test_agent_runtime.py
@@ -252,6 +252,49 @@ def test_console_budget_floor_plus_two_extra_rounds_completes():
     assert len(out.steps) == 16
 
 
+def test_console_budget_step_cap_admits_a_full_model_turn_run():
+    """The Console budget's three caps must be sized together.
+
+    ``max_model_turns`` is meant to be the PRIMARY limiter, so ``max_steps``
+    must be large enough for a full run of fence tool rounds to reach it.
+    A fence round costs 3 steps (STEP_MODEL + STEP_TOOL_CALL +
+    STEP_TOOL_RESULT) and the wrap-up reply costs 1 more, so N model turns
+    need 3*(N-1)+1 steps. Raising the turn cap without raising the step cap
+    would silently move the wall back to the step check.
+    """
+    from tldw_chatbook.Chat.console_agent_bridge import CONSOLE_RUN_BUDGET
+
+    turns = CONSOLE_RUN_BUDGET.max_model_turns
+    assert CONSOLE_RUN_BUDGET.max_steps >= 3 * (turns - 1) + 1
+
+
+def test_console_budget_reaches_its_model_turn_cap_before_step_cap():
+    """Drive a real run to the Console turn cap: it must stop on the
+    model-turn budget, not the step budget."""
+    from tldw_chatbook.Chat.console_agent_bridge import CONSOLE_RUN_BUDGET
+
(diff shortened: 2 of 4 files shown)


rmusser01:
Thanks — addressed in `8ded47a0b`. Two fixed, two declined with reasoning.

### 1 & 2. Magic numbers → named constants — **fixed**

Fair, and useful here specifically because these three values *must* stay coordinated: raising the turn cap without raising the step cap silently moves the wall back to the step check. Now `CONSOLE_MAX_MODEL_TURNS` / `CONSOLE_MAX_STEPS` / `CONSOLE_MAX_WALL_SECONDS` in `console_agent_bridge.py`, each carrying the derivation (a fence round costs 3 steps, so N turns need `3*(N-1)+1` = 58 at N=20; 64 clears it while staying a real backstop for native multi-call batches), plus `DEFAULT_MAX_MODEL_TURNS` in `agent_models.py`. `test_console_budget_step_cap_admits_a_full_model_turn_run` already fails if the relationship is ever broken.

### 3. "Subagent budget unexpectedly high" — **declined (operator decision), now documented**

The arithmetic is correct and worth stating plainly: `clamp_child_budget` clamps wall-clock but preserves turns and steps, so worst case per user message goes **24 → 60** provider turns (parent 20 + 2 children × 20). I raised this explicitly with the repo owner rather than deciding it silently, and the decision was to let children inherit the raised cap.

Two things bound it: children are depth-1 only (`max_subagents` is zeroed for them), and the whole tree is bounded in *time* by the parent's wall-clock clamp — a child can never outlive its parent's remaining budget. It is not bounded in spend, which is the accepted trade.

`clamp_child_budget`'s docstring now records the decision, the worst-case arithmetic, and an explicit "do not 'fix' this by clamping turns without checking that decision" — so the next reviewer doesn't re-derive it.

### 4. "Default turn cap changed" — **declined with evidence**

Accurate in the abstract, but not reachable in this repo. Only a caller that raises `max_steps` above 8 is affected, and the sole production construction sites of `RunBudget` are `CONSOLE_RUN_BUDGET` (which raises both, deliberately sized together) and `clamp_child_budget` (which copies the child's own values). Verified with `grep -rn "RunBudget(" tldw_chatbook/ | grep -v test` — no other caller exists. Pure engine defaults still hit `max_steps=8` first, so `max_model_turns` remains provably unreachable there; the raised default only takes effect for a caller that opts in by raising the step cap to match.

`pytest Tests/Agents/` → 263 passed.

PR #872, opened 2026-07-25
Console branching follow-ups, batch 3 (tasks 573, 543, 572)

rmusser01:
Third batch of Console branching follow-ups, implementing the three remaining To-Do tasks in the stream (IDs per the #803 collision rework).

## task-573 — carry attachments across Edit & resend
Edit & resend silently re-sent text-only when the anchor user message carried an image. The resend now applies the same vision gate as a fresh send BEFORE any node is created; builds the payload by running ancestors + a synthesized `ConsoleChatMessage` (carrying the anchor's attachments) through the one shared `_provider_message_payloads` pass (image budget / vision degradation / mime fallback identical to `submit_draft`); and forks the sibling with the attachments via `create_sibling`'s new `attachments=` parameter. Plain in-place Save unchanged.

## task-543 — record run id for stopped-via-cancel agent runs
The dominant user-Stop path (`stop_active_run` → `task.cancel()`) raised `CancelledError` before `run_reply`'s `(run_id, outcome)` ever bound, so the stopped reply's persisted id was never recorded and the run fell to the ordinal marker fallback on resume. The CancelledError branch now recovers the active run via the bridge's new `latest_unanchored_primary_run_id` seam — newest non-superseded primary, returned only while its `assistant_message_id` is still NULL, so a Stop that beats `create_run` can never overwrite the previous run's good anchor — and records the persisted id exactly like the finalizer terminal paths. New tests exercise the REAL `task.cancel()` path via a parked bridge thread (every prior stop test simulated the stop through a normally-returning `run_reply`).

## task-572 — stronger legacy-flat fingerprint for resume root-chaining
A degenerate all-USER legacy conversation (2+ parentless user rows, no replies) was mistaken for a genuine Phase-B root fork and left un-chained — resuming truncated to the last row with a phantom sibling counter. The all-USER decision now also chains when every root is childless (a genuine first-message edit-&-resend fork always hangs at least one reply subtree under a root); any root ASSISTANT row keeps chaining as the true legacy signature. The narrower residual edge is documented in the method docstring.

## Verification
- TDD throughout (red confirmed before each fix).
- 500 tests green across the edit-resend / agent-swap / controller / store / stop-reliability / resume-active-path suites plus the full `test_console_native_chat_flow.py`.

---
## Summary by cubic
Fixes three console branching gaps: Edit & resend now carries image attachments with the same vision gating as new sends; canceled runs record the stopped reply’s persisted id; resume chains degenerate all-USER legacy roots to remove phantom counters. Covers tasks 573, 543, and 572.

- **Bug Fixes**
  - Edit & resend (task-573): carries the anchor’s attachments onto the new sibling and into the provider payload; builds ancestors + a synthesized `ConsoleChatMessage` through the shared `_provider_message_payloads`; applies the same vision gate before any mutation; `create_sibling(attachments=...)` persists attachments; text-only anchors keep string payloads.
  - Stop via cancel (task-543): on `CancelledError`, recovers the active run via the bridge’s `latest_unanchored_primary_run_id` (DB’s bounded `latest_primary_run`) and records the stopped reply’s persisted id; anchored newest rows are guarded; never-persisted stops leave the run NULL (ordinal fallback); tests exercise real `stop_active_run` → `task.cancel()`.
  - Resume root-chaining (task-572): chains legacy roots when any root is ASSISTANT or when all USER roots are childless; prevents phantom sibling counters and truncation; genuine first-message forks with any subtree remain un-chained; residual edge documented.

<sup>Written for commit 5c74fd79c5f9edaefbff3df72756647c605d86d0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/872?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Chat/test_console_agent_swap.py
+++ b/Tests/Chat/test_console_agent_swap.py
@@ -1415,3 +1415,135 @@ async def test_mcp_review_hook_raise_fails_open_but_invoke_gate_still_refuses(tm
     assert (
         service.execute_calls == []
     )  # never invoked -- refused by invoke()'s own gate
+
+
+@pytest.mark.asyncio
+async def test_stopped_via_cancel_records_persisted_id_on_run(tmp_path):
+    """task-543: the dominant user-Stop path (``stop_active_run`` ->
+    ``task.cancel()``) raises ``CancelledError`` in the controller BEFORE
+    ``run_reply``'s ``(run_id, outcome)`` ever binds -- so the CancelledError
+    branch must recover the run id via the bridge's latest-unanchored-primary
+    lookup and record the stopped reply's PERSISTED id onto the run. Every
+    pre-existing stop test simulated the stop via a normally-returning
+    ``run_reply``, which made this gap invisible. RED pre-fix: the run row
+    stays NULL and falls to the ordinal fallback on resume.
+    """
+    from Tests.Chat.test_console_chat_store import FakePersistence
+
+    class _YieldThenParkGateway(_ParkingGateway):
+        """Streams ONE chunk before parking: a zero-chunk stop never persists
+        (empty rows defer -- the AC#3 NULL case, covered separately below), so
+        the persisted-id recording needs a partial reply in the buffer."""
+
+        async def stream_chat(self, _resolution, _messages):
+            yield "partial answer\n"
+            self.started.set()
+            self.release.wait(timeout=60)
+            yield "answered anyway."
+
+    gateway = _YieldThenParkGateway()
+    persistence = FakePersistence()
+    store = ConsoleChatStore(persistence=persistence)
+    db = AgentRunsDB(tmp_path / "runs.db", client_id="t")
+    bridge = ConsoleAgentBridge(agent_runs_db=db, store=store, provider_gateway=gateway)
+    controller = ConsoleChatController(
+        store=store,
+        provider_gateway=gateway,
+        provider="llama_cpp",
+        model="test-model",
+        agent_bridge=bridge,
+        agent_runtime_enabled=True,
+    )
+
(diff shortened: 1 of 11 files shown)


rmusser01:
Addressed Qodo's three findings:

1. **create_sibling docstring (fixed)** — `attachments` now documented in the Args section.
2. **Underscore test-class names (declined with evidence)** — `_YieldThenParkGateway`/`_RecordingStreamingGateway` match the established convention of the exact files they live in: `_Gateway`, `_ParkingGateway`, and `_EditResendPersistence` are pre-existing underscore-prefixed private test fakes in the same modules. Renaming only the two new classes would create the intra-file inconsistency the naming rule exists to prevent; renaming the pre-existing ones is out of this PR's scope.
3. **Unbounded run scan (confirmed + fixed)** — added `AgentRunsDB.latest_primary_run()` (single kind-filtered `LIMIT 1` query, so interleaved newer sub-agent runs can't hide the newest primary) and rewired `latest_unanchored_primary_run_id` onto it. New DB test pins newest-primary selection, subagent interleaving, superseded fallback, and the empty case — red before, green after.

Rebased on latest dev; 523 tests green across the flow/controller/agent-swap/edit-resend/store/resume/DB/stop suites (one unrelated conversation-browser test flaked under full-battery load and passes in isolation).

PR #881, opened 2026-07-25
Image-gen UAT: fix findings 620-622, re-scope 623 (ghost-paint artifact)

rmusser01:
## Summary

Live user-acceptance testing of the merged image-gen stack against a real OpenRouter key (2026-07-25) passed 7 of 8 scenario groups end-to-end. This PR files the four findings as backlog tasks **and fixes three of them**; the fourth was investigated, its presumed mechanism disproven, and re-scoped honestly.

### Fixed

- **task-620 (high)** — `DEFAULT_OPENROUTER_IMAGE_MODEL` was `openai/gpt-image-1`, which no longer exists on OpenRouter → out-of-the-box generation 404'd. New default: `google/gemini-2.5-flash-image` (verified live in the catalog; cheap/fast). A 404 from OpenRouter now raises an adapter error naming the attempted model id and pointing at `[image_generation.openrouter] default_model` instead of surfacing httpx's bare URL error; other statuses unchanged. All product/doc references updated.
- **task-621 (medium)** — keys placed directly under `[image_generation]` that match a known flat field name (e.g. `openrouter_image_default_model`) now log a warning naming the exact nested replacement (derived from the loader's own flat-map, not hardcoded); truly unknown keys get a generic warning. Fires once per load; nested shape, secrets/env precedence, and all legitimate keys (incl. `context_llm_*`) verified warning-free.
- **task-622 (high)** — the no-prompt `/generate-image` LLM-composition payload now always ends with a user-role compose instruction (appended after turns-window truncation), so it can never end on an assistant turn — which `chat_with_llama` treated as a response prefill, llama.cpp rejected with a 400, and the silent fallback masked. **Live red/green against a real llama.cpp server**: the old payload shape still 400s; the production function now returns a genuine LLM-composed prompt. Provider-agnostic; no handler changes; fallback semantics untouched.

### Re-scoped, not fixed

- **task-623 (low)** — the "stale action row inside the generation card": a systematic investigation (5 interleavings in a real Textual harness; DOM-mutation path audit) **disproved the DOM-race hypothesis** — the transcript has a single serialized DOM mutator behind a lock, and no reproduction produced a mis-parented or orphaned row. The evidence (ghost row was click-dead, had no guide line, survived Escape, cleared only on full repaint) points to a stale terminal-paint/compositor artifact instead. The task is re-scoped to a live-instrumentation investigation and stays in the backlog rather than shipping a speculative patch against a disproven mechanism.

## Testing
- `Tests/Image_Generation/` + generate/actions suites: 228 passed, 6 skipped (includes new pins: default-model flow, enriched-404 message via the real exception shape, warn-on-flat-key incl. no-false-positive sweep across every legitimate config shape, payload trailing-role invariant across all conversation shapes).
- Each task SDD-reviewed with independent verification (620/621/622 all approved zero findings; 622's reviewer re-ran the live llama red/green itself).
- `ruff` clean on touched files; `python -c "import tldw_chatbook.app"` clean.


rmusser01:
**Qodo finding (Backlog Guard will fail on pre-existing duplicate IDs): DECLINED for this PR.** The duplicates it cites (task-100/484/505 etc.) exist identically on `origin/dev` — none are introduced here, and TASK-620–623 were assigned after a sweep of every ref (global max was 619; both filename and frontmatter namespaces checked). Renumbering another program's task files inside this PR would create exactly the cross-branch collision churn the repo's backlog conventions exist to avoid; the duplicate-ID pileup is already tracked as its own cleanup task (task-544). Note also this repo intentionally cancels CI checks (verified-locally convention), so the guard is advisory — every backlog-touching PR in the current campaign merged under the same baseline.

PR #883, opened 2026-07-25
fix(skills): auto-unlock skill trust from cached keyring keys (task-624)

rmusser01:
## Summary

Closes **task-624**. `SkillTrustService.unlock_from_keyring_convenience()` existed, was unit-tested, and had **zero callers in the application** — so enabling keyring convenience persisted derived trust keys that nothing ever loaded.

Every launch reported *"Skill trust is locked for this session"* and demanded a manual unlock, while the user's keys sat unused in the keychain.

Found by live TUI verification (task-579). This is the archetypal defect unit tests cannot catch: **the method passed its own tests precisely because those tests supplied the call site production was missing.**

## Fix

A memoized `_try_cached_unlock()` now sits behind the four points that decide lockedness — `trust_posture()`, `status_for_skill()`, `trusted_file_paths()`, `_require_keys()`.

Behind the choke points rather than at app startup, so:
- every consumer benefits (Library UI, the agent worker thread, Console skill paths), not just whichever screen happened to be wired;
- there is no startup keychain cost — it stays lazy.

Design notes:
- **Latched to one attempt per trust-state generation.** Keychain reads are not free and posture is queried on every Library render; a counting-cache test pins exactly 1 load across 10 queries. The latch clears in `reset_trust` so a later re-bootstrap can cache again.
- **No new config key.** The keychain entry *is* the persistence, and `unlock_from_keyring_convenience` already flips `keyring_convenience_enabled` on success.
- **Failure is silent by design.** A missing, stale, or exploding cache leaves the session locked, so the existing Unlock path remains the fallback.
- Also adds the module's missing `loguru` import.

## Live verification

Same QA profile, same keychain state, before and after:

| | Trust header | Skill rows |
|---|---|---|
| Before | `Skill trust is locked for this session.` | ⚠ ⚠ |
| After | `Skill trust: ready.` | ✓ ✓ |

No unlock click, no passphrase. Capture committed at `Docs/superpowers/qa/skills-script-execution-2026-07-25/after-fix-trust-ready.txt`.

## Testing

`Tests/Skills/test_skill_trust_keyring_autounlock.py` (8 tests), written RED-first: 5 failed against the old code, and the 3 fallback cases (absent cache, stale salt, exploding backend) passed throughout — proving no false baseline.

Skills 357 · Agents 261 · Chat+Library 2844 (+69 skipped) · ruff clean.

The single `Tests/Library` failure is the pre-existing `rewind`/`generate-image` drift-guard baseline (`{'generate-image', 'rewind'}`), tracked as task-580 and unrelated to this change.

---
## Summary by cubic
Auto-unlocks skill trust from cached keyring keys on first trust check so users who enabled keyring convenience don’t see “locked for this session”. Makes the cached unlock retryable and wires the Settings posture, addressing task-624 without adding startup cost.

- **Bug Fixes**
  - Added `_try_cached_unlock()` behind `trust_posture()`, `status_for_skill()`, `trusted_file_paths()`, `overall_status()`, and `_require_keys()`; lazy, retryable on failure, and short-circuits after success (single keychain read).
  - On missing/stale cache or keyring errors, stay locked and never crash; flip `keyring_convenience_enabled` on successful unlock.
  - Tests cover auto-unlock, retry after transient failure, Settings posture, salt binding, fallback behavior, and single keychain read.
  - Minor: added `loguru` debug for failed cached unlock and a QA capture document.

<sup>Written for commit 5e7cb9a8dd777f39d4b41d60693d39649e8d79e5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/883?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Both findings were legitimate and are fixed in 5e7cb9a8d — thanks, the first one in particular was a genuine reliability regression I introduced.

**1. Autounlock can't recover.** Correct, and worse than it first looks: the latch was set *before* the attempt and cleared only by `reset_trust`, so a single transient keyring/IO failure would suppress every later cached-unlock for the life of the process — directly contradicting `trust_posture()`'s own documented contract that an unavailable keyring is recoverable by Retry.

The latch is now gone entirely rather than merely moved. The insight is that the happy path never needed it: a successful load sets `self._keys`, which short-circuits the method thereafter, so the keychain is still read exactly once when the cache works (pinned by a counting-cache test asserting 1 load across 10 queries). Only `key_cache is None` short-circuits now, that being the one definitively hopeless case. A failure stays retryable, which is the whole point of the Retry affordance.

**2. Settings skips autounlock.** Also correct, and a real miss on my part — `overall_status()` is a **fifth** lockedness decision point and I had only wired four. Settings could therefore report locked while valid cached keys sat available. Now wired like the rest.

Two regression tests added: a transient failure must not bar a later successful unlock, and the Settings posture must auto-unlock. Suites: Skills+Agents 620, ruff clean. Re-verified live in the same QA profile — still comes up "Skill trust: ready." with no passphrase.

PR #885, opened 2026-07-25
qa(skills): UAT results for skill script execution + local-llm provider defect (task-579/625)

rmusser01:
## Summary

Live UAT of trust-gated skill script execution, driven against the **real app** in an isolated profile with a **real local model** (llama.cpp/Gemma on `:9099`) and a **real subprocess**. Only the model is "simulated" — every gate, card and process is genuine.

Docs-only PR: UAT results, evidence captures, and one newly filed defect.

## Verified

**AC#1 — the card describes what will actually run.** Raised by a real agent tool call:

```
An agent wants to run a script from a skill:
demo-runner — scripts/hello.py (runs with /usr/bin/python3)
   Allow once    Always allow this skill       Deny
```

**AC#2 — Allow runs it; Deny does not.** Allow produced `run_skill_script → exit_code: 0` and the assistant replied with the script's genuine stdout (`QA579 HELLO FROM THE SCRIPT` — a string that exists nowhere else). A second invocation raised a fresh card; Deny left the transcript with still exactly **one** occurrence of that output, proving the denied run never executed.

Two incidental confirmations worth noting: the per-round `request_id` handshake works across sequential cards (buttons had effect rather than being silently dropped), and `run_skill_script` proved genuinely reachable by a **spawned sub-agent** — the model's first attempt delegated via `spawn_subagent`. That is the all-agents caller scope behaving as designed.

## Not verified — 579 stays open

AC#3 (always-allow suppresses the next prompt), AC#4 (content change re-prompts) and AC#6 (context switch) were not run; AC#5 is partial. **This is not a sign-off.**

## New defect: task-625

`chat_with_local_llm` reads a **top-level** `local-llm` settings key, but the provider's config lives at `api_settings.local-llm` — where the app's own documented example puts it. Every sibling local-provider function in the same file resolves via `settings["api_settings"]` correctly.

There is no config workaround: `load_settings()` does not preserve arbitrary top-level sections (verified). The provider is therefore unusable from configuration, surfacing as *"Agent run failed: provider returned HTTP 502 … configuration error"*. It blocked this UAT until the run was switched to `llama_cpp`.

---
## Summary by cubic
Adds live UAT docs and evidence verifying AC#1–#4 for trust-gated skill script execution (task-579) in the real app with a real subprocess, plus a reproducible harness and seed script. Documents the `local-llm` config bug (task-625); AC#5 remains partial and AC#6 not run; also removes a timestamp-only CSS bundle change to keep the diff clean.

- **UAT Results**
  - AC#1: Confirm card shows correct skill, script path, and runtime.
  - AC#2: Allow runs (exit_code 0, expected stdout); Deny prevents execution.
  - AC#3: Always allow persists a digest-pinned grant; next run shows no card.
  - AC#4: Editing the script invalidates the grant and re-prompts; re-approving trust does not restore it; new content runs.
  - AC#5: Partial — grant line and Revoke live under `#library-skill-trust-panel`, which only renders after opening the skill in the editor; couldn’t reach it. AC#6 not run.
  - Added harness write-up and `seed3.py`, with notes on `request_id` handshake and sub-agent calls.

- **Bug Filed**
  - `chat_with_local_llm` reads top-level `local-llm` instead of `api_settings.local-llm`; no config workaround, surfaces as configuration error/HTTP 502. UAT switched to `llama_cpp`. Tracked as task-625.

<sup>Written for commit 16556ae258505ef69c8d521b9ed529fc1057dfc9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/885?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Good catch, fixed in 16556ae25 — the CSS bundle is regenerated at app boot, so launching the TUI for this UAT swept a `Generated:` timestamp change into what should have been a docs-only PR. Reverted to dev's copy; no styling difference existed. The PR is now 9 files: the QA write-ups, the per-criterion evidence captures, the seed script, and the task-625 file.

Also folded in the harness notes and resume recipe that were stranded on the earlier `claude/task-579-live-tui-verification` branch, which had gone stale against dev — so the whole QA record (harness, traps, findings, UAT results) now lands in one place rather than being split across an unmerged branch.

PR #889, opened 2026-07-25
fix(llm): resolve local-llm config from api_settings, not a top-level key (task-625)

rmusser01:
## Summary

Closes **task-625**. The `local-llm` provider could not be configured at all.

`chat_with_local_llm` resolved its config with:

```python
cfg = settings.get("local-llm", {})      # a TOP-LEVEL key
```

The provider's configuration lives at `api_settings.local-llm` — its documented home (the commented `[api_settings.local-llm]` block in `config.py`) and where **every** sibling local provider in the same module looks. `load_settings()` does not preserve arbitrary top-level sections either, so no `config.toml` could ever populate that key.

The result: any user configuring `local-llm` as documented hit

```
ChatConfigurationError: Llamafile/Local LLM API URL (api_url) is required …
```

surfaced in the Console as an opaque *"Agent run failed: provider returned HTTP 502 … configuration error"*.

## Fix

Read `settings["api_settings"]["local-llm"]`, exactly like the siblings.

**Key-name reconciliation:** the documented block writes `api_url` while the code read `api_ip`, so resolution is now `api_url or api_ip` — the documented key wins, and anyone with a working `api_ip` setup is not broken.

**Sweep (AC#4):** grepped every `settings.get(` call across `LLM_Calls/`. This was the **only** offender — `koboldcpp`, `ooba_api`, `tabby_api`, `aphrodite_api`, `llama_cpp`, `vllm` and the rest all correctly go through `settings["api_settings"]`.

## Verification

Beyond the unit tests, a `config.toml` containing the documented `[api_settings.local-llm]` block now completes a real `chat_api_call` against a live llama.cpp server — the exact case that failed during the task-579 UAT and forced that run onto a different provider.

`Tests/LLM/test_local_llm_provider_config.py` (4 tests) written RED-first: 3 failed against the old code, while the missing-URL case passed throughout, proving the error path is untouched and the tests aren't vacuous.

Tests/LLM + Tests/Chat: 2189 passed / 69 skipped · ruff clean.

---
## Summary by cubic
Fixes the `local-llm` provider by reading config from `api_settings.local-llm` in `config.toml`, so it can be configured as documented. Addresses task-625 and avoids opaque 502s from misconfiguration.

- **Bug Fixes**
  - Read config from `api_settings.local-llm`, not a top-level key; matches docs and other local providers.
  - Support `api_url` (preferred) with `api_ip` as fallback to preserve existing setups.
  - Kept clear config error when URL is missing; added tests for both keys and the error path.

<sup>Written for commit 79f56de8c58617a3152b1a6bc16f47ca37bafac2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/889?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/LLM/test_local_llm_provider_config.py
+++ b/Tests/LLM/test_local_llm_provider_config.py
@@ -1,0 +1,94 @@
+"""task-625: the local-llm provider must resolve its config from api_settings.
+
+`chat_with_local_llm` read a TOP-LEVEL "local-llm" settings key, while the
+provider's configuration lives at `api_settings.local-llm` — which is where the
+app's own documented example puts it, and where every sibling local provider in
+the same module looks. `load_settings()` does not preserve arbitrary top-level
+sections either, so the provider was unusable from configuration entirely.
+"""
+
+import pytest
+
+from tldw_chatbook.LLM_Calls import LLM_API_Calls_Local as local_calls
+
+
+@pytest.fixture
+def captured_call(monkeypatch):
+    """Intercept the outbound request so no network is touched."""
+    seen = {}
+
+    def fake_request(*args, **kwargs):
+        seen["args"] = args
+        seen["kwargs"] = kwargs
+        return {"choices": [{"message": {"role": "assistant", "content": "ok"}}]}
+
+    # `_chat_with_openai_compatible_local_server` is the shared transport every
+    # local provider funnels through; stubbing it keeps this a pure config test.
+    monkeypatch.setattr(
+        local_calls,
+        "_chat_with_openai_compatible_local_server",
+        fake_request,
+    )
+    return seen
+
+
+def _settings_with(api_settings_block):
+    return {"api_settings": {"local-llm": api_settings_block}}
+
+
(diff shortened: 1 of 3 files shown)


rmusser01:
Thanks — agreed on the shape of the fix, and noted on the follow-up suggestion.

For the record on scope: I did sweep every `settings.get(` call across `LLM_Calls/` as part of this (task-625's AC#4), and `local-llm` was the only provider reading the wrong table — `koboldcpp`, `ooba_api`, `tabby_api`, `aphrodite_api`, `llama_cpp` and `vllm` all already go through `settings["api_settings"]`. So there is no cluster of similar breakage to consolidate today; a shared helper or load-time normalization would be worth it if a third key-name variant appears, but extracting one for a single site would be speculative right now.

PR #890, opened 2026-07-25
feat(agents): built-in tool permission gate (P1 of TASK-545)

rmusser01:
**Phase 1 of three** for TASK-545. Builds the permission **gate seam** for the agent runtime's built-in tools, reusing MCP's existing permission machinery. **No tools are ported and no config changes** — those are P2 and P3. TASK-545 stays `To Do`.

## The filed task's premise did not hold

TASK-545 said to wire `Tools/tool_executor.py`'s `ToolExecutor` into the gate, naming `write_file`/`create_note`/`update_note`. Investigation found `ToolExecutor` (System A) and `Agents/tool_catalog.py`'s `BuiltinToolProvider` (System B) are **disjoint systems with non-overlapping tool sets**:

| | System A `ToolExecutor` | System B `BuiltinToolProvider` |
|---|---|---|
| Tools | datetime, calculator, code_audit **+ all fs/note/web/RAG tools** | datetime, calculator **only** |
| Reached from | legacy `worker_events.py` / `chat_streaming_events.py` (main loop) | the agent runtime / Console (worker thread) |
| Reached by the Console | **no** — zero references to `execute_tool_call`/`get_tool_executor` | yes, this is the shipping path |

The fs/mutating tools the task names exist **only** on the deprecated legacy chat path. Gating System A would have hardened code slated for removal while leaving the live path unprotected. Per user decision, the tools move to the agent runtime (P2) and the gate is built there first (this PR). TASK-545's description and ACs were rewritten to match — CLAUDE.md requires updating the AC when deviating, not just the notes.

## What ships

1. **`Tool.risk_tags`** — concrete property, default `()`, vocabulary is the *existing* `HIGH_RISK_TAGS` (`mutates`/`process`). Concrete, not abstract, so no existing subclass breaks.
2. **`resolve_builtin_state`** — a **sibling** of `resolve_effective_state`, not a parameter on it. `resolve_effective_state` is byte-unchanged and MCP behavior is byte-identical. Precedence: tool override → server default → built-in floor `allow`, then the existing risk-flooring pass. Deliberately not the MCP global default, so changing MCP's posture never starts prompting for calculator.
3. **`BuiltinToolGate`** (`Agents/builtin_tool_gate.py`) — the impure seam: one store load per turn, per-turn stamps, session approvals. Never raises; a missing or partial service degrades to the allow floor, never allow-everything.
4. **Enforcement in `BuiltinToolProvider.invoke`** — defense in depth. `gate=None` means *build the real gate lazily*, never "ungated".
5. **Per-row approval options** — `MCPPendingCall.options` (defaulted) → `request_mcp_approvals`' dict conversion → `ChatApprovalCard`. Bulk Approve-all/Deny-all made row-aware.
6. **A run-level review hook, wired unconditionally.** Previously the hook was built only when MCP was configured, so **a user with no MCP servers had no gate at all**. It now always exists, folding MCP in when present, sharing **one** gate instance with the provider.

## Namespace: `agent:builtin`, not `builtin:tldw_chatbook`

`builtin:tldw_chatbook` is a **live** key (`readiness.BUILTIN_SERVER_KEY`) belonging to the built-in MCP *server*. Filing agent-runtime built-ins under it would put two different execution paths under one permission identity — a decision about the MCP server's `write_file` would silently govern the agent-runtime one once P2 lands overlapping names. A test pins the namespaces disjoint.

## Behavior change (one, intentional)

The kill switch is now **global**: it blocks built-in tools as well as MCP, and its label was updated. So for a user with the kill switch **on**, `calculator`/`get_current_datetime` previously executed and are now refused. Everything else is unchanged today — nothing declares `risk_tags`, so no built-in resolves to `ask` and no approval card appears for them yet.

## Two corrections captured in the spec rather than silently edited

- **`deny` was wrongly excluded** from built-in rows on the belief it wrote persistent state. It does not — `always_allow` is the sole persistent write; `deny` is a turn-scoped refusal. Excluding it protected nothing and was unsafe: with no deny option a built-in row could not be refused from the card, so **"Deny all" would have left it on "approve once" and thereby approved it.**
- **The `invoke` precedence shorthand was inverted** (`kill switch → resolve → stamp`); shipped order is `kill switch → stamp → resolve → session`, matching `MCPToolProvider.invoke`.

## Testing

Post-rebase: **710 passed** (Agents/MCP/Tools + approval UI), **2196 passed / 69 skipped** (Chat).

The repo has a pre-existing failing baseline in areas this branch does not touch. The three `Tests/UI` files implicated were run on this branch **and** on a clean `origin/dev` worktree: identical **8 failed / 29 passed / 16 skipped**, same signatures (`KeyError: 'openai'`, missing `_console_chat_store`, `library_screen.py` workers). Pre-existing, not from this work. The remaining repo-wide failures were not individually verified.

Rebased onto 9 new dev commits touching the same files (including dev's new `run_skill_script`). Textually clean; semantically checked — `run_skill_script` is a **native** runtime tool dispatched in `run_agent_loop`, not through `BuiltinToolProvider.invoke`, so this gate correctly does not touch it.

## Follow-ups filed (627-633)

**task-628 (child-run approval routing) blocks P2**: a sub-agent inherits the parent's allow-list and can call built-in tools, but a child's `begin_turn()` would clear the parent's stamps — built-ins have no equivalent of MCP's `review_state_scope`. Dormant today; a hard prerequisite before any gated mutating tool ships. Also filed: the built-in permissions UI (627, gates persistent decisions), the second instance of TASK-547's config bug (629), and four accuracy/cleanup items (630-633).

Built with subagent-driven development: 7 TDD tasks, a spec+quality review each, an adversarial whole-branch review, and a scoped re-review of its fix wave.

---
## Summary by cubic
Phase 1 of TASK-545 adds a permission gate for agent‑runtime built‑in tools using the MCP permission store. It enforces decisions per run (review hook) and per call, makes a resolved deny absolute, and the kill switch now blocks built‑ins.

- **New Features**
  - `BuiltinToolGate` enforces decisions for built‑ins via `resolve_builtin_state` under `agent:builtin` (default `allow`; high‑risk tags floor to `ask`).
  - Added `Tool.risk_tags` (default `()`); read‑only tools stay `allow` by default.
  - Run‑level review hook (`build_tool_review_hook`) is wired even without MCP and shares one gate with `BuiltinToolProvider.invoke` per turn.
  - Approval UI supports per‑row options via `MCPPendingCall.options` and `ChatApprovalCard`; built‑in rows offer session‑scoped choices by default.
  - Kill switch relabeled to “block tool calls in chat” and now also blocks built‑ins.

- **Bug Fixes**
  - `BuiltinToolGate.check` prioritizes resolved `deny` over stamps/session approvals; “Off” cannot be bypassed.
  - Run‑level review hook honors built‑in session approvals (no re‑prompt after “Approve for session”).
  - Re‑enabled `deny` in built‑in rows and made bulk Approve/Deny row‑aware to avoid invalid selections.

<sup>Written for commit 906c38d394b6a52c0c65af10a77290aa3817b7d5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/890?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Thanks — the Bug is real, and it is worse than the description suggests. Fixed in `2196b922e`.

### 1. "Stamp bypasses Off" (Bug) — **fixed**

Correct, and the reason it matters is an asymmetry I had missed: I chose the original ordering to mirror `MCPToolProvider.invoke`, but **MCP has a second layer built-ins do not**. `MCPToolProvider.compose_catalog` filters `deny`-state tools out of the catalog entirely (`Agents/mcp_tool_provider.py:215-223`), so a denied MCP tool never reaches the model *or* `invoke()`. Built-in tools are unconditional — `BuiltinToolProvider._tools` (`Agents/tool_catalog.py:193`) always contains every tool regardless of permission state, so a denied built-in **is** disclosed to the model, **is** called, and `invoke()`'s gate is the only thing between the model and execution. Mirroring MCP's precedence without MCP's catalog filtering left the deny unenforced.

`check()` now consults the resolved state for `deny` before honoring any stamp or session approval:

`kill switch → resolved deny → stamped deny → stamped permit → resolved allow → session approval → fail closed`

Everything else is preserved: the kill switch is still first and unbypassable, a stamped `deny` still refuses for the turn, `check()` still never raises, and a missing/partial service still degrades to the allow floor rather than allow-everything. The extra `resolve()` on the stamped path costs no I/O — `_load_payload` is cached per turn.

Two regression tests added. `test_effective_deny_wins_over_stamped_approve_once` was confirmed **RED** against the old ordering (`assert (None is not None)`) before the fix; `test_effective_deny_wins_over_live_session_approval` passed pre-fix (session approval already came after `resolve()`) and is kept as a lock.

### 2-4. Docstring rule violations — **fixed**

`Args:`/`Returns:` sections added to `Tool.risk_tags`, `build_builtin_gate`, and `tool_ref`, keeping the existing explanatory prose.

`pytest Tests/Agents/ Tests/MCP/ Tests/Tools/` → 629 passed.

rmusser01:
Re-review is at **Bugs (0)** — the security finding is resolved. The two remaining rule violations are both declined, with evidence.

### "build_builtin_gate docstring missing Args" — already fixed (stale marker)

This was fixed in `2196b922e` alongside the other two docstrings. The function currently carries both sections:

```
def build_builtin_gate(service: Any | None = None) -> BuiltinToolGate:
    ...
    Args:      # line 16 of the docstring
        service: The control-plane service ... `None` builds a
            service-less gate that still gates (fail-closed), not an
            ungated one.
    Returns:   # line 24
```

Its two siblings from the same commit (`tool_ref`, `Tool.risk_tags`) are both marked ✓ Resolved, so this looks like a tracker artifact rather than a real gap. No action taken.

### "Test docstrings not Google-style" — declined, contrary to repo convention

Test functions take no arguments and return nothing, so `Args:`/`Returns:` sections would be empty ceremony. More decisively, this is not what the codebase does — measured across the suite:

```
test docstrings sampled: 733; using Args:/Returns: 0 (0.0%)
```

Zero of 733 existing test docstrings use those sections. Adopting them in this PR would make these the only tests in the repository written that way, which is a consistency regression rather than an improvement. CLAUDE.md's Google-style requirement is scoped to public APIs — the production docstrings in this PR (`Tool.risk_tags`, `build_builtin_gate`, `tool_ref`, `resolve_builtin_state`, `BuiltinToolGate`'s methods) all carry the full sections.

The new tests do have descriptive prose docstrings explaining what invariant each pins, matching the surrounding style.

PR #891, opened 2026-07-25
fix(library): close the shadowed-builtin drift guard (task-580)

rmusser01:
## Summary

Closes **task-580**. Adds `rewind` and `generate-image` to `_SHADOWED_BUILTIN_NAMES`, clearing a drift-guard failure that had been carried as an *accepted baseline* across several branches — which is precisely the signal erosion the guard exists to prevent.

## Not just a red test

The set backs `skill_name_shadows_builtin()`, so a skill named `rewind` or `generate-image` was silently **not** recognised as shadowing a built-in, unlike every other console command. Small, but a real correctness gap rather than cosmetic test debt.

## Why it kept getting waved through

The guard fired correctly every single time. The problem was its message: it listed the missing strings and nothing else, so the cheapest response was always "known baseline, ignore."

All three assertions now name the file to edit and say explicitly not to accept the failure as a baseline:

```
ConsoleCommandRegistry names not covered: {'rewind', 'generate-image'}.
Add them to _SHADOWED_BUILTIN_NAMES in
tldw_chatbook/Library/library_skills_state.py -- do not accept this as a
baseline failure (task-580).
```

## Testing

Added a behaviour test asserting `skill_name_shadows_builtin()` returns the normalized name for both commands, including the whitespace/case path. It pins them by **literal name** rather than re-deriving from the registry — re-deriving would make the test tautological against the very set under test.

`Tests/Library` + `Tests/Skills`: **1025 passed**, green with no baseline caveat for the first time in this work stream. ruff clean.

---
## Summary by cubic
Fixes the shadowed-builtin drift guard by adding `rewind` and `generate-image` to `_SHADOWED_BUILTIN_NAMES` and clarifying the guard message, so skills with those names correctly shadow console commands and future misses are easy to fix (task-580).

- **Bug Fixes**
  - Added `rewind` and `generate-image` to `_SHADOWED_BUILTIN_NAMES`.
  - Updated drift-guard assertions to name `tldw_chatbook/Library/library_skills_state.py` and say not to accept as a baseline.
  - Added a behavior test for `skill_name_shadows_builtin` covering both names and case/whitespace normalization.
  - Removed an unused `monkeypatch` parameter from the new shadowing test.

<sup>Written for commit 72bb6849f29d4c3d09e41102e5e9e93ac1e9a79a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/891?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Library/test_library_skills_state.py
+++ b/Tests/Library/test_library_skills_state.py
@@ -216,19 +216,31 @@ def test_shadow_name_set_stays_in_sync_with_real_sources():
 
     # Assert RUNTIME_TOOL_NAMES (spawn_subagent, find_tools, load_tools) is a subset
     assert RUNTIME_TOOL_NAMES <= _SHADOWED_BUILTIN_NAMES, (
-        f"RUNTIME_TOOL_NAMES not covered: {RUNTIME_TOOL_NAMES - _SHADOWED_BUILTIN_NAMES}"
+        "RUNTIME_TOOL_NAMES not covered: "
+        f"{RUNTIME_TOOL_NAMES - _SHADOWED_BUILTIN_NAMES}. "
+        "Add them to _SHADOWED_BUILTIN_NAMES in "
+        "tldw_chatbook/Library/library_skills_state.py -- do not accept this "
+        "as a baseline failure (task-580)."
     )
 
     # Assert builtin tool names (calculator, get_current_datetime) are a subset
     builtin_names = {e.name for e in BuiltinToolProvider().list_catalog()}
     assert builtin_names <= _SHADOWED_BUILTIN_NAMES, (
-        f"BuiltinToolProvider names not covered: {builtin_names - _SHADOWED_BUILTIN_NAMES}"
+        "BuiltinToolProvider names not covered: "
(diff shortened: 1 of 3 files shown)


rmusser01:
Fixed in the rebased branch — removed the unused `monkeypatch` parameter rather than documenting an argument the test does not need. Rebased onto latest `dev`; module still 16/16 green.

(For contrast with the docstring finding I declined on #875: that one asked for `Args:` sections on pytest fixture parameters, which this repo's suites deliberately do not document — 0 of 21 docstring'd param-taking tests in `Tests/Skills` use them. This one is different: the parameter was genuinely unused, which is a real smell regardless of convention.)

PR #895, opened 2026-07-25
fix(agents): scope built-in gate stamps across nested sub-agent runs (TASK-628)

rmusser01:
Closes **TASK-628**, the hard prerequisite blocking TASK-545/P2. Also files **TASK-634** (agent/sub-agent settings screen).

## The bug

A spawned child run shares the parent's **exact** `BuiltinToolGate` instance and the **exact same** `review_tool_calls` closure, and `spawn()` runs the child's entire loop inline, synchronously, mid-parent-dispatch. That closure's first act is `builtin_gate.begin_turn()`, which clears the **whole** `_stamps` dict — so a child wiped the verdicts a user had just given for the parent's own same-batch calls.

This is worse for built-ins than for MCP. `MCPToolProvider.invoke` has a per-call `_approval_callback` fallback, so a lost stamp merely re-prompts. `BuiltinToolGate.check()` has no such fallback — its only approval sources are `_stamps` and a live session approval — so a clobbered stamp **fails closed outright**, making an already-approved tool unusable from inside any sub-agent.

## The task's own premise was wrong, and is corrected

TASK-628 attributed the failure to the child running "nested inside the parent's already-blocked worker thread", implying no approval round trip could reach the UI. That is not the mechanism. `spawn()` calls `_run_one` directly — same call stack, same worker thread — and `call_from_thread` only needs the *UI* thread free, which it is. Nested approval round trips demonstrably work: the existing `test_parent_deny_is_not_overridden_by_a_same_turn_spawned_childs_approval` drives two in a single turn for MCP.

There is no threading hazard. This was purely a shared-mutable-state clobber. The task file now carries that correction so the next reader doesn't re-derive it.

## The fix

`BuiltinToolGate.stamp_scope()` mirrors `MCPToolProvider.stamp_scope`: snapshot `_stamps`/`_payload` on enter, **restore — never merge** — on exit. `_payload` is included because the child's `begin_turn()` drops it too, and restoring it preserves the parent's one-store-load-per-turn property.

`AgentService.review_state_scope` holds only **one** context manager, but two components now own per-turn state a nested run would clobber. New module-level `_combine_state_scopes(scopes)` enters all present scopes on an `ExitStack`. It returns `None` for zero scopes and the single callable **unchanged** for one — so the pre-existing MCP-only wiring stays byte-identical.

## Testing

Unit: `stamp_scope` (restore, restore-on-raise, reentrant nesting); `_combine_state_scopes` (passthrough at 0/1, both entered, reverse-order unwind on raise).

Integration, through a **real spawn**, reusing the existing `ScriptedChat` harness:
- **AC#2** — parent batch `[spawn_subagent, write_thing]`; the child calls a different `"mutates"`-tagged tool through the same shared gate. Asserts the child's call resolves via its own round trip (not fail-closed for being nested) **and** that the parent's pre-child approval survives.
- **Negative control** — the identical interleave with `review_state_scope` unwired, asserting the parent's approved call fails closed. This is what proves AC#2 isn't passing trivially.
- **AC#4** — a run with no approval route at all still fails closed; this task fixes routing without weakening that guarantee.

AC#2's test was further verified discriminating by sabotage: with `stamp_scope`'s restore removed it fails (`assert not True`); restored, all 6 in the file pass.

`Tests/Agents/ Tests/MCP/ Tests/Tools/` → **637 passed**. `Tests/Chat/` → **2203 passed / 69 skipped**.

## Also in this PR

**TASK-634** — a settings screen for agent and sub-agent behavior, general and per-workspace. Every run-governing knob is currently a hardcoded constant (`CONSOLE_MAX_MODEL_TURNS`, the step/wall backstops, `max_subagents`, `max_total_tokens`, `max_tool_call_seconds`); the only user-facing agent setting is the `[console] agent_runtime` bool. Raising the round cap to 20 moved the worst case to `rounds * (1 + max_subagents)` = 60 provider turns per message with no way to see or lower it. The task notes it should share a surface with TASK-627's permissions UI.

TASK-545's P2 section is updated to record that its blocker is cleared.

---
## Summary by cubic
Scopes built-in gate approvals across nested sub-agent runs so child tool calls can be approved and parent approvals aren’t lost. Implements TASK-628 and unblocks TASK-545/P2.

- Bug Fixes
  - Added `BuiltinToolGate.stamp_scope()` to snapshot/restore `_stamps` and `_payload` around nested runs (restore, never merge).
  - Added `_combine_state_scopes()` in `console_agent_bridge.py` to compose `review_state_scope` for both MCP and built-ins; 0/1 scope remains unchanged.
  - Child runs now resolve gated built-in tools via the shared approval route; parent’s pre-child approvals are preserved. Runs with no approval route still fail closed.
  - Corrected prior diagnosis: the bug was shared state clobber, not threading; nested approval round trips work.
  - Tests cover scope behavior (restore, reentrant, on-raise), combined scopes (passthrough and reverse-order unwind), and integration via real spawn (positive + negative control).

- Refactors
  - Annotated `BuiltinToolGate.stamp_scope()` return type as `Iterator[None]`.

<sup>Written for commit dce7dfabbca5586059188d854fa7c89d3bfc4cc2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/895?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Thanks — one fixed, one declined with evidence.

### 1. `stamp_scope` missing return type — **fixed** (`ee0c9a9`-range commit)

Now `def stamp_scope(self) -> Iterator[None]:` with `from collections.abc import Iterator`. Correct for a `@contextmanager`-decorated generator.

Worth noting the precedent it mirrors, `MCPToolProvider.stamp_scope`, is also unannotated — so this PR is now stricter than the code it was modeled on rather than matching it. Happy to annotate that one too, but it's outside this PR's scope.

### 2. "Docstring missing `Returns:` (it uses `Yields:`)" — **declined**

`stamp_scope` is a generator (`@contextlib.contextmanager` + a bare `yield`). Google's Python style guide prescribes **`Yields:`** for generators and `Returns:` for ordinary returns — using `Returns:` here would be the incorrect section, not the correct one.

The codebase agrees. Measured across `tldw_chatbook/`:

```
contextmanager docstrings using Yields: 5   using Returns: 0
```

Every existing `@contextmanager` in the project that documents its yield uses `Yields:`; none uses `Returns:`. Switching would make this the only one written the other way.

The section is also substantive rather than a placeholder — it documents that the restore is unconditional and discards whatever the nested run recorded (a restore, never a merge), which is the whole point of the scope.

`Tests/Agents/test_builtin_tool_gate.py`, `test_agent_service_review_state_scope.py`, `Tests/Chat/test_console_agent_bridge.py` → 104 passed.

PR #900, opened 2026-07-26
feat(skills): retain script output and surface the file tools to agents (task-584)

rmusser01:
## Summary

Closes **task-584**. The scratch directory was deleted immediately after every run, destroying any artifact a script produced — which blocked the whole class of extract / convert / report skills.

## The premise correction that shaped this

The task was filed as "blocked on a consumer story", and I initially repeated that claim: *there is no general file-read tool*. **That was wrong.** `Tools/file_operation_tools.py` has `ReadFileTool`, `ListDirectoryTool` and `WriteFileTool`, all confined to a configurable `[tools] file_sandbox_root`. The narrow truth I'd overstated: they were registered on the global `ToolExecutor` but `BuiltinToolProvider` hardcoded Calculator + DateTime, so the agent loop never saw them — a wiring gap, not an absence.

That changed the design for the better. Instead of inventing a new contained read surface, retained output goes **where the existing contained tools already look**.

## What changed

**Output is retained.** Each run gets a directory under the file-tool sandbox root (default `<user data dir>/tool_sandbox/skill_script_output/`). Retention is bounded by **run count** — 20 most recent, pruned right after each run. A run producing nothing keeps no directory, so idle runs don't consume a slot.

**The result reports it.** `ScriptRunResult` gains `output_dir` + `output_files`, populated by the *service* (the runner stays skills-agnostic). The tool result lists **names and sizes only, never contents** — that string enters the model's context, and a script's artifact is not trust-reviewed material.

**The file tools are surfaced to the agent loop**, behind the *same* `[tools]` gates that already govern them (`read_file_enabled`, `list_directory_enabled`), both defaulting to **disabled**. This changes reachability, not the default posture. Leave them off and output is yours alone; turn one on and a skill can read its own artifacts back — no new permission dimension.

## Two things that only surfaced by building it

**The child's `HOME` was the run's cwd.** macOS Python therefore wrote `Library/Caches/com.apple.python/**.pyc` straight into it. Invisible while the directory was deleted; it would have polluted *every* artifact listing once retained. The child now gets a throwaway HOME that is always removed.

**581's cancellation test failed — legitimately.** Its script writes marker files, so 584 deliberately retains that directory. I re-pointed the test rather than deleting it: the anti-race property it exists to catch (nothing deletes the directory out from under a live child) is still asserted; only "the directory disappears" changed, which is precisely what this task changes.

## Testing

New: `Tests/Skills/test_script_output_retention.py` (5) and `Tests/Agents/test_builtin_file_tools.py` (4), both RED-first — including a test pinning that the default posture stays unchanged.

Tests/Skills + Tests/Agents: **669 passed** · ruff clean.

---
## Summary by cubic
Retains per-run skill script output under the file-tool sandbox and reports a safe name/size listing to the agent, with bounded retention. Tool exposure is unchanged; closes task-584.

- **New Features**
  - Persist per-run output under `<user data dir>/tool_sandbox/skill_script_output/`; keep the 20 most recent and prune after each run; runs with no artifacts keep no directory.
  - `ScriptRunResult` now includes `output_dir` and `output_files` (names and sizes only). The console tool reports the count, listing, and output directory. Existing `read_file`/`list_directory` can reach this location when enabled.

- **Bug Fixes**
  - Child process uses a throwaway `HOME`/`TMPDIR`; cleanup is guaranteed on all paths (including spawn failures), preventing cache junk and temp leaks.
  - Tests and docs updated; the cancellation test now asserts no cleanup race, and a new regression test ensures the throwaway home is removed.

<sup>Written for commit 23356255f29423099951ac2f4456ba581c3c27b2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/900?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Split per your call: the file-tool surfacing has moved to its own PR so the posture change is reviewed on its own merits rather than riding along with output retention.

**#900 is now retention-only** — per-run output directories under the file-tool sandbox root, count-bounded retention, the name/size listing in the tool result, the throwaway-HOME fix, and docs. It does not touch what the agent can reach.

Retention does not depend on the split-out change: output still lands under the file-tool sandbox root, which is what makes it reachable by tooling the app already has. The docs were reworded to state that reachability without claiming the tools are exposed.

rmusser01:
Fixed in the rebased branch, and it was a real flaw in code I added.

**home_dir cleanup not guaranteed** — correct. The throwaway HOME I introduced for output retention was removed only on the normal return path, so an exception after `mkdtemp` (`Popen` raising `OSError`, or any of the paths that can throw once the directory exists) leaked a temp directory that would accumulate over repeated failures. The body is now wrapped in `try/finally`, with a regression test that forces `Popen` to raise and asserts nothing is left behind.

The other two findings on this PR landed before I split the file-tool surfacing out — they apply to **#902** now, not here, and I'm addressing them there. The security one (ListDirectoryTool following symlinks out of the sandbox root) in particular belongs with the PR that makes that tool reachable.

PR #870, opened 2026-07-25
Egress: scheme/port-aware same-origin for credential decisions (task-568)

rmusser01:
## Summary

Follow-up filed from the Qodo adjudication on #862: credential-stripping decisions on redirect hops compared origins by hostname only, so a same-host HTTPS→HTTP downgrade (or same-host different-port) redirect kept `Authorization`/`Cookie`. This fixes it **centrally** in `Utils/egress.py` and every consumer of the primitive, instead of forking the policy inside Image_Generation.

**Stacked on #862** (`claude/image-gen-followups-engine`) — the Image_Generation consumers it upgrades only exist there. Merge #862 first; GitHub will retarget this PR to dev automatically.

### task-568 (`f3a058a4c`)
- New `Utils/egress.py` primitives `origin_of()` / `same_origin()`: scheme + host + **effective** port with default-port normalization (`https://h` == `https://h:443`), case-insensitive scheme/host, userinfo stripped, conservative fail-closed — any parse ambiguity is NOT same-origin, so credentials strip.
- All four `guarded_fetch_*` hop loops in egress plus `Image_Generation`'s `fetch_json`/`fetch_image_bytes` now make their `_hop_headers`/cookie decisions with the new comparison. SwarmUI's stricter local `_same_origin` (the precedent Qodo cited) now delegates to the central primitive — behavior-identical, verified against its old `parsed.port or default` semantics.
- **Blocking policy untouched**: `host_of`/`origin_set`/`evaluate_url_policy`/trust matching are byte-identical (verified by diff hunk boundaries), and zero pre-existing tests were modified — structurally confirmed (no deleted test lines in the diff). Operator-configured local backends keep credentials through their own genuinely-same-origin redirects, regression-pinned.
- Consumer sweep audited independently in review: 19 hits — 7 credential-decision sites upgraded, 15 correctly left host-based (SSRF trust grants and log labels only).

## Testing
- `Tests/Utils/test_egress.py` 67 passed (+19 new: downgrade/port-change strip, default-port-equivalence keeps, IPv6/userinfo/unparseable fail-closed); `Tests/Image_Generation/` 83 passed / 6 skipped (+7 new at the `fetch_json`/`fetch_image_bytes` level).
- Red first: git-stash revert reproduced the pre-fix `Authorization`/`Cookie` leak on a same-host HTTPS→HTTP hop.
[...]

PR #901, opened 2026-07-26
Settings ▸ Image Gen: backend/key/defaults editor with probes (spec'd, SDD-built)

rmusser01:
## Summary

A new Settings rail category that safely edits the `[image_generation]` config — the gap behind UAT findings task-620/621 (users previously hand-edited nested TOML blind). Spec: `Docs/superpowers/specs/2026-07-25-image-gen-settings-page-design.md` (user-approved); plan: `Docs/superpowers/plans/2026-07-25-image-gen-settings-page.md`. Built as 6 SDD tasks, each independently reviewed, plus a whole-branch final review and a live tmux smoke.

### What ships
- **Backends block** — all six backends with honest configured/not-configured badges (reuses the listing logic), Enabled checkboxes, default-backend selector, and per-backend **Test** probes: sanitized closed-set badges (`Reachable` / `Auth failed` / `Unreachable: connection refused|timeout|HTTP n|blocked by egress policy`), egress-policy-guarded with the adapters' exact trusted-origins shape, one-at-a-time, ephemeral, probing **current form values** (including pasted-but-unsaved keys). sd.cpp probes are filesystem-only.
- **Schema-driven backend editor** — curated v1 fields per backend; unset fields show the **resolved effective value as placeholder** (the task-620 lesson) — including on fresh installs, where a new unmerged file-table reader defeats the default-TOML deep-merge that otherwise makes every field look "set".
- **API keys** per the Providers & Models convention: masked paste-to-save, never re-echoed, source line from a new loader `key_sources` record (`env:<VAR>` / config / keyring / missing — one implementation of precedence), Clear deletes the key (swarmui also clears the legacy `api_key` spelling so a stale fallback can't resurrect a cleared credential).
- **Draft/dirty Save/Revert** joining the Settings hub's existing lifecycle (rail marker, `r` shortcut properly routed): single-write `save_sections` + key **deletions** (never `""` sentinels — emptying a saved field deletes it), config-cache reset so the running Console picks changes up immediately, enabled/default coupling validation, inline errors on all numeric fields.
- **Loader fixes along the way**: per-backend `key_sources` (additive); the swarmui config token now actually resolves (`swarm_token`, with `api_key` back-compat) — previously the documented spelling was silently inert, the exact bug class this page exists to kill.

### Review-caught before merge
- Final review (most-capable-model pass) caught a **Critical** all six task reviews missed: the loader read hardcoded `api_key` for every backend, making the panel's swarmui token writes inert — fixed with a round-trip drift test that closes the exact hole. Also: empty-string writes destroying saved keys (now deletions), stale read-only copy, order-sensitive `enabled_backends` diff.
- Live smoke caught a real-app-CSS layout defect the test harness structurally couldn't (DestinationHarness loads no app CSS): an unscoped global `Checkbox{width:100%}` rule plus implicit `1fr` starvation rendered every row's Enabled/Test at zero width. Fixed via scoped app-tier TCSS, verified live at 235 and 120 cols, and pinned with a **real-app** (`TldwCli.run_test`) width-parameterized regression test that goes red without the fix.
- Re-smoked independently after the fix: all controls visible/clickable; a live SwarmUI probe returned the exact sanitized `Unreachable: connection refused` badge in-row.

### Security posture
The load-bearing property — an env/keyring-resolved secret can never reach the config file or a log — is enforced structurally (`diff_to_sections` cannot see the effective config) and pinned by tests that were strengthened after review found the original pin trivially green (empirically red-verified by injecting the leak).

## Testing
- 427+ across the new suites and neighbors: panel+defaults 102, Image_Generation 109 (incl. 32 loader tests), settings hub 249 — all green post-rebase; `ruff` clean (2 pre-existing findings at base); app import clean; CSS bundle reproduces (two rebase conflicts on the generated bundle resolved by regeneration).
- Follow-ups filed as **task-653** (probe badge precision, stale-badge race test, Select gating, pending-deletion signaling, sibling CSS-blind-spot audit, <120-col fit, v2 template management).

---
## Summary by cubic
Adds a Settings › Image Gen category to safely edit the `[image_generation]` config with per‑backend editors, API key management, sanitized Test probes, and Save/Revert drafts. Addresses UAT findings (task‑620/621) and ensures secrets never leak into config or logs.

- **New Features**
  - Backends block: enable/disable, default selector, configured status, and one‑at‑a‑time Test probes with closed‑set badges; probes use current form values, respect egress policy; `sd.cpp` uses filesystem checks.
  - Schema‑driven per‑backend editor with clear placeholders from the unmerged user config so unset vs default is obvious.
  - API keys: masked paste‑to‑save; never re‑echoed; key source line (`env:<VAR>` / config / keyring / missing); Clear deletes keys (swarmui also removes legacy `api_key`).
  - Draft editing with Save/Revert; writes only changed fields; clearing a field deletes it; numeric inputs validated; config‑cache reset applies changes immediately.

- **Bug Fixes**
  - Loader now reads swarmui `swarm_token` correctly (with `api_key` fallback).
  - No secret copy on save: diff vs raw config only; never write empty‑string sentinels; normalized `enabled_backends` order; fixed stale read‑only copy.
  - Settings UX: revert shortcut fully recomposes; default‑backend Select refire guarded; inline errors for global numeric fields; clearing the default refreshes row markers.
  - Probes: sanitized logs (type + backend only) and badges; egress‑blocked and transport errors mapped to fixed strings.
  - URL fields validated with the shared `input_validation.validate_url`.
  - Cached raw baseline for the category session to avoid full config reloads on each keystroke.
  - CSS: scoped fixes so Enabled/Test controls render and are clickable; panel layout uses the app’s scroll owner; real‑app regression test added.

<sup>Written for commit 6840c2cb3bb44ddd229aefdc71f502c164859dc5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/901?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Qodo adjudication: **5 of 6 fixed** in `7788eee47` — stale ★ Default marker now clears on Select.NULL (test-pinned); the probe worker logs `type(exc).__name__` only (the spec's keys-never-enter-logs contract); the per-keystroke full-config deepcopy is replaced by a per-category-session baseline cache invalidated exactly at entry/save/revert (each invalidation point red-verified individually; adapter-load-count test pins N-keystrokes→1-load); docstrings completed; and the ad-hoc urlparse check now delegates to the shared `input_validation.validate_url` (strict superset — also rejects embedded credentials and out-of-range ports, consistent with the egress-side port rule from #870). **Declined**: rule 497145 (path_validation for the sd.cpp probe paths) — operator-owned local binary/model paths with no base directory to confine to; the no-confinement rationale is a standing comment in the code and matches this rule's prior declines (#862/#867/#884) plus the adapter/listing precedent using plain Path checks for the same fields.

PR #902, opened 2026-07-26
feat(agents): surface the sandboxed file tools to the agent loop (task-654)

rmusser01:
## Summary

Split out of **#900** so the posture-relevant part is reviewed on its own merits.

`Tools/file_operation_tools.py` provides `ReadFileTool` and `ListDirectoryTool`, both confined to a configurable `[tools] file_sandbox_root` and registered on the global `ToolExecutor` — but `BuiltinToolProvider` hardcoded `CalculatorTool` + `DateTimeTool`, so **the agent loop never saw them**.

That gap has a concrete cost: skill-script output is retained under that same sandbox root (task-584) precisely so the app's existing contained tooling can reach it, and today nothing in the agent loop can.

## Scope of the change

| | |
|---|---|
| **Changes** | Reachability — the tools can now appear in the agent's catalog |
| **Does not change** | The default posture. Both gates (`read_file_enabled`, `list_directory_enabled`) already existed and already default to **disabled** |

With no config, the agent sees exactly what it saw before — pinned by a test. Gates are independent, and a tool that fails to construct is simply absent rather than breaking provider construction.

`WriteFileTool` is **deliberately not surfaced**: granting the agent loop filesystem writes is a larger decision than making retained output readable, and nothing here needs it.

## The non-obvious part

Both names are pinned in `_SHADOWED_BUILTIN_NAMES`, and this is worth reviewing carefully:

The shadowed-builtin drift guard (task-580) constructs a `BuiltinToolProvider` with **default** config — so a *config-gated* tool is invisible to it. The guard would never have flagged these names, and it stays green either way. Without the explicit pin, a skill named `read_file` would silently shadow a real builtin the moment a user enabled the gate.

A test enables both gates and asserts coverage, which is the only way to catch this class of gap.

## Testing

`Tests/Agents/test_builtin_file_tools.py` (5), RED-first — including one asserting the default posture is unchanged.

Tests/Agents + Tests/Library: **962 passed** · ruff clean.

---
## Summary by cubic
Expose the sandboxed file tools to the agent loop behind existing `[tools]` gates so agents can read retained script output, and harden `ListDirectoryTool` to block symlink traversal outside the sandbox; default posture remains unchanged (task-654).

- **New Features**
  - `BuiltinToolProvider` now surfaces `ReadFileTool` and `ListDirectoryTool` when `read_file_enabled` / `list_directory_enabled` are enabled.
  - Tools construct defensively; pin `read_file`/`list_directory` in `_SHADOWED_BUILTIN_NAMES`; `WriteFileTool` is not surfaced.

- **Bug Fixes**
  - `ListDirectoryTool` no longer descends into symlinked directories and re-validates that children resolve under the sandbox root.

<sup>Written for commit 67cfff5151aca775898437359fbfa5867990c702. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/902?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Agents/test_builtin_file_tools.py
+++ b/Tests/Agents/test_builtin_file_tools.py
@@ -74,3 +74,32 @@ def test_gated_tool_names_are_covered_by_the_shadow_guard(tools_config):
     assert gated <= _SHADOWED_BUILTIN_NAMES, (
         f"gated builtin tool names not covered: {gated - _SHADOWED_BUILTIN_NAMES}"
     )
+
+
+@pytest.mark.asyncio
+async def test_list_directory_does_not_follow_symlinks_out_of_the_sandbox(
+    tmp_path, monkeypatch
+):
+    """Containment: a planted symlink must not enumerate outside the root.
+
+    Surfacing this tool to the agent loop makes the escape reachable by an
+    agent, so it is fixed here rather than left to the tool's own PR: a
+    symlink created inside file_sandbox_root previously let the recursive
+    walk list files anywhere on disk.
+    """
+    import tldw_chatbook.Tools.file_operation_tools as fot
+
+    sandbox = tmp_path / "sandbox"
+    outside = tmp_path / "outside"
+    sandbox.mkdir()
+    outside.mkdir()
+    (outside / "OUTSIDE_SECRET.txt").write_text("x", encoding="utf-8")
+    (sandbox / "escape").symlink_to(outside)
+    (sandbox / "legit").mkdir()
+    (sandbox / "legit" / "inside.txt").write_text("y", encoding="utf-8")
+
+    monkeypatch.setattr(fot, "_resolve_sandbox_config", lambda: str(sandbox))
+    listing = str(await fot.ListDirectoryTool().execute(directory_path=".", recursive=True))
+
+    assert "OUTSIDE_SECRET" not in listing, "symlink escaped the sandbox root"
+    assert "inside.txt" in listing, "legitimate nested listing must still work"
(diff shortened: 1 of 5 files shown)


rmusser01:
Both findings that moved here from #900 (they were raised before the split) are addressed.

**Symlink escape — confirmed and fixed.** I reproduced it before touching anything: a symlink planted inside `file_sandbox_root` let `ListDirectoryTool`'s recursive walk enumerate an out-of-sandbox file, which duly appeared in the listing. This is exactly why splitting the PR mattered — the flaw was latent while nothing in the agent loop could call the tool, and **this PR is what makes it reachable**, so it gets fixed here rather than deferred to the tool's own history. The walk no longer descends into symlinked directories and re-checks that each child resolves under the root; a regression test asserts the out-of-sandbox file is gone while legitimate nested listing still works.

**Test docstrings — declined, with measurements.** The finding asks for Google-style `Args:` sections documenting pytest fixture parameters. Measured in the suite this PR touches: of **106** parameter-taking tests in `Tests/Agents`, 40 carry a docstring and **0** document `Args:`. (`Tests/Skills` is the same shape: 0 of 21.) Adding them here would make these files the sole outliers rather than bringing them into line.

The CLAUDE.md docstring rule is applied to public callables, and it *is* honoured in this PR's production code — `_is_within` carries full Args/Returns. Fixtures are injected by the framework rather than passed by callers, so documenting them describes the harness, not a contract. Happy to reverse if the maintainer wants the convention changed, but that should be a suite-wide sweep, not one file.

PR #906, opened 2026-07-26
feat(mcp): built-in tool permissions UI — persistent allow/deny for agent:builtin (TASK-627)

rmusser01:
Closes **TASK-627**. Surfaces the `agent:builtin` namespace in the MCP workbench's Permissions matrix and makes persistent allow/deny writable **and reversible** — unblocking TASK-545/P2 from session-scoped decisions only.

## Why this was blocked

P1 shipped built-in tool gating deliberately restricted to session scope, because the workbench renders rows from the live MCP catalog: a synthetic `agent:builtin` key never appeared there, so a persistent decision could not be undone in-app short of hand-editing JSON. Three concrete gaps stood in the way, all verified:

1. **Rows come from the MCP catalog only.** `_build_permission_rows` groups by `tool.server_key` from the `tools: list[HubTool]` argument — a key absent from that list is never a candidate. No filter to relax; it simply never enters the loop.
2. **The resolver is hardcoded.** `effective_tool_states()` applies `resolve_effective_state` to everything. Feeding built-ins through it would resolve them with MCP semantics (ask-floor + hash check) — *wrong*, not merely absent.
3. **Persistent `allow` was blocked at two layers**, both demanding a `definition_hash` that `resolve_builtin_state` never reads.

## What ships

**A hash-free-namespace exemption.** `HASH_FREE_SERVER_KEYS = frozenset({BUILTIN_TOOL_SERVER_KEY})`; the store and service waive the `definition_hash`/`tool=` requirement for `allow` **only** for keys in that set. The hash is a rug-pull guard against a *remote* server mutating a tool after you trusted it — meaningless for in-process code (an attacker who can change it already has execution) and actively harmful, since a stored hash would force a re-prompt on every release that edits a docstring. MCP's guard is unchanged for every other key, and the set's *contents* are pinned by test: adding a remote namespace there is the one way this becomes a real weakening.

**A settings-time enumerator** (`builtin_permission_rows`) resolving via `resolve_builtin_state` only, listing **orphaned** stored entries so a decision for a since-removed tool stays clearable, deterministic, never raising on a malformed payload, and never constructing the lazy gate.

**A built-in section in the Permissions matrix**, built by a **sibling** method whose rows are *appended after* `_build_permission_rows()` returns. Built-ins never enter the `effective` map or `_last_hub_tools`, so it is structurally impossible for one to reach `effective_tool_states()` — not merely avoided by convention.

**Persistent allow/deny from the UI** — which fixed a real dead end (below).

**`needs-decision` flagging** for approval rows a bulk action could not apply to, plus a new `Select.Changed` handler to clear it.

## A dead end this fixed

`cycle_ui_state` is a strict ring **Inherit → Allow → Ask → Off**. From the default state the **first** Space press on a built-in row landed on `allow`, which the `_tool_for() is None` guard rejected before any write — with the toast *"Tool is no longer in the catalog"*, for a tool plainly present. Because you could not step past `allow`, `ask` and `deny` were unreachable too: every built-in row was a permanent dead end, not a half-wired one. A reviewer found this by probing three consecutive presses (no state change, no store write) after an implementer's report claimed the opposite. The regression test asserts the pre-state is `None`, so it exercises the real transition rather than a pre-seeded one.

## Informed by a code-level spike

Two third-party agent frameworks were read at source level (not README level) before designing: **Pi** and **CheetahClaws** (a Python reimplementation of Claude Code's loop with ~28 built-in tools plus MCP).

CheetahClaws decides permissions by **hardcoded built-in tool-name matching**, so MCP tools (`mcp__<server>__<tool>`) match nothing and inherit each mode's fall-through: over-cautious in `auto`, and in **`plan` mode silently auto-approving mutating external tools** — breaking the one guarantee that mode exists to provide. It *has* the metadata to do better (`ToolDef.read_only` from `readOnlyHint`) but consumes it only for result caching.

That is the failure this PR's architecture prevents, and it is why the per-namespace resolver is worth its plumbing. Neither framework has a definition hash at all, which is what settled the exemption above rather than synthesizing a meaningless one.

## Testing

`Tests/MCP/ Tests/UI/test_mcp_workbench.py Tests/UI/test_mcp_permissions_mode.py Tests/UI/test_console_mcp_approval.py Tests/UI/test_chat_approval_card.py Tests/Agents/` → **908 passed** pre-rebase (baseline 854).

Post-rebase, one test fails — `test_batch_row_widgets_have_nonzero_geometry_and_do_not_overlap_under_bundled_css`. It fails **identically on a clean `origin/dev` worktree**, so it is pre-existing and not from this branch. The final reviewer additionally ran five adjacent suites this branch's run set omitted: 533 passed.

The rebase surfaced two real conflicts. The **CSS bundle** is generated, so rather than hand-merge it I regenerated from the merged component sources and verified with `check_bundle_sync.py`. The **agent-settings task** had been renumbered three times by concurrent sessions (634 → 635 → 659); dev's canonical is 659, so this branch adopts it and skips its own obsolete renumber — leaving the unrelated `task-635 - ensure_imported_profile` untouched.

## Follow-ups filed (660-663)

**660** is a defect a reviewer found and no existing task owned: selecting a built-in row calls `show_tool(None)` and wipes the inspector, and the preview sentence's override count is computed before the built-in section is appended — so a persistent built-in override changes the table cell but not the summary line. **661** a hard-floor never-allowable category (with task-582's `MAX_… + min()` clamp as the in-repo shape to copy). **662** permission modes — which, if built, must be driven by `risk_tags` and fail closed, the two things CheetahClaws got wrong. **663** a standing constraint that the model must never be able to widen its own permission posture (CheetahClaws' `EnterPlanMode`/`ExitPlanMode` are always-auto-approved and mutate the mode directly).

Built with subagent-driven development: 6 TDD tasks, a spec+quality review each, an adversarial whole-branch review, and a fix wave.

---
## Summary by cubic
Adds a built-in tool permissions UI to surface `agent:builtin` in the MCP workbench and enable persistent, reversible allow/deny for built-in tools. Implements TASK-627, unblocks TASK-545/P2, and fixes inspector/preview bugs for built-in rows.

- **New Features**
  - Show a “Built-in (agent runtime)” section in Permissions mode listing built-in tools and orphaned stored entries; rows resolve via the built-in resolver only.
  - Persist allow/deny for built-in tools from the UI; changes take effect on the next turn.
  - Introduce `HASH_FREE_SERVER_KEYS = {BUILTIN_TOOL_SERVER_KEY}` so `allow` for `agent:builtin` needs no `definition_hash` in the store or service.
  - Flag approval rows a bulk action couldn’t apply to with a `needs-decision` badge; cleared by the row’s own change; added CSS and handler.
  - Add a settings-time enumerator (`builtin_permission_rows`) to feed the built-in section safely and deterministically.

- **Bug Fixes**
  - Built-in rows no longer dead-end on the first `allow` step; allow/ask/deny are reachable and persist correctly.
  - Selecting a built-in row now shows inspector details instead of clearing the panel.
  - The preview sentence now counts built-in overrides, keeping the summary consistent with the full matrix.

<sup>Written for commit 0055afdc5a80e2e0c6c925fadced83ea623dc3ae. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/906?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Thanks — both Bugs fixed in `0055afdc5`, rule violation declined with evidence.

Worth noting both Bugs were independently found by our own whole-branch review before this one, and were filed as TASK-660. Two reviewers converging on them changed my read: they are visible in the feature this PR ships, not deferrable polish. So they are fixed here and 660 narrows accordingly.

### 1. "Inspector clears built-in rows" — **fixed**

Correct, and worse than it reads: before this branch no row could blank the inspector, so the newest and most interactive rows became the only ones that do — and the persistent decisions this PR exists to enable were the one thing not inspectable. `on_mcp_permissions_mode_row_selected` resolved via `_tool_for()`, which only searches `_last_hub_tools`, so a built-in row always yielded `None` → `show_tool(None)`.

Built-in rows now route to the inspector's permission view using the `EffectiveToolState` the section already resolved (cached alongside the rows, so the handler does not re-read the store), with `cascade=None` — built-ins have no MCP cascade. MCP row selection is unchanged.

### 2. "Preview ignores built-in overrides" — **fixed**

Also correct. The count was computed inside `_build_permission_rows()` on the MCP-only list, *before* the built-in section was appended, so setting a persistent built-in override changed the table cell but not the summary line — and falsified `update_matrix()`'s documented "ALWAYS summarizes the full, UNFILTERED matrix" contract.

Fixed by threading built-in rows into that call's `extra_override_rows` so they are counted, **without** entering the row list or the `effective` map. That distinction is load-bearing: built-ins must never become members of any structure MCP code reads, because that is what makes it structurally impossible for one to reach `effective_tool_states()` (which applies MCP semantics *and* sets a rug-pull marker `resolve_builtin_state` ignores). Merging them in to fix the count would have traded a cosmetic bug for the architectural one this design exists to prevent.

Four tests added: built-in row selection shows its permission block, a pinned row still clears, the override count includes a persistent built-in override, and the count is unchanged when none is set.

### 3. "payload parsed without Pydantic" — **declined**

The payload here is the MCP permission store's own JSON, and this enumerator deliberately reads it exactly the way the two resolvers that own it already do — `_as_mapping` + `.get()`, the same isinstance-guard chain, imported from `permission_store` specifically so malformed-payload behavior is *identical* rather than re-derived.

Measured:

```
_as_mapping uses in permission_store.py: 18
pydantic uses in permission_store.py / builtin_tool_gate.py: 0
pydantic models anywhere under MCP/: 0
```

There is no Pydantic model for this payload anywhere in the subsystem. Introducing one just for the enumerator would make it the only component parsing this store differently from `resolve_effective_state` and `resolve_builtin_state` — a consistency regression, and a second source of truth for what a valid payload looks like.

The robustness concern behind the rule is already met: a reviewer probed nine malformed shapes directly against `builtin_permission_rows` (`profiles: None`, `default: None`, `servers: None`, a non-dict server entry, `tools` as a list, a bare-string tool entry, junk `state`, and others) — none raise, all degrade to the documented built-in `allow` floor with risk-flooring intact. That is the same guarantee the rest of the store gets.

`Tests/UI/test_mcp_workbench.py Tests/UI/test_mcp_permissions_mode.py Tests/UI/test_mcp_inspector.py Tests/MCP/` → 687 passed.

PR #915, opened 2026-07-26
Image-gen backends: fal.ai + Gemini (nano-banana) + reference-image engine seam

rmusser01:
## Summary

Two new image-generation backends on the established adapter contract — **fal.ai** (queue+poll, default `fal-ai/flux/schnell`) and **Gemini / AI Studio** (`generateContent`, default `gemini-2.5-flash-image` — nano-banana) — plus the **engine-level reference-image seam** (image-to-image input for the new backends). Spec and plan under `Docs/superpowers/` (user-approved); built as SDD tasks with per-task adversarial review, a whole-branch final review (merge-ready, zero Critical/Important), and live UAT to the extent current keys allow.

**Fireworks was requested and then dropped mid-program**: the implementer discovered — and I verified from two independent sources — that Fireworks deprecated its image-generation API entirely on 2026-06-10 (endpoints 401 for everyone). The adapter built against the last-known schema was reverted; spec/plan carry dated drop annotations. FLUX remains served by the existing Together backend.

### Security decisions the branch is built around
- **fal never requests a server-provided URL.** Its queue poll/result URLs are self-built from the configured base + app id (first two model-path segments, cross-verified against fal's own client SDK source) + a charset-validated request id; the vendor's `status_url` is used only as a parsed cross-check that fails loudly on shape drift. This closes the credential-exfiltration channel that naively following queue responses would open.
- **Model ids are validated before URL construction** (new attack surface — the six existing adapters carry the model in the JSON body; these put it in the URL path). Charset allowlists with `..`/query rejection, per backend.
- **Gemini's key lives in the `x-goog-api-key` header only** (never the `?key=` query form Google also accepts — URLs get logged, headers don't; the live-UAT 429 confirmed the error text carries no credential). The header was also added to the central egress `_STRIP_HEADERS` — it's the app's first non-`Authorization` auth header and was silently outside the cross-origin credential-strip net.
- **Reference images validated at the choke point** — capability flags (`fal`/`gemini` true, six legacy false), mime allowlist, 10MB cap, bytes-in-memory only; unsupported backends refuse loudly. Gemini encodes as an `inline_data` part, fal as a base64 data URI. Console attach UX is a follow-up spec; the seam is engine-complete.
- New guarded `fetch_bytes_via_post` helper with a **streaming byte cap** (Content-Length pre-check + mid-read abort — review rejected read-then-check because httpx eagerly buffers with no default limit); the existing streaming helpers were relocated to `http_client` as the shared home.

### Verified-from-live-docs discipline (the task-620 lesson)
Every vendor shape was verified against live documentation during implementation and pinned in tests — including the non-obvious `responseModalities: ["TEXT","IMAGE"]` requirement (image-only silently returns empty parts) and fal's `image_size` object form. Reviewers independently re-verified the claims (one fetched fal's client source to confirm the app-id rule byte-for-byte).

### Live UAT so far (user keys, continuing post-merge)
Settings page renders all eight backends; Gemini auto-configures from `GOOGLE_API_KEY` (the designed env fallback) and its **auth-verified probe returned Reachable against the live models endpoint**; fal's reachability probe returned its exact spec'd badge. Gemini generation reached Google correctly but the key's image-quota bucket is exhausted (429) — full generation + nano-banana reference-edit UAT completes when a quota-bearing key is available; fal end-to-end awaits its key. Opt-in live suites (`TLDW_LIVE_FAL_API_KEY`/`TLDW_LIVE_GEMINI_API_KEY`) are in the tree for exactly that.

## Testing
- `Tests/Image_Generation/`: 237 passed / 9 skipped (the skips are the opt-in live suites) — includes 43 fal tests (poll lifecycle with the self-built-URL pins), 32 Gemini tests (parts iteration, safety-block mapping, header-not-query pin), the streaming-cap matrix, reference-seam validation matrix, and config/key_sources rows.
- Settings suites 113 passed (drift test passed **unmodified** when the new rows landed — the schema-driven design working as intended); egress suite 70 passed (central strip addition, pre-existing tests untouched).
- Follow-ups filed as **task-686** (cap hardening before Console attach UX, fal namespace grammar, the pre-existing chat-side `x-goog-api-key` redirect gap, keyring tests, docstring reword, friendlier 429).

---
## Summary by cubic
Adds two new image-generation backends — fal.ai and Gemini (`gemini-2.5-flash-image`) — with engine-level reference-image support. Drops Fireworks (vendor deprecated), updates settings/config, adds a streaming-safe HTTP helper, and hardens Gemini reference-image handling.

- **New Features**
  - `fal` and `gemini` adapters on the existing contract; registered in the adapter registry.
  - Reference-image seam: backend capability flags, mime allowlist (png/jpeg/webp), and a 10MB cap enforced at the engine choke point.
  - Settings: new rows and reachability/auth probes; env fallbacks (`FAL_KEY`, `GEMINI_API_KEY`/`GOOGLE_API_KEY`).
  - Config defaults and listing checks added; live test opt-ins (`TLDW_LIVE_FAL_API_KEY`, `TLDW_LIVE_GEMINI_API_KEY`).
  - `fetch_bytes_via_post` with streaming byte cap for safe large responses.

- **Security**
  - fal: self-built poll/result URLs (never follow vendor-provided URLs); strict model-id validation.
  - Gemini: `x-goog-api-key` header only; stripped on cross-origin redirects; added to central `_STRIP_HEADERS`.
  - Reference images validated centrally (capability, mime, size, non-empty) before adapter dispatch.
  - Gemini: enforces the bytes-only reference-image contract (no disk reads); raises on contract violations.

<sup>Written for commit 040aa4de31b111fe502807c600f1c63abb12b651. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/915?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/plans/2026-07-26-imagegen-fal-gemini-fireworks.md
+++ b/Docs/superpowers/plans/2026-07-26-imagegen-fal-gemini-fireworks.md
@@ -1,0 +1,193 @@
+# fal.ai / Gemini / Fireworks Image Backends Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Three new image-generation backends (`fal`, `gemini`, `fireworks`) on the established adapter contract, plus the engine-level reference-image seam, wired through config, listing, registry, and the Settings page.
+
+**Architecture:** One adapter file per service; everything else is rows in existing tables (`_SECRETS`/`_NON_SECRET`, registry lazy-map, `BACKEND_IDS`/`FIELD_SCHEMA`, probe dispatch). One new guarded HTTP helper (`fetch_bytes_via_post`) for Fireworks' bytes-returning POST. Reference images open the dormant `ResolvedReferenceImage` seam with per-backend capability flags and fail-loudly choke-point validation. Spec: `Docs/superpowers/specs/2026-07-26-imagegen-fal-gemini-fireworks-design.md` — read it before any task.
+
+**Tech Stack:** Python 3.11+, httpx (sync, via the package's guarded helpers), pytest with fake clients.
+
+## Global Constraints
+
+- All HTTP through the package's guarded helpers (`http_client.fetch_json`, `image_format_utils.fetch_image_bytes`, and Task 1's `fetch_bytes_via_post`): egress `check_url_or_raise` pre-request, self-built URLs carry `trusted_origins=origin_set(url)`, response-extracted URLs are fully enforced (no trust, no credentials), no auto-redirects.
+- **Never follow server-provided URLs with credentials attached** (fal poll URLs are self-built; response URLs are cross-check only).
+- **Model ids are validated before URL construction**: gemini/fireworks `^[A-Za-z0-9._-]+$`; fal additionally allows `/` but rejects `..` segments, leading/trailing `/`, and any of `?#%` or whitespace.
+- API keys never in logs, error text, or URLs (gemini key goes in the `x-goog-api-key` HEADER, never `?key=`).
+- 404/400-model-unknown errors name the attempted model id and the config key (`[image_generation.<backend>] default_model`) — the task-620 pattern; copy `openrouter_image_adapter.py`'s enrichment shape.
+- Negative prompt is appended to the prompt text (openrouter precedent), never dropped.
+- `ImageGenResult.resolved_model`/`resolved_seed` only where the backend genuinely reports them — never fabricate.
+- Defaults (exact): fal base `https://queue.fal.run`, model `fal-ai/flux/schnell`, poll 2s, timeout 120; gemini base `https://generativelanguage.googleapis.com/v1beta`, model `gemini-2.5-flash-image`, timeout 120; fireworks base `https://api.fireworks.ai/inference/v1/workflows/accounts/fireworks/models`, model `flux-1-schnell-fp8`, timeout 120.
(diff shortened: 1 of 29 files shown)


rmusser01:

**Finding 3 (temp_path read without path_validation) + the temp_path half of finding 2: FIXED at the root** in `040aa4de3` — the Gemini adapter's `temp_path` read branch is REMOVED entirely. The spec's engine rule is bytes-in-memory only, and the choke-point validator refuses `content=None` before any adapter runs, so that branch was spec-divergent dead code; the adapter now raises a contract-violation error if a content-less reference somehow reaches it (red-verified, plus an assertion-stub proving no network call follows). No file read exists, so the path-validation question is moot. Adapter/class docstrings (finding 4) also **FIXED** in the same commit.

**Finding 1 (ModelStudio reference-image "regression"): DECLINED.** The flow the finding describes has been dormant since Phase 1: zero production code paths populate `request.reference_image` for any backend (the only constructor is `worker.build_request`, whose two callers never pass it — verified in the Task-3 review record), so nothing reachable regressed. The spec explicitly scopes reference input to the new backends and lists "reference images for the six existing backends" as a non-goal, keeping ModelStudio's per-model map dormant; ModelStudio's own direct capability call is byte-for-byte unchanged.

**Finding 2's bytes_len/empty-content hardening: DECLINED as already-triaged** — the whole-branch final review found the same two edges (cap trusts `bytes_len` rather than `len(content)`; `content=b""` passes the choke point while both adapters still fail loudly) and filed them as **task-686 AC1**, to be hardened before the Console attach-UX program creates the first production constructor. No reachable path exists today.

PR #917, opened 2026-07-26
Watchlists rebuild Phase A: data foundation

rmusser01:
Phase A of the Watchlists screen rebuild. **Data layer only — no UI file is touched anywhere in this branch.** Spec and plan are included as documents.

Design: `Docs/superpowers/specs/2026-07-25-watchlists-console-rebuild-design.md`
Plan: `Docs/superpowers/plans/2026-07-25-watchlists-rebuild-phase-a-data-foundation.md`

## What this adds

- `watchlists` + `watchlist_sources` tables — a watchlist is a named bundle of sources, many-to-many, the unit of organisation and checking.
- `WatchlistBundleService` — CRUD, membership, case-insensitive name collision with auto-suffixing, and a one-shot `folder` → watchlist migration.
- Durable body text on items (`content`, `content_kind`, `content_format`) plus `is_flagged`, and an FTS5 index over title/content/author with a chunked, resumable backfill.
- `persist_subscription_item` — one persistence path replacing two divergent INSERTs.
- `get_watchlist_item_counts` — per-watchlist plus Unassigned/All-sources counts in a single query.
- `local_watchlist_runs` relocated into `SubscriptionsDB` with a `batch_id` column.

## Pre-existing bugs found and fixed along the way

Design and review turned up several defects that predate this work:

- **Foreign keys were never enforced.** The `PRAGMA foreign_keys = ON` ran on a connection closed immediately afterwards, and `BaseDB._get_connection` sets only `row_factory`. Every `ON DELETE CASCADE` in this schema was inert — deleting a subscription silently orphaned its items. Fixed in its own commit.
- **Two item-INSERT paths wrote disjoint column sets**, and neither wrote body text, so what the reader could display depended on which code path fetched the item.
- **`cursor.lastrowid` returned the wrong row** on the upsert's conflict branch.
- **A failed `subscription_filters` rebuild could brick the database** — `CREATE TABLE` runs in autocommit, so the rollback left the table behind and the next open died inside `_initialize_schema`.
- **The FTS delete triggers rejected mutations of pre-existing rows.** External-content FTS5 only accepts `'delete'` for indexed rowids; the index is created empty over existing data, so update/delete of any legacy item — and cascade-delete of its source — failed with `database disk image is malformed`. The delete legs are now membership-guarded.

## Testing

`Tests/DB Tests/Subscriptions Tests/Watchlists Tests/Scheduling Tests/Home` → **671 passed, 1 failed**.

The single failure is `Tests/Watchlists/test_watchlists_navigator.py::test_navigator_has_all_section_buttons`, which fails on this branch before any of these commits — a pre-existing UI-test failure, out of scope for a data-layer change.

## Known follow-ups (not in this PR)

- **Blocks Phase B:** `backfill_items_fts` has no production caller, so pre-existing items are never indexed. Wiring was deliberately deferred so Phase A stays data-layer only.
- `SubscriptionsDB(":memory:")` is non-functional — schema is built on a throwaway connection while `.conn` opens a different, empty database. `WatchlistPreviewService.preview()` uses `:memory:`, so source Preview/dry-run is very likely broken today. Pre-existing, orthogonal to this work.
- `local_watchlist_alert_rules` is still created lazily from a service, contrary to the plan's "no lazily-created tables" constraint. Same treatment as `local_watchlist_runs`, next phase.
- Server watchlist membership has no wire path (`SourceUpdateRequest` has no `group_ids`; group requests are `extra="forbid"`), so watchlist creation is disabled on the server backend.

---
## Summary by cubic
Builds the data foundation for the Watchlists rebuild: new watchlist storage/services, item content + FTS search, and a single, reliable item-persist path; also adds and corrects the Watchlists design/specs and includes the Phase B workbench/shell plan. No UI changes.

- **New Features**
  - Add `watchlists` and `watchlist_sources` (many-to-many) and a `WatchlistBundleService` (CRUD, membership, case-insensitive naming with auto-suffix, folder→watchlist migration, paginated `list_watchlists`).
  - Store durable item body via `content`, `content_kind`, `content_format`, plus `is_flagged`; add FTS5 over title/content/author with chunked, resumable backfill.
  - Introduce `persist_subscription_item` to replace divergent INSERTs; preserves `reviewed`/`ignored`/`ingested` and returns the correct row id.
  - Provide `get_watchlist_item_counts` for per-watchlist + Unassigned/All-sources in one query.
  - Move `local_watchlist_runs` into `SubscriptionsDB` and add `batch_id`.

- **Bug Fixes**
  - Enable foreign-key enforcement on `SubscriptionsDB` runtime connections; guard the `subscription_filters` rebuild (FK OFF/ON with DROP IF EXISTS) to avoid data loss and startup lockups.
  - Fix upsert id reporting with RETURNING; previous `cursor.lastrowid` could return a wrong row.
  - Guard FTS5 delete triggers to tolerate legacy, unindexed rows and recreate triggers to upgrade in place.
  - Item persistence: route scheduled-checks through `persist_subscription_item`; collapse empty JSON fields to NULL to prevent phantom values.
  - Anchor counts on `watchlists` so empty watchlists and sources with no items report zero.

<sup>Written for commit 803bd0ccec3d7e9598402f2c2cb6b2ca7e9272db. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/917?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/plans/2026-07-25-watchlists-rebuild-phase-a-data-foundation.md
+++ b/Docs/superpowers/plans/2026-07-25-watchlists-rebuild-phase-a-data-foundation.md
@@ -12,7 +12,7 @@
 
 - Spec: `Docs/superpowers/specs/2026-07-25-watchlists-console-rebuild-design.md`. Every requirement below is traceable to it.
 - All schema changes are **idempotent** — running twice is a no-op — and live in `SubscriptionsDB`: table creation in `_initialize_schema`, additive migration in `_ensure_watchlists_schema`. No table is created lazily from a service. (Ruling, 2026-07-25: this is why Task 2 relocates `local_watchlist_runs`.)
-- `PRAGMA foreign_keys = ON` is already set (`Subscriptions_DB.py:82`). Cascades are live; do not disable.
+- Foreign-key enforcement is enabled by **Task 1a**. It was previously off at runtime — see that task. Do not disable it.
 - `SubscriptionsDB` uses `threading.local()` connections (`:75`, `:368-370`). Never share a connection across threads.
 - Use `db.transaction()` (`:373`) for writes; it commits on success and rolls back on exception.
 - `subscriptions.tags` is **comma-joined**, not JSON (`:422`). `watchlists.tags` matches.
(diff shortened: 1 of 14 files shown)


rmusser01:
**Full-suite result** (promised in the PR description).

`pytest Tests -q -x` over all 16,373 collected tests: **3,963 passed, 78 skipped**, stopping at one failure after 35m12s.

The failure is `Tests/Evals/test_evaluation_metrics.py::TestEvaluationMetrics::test_instruction_adherence_basic`, and it is **not caused by this branch**. Evidence:

- Passes in isolation on the pre-branch baseline (`61205b94d`).
- Passes in isolation on this branch.
- Passes when the whole `test_evaluation_metrics.py` file is run on this branch (13 passed).
- `Tests/Evals` and `tldw_chatbook/Evals` contain no reference to `Subscriptions`, `Subscriptions_DB`, `item_persist`, or `watchlist` — and this branch changes only four files, all under `DB/` and `Subscriptions/`.

So it is pre-existing cross-test pollution that only manifests in a full-suite run, not a regression here. Worth its own investigation, but out of scope for this PR.

The targeted evidence in the PR description stands unchanged: `Tests/DB Tests/Subscriptions Tests/Watchlists Tests/Scheduling Tests/Home` → 671 passed, 1 pre-existing unrelated UI failure.

rmusser01:
## Review adjudication

**Fixed (`ca9adea2e`)**

1. **Empty JSON containers persisted as `"[]"` instead of `NULL`** — real bug, and mine. Unifying the two INSERT paths silently adopted the looser of two different behaviours: the old DB-side path used truthiness (`[]` → `NULL`), the service-side path used `is not None` (`[]` → `"[]"`). Since `_parse_categories` treats any string as comma-separated, `categories=[]` produced a phantom category `["[]"]`. `_json_or_none` now serialises empty containers to `NULL`. Verified: `categories=[] → None`, `["ai","ml"] → '["ai", "ml"]'`.
2. **`sqlite3.connect` without a context manager** — fixed with `contextlib.closing`. Note the suggested form, `with sqlite3.connect(path) as conn:`, would **not** have fixed it: sqlite3's context manager wraps a *transaction*, not the connection's lifetime, so the connection would still leak.
3. **`list_watchlists` pagination** — added `limit=100, offset=0`, matching `get_subscriptions_by_folder`'s existing signature and defaults. No current caller changes behaviour.
4. **Google-style docstrings** — `create` and the other public methods on `WatchlistBundleService` now carry Args/Returns/Raises, per `CLAUDE.md`. `add_source`'s `sqlite3.IntegrityError` was verified live rather than assumed.

**Declined, with evidence**

5. **`_ensure_watchlists_schema` adds DDL inline** — that helper and its inline-DDL pattern pre-date this branch (3 occurrences on the pre-branch baseline `61205b94d`); it is ADR-018's established mechanism, and this PR follows it rather than introducing it. Changing it is a separate refactor of existing migration architecture.
6. **`list_watchlists` bypasses `transaction()`** — it is a read. `SubscriptionsDB` reads via `self.conn` at 6+ existing sites (`:679`, `:850`, `:1003`, `:1065`, `:1092`, `:1263`). Writes in the new service do use `transaction()`. This matches the file's own convention.
7. **`PRAGMA table_info` uses an f-string** — SQLite does not accept bind parameters in `PRAGMA`, so this is not expressible as a parameterised query. The interpolated value is a hardcoded literal in test-helper code, never user-derived. There is an explicit ruling on this in the plan's Global Constraints.

## CI failures are pre-existing

The three reported failures — `test_shared_collapsible_header_hover_uses_non_obscuring_surface_contract` and both `test_chat_tab_action_hover_states_use_neutral_readable_surface` parametrisations — reproduce on **latest dev (`006b31ed2`)** with no changes applied. They are CSS hover-contract tests; this PR changes four files, all under `DB/` and `Subscriptions/`, and touches no UI or CSS.

PR #921, opened 2026-07-26
TASK-545 P2: port mutating tools behind the built-in permission gate

rmusser01:
## Summary

TASK-545 **Phase 2**: puts real mutating tools behind the built-in permission gate, and makes that gate live for the first time.

The gate (shipped in P1, #890) has been **inert in production by design**. `resolve_builtin_state` floors an inherited `allow` to `ask` only when a tool's `risk_tags` intersect a high-risk set — and no shipped tool declared any tags. So `BuiltinToolGate.check()`'s `ask` branch, the review hook's built-in rows, the approval card, and task-628's `stamp_scope` were all exercised **solely by tests**. Tagging tools is what makes that machinery reach real users. That, not the mechanical registration, is the risk this PR is organized around.

## What changed

1. **`BUILTIN_HIGH_RISK_TAGS = HIGH_RISK_TAGS | {"reads"}`**, consulted **only** by `resolve_builtin_state`. `resolve_effective_state` (MCP) is untouched — widening the shared set would make remote tools carrying the tag start prompting, which is not this phase's call. The asymmetry is deliberate and tested in both directions.
2. **Five tools tagged**: `write_file`/`create_note`/`update_note` → `("mutates",)`; `read_file`/`list_directory` → `("reads",)`.
3. **Three tools registered** in `BuiltinToolProvider` behind the **pre-existing** default-off `[tools]` gates (`tool_executor.py:735/763/789`) — task-584's precedent verbatim. Names added to `_SHADOWED_BUILTIN_NAMES`, whose drift guard builds a *default-config* provider and so is structurally blind to gated names.
4. **Note tools write as the configured user**, resolved from `load_settings()["USERS_NAME"]` — the same source `app.notes_user_id` uses. Not `get_cli_setting("general", "users_name")`, which reads only TOML and diverges whenever the `USERS_NAME` env var is set.
5. **Integration tests** driving the real tools through the gate, replacing P1's synthetic double.

### A live gap closed in passing

`read_file`/`list_directory` were already registered (task-584) behind default-off gates but carried **no** `risk_tags`. Enabling either gave you a filesystem-read tool that resolved to the `allow` floor and ran **with no prompt at all**. That was live on `dev`, independent of this port.

## Two pre-existing bugs found

**`note_management_tools` could not be imported at all.** It did `from ..config import USER_DB_BASE_DIR` — a name that exists only inside `config.py`'s `CONFIG_TOML_CONTENT` **string literal**, never as a module-level binding. That is why these tools have no tests anywhere: the module never loaded. Fixed to `get_chachanotes_db_path().parent`, mirroring `app.py`.

**The registration loop hid it.** `except Exception: continue` made a genuine import breakage indistinguishable from "the gate is off" — no log, no error. It now logs (gate-off stays quiet, so a default startup emits nothing). Sabotage-verified in both directions.

## Correction to the design spec

The spec originally justified change 4 by claiming a configured user's agent-created notes "land in a bucket their Notes UI never reads." **That was wrong.** The `notes` table has no user column (`ChaChaNotes_DB.py:490`), and `NotesInteropService.add_note` documents that "the user_id will be used as the client_id". Notes were always *visible* — they were **misattributed** to a fabricated `default_user` client, which is what sync and last-write-wins conflict resolution key off. The change is still correct; the stated reason was not. Corrected in the spec, resolver docstring, and test docstring, since a maintainer trusting the old wording could write a "user's notes" query that silently returns everyone's.

## Verification

- `Tests/Agents/ Tests/MCP/ Tests/Tools/ Tests/Library/test_library_skills_state.py` → **709 passed**
- Default posture unchanged: with no `[tools]` keys, the catalog is exactly `calculator` + `get_current_datetime`
- No bypass: the final review confirmed only two sites in the repo execute a `Tool`, and `ToolExecutor.execute_tool_calls` has **zero** production callers
- Discriminating tests sabotage-verified: resolved `deny` beating a permitting stamp, MCP not learning the new tag, and the new registration-failure log

## Scope

**Out of scope, now tracked by TASK-694:** `rag_search`, `web_search`, `search_notes`, `code_audit`. P2's AC originally named eight tools; narrowing it to the five delivered would have dropped those four off the board, so they get their own task — including the risk-tag decision each needs (`web_search` has no existing tag describing outbound network access).

TASK-545 stays **open**: P3 (the dead `[tools]` config, and System A's fate) remains.

## Follow-ups filed

691 `expected_version` default-of-1 · 692 per-call `CharactersRAGDB` + import-time `mkdir` · 693 surface the sandbox root · 694 the remaining four tools · 695 approval card can truncate `write_file`'s target path out of view · 696 review-hook and real-DB worker-thread coverage gaps

> Note on IDs: these were first filed as 687–690, which collided with `feat/library-server-ingest-684-1` and `feat/watchlists-phase-a-data-foundation`. `origin/dev`'s max was 686, but the max across all 101 remote branches was 690. Scanning dev alone is not sufficient.

---
## Summary by cubic
Ports `write_file`, `create_note`, and `update_note` into the built-in tool provider behind the permission gate and makes the gate live by tagging tools (including `read_file`/`list_directory`) with risk classes. Default posture is unchanged; users now get approval prompts for reads and mutations as required by TASK-545 P2.

- **New Features**
  - Added `BUILTIN_HIGH_RISK_TAGS = HIGH_RISK_TAGS ∪ {"reads"}` used only by `resolve_builtin_state` (MCP stays unchanged).
  - Tagged tools: `write_file`/`create_note`/`update_note` → `("mutates",)`; `read_file`/`list_directory` → `("reads",)`.
  - Registered mutating tools behind existing `[tools]` gates and added to `_SHADOWED_BUILTIN_NAMES`; tests cover approval flow and worker-thread execution.

- **Bug Fixes**
  - Closed a live gap where enabling `read_file`/`list_directory` executed without prompting.
  - Fixed `note_management_tools` import and deferred the notes DB base dir to call time via `get_chachanotes_db_path().parent` to avoid import-time side effects.
  - Note tools now attribute to the configured `USERS_NAME`; tool registration now logs build failures instead of silently skipping.

<sup>Written for commit 8927ed501be05b5dea93bf9a3717d25e069a67e1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/921?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## Review response

**1. Import-time mkdir breaks tool import 🐞 — FIXED** (`e8867537b`)

Agreed, and this was independently flagged by the whole-branch review before Qodo ran, which raises my confidence it's real. The severity is exactly as described: `get_chachanotes_db_path()` → `get_user_data_dir()` → `mkdir(parents=True, exist_ok=True)`, so *importing* the module created a directory, and an unwritable `$HOME` or `[paths] data_dir` raised **during import** — which `tool_catalog`'s registration loop catches and skips, surfacing as "`create_note` is missing" rather than as a normal error from an actual call.

Replaced the module-scope constant with `_notes_db_base_dir()`, resolved per call, matching the sibling convention already in `file_operation_tools._tool_sandbox_root`. Two regression tests added, both **sabotage-verified** — restoring the module-scope form makes them fail with `AssertionError: get_chachanotes_db_path() called at import time` and `PermissionError: read-only file system` respectively, and both pass on revert.

**2. `risk_tags` docstrings missing `Returns:` 📘 — declined, with a measurement**

This repo does not follow that convention for properties. Measured across `tldw_chatbook/` by AST:

| `@property` docstrings | count |
|---|---:|
| **with** `Returns:` | 10 |
| **without** `Returns:` | 124 (**93%**) |

The closest precedent is decisive: `Tool.name` and `Tool.description` — the direct siblings of `risk_tags` on the same ABC in `tool_executor.py` — are single-line docstrings with no `Returns:` section. The `risk_tags` contract *is* documented with a full Google-style `Returns:` block, once, on the base class where it belongs; the five overrides are one-line statements of why a particular tool carries its tag, and the return value is visible on the line below.

Adding `Returns: ("reads",).` under `"""Reading arbitrary sandbox files is a disclosure risk."""` would restate the next line of code and diverge from 93% of the codebase.

Happy to apply it if the project wants the convention changed — but that's a repo-wide decision, not one to make in five new properties.

PR #923, opened 2026-07-26
Bring server ingestion and web clipping into the Library ingest canvas (tasks 684.1-684.3)

rmusser01:
Brings server-backed ingestion and web clipping into the Library ingest canvas, so the second ingestion window has nothing left that the canvas cannot do. Tasks **684.1**, **684.2**, **684.3** complete; **684.4** (deleting that window) is deliberately left for its own PR.

Follows #907 and #910, which fixed the P0/P1 findings from the ingest UAT.

## What this adds

**Server-backed file ingest (684.1)** — a Library ingest can run on a configured server, reusing the whole existing canvas: source picking, per-type options, pre-flight, queue. `Library/server_ingest_request.py` maps a source plus the form's option snapshot onto the kwargs the service already takes, so the request shape is testable with no UI, no I/O and no server.

**Remote jobs in the same queue (684.2)** — schema v3 adds `origin`/`remote_job_id`/`batch_id` and admits the server's `cancelled` state. `Library/server_ingest_reconcile.py` folds server statuses back into the local registry and is Textual-free, so the whole decision layer is exercised without an app, a worker or a server; the poller only fetches and hands over.

**Web clipping (684.3)** — a URL is now just a source in the existing path field rather than a fourth mode. Routing splits on what the source *is*: a page goes to the clipper endpoint, a file to the ingest-jobs API, under the one backend switch.

## Two findings worth a reviewer's attention

**Server ingest was unusable as first built.** Runtime policy declares `media.ingestion_jobs.launch.server` with `required_source="server"`, so the service refuses the launch while the Library runtime is local. The switch was offered, the user could pick it, and every submission failed with *"requires server mode"*. 941 unit tests missed it, and direct API probes succeeded because they bypass the app's enforcer — only driving the real TUI exposed it. The opt-in is now necessary but not sufficient: both the resolver and the canvas apply the gate, and an unmet precondition falls back to a **local** ingest (the file stays where it is) with the canvas explaining why.

**Web clipping is not server-only.** `classify_ingest_source` already calls a page an "article" and the pipeline already extracts it via `web_article_ingestion`, so local clipping needed **no routing code at all** — no server, no policy gate. 684.3 had been scoped as porting a server-only feature.

## Six defects the test suite structurally could not catch

All found by contact with a live server, and all one shape: **something typed or faked from assumption rather than from the wire.**

| | Defect | Effect |
|---|---|---|
| 1 | `MediaIngestJobStatus.result` typed as `ReadingImportResponse` — the *reading-list* import result, a different domain reused by mistake | Validation failed with four missing fields on **every completed job**. The poller treats a raised list call as transient, so it logged at debug and retried forever: jobs would have sat **"queued" in the UI indefinitely**. Only a *finished* job triggers it. |
| 2 | Neither the client nor the service accepted an `offset` the poller sent | `TypeError` on every real call → silent one-page fallback. **Pagination was dead in production**, though the endpoint has always supported it. |
| 3 | A broad `except TypeError` around that call | Cannot distinguish "no such parameter" from "the callable raised internally", so a real bug read as a missing feature. Replaced with an explicit signature check. |
| 4 | `/ingest-web-content` requires a lowercase `token` header the client never sent | **Every web clip 422'd** before doing any work, for every API-key user. |
| 5 | `get_type_group` decided by file extension alone | A **YouTube URL pre-flighted as an unsupported *file***, while the pipeline called the same URL "video" and would have ingested it. |
| 6 | The pre-flight probe turned its own 403 into "URL unreachable" and dropped the source | Blocked a Wikipedia article that a tldw server clipped at **200** — its browser-based stack succeeds where ours is refused. (The User-Agent is not the cause; verified with a browser UA.) |

Also corrected: `mark_remote_done`'s docstring claimed the server "carries only counts, no media id" — false, and itself derived from defect 1. Behaviour is unchanged and still right for a better reason: that id addresses a row in the **server's** library while `media_id` locally means a row in this machine's, so storing it would point "Open in Library" at a wrong or absent local row.

**One trap worth naming**, since it costs time: `token` alone returns **429 `rate_limited`**, which reads as "back off and retry" and never succeeds — it is this server's shape for a *denial*, not throttling. A control request against a route known to accept `X-API-KEY` is what distinguishes them.

## On test quality

The same failure mode recurred three times on this branch: **a fake written to match my own call site validated the wrong call.** The keyword-only `cancel` fake took it positionally; the pagination fake declared an `offset` the real service lacked; every `result` fixture matched the wrong model.

The guards are therefore contract tests against **real signatures and verbatim captured payloads**, not better fakes — `_accepts_keyword(RealClass.method, "offset")`, a live list response pasted in whole, and a check that every kwarg the clip mapping emits is one the real service accepts. Each new guard is mutation-checked: removing it fails its test.

## Verification

- **1,479 passed, 0 failed** across `Tests/Library`, `Tests/tldw_api`, `Tests/Media`, run after the rebase onto dev
- Live-verified against a real tldw server on :8000 — submit → `queued` → `completed` → `DONE`, batch cancel, and a real clip returning 2,801 characters
- Live TUI run: with the opt-in set but the runtime local, the canvas explains the precondition and the ingest genuinely runs locally (`origin=local`, real media row); the backend choice persists across a restart
- **Status spellings confirmed on the wire**: `queued`, `completed`, `cancelled`. `running`/`failed` are mapped but unobserved — the endpoint types `status` as a bare string with no enum, so only a live run can learn the vocabulary, and an unmapped spelling is ignored by design.

`Tests/UI/test_library_shell.py` fails ~6 tests per run **on dev as well as here** — same counts (6 failed/251 passed), 5 of 6 identical, each tree failing a different flaky sixth, every varying test passing in isolation. Filed as **task-699**; no commit here touches that file.

## Follow-ups filed

- **699** — Library shell tests fail nondeterministically, hiding real regressions
- **700** — open a server-ingested item from the ingest queue (needs a server-aware affordance)
- **701** — `TLDW_CONFIG_PATH` does not isolate the runtime-policy state file, so a scratch profile shares the real user's local/server mode
- **702** / **697** — the two URL-handling bugs above, fixed here
- **698** — ingest form does not reset to defaults on Library re-entry (pre-existing on dev)

IDs 688/689/690 were renumbered to 700/701/702: `feat/watchlists-phase-a-data-foundation` already held them, which only a sweep of **every remote branch** reveals — checking `origin/dev` alone is the half-measure that produced this repo's recurring collisions.

---
## Summary by cubic
Brings server-backed ingestion and web clipping into the Library ingest canvas, unifying local and remote jobs in one queue with clear backend selection and policy gating. Completes tasks 684.1–684.3; removing the old ingest window (684.4) will follow.

- **New Features**
  - Start server ingests from the canvas (684.1). A backend switch appears only when server submission is available, respects runtime policy, persists the choice, and falls back to local with an explanation if server mode isn’t active.
  - Remote jobs in the same queue (684.2). Schema v3 adds `origin`, `remote_job_id`, `batch_id`, and `cancelled`; an async poller reconciles statuses; batch cancel is supported; polling stops when all batches settle.
  - Clip web pages from the canvas (684.3). URLs route to the clipper; local clips work without a server; server clips settle synchronously and are recorded in the queue; the new “web” group exposes scrape method and limits.

- **Bug Fixes**
  - URL handling: video hosts map to audio/video; links with extensions map to file types; other HTTP URLs map to web. Preflight no longer blocks on 403/405; only 404/410 refuse.
  - Server API correctness: send the required `token` header for the clip endpoint; expose pagination (`offset`, `has_more`, `next_offset`) in clients/services; call keyword-only cancel correctly.
  - Status robustness: relax `MediaIngestJobStatus.result` to a free-form object; map server statuses to local states (including `cancelled`); ignore unknown statuses without declaring success.
  - Safety and UX: ingest target is an explicit preference (not browse scope); canvas states “Imports run on this machine/server” and shows the switch only when real; Workspaces “Import sources” opens the canvas.
  - Remote tracking reliability: require both `batch_id` and `remote_job_id` before attaching remote ids and starting polling; do not over-count “parsing” as a state transition while still updating progress; clarified cancel batch docs.

<sup>Written for commit 7c424ff080715a21343e2fb7ae83b6749d371b5c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/923?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## Review addressed — 3 fixed, 1 declined with evidence

Each finding was checked against the code before acting.

### 1. Missing remote ids validation — **fixed** (`7c424ff`)

Confirmed, and the most serious of the four. Following a remote job needs **both** ids: `pending_remote_batches` decides what to poll from `batch_id`, and `reconcile_remote_ingest_jobs` matches statuses to jobs by `remote_job_id`. A response missing either was attached anyway and polling started regardless, so:

- no `batch_id` → the job is never polled at all;
- `batch_id` but no job `id` → the batch is polled forever while no status can ever match.

Either way the row sits at `queued` indefinitely. That is the same never-resolves failure the mistyped `result` field caused earlier in this branch, so the diagnosis is well taken.

Now fails fast rather than queueing forever. The message deliberately admits the server may still be working — it *did* accept the submission; what is lost is our ability to follow it — so the user checks the server instead of blindly re-importing. Four cases covered (no `batch_id`, no `jobs`, a job without an `id`, empty response), each also asserting the batch is never polled.

### 2. Reconcile counter miscounts parsing — **fixed** (`7c424ff`)

Confirmed. The already-`PARSING` branch handed back the `job` snapshot, and a non-`None` value was read as "something moved" — but that snapshot is a `replace()` copy taken *before* the call, so it proves nothing, and the documented transition count overstated itself on every poll of a still-running job.

Returns `None` there now. Kept the documented contract rather than redefining it, since nothing operational reads the count (the caller discards it). Added a second test asserting **progress still lands** while a job stays running, so this fix cannot quietly swallow progress updates: a running job reporting 40% has not changed state, but the queue should still show 40%.

### 3. `cancel_remote_ingest_batch` docstring missing `Args` — **fixed** (`7c424ff`)

Fair — `CLAUDE.md` mandates Google style with `Args`, and sibling methods in this branch comply. Documented, including *why* an empty `batch_id` returns early: a queue row that never received a batch id must not send a cancel for every job on the server.

### 4. Schema migration outside `migrations/` — **declined**, evidence below

The rule is sound but does not apply to this database:

- **`tldw_chatbook/DB/migrations/` is ChaChaNotes-specific.** Every file in it is `chachanotes_v*_to_v*.sql` (plus two legacy ones), and it is referenced only from `ChaChaNotes_DB.py`. There is no repo-level `migrations/` directory at all.
- **`Library_Ingest_Jobs_DB.py` already migrates inline on `dev`**, before this PR: `_migrate_v1_to_v2()` runs `ALTER TABLE ingest_jobs ADD COLUMN` directly, alongside inline `CREATE TABLE`. `_migrate_v2_to_v3()` follows that established pattern exactly.
- Moving only v3 out would **split this DB's migration logic across two places**, leaving v1→v2 inline and v2→v3 elsewhere — less auditable, which is the opposite of what the rule is protecting.

Adopting the SQL-file convention for this database is a reasonable idea, but it is a change to the whole file's approach and belongs in its own change rather than being applied to one version of one table mid-feature. Happy to file it if you'd like it tracked.

### Verification

- 85 tests across the reconcile and runner suites pass
- Both bug fixes are **mutation-checked**: reverting either guard fails its tests (`if not batch_id or remote_job_id is None:` → `if False:` fails all four; `else None` → `else job` fails the counter test)
- `No duplicate backlog task IDs` check passes

PR #924, opened 2026-07-26
feat(evals): word bench engine (PR 2 of 3)

rmusser01:
PR 2 of 3 in the Evals Console rebuild. Builds `tldw_chatbook/Evals/word_bench/` -- an engine with
**no UI**. PR 1 (#922, merged) retired the unreachable Evals UI; PR 3 builds the screen that renders
this engine's output.

## What a word bench measures

For each (snippet, target) cell it captures the model's next-token distribution at one position.
Snippets are text fragments; targets are (provider, model, steering) columns. The grid pivots: hold a
target fixed to compare snippets (directional prompting), hold a snippet fixed to compare targets
(which model is more steered). The point is making censorship and directional prompting visible as a
number.

| Module | Responsibility |
|---|---|
| `models.py` | Typed contract shared by the other five |
| `normalizer.py` | Provider response to distribution; control-token skipping |
| `analysis.py` | Entropy, Jensen-Shannon divergence, probe resolution, spread, group means |
| `capture_client.py` | The one HTTP seam; neutral sampler; preflight + sanity canary |
| `storage.py` | Grid as a pivot of existing tables; run snapshot |
| `runner.py` | Row-major fill, cancel, canary stamping |

Schema goes v3 to v4: `eval_runs.run_group_id` plus index and migration. Verified on both the
fresh-create and upgrade paths, and idempotent. Re-checked at push time that no other branch claims
version 4.

## The spec was wrong three ways, and live probing caught it

The design spec required pinning response shapes to captured fixtures rather than documentation.
Probing a live llama.cpp before writing the parser proved that necessary:

1. **Both endpoints return the same modern shape**, carrying a token `id`. The spec predicted a
   legacy token-to-logprob dict on `/v1/completions`.
2. **Chat mode's first token is a control token** -- observed `<|channel>` at probability 1.0, with
   alternatives at -21. Measuring it would have produced a grid of identical zero-entropy cells
   reading "no difference between any of these prompts". Chat mode now requests a window and skips
   to the first content token, recording `content_offset`.
3. **Raw mode can be out-of-distribution.** The same model continued "The capital of France is" with
   `"thought"`, not `" Paris"`. Well-formed logprobs are not necessarily meaningful ones, so
   preflight gained a distribution sanity canary that **warns without blocking** -- a degenerate
   target may be exactly what a user wants to study -- and whose verdict is stamped onto every cell.

Fixtures are committed under `Tests/Evals/fixtures/word_bench/`. A provider without a captured
fixture is not a supported provider.

**Also found: top-1 is an unstable reading.** Two identical requests seconds apart returned the top
two tokens in opposite rank order, magnitudes stable to ~0.002. Divergence over the full distribution
is unaffected, which is the empirical argument for it rather than top-1 being load-bearing. Pinned by
a regression test and flagged for PR 3's Top-1 lens.

## Two defects that would have produced confidently wrong output

Both were caught by implementers refusing to bend a test to make it pass:

- **`identity()` matched on model-local token ids.** Comparing across models -- the grid's entire
  column purpose -- would have reported divergence 0 for completely disjoint distributions: "no
  difference" precisely when two models differ most. Identity is now bytes-based.
- **An empty `top_logprobs` array yielded a "Ready" target and an all-zero grid.** This is the most
  likely real-world form of "this server ignores the logprobs parameter": the endpoint honours the
  request shape and returns no distribution. It now raises and preflights as **Blocked**.

The whole-branch review also disproved the analysis module's own headline claim. It stated divergence
was a lower bound; it is not, because crediting an absent token with exactly 0 pulls the value up
against the lumping approximation that pulls it down, and neither dominates (0.291 reported against a
feasible 0.121). The docstrings and the spec's grid mock are corrected -- the number is comparable and
reproducible, but it is not a bound and must not render with a leading `>=`.

## Verification

- `Tests/Evals/word_bench`: 87 passed
- `Tests/Evals`: 356 passed, 13 skipped
- PR 1's deletion guard: 38/38
- `Tests/UI` collection: 5,210 tests, zero collection errors
- `import tldw_chatbook.app` clean
- Zero files touched under `tldw_chatbook/UI/`
- One live capture against a real llama.cpp: `preflight: ok K=5 canary=degenerate warned=True`

## Follow-ups

TASK-703..709 are filed, not merely noticed. Most consequential: `run()` discards every
`PreflightResult` except the canary, so PR 3 cannot render the readiness surface the spec requires
without re-running preflight (703); `load_grid` silently truncates at 1,000 cells per target while
the index this PR adds goes unused (704); and `concurrency` is defined, persisted, and ignored -- a
gap in this PR's own plan (707).

Spec: `Docs/superpowers/specs/2026-07-25-evals-console-rebuild-design.md`
Plan: `Docs/superpowers/plans/2026-07-26-evals-rebuild-pr2-word-bench-engine.md`

---
## Summary by cubic
Build the word bench engine in `tldw_chatbook/Evals/word_bench/` to capture and analyze next‑token distributions across snippets and targets (no UI). Upgrades the Evals DB to schema v4 with `eval_runs.run_group_id`, and fixes grid loading and run status handling for reliable renders in PR 3.

- **New Features**
  - Engine modules: `models.py` (typed contracts), `normalizer.py` (chat control-token skip via `content_offset`; bytes‑first token identity), `analysis.py` (entropy K‑normalized, JSD with shared‑K, probes, spread, `effective_k`), `capture_client.py` (neutral sampler; preflight with canary), `storage.py` (bench snapshot; `run_group_id` save/load), `runner.py` (row‑major fill; cancel; stamps canary).
  - Spec-aligned capture: chat measures first content token; fixtures pinned to live `llama.cpp` payloads.
  - DB: schema v4 adds `eval_runs.run_group_id` + index; benches snapshot fully resolved config for durable renders.

- **Bug Fixes**
  - Grid loading no longer truncates at 1,000 cells and now filters by `run_group_id` in SQL (uses the new index) instead of rescanning pages in Python.
  - Runs correctly transition to `running`/`completed`/`cancelled` (no more rows stuck at `pending`); completed cells are preserved on cancel.
  - Token identity is bytes-based across models with a UTF‑8 fallback when `bytes` is absent, ensuring cross‑provider correctness.
  - Empty/missing `top_logprobs` is Blocked (`no_logprobs`) via explicit normalizer error codes; divergence documented as an estimate; duplicate identities accumulate.
  - Capture returns `CellError` for non‑JSON or malformed responses instead of raising; logging uses brace‑style placeholders; progress reports each step.

<sup>Written for commit 6090d3354af6236696d8f2bf0ace793b21f165f8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/924?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Addressed the review. Six of seven findings fixed in `c560b89d8`; one declined with evidence.

**Fixed**

| # | Finding | Resolution |
|---|---|---|
| 4 | Grid truncated at 1000 | `load_grid` pages `get_run_results` until exhausted |
| 5 | Run group lookup row-limited | `Evals_DB.list_runs` gained an optional `run_group_id` filter that filters **in SQL**, so `idx_eval_runs_group` is finally used |
| 6 | Bench edits drop fields | `description` now persists; `dataset_id` made explicitly immutable (see below) |
| 7 | Runs never leave pending | `running` at launch, `completed` at the end, `cancelled` on the cancel path, `end_time` on terminal transitions |
| 2 | `load_bench()` missing docstring | Added, plus one for `_cell_from_payload` |
| 3 | File-backed test SQLite | Switched to `:memory:` |

Two of those deserve a note.

**`dataset_id` (finding 6)** is now immutable after bench creation rather than editable. Extending `update_task` to accept it looked trivial but broke existing edit-path tests with a real foreign-key failure, so the constraint is documented in `save_bench`'s docstring instead of half-implemented. PR 3 will disable the control on an existing bench.

**`:memory:` (finding 3)** was verified rather than assumed. `EvalsDB._get_connection` is thread-local, and `sqlite3.connect(":memory:")` gives each new connection its own separate empty database — so this is only safe because nothing in the word-bench code or tests touches a second thread. Confirmed by grep and two clean full-suite runs.

**Declined: finding 1 — v4 migration should live under `migrations/`**

`tldw_chatbook/DB/migrations/` contains `chachanotes_*.sql` files, consumed by `ChaChaNotes_DB.py` and `Library_Ingest_Jobs_DB.py`. `Evals_DB.py` does not use that directory at all — its own v1→v2 and v2→v3 migrations are inline Python in the same `_migrate_schema` method this change extends.

Adding a lone `.sql` file for Evals would be the novel pattern here, and would leave one method half-inline and half-external. Following the file's established convention is the more consistent choice. Happy to revisit if the intent is to migrate all of `Evals_DB` to the SQL-file approach, but that is a separate change.

**Follow-ups:** TASK-704, 705, 706 are closed as fixed here. TASK-703 (return preflight results), 707 (concurrency), 708 (migration test), and 709 (polish) remain open. One new gap recorded on 706: an uncaught exception mid-run leaves a row at `running` — there is no `failed` transition yet.

Tests: word bench 92 passed, `Tests/Evals/test_evals_db.py` 35 passed, PR 1's deletion guard 38/38, `import tldw_chatbook.app` clean.

PR #926, opened 2026-07-26
Watchlists rebuild Phase B: collapsible workbench

rmusser01:
Phase B of the Watchlists rebuild. **Stacked on #917** — base is `feat/watchlists-phase-a-data-foundation`, so review that first. No data-layer file is touched here.

Plan: `Docs/superpowers/plans/2026-07-25-watchlists-rebuild-phase-b-workbench-shell.md`

## What this does

The screen's three fixed columns become a collapsible workbench: two rails around a vertically stacked centre, five independently collapsible regions (`z`), a solo mode (`Z`), rail toggles (`[` / `]`), and collapse state persisted per user.

**Every existing pane is re-hosted, not replaced.** `OverviewPane`, `SourcesPane`, `RunsPane`, `ItemsPane`, `RulesPane`, `NotificationsPane`, the navigator, the backend selector and the inspector all keep working and keep their stable selectors. `compose_content` drops from ~250 lines of nested composition to ~20 lines delegating to named builders.

New modules: `region_layout.py` (pure collapse/solo/restore state machine, no Textual import), `region_layout_store.py` (config persistence), `watchlists_workbench.py` (the container), `watchlists_console_handoff.py` (Console staging/follow, extracted from the 1,219-line screen).

## A correction worth reading

This plan originally said the screen was placeholder scaffolding to be replaced with stubs. **That was wrong.** It came from a screenshot in `Docs/static/watchlists-screenshots/` captured during Slice 1A (`90d4da5b2`); the real panes landed the same day in `33fcf3f66`, so the image was stale within hours, and it was trusted over the code.

The Task 5 implementer refused to execute the plan and traced which 11 of 16 tests a literal execution would break. The spec was corrected (see its "Starting point — corrected 2026-07-26" section) and the task rewritten from *replace* to *re-host*. The redesign case survives on density: measured live at 235×52, the shipped screen spent ~55 of 235 columns on a rail showing six words, a full-width three-row `Select` on one value, and an Overview grid of seven empty bordered cards.

## Defects found and fixed during review

- **Collapsing a rail pushed the whole workbench off-screen.** `.watchlists-region-header { width: 100% }` is right for a centre header (child of a `Vertical`) and catastrophic for a rail header (child of the `Horizontal`). Measured at 160×42: the header spanned the full 160 columns and three regions landed outside the viewport, with the right rail's handle unreachable by mouse. Every task-level test used a bare `App` with no stylesheet, and the one production-CSS test forced all regions expanded first.
- Any region toggle **emptied the Sources, Items and Rules tables** (2→0, 1→0, 1→0 rows) — unseeded panes rebuilt by factories. The same gap also blanked them on first navigation to a section.
- The **Inspector lost its selection while the screen kept it**, so `d`/`c`/`p` kept acting on an invisible selection.
- An **open Rules edit form was destroyed** by an unrelated toggle, and then — after the first fix — **saving a rule left a phantom form reopened** pre-filled with the just-saved rule, because the save path lacked the synchronous clear its Sources sibling already had.
- The centre's equal `1fr` heights gave **Items 11 of 34 rows** while Feeds got an equal 11 for a ≤5-line summary; Items hosts every real pane. Rebalanced CSS-only.

## Testing

`Tests/Watchlists` + the three guard files + visual parity → **351 passed, 1 skipped, 1 failed**.

The single failure, `Tests/Watchlists/test_watchlists_navigator.py::test_navigator_has_all_section_buttons` (asserts 6 buttons, finds 5), fails on this branch before any of these commits and is unrelated.

No existing test body was edited in any of the three guard files — additions only.

## Known follow-ups

- Pane-local state other than the Sources draft and Rules form (search filters, table cursor) still resets on a region toggle — inherent to the whole-widget recompose.
- `focused_region` is stale when focus leaves the workbench, so `z` acts on the last-visited region.
- A deep link into a collapsed region selects correctly in state but shows nothing.
- Region titles are the aspirational Phase C names ("Feeds" hosts a pane titled "Sources").
- `save_region_layout` runs a synchronous whole-file config write on every toggle and on mount.

---
## Summary by cubic
Rebuilt the Watchlists screen into a collapsible workbench with two rails around a vertically stacked centre, adding five independently collapsible regions, solo mode, rail toggles, focus-aware collapse, and per-user persisted layout while re-hosting all existing panes and selectors.

- **New Features**
  - Added `WatchlistsWorkbench` with five regions (left rail, feeds, items, content, right rail); supports `z` collapse, `Z` solo, and `[`/`]` rail toggles; collapsed regions render as clickable, focusable headers.
  - Introduced pure state machine `region_layout.py` and persisted settings in `region_layout_store.py` (per-user collapse state; CONTENT collapsed by default on first run); wired focus tracking so `z` targets the region with real focus.
  - Re-hosted existing panes (Overview, Sources, Runs, Items, Rules, Notifications), navigator, backend selector, and inspector with stable selectors; `compose_content` simplified via region factories.
  - Extracted Console staging/follow to `watchlists_console_handoff.py`; CSS rebalance stacks FEEDS/ITEMS/CONTENT vertically with FEEDS auto height so ITEMS/CONTENT split remaining space.

- **Bug Fixes**
  - Fixed rail header width that pushed the workbench off-screen.
  - Prevented data loss on toggle by seeding Sources/Items/Rules from screen state; preserved Inspector selection across rebuilds.
  - Mirrored transient pane state: Sources create-form draft and Rules edit-form visibility/editing now survive recomposes; cleared races that could reopen a saved rule form.
  - Persisted the pre-solo baseline (never the solo-derived state); moved layout writes off the UI thread and skipped no-op writes to avoid blocking and churn.

<sup>Written for commit 61b06ceaee14a4bbb50186c008d5c631e17a178b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/926?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## Review adjudication

All eight findings accepted and fixed in `f5f38a881`. Notably, both bugs had already been caught by this branch's own whole-branch review and *deferred* as minors — grading them as bugs is the fairer call, so they are fixed rather than carried.

**Bug — solo restore lost on restart.** The store deliberately does not persist solo, on the grounds that restoring "soloed" would strand the user in a layout they never chose. But it persisted `layout.collapsed`, which *while soloed* is the solo-derived collapse of the other two centre panes. A restart therefore restored those as **plain** collapses with no baseline to restore from, so `Z` could never get back — the stored state contradicted the module's own docstring.

Now the pre-solo baseline is persisted instead. Verified: with the right rail deliberately collapsed and Items soloed, the live layout is `['content', 'feeds', 'right_rail']` and what gets written is `['right_rail']` — what the user actually configured.

`test_save_never_persists_solo` had encoded the buggy behaviour and was corrected.

**Bug — blocking config writes on every toggle.** `_apply_layout` wrote unconditionally, including from `on_mount` when nothing had changed, and `save_setting_to_cli_config` is a synchronous whole-file read-modify-write plus cache reload on the UI event loop — once per `z`/`Z`/`[`/`]` keypress. This repo has form here: a performance audit found the Console's 0.2s tick doing SQLite on the event loop. Now the write is skipped when the layout is unchanged and moved to a thread worker, with last-write-wins under a burst.

**Six rule violations** — Google-style docstrings and annotations added per `CLAUDE.md`, preserving the existing prose that explains the non-obvious *why* (the `Widget.layout` collision, the factory-vs-instance recompose finding, the `None`-vs-`[]` config distinction) rather than replacing it.

## One correction to my own fix report

The fix agent reported a pre-existing unrelated failure in `Tests/UI/test_console_live_work_handoffs.py`. That does not reproduce: the file is **48/48 green** both on this branch and on latest dev (`006b31ed2`). No such failure exists.

The one genuine pre-existing failure remains `Tests/Watchlists/test_watchlists_navigator.py::test_navigator_has_all_section_buttons` (asserts 6 buttons, finds 5), which predates this branch.

rmusser01:
Rebased onto latest dev (now including Phase A, merged as #917) and retargeted from the Phase A branch to `dev`.

Two conflicts, both semantic rather than textual — dev had added the `RecomposeCaptureGuard` mixin to `SourcesPane` and `RulesPane`, the same two panes this branch extends with state-mirroring messages. Resolved by keeping **both**: the new `CreateFormDraftChanged` / `CreateFormVisibilityChanged` / `RuleFormVisibilityChanged` messages, and dev's `class SourcesPane(RecomposeCaptureGuard, Vertical)` / `class RulesPane(RecomposeCaptureGuard, Vertical)` signatures. The generated CSS bundle conflicted twice and was regenerated via `build_css` rather than hand-merged.

Post-rebase: **359 passed, 1 skipped, 1 failed**.

The failure is `Tests/Watchlists/test_watchlists_navigator.py::test_navigator_has_all_section_buttons`, which fails on `dev` itself — confirmed by running it there directly after the Phase A merge. It is not introduced by this branch.

PR #927, opened 2026-07-26
TASK-545 P3: retire System A's tool executor and make the tools config live

rmusser01:
## Summary

Third and final phase of the built-in tool gate program (TASK-545), plus the two config bugs it flushed out (TASK-547, TASK-658).

- **Single source of truth for gateable built-in tools** — `Agents/tool_catalog.py` now owns the gateable-tool registry; the Settings ➜ Tools section and the runtime read the same list instead of two divergent ones.
- **Settings tools section controls the real runtime** — `UI/Tools_Settings_Window.py` no longer writes config keys that nothing reads (`[tools]` was unreachable via `get_cli_setting`, TASK-547); toggles now land where the runtime actually looks.
- **System A retired** — `ToolExecutor`/`ToolResultCache`/`get_tool_executor` had zero production callers; their implementations (653 lines) are removed from `Tools/tool_executor.py`. The `Tool` ABC and its two concrete built-ins stay because System B's tool classes derive from the ABC. A regression test (`test_system_a_is_retired.py`) pins the retirement.
- **TASK-658** — `quick_ingest` honors `[database] media_db_path` (the `get_cli_setting` bare-section silent-default bug).
- **Config bug-class sweep** — the same `get_cli_setting` defect surfaced four more times in unrelated subsystems; filed as TASK-715/716/717 plus TASK-719 for the bug class itself, and TASK-718 to rehome file-operation auditing off the deleted Settings side effect. (Originally filed as 699–703; renumbered past an origin/dev collision.)

## Test plan

- `pytest Tests/Tools Tests/Agents/test_gateable_builtin_tools.py Tests/UI/test_settings_tools_section.py Tests/Local_Ingestion/test_quick_ingest_db_path.py` — 40 passed
- Branch merges clean onto current origin/dev (`git merge-tree`)

---
## Summary by cubic
Retires System A’s tool executor and wires the Settings → Tools switches to the live runtime. Consolidates gateable built-ins into one source of truth and fixes `quick_ingest` to honor and validate `[database].media_db_path`. Completes TASK-545 (P3) and addresses TASK-547 and TASK-658.

- **Refactors**
  - Removed `ToolExecutor`/`ToolResultCache`/`get_tool_executor` from `tldw_chatbook.Tools.tool_executor`; kept the `Tool` ABC and always-on tools; added a regression test.
  - Centralized gateable built-ins as `GateableTool` rows in `tldw_chatbook.Agents.tool_catalog`; both `BuiltinToolProvider` and the Settings UI derive from it; exposed `ALWAYS_ON_BUILTIN_NAMES`.
  - Rewrote `tldw_chatbook.UI.Tools_Settings_Window` to list all gateable tools (incl. disabled), save `[tools] <name>_enabled`, and display risk tags; removed the old executor dependency and the `install_claude_code_hooks()` side effect (follow-up TASK-743).
  - Updated `CLAUDE.md` to the new registration path; added P3 spec/plan docs; renumbered and filed follow-ups TASK-740–744 and TASK-760.

- **Bug Fixes**
  - TASK-547: Tools settings now write the keys the runtime reads; fixed a save-path KeyError in the Tools UI.
  - TASK-658: `tldw_chatbook.Local_Ingestion.local_file_ingestion.quick_ingest` reads `[database].media_db_path` via `get_cli_setting("database", "media_db_path", ...)`, expands `~`, and validates with `validate_path_simple`.
  - Added tests for tool-gate agreement, settings round-trip, media DB path, and System A’s retirement; removed the obsolete cache test and fixed a `note_management_tools` reload leak in tests.

<sup>Written for commit 449ba9aba164b817376d062ce77e0f9da2c4705a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/927?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## Review response — all 3 Qodo findings accepted and fixed

No declines this round. Each was verified against the repo before implementing, and one of them turned up a further defect that is now filed separately.

### 1. `quick_ingest()` uses unvalidated `db_path` — FIXED

Accepted. I checked whether this is actually repo convention before taking it, because `get_media_db_path()` (the accessor the app itself uses for this exact key) does *not* validate. The measurement settles it in Qodo's favour: `config.py` already imports `validate_path_simple`, and the four most recently added db-path accessors — `get_library_collections_db_path`, `get_library_ingest_jobs_db_path`, `get_workspaces_db_path`, `get_scheduled_tasks_db_path` — all validate config-sourced paths in exactly this shape. The unvalidated ones are the older accessors. So this is the direction the codebase is moving, not an exception to it.

Fixed by matching those accessors' call shape.

**Ordering trap worth recording:** `validate_path_simple` lists a literal `"~/"` among its dangerous patterns, and the default value here is `"~/.local/share/tldw_cli/tldw_cli_media_v2.db"`. Validating before expanding would raise `ValueError` on the default path for every user. `expanduser()` therefore has to come first — which is also what the four accessors above do.

The validator returns the path **unresolved** (it only warns when resolution differs), so the TASK-658 regression tests' exact-path assertions still hold. A config path containing traversal now raises rather than being silently opened; that failure propagates out of `quick_ingest` deliberately.

### 2. Traceback dropped in tool logs — FIXED

Accepted, both sites (`tool_catalog.py` and `Tools_Settings_Window.py`). This one matches a defect this repo has hit before: loguru ignores `exc_info=True`, so `logger.opt(exception=True)` is the only form that attaches the traceback. Both handlers deliberately swallow the exception (absent tool / degraded row), which is exactly the case where the traceback is the only diagnostic left — the original `note_management_tools` breakage cited in the surrounding comment went unnoticed for an unknown period for this reason.

### 3. Audit docs reference deleted executor — FIXED

Accepted, and it was worse than "stale": step 3 of `INTEGRATION_INSTRUCTIONS` told the reader to call `_global_executor.register_tool(...)`, which this PR deletes, so following it now raises `AttributeError`. Rewritten to state plainly that there is currently no seam, that auditing is consequently installed by no code path, and to point at the live `BuiltinToolProvider` path with the follow-up (TASK-743) that decides between rehoming and deleting.

---

### Additional defect found while verifying finding 1 — filed as TASK-760

Checking whether `quick_ingest` should just call `get_media_db_path()` surfaced a real divergence. Measured on a default config:

| | resolved path |
|---|---|
| app (`get_media_db_path()`) | `~/.local/share/tldw_cli/default_user/tldw_chatbook_media_v2.db` |
| `quick_ingest()` fallback | `~/.local/share/tldw_cli/tldw_cli_media_v2.db` |

Different per-user subdirectory *and* different filename stem, so anything ingested via `quick_ingest` would land in a database the app never opens. It predates this PR (TASK-658 only fixed the custom-path case) and is latent today because `quick_ingest` has no in-tree callers. Filed as TASK-760 rather than widened into this PR, per the repo's rule about not implementing beyond a task's acceptance criteria.

### Also in this push

The follow-up tasks originally filed as 699–703 were renumbered twice: origin/dev claimed 699–709 while this branch was in flight, and PR #928 then claimed 712–723, colliding with the 715–719 block exactly. They now sit at **740–744**, past the observed maximum.

PR #929, opened 2026-07-26
Watchlists follow-ups: wire the FTS backfill, fix in-memory SubscriptionsDB

rmusser01:
The two blocking follow-ups filed from the Watchlists Phase A/B work (#917, #926).

## task-689 — `SubscriptionsDB(":memory:")` was non-functional

Verified before the fix: an in-memory instance had **zero tables** on the connection everything uses, and `add_subscription` raised `no such table: subscriptions`.

`_initialize_schema` built the schema inside `with closing(self._get_connection())` — a connection closed immediately after — while the thread-local `.conn` property opened a *different* one. For a file database both point at the same file, so it worked. For `:memory:`, every connection is a separate private database, so the schema landed somewhere nothing could see.

This mattered in production: `WatchlistPreviewService.preview()` deliberately constructs an in-memory DB so URL snapshots aren't persisted, and its execute path calls `record_check_result`/`record_check_error`, which write to tables. **Source Preview / dry-run was very likely broken.**

Fixed by initialising the schema on the thread-local connection, matching `ChaChaNotes_DB`'s existing pattern. The documented trade-off — a second thread touching the same in-memory instance still gets a schema-less connection — is the same limitation `ChaChaNotes_DB` already accepts, and the only `:memory:` caller never crosses threads.

Also fixed a second failure on that same path: Phase A enabled FK enforcement, and preview used a synthetic `subscription_id` of `-1`, so any FK-bearing write would raise `IntegrityError` once the schema actually existed. Preview now seeds a real parent row and still persists nothing to the user's real database.

Verified after: 17 tables present, writes succeed, and two in-memory instances stay isolated from each other.

## task-688 — `backfill_items_fts` had no caller

The FTS index is created empty over a table that may already hold rows, and only the insert/update triggers populate it going forward — so every item scraped before the upgrade was permanently unsearchable while search appeared to work.

Now driven to completion from `on_mount` via `run_worker(..., thread=True, exclusive=True, group=…)`, so a large backlog never blocks startup or screen mount. The worker builds its own `SubscriptionsDB` because connections here are thread-local, and logs rather than crashing if indexing fails.

## Testing

`Tests/DB Tests/Subscriptions Tests/Watchlists Tests/App` → **434 passed, 1 failed**.

The failure is `Tests/Watchlists/test_watchlists_navigator.py::test_navigator_has_all_section_buttons`, which fails on `dev` itself.

## Noted, not fixed

`Library_Collections_DB`, `Workspace_DB` and `AgentRuns_DB` share the same throwaway-connection pattern in `_initialize_schema`. They have no `.conn` cache and no production `:memory:` caller, so they aren't broken today — but they would fail the same way if ever pointed at `:memory:`. Left alone deliberately rather than fixed blind.

---
## Summary by cubic
Fixes the broken in-memory `SubscriptionsDB` used by Watchlist Preview and wires a startup FTS backfill so pre-existing `subscription_items` are searchable, with safer resource handling and better logs. Completes task-689 (in-memory DB + preview) and task-688 (FTS backfill).

- **Bug Fixes**
  - Initialize schema on `self.conn` (via `self.transaction()`); `_ensure_watchlists_schema` reuses the same connection. `WatchlistPreviewService.preview()` seeds a real `subscriptions` row to satisfy FKs, persists nothing, and now closes its in-memory DB in a try/finally (also on errors).
  - Backfill worker logs the DB path and rows indexed (via `FTSBackfillError`) and always closes its DB; added tests for in-memory schema/isolation, close-on-success/error, and an end-to-end preview URL path.

- **New Features**
  - Added `tldw_chatbook/Subscriptions/fts_backfill.py::backfill_subscription_items_fts` to loop `db.backfill_items_fts` to completion.
  - Started a background worker in `on_mount` (`thread=True`, `exclusive=True`, `group="subscriptions-fts-backfill"`) that opens its own `SubscriptionsDB`; runs off the UI thread, is idempotent/resumable, and logs progress.

<sup>Written for commit 6b1e97f436dbc837216380bbd062a6fbfb9fc35a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/929?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/DB/test_subscriptions_db.py
+++ b/Tests/DB/test_subscriptions_db.py
@@ -354,3 +354,53 @@ def test_rebuild_recovers_from_stray_subscription_filters_new_table(tmp_path):
         for r in migrated.conn.execute("SELECT name FROM sqlite_master WHERE type='table'")
     }
     assert "subscription_filters_new" not in tables
+
+
+def test_in_memory_db_has_usable_schema():
+    """Regression for task-689. Before the fix, ``_initialize_schema`` built
+    the schema on a connection from ``with closing(self._get_connection())``
+    that was closed immediately after, while the ``.conn`` property used by
+    every other method opened a *different* ``:memory:`` connection -- an
+    entirely separate, empty database in SQLite's model. ``.conn`` therefore
+    had zero tables and any write raised ``OperationalError: no such table``.
+    """
+    db = SubscriptionsDB(":memory:", client_id="probe")
+    tables = {
+        row[0]
+        for row in db.conn.execute("SELECT name FROM sqlite_master WHERE type='table'")
+    }
+    assert "subscriptions" in tables
+    assert "subscription_items" in tables
+    assert "watchlists" in tables
+
+    # A basic write must succeed against the connection callers actually use.
+    source_id = db.add_subscription(name="ArXiv", type="rss", source="https://a.example/f")
+    assert db.get_subscription(source_id)["name"] == "ArXiv"
+
+
+def test_in_memory_db_instances_stay_isolated():
+    """Two separate ``:memory:`` instances must not see each other's data --
+    each opens its own private SQLite database, and nothing here shares a
+    cache or file between them."""
+    db_a = SubscriptionsDB(":memory:", client_id="a")
+    db_b = SubscriptionsDB(":memory:", client_id="b")
+
+    db_a.add_subscription(name="ArXiv", type="rss", source="https://a.example/f")
+
+    assert db_a.conn.execute("SELECT COUNT(*) FROM subscriptions").fetchone()[0] == 1
+    assert db_b.conn.execute("SELECT COUNT(*) FROM subscriptions").fetchone()[0] == 0
+
+
+def test_ensure_watchlists_schema_idempotent_on_in_memory_db():
+    """The ``conn=None`` standalone-call path (used directly by
+    test_schema_migration_is_idempotent's file-backed counterpart) must also
+    stay correct against an in-memory instance rather than silently
+    operating on a throwaway, discarded connection."""
+    db = SubscriptionsDB(":memory:", client_id="probe")
+    db._ensure_watchlists_schema()
+    db._ensure_watchlists_schema()
+    tables = {
+        row[0]
+        for row in db.conn.execute("SELECT name FROM sqlite_master WHERE type='table'")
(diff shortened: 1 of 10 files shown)


rmusser01:
## Review adjudication

**Fixed (`6b1e97f43`)** — five of six accepted.

- **Preview DB never closed** (bug). Correct, and this PR made it worse rather than better: before the in-memory fix that object was inert, so nothing meaningful was allocated. Now it holds a live thread-local connection and a real in-memory database, so every preview accumulated one until GC. Closed deterministically in a `try`/`finally`, including the exception path — which is exactly when you want the resource back.
- **Close errors silently ignored** (bug). The `except Exception: pass` around `db.close()` now logs via `logger.opt(exception=True)` — note loguru ignores `exc_info=True`, which this repo has been bitten by before.
- **Backfill error log lacks context** — the failure log now carries the database path and how many rows were indexed before the failure, so it is actionable rather than merely alarming.
- **`_initialize_schema` commits outside `transaction()`** — routed through the class's own context manager. `transaction()` being callable from `__init__` was verified by tracing the `SubscriptionsDB.__init__`/`BaseDB.__init__` ordering, not assumed.
- **Test helper bypasses `transaction()`** — `_drop_ai_trigger` now uses it.

**Declined, with evidence**

- **Unvalidated `get_subscriptions_db_path()`** in the new worker. That call is used exactly this way at two pre-existing sites — `app.py:4809` and `:4860` — both predating this PR. Validating only the third would be inconsistent and would not close the hole. If the concern is real it needs its own task covering all three call sites; fixing one is theatre.

## Rebased

Rebased onto the newest `dev` (`113654940`) — it advanced twice while this PR was open. No conflicts.

`Tests/DB Tests/Subscriptions Tests/Watchlists Tests/App` → **436 passed, 1 failed**, the failure being `test_watchlists_navigator.py::test_navigator_has_all_section_buttons`, which fails on `dev` itself.

PR #933, opened 2026-07-26
fix(tests): repair Library shell precedence and reset assertions (tasks 687, 698)

rmusser01:
Fixes the two Library shell config-precedence tests (**687**) and the ingest deep-link reset test (**698**), and closes out most of **699**.

Both tasks were filed as product bugs, from tests that were failing on `dev`. **Neither is a product bug.** The screen behaves correctly in all three cases; the tests were measuring the wrong thing. Established by probing, not by reading the assertions.

`Tests/UI/test_library_shell.py` goes from **9 failures on clean dev** to **0–1**.

## 687 — the precedence tests

Each seeds one setting into `app_config`, then patches `get_cli_setting` to raise on **any** call. Recording the calls instead of raising gives a symmetric answer:

| Test | Value from app_config | Its CLI key read? | What tripped the raise |
|---|---|---|---|
| search history | `('from-app-config',)` ✓ | `library.search` **never** ✓ | `library.rail_state` — not seeded here, so its fallback is correct |
| rail preferences | `details_open=True` ✓ | `library.rail_state` **never** ✓ | `library.search`, plus 20 `library.ingest_options.*` reads |

So precedence holds in both. What tripped each trap was the *other* setting — the one that test doesn't seed — plus the ingest canvas loading its persisted per-type options.

**My first fix was vacuous, and mutation testing caught it.** Narrowing each trap to raise only on its own key looked right, but the rail fallback sits inside an `except Exception:` block that swallows an `AssertionError` raised from the patch — so the test could not fail at all. Deleting the entire `app_config` branch still passed.

That also explains the original symptom: these failed with *"Library shell never loaded"* rather than an assertion message, because the raise that actually escaped came from a call site with no `try`/`except` and killed the load worker.

Both tests now **record** the calls and assert afterwards, outside the screen's exception handling.

## 698 — the deep-link reset

`type_options` survived re-entry holding `{'analyze': False, 'chunk': True, 'chunk_size': '1000'}` — the schema defaults, with `chunk_size` a string because number inputs round-trip through display text. The test's pre-fill never set `type_options`, so those values cannot be "from the previous visit": a mounted canvas reports what its widgets rendered, and so can never equal a never-mounted `LibraryIngestFormState`. The assertion was testing the mount, not the reset.

Probed the stronger question it never asked — pre-fill options the user actually *changed* (`chunk_size='500'`, `chunk=False`) — and they come back at their defaults. The reset discards user overrides, not merely typed text.

The test now asserts every other field equals a fresh form, and separately that `type_options` holds only defaults, with the pre-fill strengthened to include those changed options so it exercises the invariant its own docstring claims.

## 699 — reduced, not closed

Three of the nine failures were **stable across every run** and were misfiled by me as nondeterminism; they are the 687 and 698 cases above. Full-file runs since: **1 failed / 256 passed, then 257 passed, then 257 passed.**

A real order-dependent tail remains — `note_conflict_shows_overwrite_reload_and_keeps_user_text` passes alone, passes with its siblings, and fails only in some full-file runs. Its wait carries a 15-second timeout, so slowness does not explain it; the cause is accumulated state across 257 Textual app instances. **699 stays open** for that.

## Verification

- Mutation-checked, both directions: disabling the deep-link reset fails on `path`/`title`/`author`/`keywords`; removing the `app_config` branch from `_library_rail_preferences` fails with *"rail preferences fell back to the CLI config despite app_config"*
- Three full-file runs after the fixes: 256/257, 257/257, 257/257
- Rebased onto current dev with zero overlapping files

Measurement note: this machine ran 14 concurrent pytest processes at load ~12 throughout, which changes which tests lose a race. Compare failure **sets** from identical commands, never counts across different invocations.

---
## Summary by cubic
Fixes three failing Library shell tests by correcting assertions and recording CLI lookups with normalized keys so tests reflect real behavior. Addresses Linear 687 and 698; trims the remaining failure pool in 699.

- **Bug Fixes**
  - Config precedence (687): replaced raising `get_cli_setting` patch with call recording that normalizes to `(section, key)`; assert no reads of `library.search`/`library.rail_state` when `app_config` provides them; covers dotted, multi-arg, and keyword calls.
  - Deep-link reset (698): verify text fields match a fresh `LibraryIngestFormState`; compare `type_options` against capability defaults and pre-fill user-changed options to ensure they reset.
  - Stability (699)/hygiene: full-file runs drop from 9 failures to 0–1; fixed import groups; the remaining order-dependent test stays tracked in 699.

<sup>Written for commit b6c04fe834fb9f987b7259a3ebd969cf34d0a9dc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/933?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_library_shell.py
+++ b/Tests/UI/test_library_shell.py
@@ -9,7 +9,11 @@ from pathlib import Path
 from types import SimpleNamespace
 from unittest.mock import Mock
 
+import dataclasses
+
 import pytest
+
+from tldw_chatbook.Library.ingest_capabilities import get_capabilities
 from textual.app import App, ComposeResult
 from textual.screen import Screen
 from textual.widgets import Button, Collapsible, Input, Markdown, Static, TextArea
@@ -855,13 +859,15 @@ async def test_library_shell_search_history_prefers_app_config_over_cli_config(
     app.app_config["library"] = {"search": {"history": ["from-app-config"]}}
     _seed_conversations(app, _two_conversations())
 
-    def raising_get_cli_setting(*args, **kwargs):
-        raise AssertionError(
-            "get_cli_setting should not be called when app_config already has history"
-        )
+    real_get_cli_setting = library_screen_module.get_cli_setting
+    cli_reads: list[tuple] = []
+
+    def recording_get_cli_setting(*args, **kwargs):
+        cli_reads.append(tuple(args[:2]))
+        return real_get_cli_setting(*args, **kwargs)
 
     monkeypatch.setattr(
-        library_screen_module, "get_cli_setting", raising_get_cli_setting
+        library_screen_module, "get_cli_setting", recording_get_cli_setting
     )
 
     host = LibraryHarness(app)
@@ -870,6 +876,16 @@ async def test_library_shell_search_history_prefers_app_config_over_cli_config(
         await _wait_for_library_shell(screen, pilot)
 
         assert screen._library_search_history == ("from-app-config",)
+
+    # Recorded rather than raised. Two reasons: the screen legitimately reads
+    # OTHER settings during load (rail state, which this test does not seed, and
+    # the ingest canvas's persisted per-type options), so raising on every call
+    # reports a precedence bug that is not there; and the fallbacks sit inside
+    # `except Exception:` blocks, which swallow an AssertionError raised from
+    # the patch and leave the test unable to fail at all (task-687).
+    assert not [read for read in cli_reads if str(read[0]).startswith("library.search")], (
+        f"search history fell back to the CLI config despite app_config: {cli_reads}"
+    )
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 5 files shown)


rmusser01:
## Both findings addressed (`b6c04fe8`)

**Fragile CLI read recording** — valid, and the more important of the two. `get_cli_setting` accepts both `get_cli_setting("library", "search", default)` and the dotted `get_cli_setting("library.search")`, so matching on the raw first argument only catches whichever shape the code happens to use today. A switch to the two-argument form would have made these tests **silently stop detecting the very fallback they exist to detect** — the same failure mode as the vacuous version they replaced. A keyword-only call would also have raised `IndexError` rather than failing usefully.

Calls are now reduced to a canonical `(section, key)` pair before comparison. Verified across every shape:

| call | normalised |
|---|---|
| `("library.search")` | `("library", "search")` |
| `("library.rail_state", None)` | `("library", "rail_state")` |
| `("library", "search")` | `("library", "search")` |
| `("library", "rail_state", {})` | `("library", "rail_state")` |
| `(section=..., key=...)` | `("library", "search")` |
| `()` | `("", "")` — matches nothing, does not raise |

Still mutation-checked afterwards: removing the `app_config` branch from `_library_rail_preferences` fails with *"rail preferences fell back to the CLI config despite app_config"*.

**Import groups** — fair; my scripted edit had stranded `import dataclasses` in its own block and put a local import above the `textual` ones. Restored to stdlib / third-party / local.

Rebased onto current dev.

PR #934, opened 2026-07-26
fix(runtime-policy): isolate the state file per config profile (task 701)

rmusser01:
The runtime-policy state file records whether the app runs **local or server**. Its path came from `DEFAULT_CONFIG_PATH` — a hardcoded home directory — even though the config path itself already had a resolver honouring `TLDW_CONFIG_PATH`.

So a scratch profile read and wrote the **real user's file**: putting a test profile into server mode left the default profile in server mode afterwards.

That is not merely untidy. It is what stopped task-684.1's server-ingest path being verified live — flipping a scratch profile would have flipped the user's own app, and their config has a server bound, so the self-healing demote would not have fired either.

## The fix

`default_runtime_policy_path()` resolves from the config path in effect, **per call** rather than captured at import. The override is an environment variable a launcher or a test can set *after* this module is first imported; a module-level constant freezes whichever value happened to be present then.

`DEFAULT_RUNTIME_POLICY_PATH` stays as the no-override answer and as the fallback when an override cannot be resolved, so an unreadable path cannot stop the app booting.

## Verification

**Live, end to end** — a scratch profile switched to server mode wrote `"server"` into its *own* `runtime_policy.json`, while the real profile stayed `"local"` and byte-identical to a backup taken beforehand.

**Mutation-checked** — reverting to the hardcoded path fails three of the four tests. The fourth deliberately guards the *unchanged* default, so existing installs cannot silently lose the mode they had persisted.

`Tests/RuntimePolicy`: **252 passed**.

## Also: a regression I shipped in #930

Deleting `tldw_api_events.py` with the ingestion window left it listed as the sole holdout in `Docs/Development/server-client-provider-migration-audit.md`, and that document is enforced by a test in `Tests/RuntimePolicy`, which reported inventory drift.

My verification for #930 was full-tree `--collect-only` plus the touched suites. Neither could catch this: collect-only only sees import errors, and `Tests/RuntimePolicy` was not among the suites I ran. The holdout is now recorded as **resolved by retirement** rather than by migration, which is what actually happened.

Worth noting as a method correction — a deletion needs the suites that assert *about* the codebase (inventories, audits, contracts), not only those that import it.

---
## Summary by cubic
Isolates the runtime-policy state file per config profile so profiles no longer overwrite each other’s local/server mode (TASK-701). Fallback now catches only bad path errors and warns; unexpected failures surface.

- **Bug Fixes**
  - Added `default_runtime_policy_path()` to resolve beside the active `TLDW_CONFIG_PATH` profile on each call; falls back to `DEFAULT_RUNTIME_POLICY_PATH` only on `ValueError`/`OSError` and logs at warning.
  - Updated `load_runtime_policy_for_app()` to use the new resolver so each profile persists its own `runtime_policy.json`.
  - Tests cover: per-profile isolation, unchanged default without override, loader writing to the profile path, warned fallback on a rejected override, and propagation of unrelated errors.
  - Updated the migration audit to mark `tldw_chatbook/Event_Handlers/tldw_api_events.py` as retired.

<sup>Written for commit 58154962eaf33cbedad65c56a07469fdcc172461. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/934?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/Development/server-client-provider-migration-audit.md
+++ b/Docs/Development/server-client-provider-migration-audit.md
@@ -96,7 +96,7 @@ These are direct helper call sites rather than service classes. They should be r
 
 | Module | Audit lines | Notes |
 | --- | ---: | --- |
-| `tldw_chatbook/Event_Handlers/tldw_api_events.py` | 572 | Explicit UI/event helper holdout. Direct endpoint/auth form flow is not safely replaceable by the current app provider without a broader event/UI state refactor. Semantic match: `api_client = build_runtime_api_client(`. |
+| _(none)_ | — | The sole holdout, `tldw_chatbook/Event_Handlers/tldw_api_events.py`, was deleted with the standalone ingestion window it served (task-684.4). Its direct endpoint/auth form flow -- the reason it could not be migrated to the app provider without a broader event/UI refactor -- went with it, so the holdout resolved by retirement rather than migration. |
 
 ### Intentional Current Provider And Bootstrap Usage
 
--- a/Tests/RuntimePolicy/test_runtime_policy_bootstrap.py
+++ b/Tests/RuntimePolicy/test_runtime_policy_bootstrap.py
@@ -801,3 +801,45 @@ class TestRuntimePolicyPathIsolation:
         assert (scratch.parent / "runtime_policy.json").exists(), (
             "the profile's own policy file was never written"
         )
+
+    def test_a_rejected_override_falls_back_loudly(self, monkeypatch, caplog):
+        """A bad path may not stop the app booting, but must not be silent.
+
+        Falling back means this profile's local/server mode is read from and
+        written to the DEFAULT profile -- the very cross-profile leak this
+        function exists to prevent -- so it is logged at warning, not debug.
+        """
+        from tldw_chatbook.runtime_policy import bootstrap
+
+        def rejecting(*args, **kwargs):
+            raise ValueError("rejected path")
+
+        monkeypatch.setattr(
+            "tldw_chatbook.config._get_effective_config_path", rejecting
+        )
+
+        assert bootstrap.default_runtime_policy_path() == (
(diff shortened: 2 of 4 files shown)


rmusser01:
## Addressed (`58154962`)

**Broad catch masks misrouting** — valid, and the serious one. Catching every `Exception` meant an unexpected failure in path resolution silently sent runtime-policy state to the **default profile**: a scratch profile quietly editing the real user's local/server mode, which is precisely the leak this change exists to stop. Logging it at debug would have hidden that in a normal run.

Now catches only what the resolver actually raises for a bad path (`ValueError`, `OSError`) and logs the fallback at **warning**, naming the profile that will be affected. Anything else is a defect rather than a bad path and is left to surface, since swallowing it routes state to the wrong profile.

Two tests pin both halves: a rejected override falls back to the default, and an unrelated `RuntimeError` propagates instead of being absorbed.

**Import groups** — checked; `loguru` sits with the third-party imports.

### On the two alternatives

Both were offered as options with this PR's approach explicitly recommended over them, and I've kept it. For the record:

- *A public "profile data dir" helper on config* — reasonable, and it would remove the private import. It's a config API change that affects more than this file, so it belongs in its own change rather than riding along here.
- *A dedicated `TLDW_RUNTIME_POLICY_PATH` override* — I'd argue against it. It adds a second source of truth for something that should simply follow the profile, and the failure it invites is a policy file pointing at a different profile than the config it belongs to — a subtler version of the bug being fixed.

`Tests/RuntimePolicy`: 252 passed. Rebased onto current dev.

PR #936, opened 2026-07-26
feat(agents): built-in tool packs — Phase 0 substrate + Phase 1 files pack

rmusser01:
Expands the agent runtime's built-in tool surface from two tools (`calculator`, `get_current_datetime`) toward a pack-based architecture, in two phases. Design: `Docs/superpowers/specs/2026-07-25-builtin-tool-packs-design.md`. Plan: `Docs/superpowers/plans/2026-07-25-builtin-tool-packs-phase-0-1.md`.

## Phase 0 — runtime substrate

Six properties that made a larger catalog unsafe or unreachable:

- **Tool ceiling 8 → 24, disclosure threshold 8 → 16.** `max_active_tools` was a one-way ratchet with no eviction — `load_tools` returned `"no room"` and nothing ever unloaded. The threshold rises with it because `initial_disclosure` runs per-run, so a catalog above it costs `find_tools`+`load_tools` round trips before any work, on every message.
- **Tool results capped at the history-append seam.** Only the persisted step record was bounded (2000 chars); the history append was unbounded. New `RunBudget.max_tool_result_chars` (16000, `0` = unlimited) enforced in `agent_runtime`, so built-in, MCP and skill results share one rule.
- **Per-tool timeout override**, resolved through the registry and duck-typed so MCP/skill providers keep the run budget.
- **`network` added to `HIGH_RISK_TAGS`**, so egress tools floor to `ask`. Deliberate behaviour change: an MCP tool already declaring `network` begins prompting.
- **MCP name-shadowing made visible.** Built-ins still win collisions (inverting that would let a hostile server name-squat an audited built-in), but the shadowing is now logged rather than silent.
- **Console budget 20 → 30 turns**, with `max_steps` 64 → 96 (`3*(N-1)+1` = 88 at N=30) and wall 1200 → 1800s, plus a 1,000,000-token spend ceiling. Sub-agents still inherit turns per the documented 2026-07-25 operator decision, so worst case is 90 provider turns — the token ceiling bounds spend instead of clamping turns.

## Phase 1 — the pack seam

- `Agents/builtin_services.py` — injection seam, with a written thread-safety/no-event-loop contract (tools run on a fresh per-call daemon thread under `asyncio.run`).
- `Agents/builtin_packs/` — registry plus a read-only `files` pack: `read_file`, `list_directory`, `glob_files`, `grep_files`. Packs whose optional deps are unmet are absent from the catalog rather than failing at invoke.
- `Agents/builtin_pack_config.py` — reads `[agent_tools] enabled_packs`.
- `Utils/sensitive_paths.py` — credential and app-state denylist, refused regardless of sandbox root.
- `Tool` ABC relocated to `Tools/base.py`; dead `ToolExecutor`, `code_audit_tool.py`, `file_operation_hooks.py` and the Tools Settings switches deleted.

**Default posture is unchanged: built-in file tools remain disabled by default.**

## Corrections to the task record

- **TASK-547 closed.** Its defect was confined to `get_tool_executor()`'s two-arg `get_cli_setting("tools", {})`, which mis-slots `{}` into the *key* parameter. `[tools]` itself was always reachable via the three-arg form — which TASK-584's gates and `_resolve_sandbox_config` both used. Removing the executor removed the only broken caller.
- **TASK-545 P3 satisfied**, P2 explicitly **not** — only the read-only `files` pack shipped. `write_file`, `edit_file`, the note tools, `rag_search` and `web_search` remain P2 work.
- TASK-545's description named `worker_events.py` and `chat_streaming_events.py` as the legacy path; both were already gone.
- Legacy `[tools]` flags grant **exactly** the tool each names (`read_file_enabled` → `read_file` only). An earlier revision mapped either flag to the whole pack, which would have silently granted three extra tools during a compatibility migration.

## Review notes

Every task was reviewed in isolation and then the whole branch was reviewed together. Several defects originated in the plan's own code snippets rather than implementation error, and two were only visible across tasks:

- The sensitive-path denylist was wired into `is_within()`, which `read_file` never calls — it routes through `validate_path()`. The denylist was dead code for the tool that motivated it. Fixed at each tool's `execute()`, deliberately **not** in the shared validator (~40 first-party call sites legitimately validate the app's own config and DB paths).
- `clamp_child_budget` dropped `max_tool_result_chars`, silently resetting a caller's explicit value for every sub-agent. Cross-task: one task added the field, another edited that function's docstring without noticing.
- `grep_files` called the intentionally-uncached denylist once per candidate — ~37s per call at the candidate bound. Fixed with a per-call context rather than global memoization, which would defeat the uncached-by-design property. 3.12s → 0.28s over a 1,530-file tree.
- `Path.glob()` is lazy, so invalid patterns raised on first iteration, outside the guard — breaking the never-raise tool contract.
- SQLite WAL/SHM sidecars hold the same recent data as the databases and needed explicit denial.

## Testing

`Tests/Agents/ Tests/Tools/ Tests/Utils/ Tests/Chat/test_console_agent_bridge.py` → **868 passed**.

The full suite takes 4h17m in this environment and cannot fully collect (`numpy` and `pytest-mock` are absent from the venv, causing 21 collection errors in `Tests/TTS`/`Tests/Transcription` and 6 errors in `test_chat_functions.py::TestChatApiCall`). A full run reported 74 failures; 6 were bisect-confirmed pre-existing and the remainder are in areas this branch does not touch, but **they were not individually attributed** — a baseline run on `dev` would be needed to close that out.

## Merge notes

Conflicts with in-flight TASK-656/627 work: `UI/Tools_Settings_Window.py` loses 301 lines (the real conflict), `builtin_tool_gate.py` has a one-line import change, and `HIGH_RISK_TAGS` gains `network` — semantically, a permissions UI rendering risk badges will start flagging network-tagged MCP tools.

`shadowed_mcp_names` is public but unconsumed on purpose — it is TASK-656's seam.

With the Tools Settings switches removed and `[agent_tools] enabled_packs` not yet in the config template or user docs, packs are currently enable-able only by hand-editing `config.toml`. Surfacing that is TASK-656/659 scope.

Also flagged: `backlog/tasks/task-635` is duplicated in this worktree; `origin/dev` already renumbered it to 662, so take dev's version on rebase.

---
## Summary by cubic
Adds a pack-based tool architecture to the agent runtime and ships a read‑only `files` pack, with runtime hardening (higher tool limits, result caps, per‑tool timeouts, clearer network risk). Default posture is unchanged: file tools stay off until explicitly enabled.

- **New Features**
  - Pack registry (`Agents/builtin_packs/`) with an injection seam (`BuiltinToolServices`); unavailable packs are omitted from the catalog.
  - `files` pack: `read_file`, `list_directory`, `glob_files`, `grep_files` — all sandbox‑rooted and refusing sensitive paths (credentials, app state, DBs and WAL/SHM/journal sidecars). Hardened search: hidden‑file rule now enforced for glob/grep; traversal guard rejects absolute/parent patterns (incl. Windows); `grep_files` streams reads with a per‑file byte cap; candidate bounds and invalid‑pattern handling prevent runaway scans.
  - Runtime substrate:
    - Tool limits: `max_active_tools` 24; `DIRECT_DISCLOSE_THRESHOLD` 16.
    - Result cap: `RunBudget.max_tool_result_chars` (default 16000) enforced at history append.
    - Per‑tool timeout overrides resolved via the registry.
    - `network` added to `HIGH_RISK_TAGS` so egress tools floor inherited allow → ask (affects MCP tools declaring `network`).
    - MCP/built‑in name collisions surfaced in logs; built‑ins still win.
    - Console budget: 30 turns, 96 steps, 1800s wall, 1,000,000‑token ceiling.
  - Closes TASK-547; advances TASK-545 (P3 satisfied; P2 deferred — write/edit and search tools not included).

- **Migration**
  - Enable packs via `[agent_tools] enabled_packs = ["files"]` in `config.toml`. Default remains off.
  - Legacy `[tools]` flags still work as a fallback but now grant only the named tool (`read_file_enabled` → `read_file`, `list_directory_enabled` → `list_directory`); an explicit `enabled_packs` list always wins.
  - `Tool` ABC moved to `tldw_chatbook/Tools/base.py`; the legacy `ToolExecutor`, code audit tool, and Tools Settings UI switches were removed. Expect some MCP tools with `network` risk to prompt where they previously executed silently.

<sup>Written for commit 1648e02943248c76eb09a795ef47507a3849a1ff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/936?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## Qodo findings addressed — `1648e0294`

Five accepted as-is, one resolved differently. 879 targeted tests green.

**1+2 (glob/grep bypass `validate_path`) — accepted, and it was exploitable.** Reproduced before the fix:

```
read_file('.env')                      -> {'error': 'Access to hidden files/directories is not allowed'}
grep_files('API_KEY', glob='**/.env')  -> 'API_KEY=supersecret123'
```

`validate_path` refuses any dot-prefixed component in the user-supplied portion, so `read_file` already rejected `.env`; `glob_files`/`grep_files` used only `is_within` and did not. A secret was readable through one tool and refused by the other.

Fixed with an inline hidden-component check rather than by calling `validate_path` per candidate, for two reasons: `validate_path` raises `ValueError` and these tools must never raise, and it costs measurably more per candidate (+46.6% vs +19.9% over a 1,500-file tree). The check was deliberately *not* folded into `is_within`, because `ListDirectoryTool` has an `include_hidden` opt-in that must keep working. `is_within` is still called — it is what applies the credential/app-state denylist, which `validate_path` does not (deliberately: ~40 first-party call sites legitimately validate the app's own config and DB paths).

**5 (grep reads whole files) — accepted.** Now streams line-by-line with early exit at `_MAX_MATCHES`, plus a `_MAX_GREP_FILE_BYTES` per-file cap.

**6 (Windows absolute glob bypass) — accepted.** Drive-letter and UNC forms now detected via `PureWindowsPath`. Note this was a cost/consistency issue rather than an escape, since `is_within` guards every candidate regardless.

**3 (import grouping) — accepted.**

**4 (`to_openai_format` docstring) — observation accepted, fix declined.** The method has zero callers anywhere in `tldw_chatbook/` or `Tests/`; the `ToolExecutor` that consumed it was deleted on this branch. Deleted the method rather than adding a `Returns:` section to dead code.

Thanks — 1+2 is the best finding on this branch and 15 prior review passes missed it. They kept asking whether the explicit denylist covered a path, and never asked whether the new tools matched the rules `read_file` already enforced.

rmusser01:
Superseded by #953.

This branch was built before TASK-545 P2 and TASK-656 landed on dev, and is 418 commits behind. Dev solved built-in tool registration with the `_GATEABLE_BUILTINS` table rather than the pack architecture proposed here, so the two designs compete rather than compose.

#953 re-homes everything dev does **not** have onto current dev: the six runtime-substrate changes (tool ceiling and disclosure threshold, tool-result cap at the history-append seam, per-tool timeout override, `network` risk tag, MCP shadow visibility, 30-turn Console budget), `Utils/sensitive_paths.py`, and `glob_files`/`grep_files` as `GateableTool` entries.

Dropped deliberately: the pack architecture (superseded), the `ToolExecutor` deletion (dev already did it), and the `Tools_Settings_Window` deletion — that last one would have regressed a surface dev has since improved to derive from `_GATEABLE_BUILTINS`.

Worth carrying forward from the review history here: the sensitive-path denylist as written on this branch **never actually protected the permission store** — it hardcoded `~/.config/tldw_cli/mcp_permissions.json` while the app builds that file under `get_user_data_dir()`, so the check never matched and `write_file` could overwrite the real store. Fixed in #953 by deriving the path from the app's own accessor. The design doc and plan from this branch are retained there.

PR #941, opened 2026-07-26
feat(evals): Console-styled Evals workbench (PR 3a of 3)

rmusser01:
PR 3a of the Evals rebuild. Replaces the broken card hub with a Console-styled three-pane workbench
you can inspect and configure a bench in. PR 1 (#922) retired the unreachable Evals UI; PR 2 (#924)
built the word bench engine; **PR 3b adds the results grid.**

## The screen was rendering nothing

The Evals destination showed a header and mode strip with an **empty body**. Confirmed during PR 1 by
capturing the real screen on a branch and on a baseline: identical, both empty.

The cause was architectural. The hub mounted Textual `Screen` objects **inside a `Container`**, and a
nested `Screen` — plus every descendant — collapses to `region=Region(0,0,0,0)` while staying fully
queryable. The old suite passed throughout: `test_evals_screen_shell.py` asserted the hub rendered six
nav-card buttons, and it did, because that test drove a minimal host app rather than the real shell.

Selection state replaces the hand-rolled screen stack. No `Screen` is mounted inside a container.

## What lands

| | |
|---|---|
| `evals_state.py` | `EvalsSelection`, `EvalsViewModel` over `EvalsDB` |
| `library_rail.py` | Benches / Datasets / Runs, collapsible, with a labelled `─ classic ─` subgroup |
| `bench_editor.py` | Bench detail, target table, read-only classic-task detail |
| `snippet_editor.py` | Snippet table, whitespace flags, three import formats |
| `inspector.py` | Per-target readiness, estimate, run action |
| `_evals.tcss` | New sheet, `$ds-*` tokens only |

Also: the runner now returns its per-target `PreflightResult` set (TASK-703) instead of discarding
everything but the canary, so readiness renders from the run's own snapshot rather than a fresh
preflight that could disagree with what the run actually used.

The card hub and its 8 test files are retired — ~3,900 lines. The deletion guard goes 19 → 25 entries.

## Two Criticals caught in review, both worth knowing

**The workbench collapsed to content height.** `.ds-panel { height: auto }` beat
`.destination-workbench { height: 1fr }` on manifest order, and `#evals-workbench` was missing the
id-level override all eight sibling workbenches carry. Measured at 160×45: the workbench occupied
**11 of 35 rows**, the snippet editor's only control laid out **outside its pane's clip region** so
`pilot.click` didn't fire, and the primary action rendered at **y=45 on a 45-row terminal** — below the
screen. Every test passed against it, because they asserted `region.width > 0`, which proves "laid
out", not "visible".

The fix adds the id-level rules and, more importantly, upgrades the assertions to
`pane.region.contains_region(child.region)`. The live gate was also wrong and is now stronger: it
asserts the workbench spans the body, not merely that the capture differs from `dev`.

**Unvalidated JSON ids became widget ids.** An id containing a space raised `BadIdentifier`;
re-importing an export — the round-trip the code advertises — raised `MountError`. In both cases the
write landed *before* the crash, leaving the dataset permanently un-openable with no in-app recovery.
Ids are now validated at the parse boundary and de-duplicated on import.

Four Importants also fixed: a Latin-1 CSV (what Excel exports) crashed the app; readiness drained and
JSON-decoded the entire run group twice per selection in a render path; classic tasks were unreachable
from the rail, leaving a spec requirement and ~200 lines dead; and panes never scrolled.

## Verification

- 67 Evals UI tests, 60 deletion-guard + CSS-integrity tests, `Tests/Evals` 384 passed / 13 skipped
- `Tests/UI` collects 5,317 with zero errors; `import tldw_chatbook.app` clean
- **Live gate:** workbench spans **43 of 43** available rows (it was 12 when squashed, and 1 on `dev`).
  All three panes render with empty states. Zero CSS warnings.
- The CSS bundle is regenerated by `build_css.py`, never hand-edited — including through 126 commits
  of rebase conflicts, each resolved by regeneration rather than by merging generated output.

## Scope

No results grid, no lenses, no export — those are PR 3b. On a fresh install the screen is three empty
sections and a primary action **deliberately disabled with a stated reason**, since running a bench
needs the grid to render into. Empty states and the sample bench are PR 3b Task 2.

Follow-ups: TASK-761 (`Utils/cost_estimation.py` is newly orphaned by PR 1 and its price table lacks
this feature's models). TASK-707/708/709 remain from PR 2.

Spec: `Docs/superpowers/specs/2026-07-25-evals-console-rebuild-design.md`
Plan: `Docs/superpowers/plans/2026-07-26-evals-rebuild-pr3a-shell-and-editors.md`

---
## Summary by cubic
Replaces the empty Evals hub with a Console‑style three‑pane workbench (library rail, detail pane, readiness inspector) for inspecting and configuring benches. Adds bench and snippet editors and snapshots preflight in runs so readiness renders from what the run actually used; legacy hub/screens are removed.

- **New Features**
  - Three‑pane `EvalsScreen` driven by selection state (no nested `Screen`s).
  - Bench editor: metadata, target table; classic tasks render read‑only with history.
  - Snippet editor: char counts, whitespace markers, exact‑duplicate flags; import text/CSV/JSON with group support.
  - Readiness inspector: per‑target status and call/time estimate from stored preflight.
  - Runner now returns `RunOutcome` with `preflight`; storage snapshots and `load_run_preflight` added.
  - New `_evals.tcss` and id‑level rules; panes scroll; results grid lands in PR 3b (run action stays disabled until then).

- **Bug Fixes**
  - Fixed empty body caused by mounting a `Screen` inside a `Container`; workbench now renders and spans the body with stronger region tests.
  - Fixed collapsed height via id‑level CSS and containment assertions; descendants lay out inside pane clip regions.
  - Validated/de‑duplicated imported JSON/CSV ids to prevent crashes; render defensively against unsafe stored ids (all widget ids are index‑derived); import paths validated via `validate_path_simple`; non‑UTF‑8 imports surface a clear error; notifications name skipped counts.
  - Eliminated redundant preflight reads; readiness uses the run snapshot once per selection.
  - Classic tasks are visible under a labelled subgroup in the rail; panes now scroll for long lists.
  - Retired the card hub and related modules/tests; deletion guard extended; removed dead embedded CSS from `Constants.py` and cleaned up dangling app/perf guards.

<sup>Written for commit d24f3d6adb7e95c3de5b7ea0272494549c8fae85. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/941?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Addressed the review in `8ad52299d`, and ran down the red check.

## The four findings

| # | Finding | Resolution |
|---|---|---|
| 1 | Snippet id collisions crash | Every widget id in the row is now index-derived |
| 2 | Duplicate target ids crash | Same, in `bench_editor` and `inspector` |
| 3 | Unvalidated `read_text()` | Routed through `validate_path_simple` |
| 4 | `select()` docstring lacks `Args:` | Added |

**Findings 1 and 2 exposed a real gap in an earlier fix in this same PR**, and the distinction is worth recording.

That fix stopped unvalidated snippet ids becoming widget ids — a space raised `BadIdentifier`, a duplicate raised `MountError`, and since the write landed *before* the crash the dataset became permanently un-openable. But it fixed the **write** path and the **outer row** id only. A dataset already holding bad or duplicate ids — written before this PR, or by anything other than our importer — still crashed on display, and the row's descendants still derived ids from stored data. The same shape existed for bench targets in two more modules.

Validating on write does not protect data that is already there. **A rendering path must never trust stored data to be widget-id-safe.** Every widget id under `UI/Evals/` is now index-derived, and the new tests compose genuinely bad datasets through the real screen rather than exercising the sanitizer in isolation.

## The red check is pre-existing, and it found a separate bug

`Artifact leases — Python 3.11 — windows-latest` fails at:

```
Tests/Model_Artifacts/test_operation_leases.py::test_lease_constructors_canonicalize_lock_root
ValueError: Path contains dangerous pattern
tldw_chatbook/Utils/path_validation.py:327
```

This PR touches neither that test nor its subject, and the failing run predates the `validate_path_simple` call added for finding 3. Locally that file is **48 passed** — it fails on Windows only.

The cause is genuine though: `validate_path_simple` scans the **raw** path string for a Windows parent-directory segment among its dangerous patterns. On Windows that substring appears in ordinary paths awaiting canonicalization — precisely what a *canonicalize* function is handed. Filed as **TASK-838**.

It also made me check my own finding-3 fix, since it adds a consumer of that helper: the call is wrapped in `try/except ValueError` and degrades to a notification, so it cannot crash on Windows for the same reason. Worth stating explicitly given the rule tells contributors to keep adding consumers.

## Verification after rebase

Rebased onto `1f3788a92`. The generated CSS bundle conflicted on every CSS-touching commit; each was resolved by re-running `build_css.py`, never by merging generated output, and the loop was guarded so any conflict outside the bundle would have stopped it.

- 70 Evals UI tests, 60 deletion-guard + CSS-integrity tests
- `Tests/Evals` 384 passed / 13 skipped
- `Tests/UI` collects 5,354 with zero errors; `import tldw_chatbook.app` clean

PR #951, opened 2026-07-27
docs: project-level lessons-learned files + a DoD hook to keep them fed

rmusser01:
Working knowledge about this repo has been living in **per-agent memory** — invisible to you, to Codex, and to the next session. This moves it into the repo.

## Three files, split by theme

| File | Covers |
|---|---|
| `lessons-testing-evidence.md` | what actually counts as evidence a change works |
| `lessons-live-verification.md` | running the app, talking to a real server |
| `lessons-backlog-hygiene.md` | task IDs, CLI quirks, git plumbing |

Deliberately **not** ADRs and **not** audits. `backlog/decisions/` records decisions with alternatives weighed; `backlog/docs/` has held point-in-time reports. A lesson is neither — it is a trap that recurs, and it needs to accumulate.

## Three choices worth arguing

**Split by theme, not one `LESSONS.md`.** This repo routinely has several agents committing at once, and a single hot file would conflict constantly — that happened repeatedly during the work these lessons came from.

**Evidence-first entries**: claim → the incident that produced it → the action. A lesson without its incident decays into folklore and stops being trusted.

> "Fakes can validate your own mistake" is arguable.
> "The fake declared `offset` because I wrote it to match my call site, and pagination was dead in production while 900+ tests were green" is not.

**A lifecycle hook, or none of this happens.** DoD gains item 9: if a task surfaced knowledge that generalises, record it — with the incident. It explicitly says most tasks produce nothing and not to invent one, because a file padded with platitudes is worse than no file.

## Both guidance files

`CLAUDE.md` and `AGENTS.md` each get the DoD item and a pointer near *Source of Truth*, so Claude and Codex get the same instruction. Edited byte-safe — these two have had CRLF churn before; the diff is 26 added lines with no whitespace noise.

## Where the content came from

The Library ingest UAT (tasks 673–702, 17 PRs). Every entry is an incident from that work: seven defects found only by live contact, three fixes that were vacuous until mutation testing caught them, two regressions that slipped through because the suites run weren't the suites the change could reach, and ten-plus task-ID collisions.

## Note for backlog-md-py

`backlog doc list` renders every entry blank — including the five docs that predate these. The docs carry no frontmatter (they open with an H1), so the lister has no title to read. Not fixed here, but it's a small gap worth closing in your port.

---
## Summary by cubic
Adds three incident-backed lessons-learned docs and a Definition of Done hook to capture new lessons. Also fixes DoD numbering, corrects the task-ID sweep command, and replaces a dead verify link with an inline recipe.

- **New Features**
  - Added `backlog/docs/lessons-testing-evidence.md`, `backlog/docs/lessons-live-verification.md`, and `backlog/docs/lessons-backlog-hygiene.md`.
  - Added DoD item: when a task surfaces reusable knowledge, add/update a `lessons-*.md` entry and include the incident; most tasks won’t add one.
  - Linked from `CLAUDE.md` and `AGENTS.md`; read the relevant lessons file before starting.

- **Bug Fixes**
  - Fixed DoD numbering in `AGENTS.md` (Lessons learned is 9; ADR hygiene is 10).
  - Corrected the task-ID sweep to use numeric sort; added a measured example showing why.
  - Replaced a dead verify link with an inline `tmux` recipe and notes on `TLDW_CONFIG_PATH` isolation and Ctrl+digit limits.

<sup>Written for commit fff98616702c5eafd4c6ac30ffa1b7d612da1ad4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/951?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -435,7 +435,7 @@ A task is **Done** only when **ALL** of the following are complete:
    just the rule**: a lesson without the evidence that produced it decays into folklore
    and gets ignored. Most tasks produce nothing here, and that is fine; do not invent
    one to fill the slot.
-9. **ADR hygiene**: ADR check completed; any new or superseded ADRs are linked from the task Implementation Plan and Implementation Notes.
+10. **ADR hygiene**: ADR check completed; any new or superseded ADRs are linked from the task Implementation Plan and Implementation Notes.
 
 ⚠️ **IMPORTANT**: Never mark a task as Done without completing ALL items above.
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -237,6 +237,13 @@ Critical files for common tasks:
 - Every implementation decision starts with reading the corresponding Markdown task file.
 - Project documentation is in **`backlog/docs/`**.
 - Project decisions are in **`backlog/decisions/`**.
+- Hard-won working knowledge is in **`backlog/docs/lessons-*.md`** -- traps that have
+  actually cost time in this repo, each recorded with the incident that produced it.
+  **Read the one covering your area before starting**; they are short, and they exist
+  because these mistakes recur:
+  - `lessons-testing-evidence.md` -- what actually counts as evidence a change works
+  - `lessons-live-verification.md` -- running the app and talking to a real server
+  - `lessons-backlog-hygiene.md` -- task IDs, CLI quirks, git plumbing
 - Canonical Architecture Decision Records (ADRs) live in **`backlog/decisions/`**. Historical ADR-like material elsewhere is reference-only unless a canonical ADR imports or supersedes it.
 - Before implementation planning, read relevant ADRs and decide whether the task requires a new ADR.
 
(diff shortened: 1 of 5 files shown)


rmusser01:
## All three findings were real — fixed in `fff9861`

**Duplicate DoD numbering.** `AGENTS.md` already carried a ninth item (ADR hygiene) that `CLAUDE.md` does not, so inserting "Lessons learned" as 9 produced two. ADR hygiene is now 10.

**The task-ID sweep sorted lexicographically** — and this is the one worth dwelling on. The documented command used `sort -u`, which puts `task-100` before `task-99`. Run against this repo:

| | reported max |
|---|---|
| `sort -u` (as documented) | **99** |
| `sort -n -u` (fixed) | **838** |

A file whose entire subject is ID collisions was teaching a command that guarantees them. The measured before/after is now in the entry, because the number is more convincing than the rule.

**Dead link to the verify recipe.** `lessons-live-verification.md` pointed at `.claude/skills/verify/`, but `.gitignore` line 7 ignores `.claude` and none of it is tracked — the reference resolves to nothing on a fresh clone. The tmux recipe is inlined instead, with the profile-isolation flag and the note that Ctrl+digit cannot be sent through tmux.

On the strategic suggestion about frontmatter: agreed, and it is the same gap I flagged in the PR body — `backlog doc list` renders all eight docs blank, including the five that predate these. Adding frontmatter changes the format of existing docs, so it belongs in a separate change (and arguably in backlog-md-py rather than here).

Worth noting a documentation PR produced three genuine correctness bugs, one of which inverted the advice it was giving.

PR #952, opened 2026-07-27
fix(audio): macOS gets its preferred engine, and audio ingest stops crashing (tasks 839, 840)

rmusser01:
Two bugs, the first of which was **hiding** the second. Audio and video ingest could never complete.

## 839 — macOS never gets its preferred engine

`config.py`'s darwin branch prefers `parakeet-mlx`, then `lightning-whisper-mlx`, then `faster-whisper`. That logic is correct. But **nothing installed either Apple Silicon engine**: the `audio` and `video` extras pinned `faster-whisper` and mentioned the others only in a comment suggesting a second manual install.

So the preference could never engage. On a machine with `soundfile`, `scipy` and `numpy` all present, audio ingest failed with:

```
FileIngestionError: Audio processing failed:
faster-whisper is not installed. Install with: pip install faster-whisper
```

…while the app had logged *"No macOS-specific STT providers found, using faster-whisper as default"* at startup — the fallback firing exactly as written, for want of a package no install path provided.

Both extras now carry `parakeet-mlx; sys_platform == 'darwin'`. `faster-whisper` stays for every other platform and as the darwin fallback, so **non-macOS installs are unchanged**. Confirmed after installing: the log flips to *"Detected parakeet-mlx available on macOS, setting as default STT provider."*

## 840 — the crash that gap was hiding

With an engine finally present, transcription succeeded and **chunking died**:

```
Failed to chunk text using method 'words':
'<=' not supported between instances of 'dict' and 'int'
```

`ChunkingService.chunk_text` takes flat keyword arguments `(content, chunk_size, chunk_overlap, method)`. The audio processor passed an options **dict positionally**, so it landed in `chunk_size` and was compared against an integer.

A second defect sat underneath: `chunk_text` returns records like `{'text': ..., 'start_char': ..., 'word_count': ...}`, not bare strings, but the caller wrapped each as `{"text": chunk}` — nesting the whole record under another `text` key. Fixing only the call would have written dictionaries into the media content.

## Verified with a real transcript

Not by absence of an exception. macOS `say` synthesised a Gettysburg line, `afconvert` made it 16 kHz mono, and after the fix the pipeline returns 82 characters:

> `4-Score and 7 years ago our fathers brought forth on this continent to new nation.`

Recognisably the input — so transcription, chunking and persistence all work. `Tests/Local_Ingestion`: **77 passed**.

## Why neither was caught before

The missing engine **masked** the crash: without any transcriber the path failed earlier, so nothing reached chunking. Fixing packaging is what made it reachable.

And my first fixture was a 440 Hz sine tone, which failed with *"No text could be extracted"* — that is task-677's empty-extraction guard behaving **correctly** on audio containing no speech. A tone cannot distinguish a working transcriber from a broken one; only synthesized speech can.

## Pattern

840 is the same shape as the pagination defect in 684.2 — a caller assuming a signature the callee does not have — which is the first entry in `backlog/docs/lessons-testing-evidence.md`, added earlier today. It recurred within the same session, in a different subsystem.

---
## Summary by cubic
Fix macOS audio/video ingest by shipping `parakeet-mlx` on macOS and repairing chunking and storage so transcription completes. Normalize chunk shapes and keep real offsets to stop JSON/XML crashes and overlapping range errors. Addresses Linear 839, 840, and 841.

- **Bug Fixes**
  - Audio/video: call `chunk_text` with named args (`content`, `chunk_size`, `chunk_overlap`, `method`), preserve `start_char`/`end_char`/`chunk_index`, and avoid nesting. Storage uses real offsets with a length-sum fallback only when missing. Added seam tests for signature and offset handling.
  - RAG chunking: add `_chunk_to_text` to handle string and dict chunks; serialize structured chunks missing `text`. Contract tests across methods; skip `semantic` when NLTK data is absent.

- **Dependencies**
  - Add `parakeet-mlx` to `audio` and `video` extras for `sys_platform == 'darwin'`; keep `faster-whisper` as fallback and the non-macOS default. macOS now gets its preferred engine as defined in config.

<sup>Written for commit e3ac7dcd716064c8fd188dadd8756ca9b3c4fb41. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/952?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Local_Ingestion/test_audio_chunking_seam.py
+++ b/Tests/Local_Ingestion/test_audio_chunking_seam.py
@@ -1,0 +1,66 @@
+"""Contract tests for the audio processor's chunking seam (tasks 840, 952 review)."""
+
+from __future__ import annotations
+
+import inspect
+
+import pytest
+
+
+class TestAudioChunkingSeam:
+    """The audio path must call the chunking service the way it is declared.
+
+    Every audio and video ingest that reached chunking used to die with
+    "'<=' not supported between instances of 'dict' and 'int'": an options dict
+    was passed positionally into ``chunk_size``. A signature check is the guard
+    that would have caught it, since a hand-written double would have been
+    written to match the wrong call (task-840).
+    """
+
+    def test_the_service_signature_is_the_one_the_caller_assumes(self):
+        from tldw_chatbook.RAG_Search.chunking_service import ChunkingService
+
+        params = inspect.signature(ChunkingService.chunk_text).parameters
+        for name in ("content", "chunk_size", "chunk_overlap", "method"):
+            assert name in params, f"chunk_text lost its {name} parameter"
+        # The defect: a dict landing where an int is declared.
+        assert params["chunk_size"].annotation in (int, "int")
+
+    def test_chunk_text_returns_string_text_and_keeps_offsets(self):
+        """Offsets must survive: the storage path otherwise re-derives them by
+        summing lengths, which double-counts whenever chunks overlap."""
+        from tldw_chatbook.Local_Ingestion.audio_processing import LocalAudioProcessor
+
+        proc = LocalAudioProcessor.__new__(LocalAudioProcessor)
+
+        class _Service:
+            def chunk_text(self, content, chunk_size=400, chunk_overlap=100, method="words"):
+                assert isinstance(chunk_size, int), "an options dict reached chunk_size"
+                return [
+                    {"text": "alpha beta", "start_char": 0, "end_char": 10, "chunk_index": 0},
+                    {"text": "beta gamma", "start_char": 5, "end_char": 15, "chunk_index": 1},
+                ]
+
+        proc.chunking_service = _Service()
+        out = proc._chunk_text("alpha beta gamma", method="words", max_size=2, overlap=1)
+
+        assert [c["text"] for c in out] == ["alpha beta", "beta gamma"]
+        assert all(isinstance(c["text"], str) for c in out)
+        # Overlapping chunks: summing lengths would give 0 and 10, not 0 and 5.
+        assert [c["start_char"] for c in out] == [0, 5]
+        assert [c["end_char"] for c in out] == [10, 15]
+
+    def test_a_string_chunker_result_still_works(self):
+        from tldw_chatbook.Local_Ingestion.audio_processing import LocalAudioProcessor
+
+        proc = LocalAudioProcessor.__new__(LocalAudioProcessor)
+
+        class _StringService:
+            def chunk_text(self, content, chunk_size=400, chunk_overlap=100, method="words"):
+                return ["one two", "three four"]
+
+        proc.chunking_service = _StringService()
+        out = proc._chunk_text("one two three four", method="words")
+
+        assert [c["text"] for c in out] == ["one two", "three four"]
+        assert all("start_char" not in c for c in out), "offsets must not be invented"
--- a/Tests/RAG/test_chunking_service.py
+++ b/Tests/RAG/test_chunking_service.py
@@ -1,0 +1,71 @@
+"""Contract tests for ChunkingService across every supported method."""
+
+from __future__ import annotations
+
+import pytest
+
+
+class TestEveryChunkingMethodReturnsUsableText:
+    """Every supported method must yield dicts whose ``text`` is a string.
+
+    ``Chunker.chunk_text`` is not uniform: the text methods (words, sentences,
+    paragraphs, tokens, semantic) return plain strings, while the
+    structure-aware ones (json, xml, ebook_chapters) return dicts carrying their
+    text alongside metadata. ``ChunkingService`` assumed strings and called
+    ``.split()`` on them, so **every json and xml chunk request died** with
+    "'dict' object has no attribute 'split'" (task-841).
+
+    Found by exercising each method rather than by reading the dispatch, which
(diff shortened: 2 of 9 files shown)


rmusser01:
## Added: UAT of every chunking method (`1c467bf`)

Prompted by 840 being a chunking bug, I exercised **all seven** supported methods through the real service. **Two were completely broken.**

| Method | Before | After |
|---|---|---|
| words, sentences, paragraphs, tokens | ✅ | ✅ |
| semantic | ✅ | ✅ (skips without NLTK data) |
| **json** | ❌ crash | ✅ |
| **xml** | ❌ crash | ✅ |

### The bug — same seam, third time today

`Chunker.chunk_text` is **not uniform**: text methods return plain **strings**, structure-aware ones (`json`, `xml`, `ebook_chapters`) return **records** carrying text plus metadata. `ChunkingService` post-processed assuming strings and called `.split()`, so every JSON and XML chunk request died with `'dict' object has no attribute 'split'`.

A `_chunk_to_text` normaliser now accepts either shape, and a structured record with no text field is **serialised rather than dropped** — content still reaches the index instead of vanishing silently.

### I nearly filed this as two bugs

JSON first failed with *"Chunkable key 'data' not found"*, which reads like a deliberate constraint — I almost recorded it as by-design. That was the JSON chunker's **own input validation firing before the real defect could**. Retesting *with* a `data` key produced the same `.split()` error as XML, revealing one bug rather than two.

### Second finding, filed not fixed

**task-842**: `semantic` needs NLTK `punkt` data that is a **runtime download, not a declared dependency** — the same call succeeded in one environment and raised `LookupError` in another, purely on what was cached. The new test skips explicitly on that condition; letting a data gap masquerade as a shape bug would bury both.

### Pattern

This is the **third instance in one session** of a caller assuming a shape its callee does not produce — after the remote-ingest pagination defect (684.2) and the audio chunking call (840). Same lesson each time, and it is the first entry in `backlog/docs/lessons-testing-evidence.md`.

`Tests/RAG` + `Tests/Local_Ingestion`: **792 passed, 9 skipped**. Regression test parametrises over every method.

rmusser01:
## Both findings addressed (`e3ac7dc`)

**Chunk offsets discarded** — valid, and a correctness bug I introduced while fixing the crash. My 840 fix stripped chunk records down to just `text`, throwing away the `start_char`/`end_char`/`chunk_index` the service actually provides. `_store_in_database` then re-derives offsets by summing prior chunk lengths, which **double-counts whenever chunks overlap** — and overlap is on by default.

The existing comment there even says *"This is approximate since we don't have exact character positions"*. But it does have them; my normaliser was discarding them one function earlier.

Offsets are carried through and used now, with length-summing kept only as a fallback for chunkers returning bare strings.

Worth naming the shape of this: the crash I was fixing was **loud**, and the bug I put in its place would have **silently** stored wrong character ranges for every overlapping audio chunk. Good catch.

**No tests for `_chunk_text`** — fair. I had verified by live speech ingest but locked nothing in. Three tests now:

| Test | Guards |
|---|---|
| Service signature still declares chunk_size as an int | The original 840 defect — a hand-written double would have been written to match the wrong call, which is exactly how the pagination bug in 684.2 survived |
| Offsets survive, using **overlapping** chunks (0–10, 5–15) | Length-summing would give 0 and 10 rather than 0 and 5 |
| String-returning chunker still works | And offsets are **not invented** when none are supplied |

Mutation-checked: restoring the positional-dict call fails the first; dropping the offsets fails the second.

Re-verified end to end — the same synthesised speech still transcribes to *"4-Score and 7 years ago our fathers brought forth on this continent to new nation."* `Tests/Local_Ingestion`: **80 passed**.

On the third suggestion (transcription engines as dedicated extras): offered as a later refactor, and not taken here — this PR already carries tasks 839, 840 and 841.

rmusser01:
## Both findings are addressed on the reviewed commit

**Chunk offsets discarded** — marked ✓ Resolved. Agreed, and thank you: that was a correctness bug I introduced while fixing the crash, and it would have failed silently.

**No tests for `_chunk_text`** — this one appears stale. The tests landed in `e3ac7dc`, which is the commit that was reviewed:

```
$ git show origin/fix/macos-stt-default-839:Tests/Local_Ingestion/test_audio_chunking_seam.py | grep -c "def test_"
3
```

`Tests/Local_Ingestion/test_audio_chunking_seam.py` is present on the remote at that exact SHA, with three tests, two of which call `proc._chunk_text(...)` directly:

| Test | Guards |
|---|---|
| Signature contract | The original defect — that the service still declares its chunk-size parameter as an int, which is what a hand-written double would have hidden |
| Offsets survive | Uses genuinely overlapping chunks (0–10, 5–15), where the old length-summing gives 0 and 10 instead of 0 and 5 |
| String chunker still works | And that offsets are not invented when the chunker supplies none |

Both are mutation-checked: restoring the positional-dict call fails the first, dropping the offsets fails the second. `Tests/Local_Ingestion`: 80 passed.

Happy to add more if there is a specific path you think is uncovered — but the two behaviours named in the finding (the call shape and the output shaping) are each pinned by a test that fails when reverted.

PR #953, opened 2026-07-27
feat(agents): agent-runtime substrate + sensitive-path denylist + glob/grep tools

rmusser01:
Supersedes #936. Same substrate work, re-homed onto current `dev` after `TASK-545 P2` and `TASK-656` landed there independently — see "Relationship to #936" below.

## Runtime substrate

- **Tool ceiling 8 → 24, disclosure threshold 8 → 16.** `max_active_tools` is a one-way ratchet: `load_tools` returns `"no room"` and nothing ever unloads, so an agent that loads early is locked out for the rest of the run. The threshold rises with it because `initial_disclosure` runs *per run* — a catalog above it costs `find_tools`+`load_tools` round trips before any real work, on every message.
- **Tool results capped at the history-append seam.** Only the persisted step record was bounded (2000 chars); the history append was unbounded. New `RunBudget.max_tool_result_chars` (16000, `0` = unlimited) applied once, unconditionally, before `add(STEP_TOOL_RESULT, ...)` — so it covers the review-hook refusal path too, not just the normal one. Uniform across built-in, MCP and skill results.
- **Per-tool timeout override**, resolved through the registry and duck-typed so MCP/skill providers keep the run budget. `0` resolves to `None`, never a literal zero-second timeout.
- **`network` in `BUILTIN_HIGH_RISK_TAGS`.** Egress from a built-in floors to `ask`. Deliberately *not* added to the shared `HIGH_RISK_TAGS`: dev's comment there states widening it would make remote MCP tools start prompting and that this is not the current phase's call. This respects that.
- **MCP name shadowing surfaced.** Built-ins still win collisions — inverting that would let a hostile MCP server name-squat an audited built-in — but a dropped tool is now logged instead of vanishing silently. `_partition_mcp_catalog_by_collision` computes both sides in one walk so the filter and the reporter cannot drift.
- **Console budget 20 → 30 turns**, `max_steps` 64 → 96 (`3*(N-1)+1` = 88 at N=30), wall 1200 → 1800s, plus `CONSOLE_MAX_TOTAL_TOKENS`. Sub-agents still inherit turns per the documented operator decision; the token ceiling bounds spend instead of clamping them.

## Path hardening

`Utils/sensitive_paths.py` — credential and app-state paths refused **regardless of sandbox root**, matched by resolved ancestry (never string prefix), failing closed. Wired into every file tool after `validate_path` and before any filesystem access — deliberately *not* inside `validate_path` itself, which has ~40 first-party call sites that legitimately validate the app's own config and DB paths.

Review found the first cut of this **did not work**: the permission store was hardcoded as `~/.config/tldw_cli/mcp_permissions.json`, but the app builds it under `get_user_data_dir()`. The literal never existed, so the check never fired — `write_file` was demonstrated overwriting the real store with `{"global_default": "allow"}`, the exact permission-gate bypass the denylist exists to prevent. Now derived from the app's own accessor, with a regression test that re-derives it the same way rather than asserting a string that can drift back to matching nothing.

Related coverage that followed from it: state files created outside `config.py` (`agent_runs.db`, `evals.db`, `local_mcp_store.json`, logs) are covered by a rule rather than an enumeration — every *file* directly in `get_user_data_dir()` is refused, while container subdirectories including the default sandbox root stay reachable. `config.toml` denial honours `TLDW_CONFIG_PATH`. SQLite `-wal`/`-shm`/`-journal` sidecars are denied per-file (a directory-wide deny would break the shipped default, since the sandbox root is their sibling).

## Two new tools

`glob_files` and `grep_files`, registered as `GateableTool` entries and defaulting OFF like every sibling. The Settings UI picked them up with no changes — dev's `_GATEABLE_BUILTINS` really is the single source of truth it documents itself as.

Bounded deliberately: `_MAX_MATCHES` caps results and `_MAX_CANDIDATES` caps entries *examined*. Both are needed — `Path.glob("../**/*")` does not raise, it yields ~1.4M paths, and out-of-root candidates never match, so a match-only bound never trips. `grep_files` streams line-by-line with a per-file byte cap rather than reading whole files. `Path.glob()` is lazy, so invalid-pattern `ValueError` fires on first iteration and is guarded there.

They enforce the same hidden-file rule as `read_file`. Without it, verified exploitable:

```
read_file('.env')                       -> {'error': 'Access to hidden files/directories is not allowed'}
grep_files('API_KEY', glob='**/.env')   -> 'API_KEY=supersecret123'
```

`.env` is not on the denylist and does not need to be — `read_file` already refused it structurally. A dotted sandbox root inverted the same protection one level up and is fixed the same way.

## Relationship to #936

#936 was built before `TASK-545 P2` and `TASK-656` landed and is 418 commits behind. Dev solved built-in tool registration with the `_GATEABLE_BUILTINS` table rather than the "packs" architecture that branch proposed, so this PR drops packs entirely and keeps only what dev does not have. Also dropped: the `ToolExecutor` deletion (dev already did it) and the `Tools_Settings_Window` deletion — that one would have *regressed* a surface dev has since improved. #936 will be closed.

## Testing

`Tests/Agents/ Tests/Tools/ Tests/Utils/ Tests/Chat/test_console_agent_bridge.py Tests/UI/test_tools_settings_window.py` → **972 passed**, 16 skipped, 6 failed (all pre-existing `test_chat_api_key_*`, confirmed by stash bisection).

The full suite takes ~3h40m here and cannot fully collect (`numpy`/`pytest-mock` absent from the venv). A full run on the predecessor branch gave 73 failures across `Tests/UI`, `TTS`, `RAG`, `Audio`, `Local_Ingestion`, `Web_Scraping`, `Watchlists`, `LLM_Provider_Catalog` and `DB` — **zero** in `Agents`, `Tools`, `Utils` or `Chat`, the only areas touched here.

## Known follow-ups (not blockers)

- `is_sensitive_path` docstring promises fail-closed on unresolvable paths, but `_resolved` catches only `OSError`/`RuntimeError`; a NUL byte raises `ValueError`. Unreachable today — `validate_path` rejects first and the tools' outer catch contains it.
- The direct-child rule covers the active user folder only; under a deliberately widened root, `chromadb/chroma.sqlite3` and sibling profile folders stay readable. Disclosure asymmetry, not a gate bypass — skill trust manifests are HMAC+keyring authenticated and script grants digest-pinned.
- An agent can create a *directory* named after a not-yet-created state file, which would later break the app's open of it. DoS only.
- `glob_files`/`grep_files` are scoped to the sandbox root and do not honour dev's `allowed_file_roots` workspace folders, so they are strictly narrower than their siblings.

---
## Summary by cubic
Adds an agent-runtime substrate with safer file tools and two new gated tools (`glob_files`, `grep_files`). Increases budgets, caps tool-result text, adds per-tool timeouts, hardens path checks to prevent credential/app-state leaks, mitigates `grep_files` regex backtracking, and tracks follow-ups.

- **New Features**
  - Raise tool capacity to 24 and disclosure threshold to 16; increase console limits (30 turns, 96 steps, 1800s) and add CONSOLE_MAX_TOTAL_TOKENS.
  - Cap tool result text at history-append via RunBudget.max_tool_result_chars (16k; 0 = unlimited); per-tool timeouts via `Tool.timeout_seconds` resolved through the registry.
  - Add gateable `glob_files` and `grep_files`, scoped to the sandbox, with hidden-file rules, strict candidate/match bounds, dotted-root refusal, and catch-all error handling.
  - Treat `network` as high risk for built-ins (`BUILTIN_HIGH_RISK_TAGS`) and warn once on MCP tools shadowed by built-in names (single catalog walk).

- **Bug Fixes**
  - Enforce a sensitive-path denylist (credentials, config, DBs, permission store, logs) resolved via app accessors (e.g., `get_user_data_dir()`), independent of sandbox root, and filter sensitive files from directory listings.
  - Apply denylist and hidden-file checks across `read_file`/`write_file`/`list_directory`/`glob_files`/`grep_files`; block dotted sandbox roots in all tools; add outer catch-all in `glob_files`/`grep_files`.
  - Mitigate regex ReDoS in `grep_files` with per-line search and total-line caps plus a 20s per-tool timeout.
  - Deduplicate the shadowed-MCP-name warning and avoid duplicate catalog walks during run composition.
  - Track remaining hardening items as backlog tasks (TASK-839–846; moved the cost_estimation task to 844; added decisions/audits for the `network` risk tag set and path-checks against actual app write locations).

<sup>Written for commit 790dc41740fc979af72bd14df707d30d79292cf1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/953?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## Qodo findings addressed — `2fcc4a827`

Two fixed, one declined with evidence. 879 targeted tests green.

**3 (regex runaway threads) — accepted, and it was the best finding here.** Confirmed live before the fix: `re.compile(r'(a+)+$').search('a'*60 + 'X')` still running after 3s on a **61-character** line, and the code searched the *full* line while truncating only the stored result. The cross-component reasoning is what makes this sharp — `_call_with_timeout` **abandons** the daemon thread on timeout, so a pathological pattern keeps burning CPU after the agent has already reported failure, and repeated calls stack.

Mitigations, stated for what they actually buy:
- `_MAX_GREP_LINE_SEARCH_CHARS = 500` — the slice handed to `regex.search`. This is the only bound that genuinely constrains the worst case, since the thread cannot be killed.
- `_MAX_GREP_LINES_SCANNED = 200_000` — total lines per call, so a huge file cannot extend exposure.
- `GrepFiles.timeout_seconds = 20.0`, via the per-tool override seam this PR adds.

Measured: `(a+)+$` over a 4,000-char line went from unbounded (11.64s and climbing, no match) to **0.136s**; ordinary searches unaffected.

This does **not** eliminate catastrophic backtracking — a pathological pattern against a capped line can still be costly, and a timeout still will not stop it. It makes the worst case finite and small instead of unbounded. A complete fix needs a regex engine with timeout support or a killable subprocess. `grep_files` also carries the `"reads"` risk tag, so it floors to `ask` and a human approves each call — which is part of why the partial mitigation is proportionate here rather than a blocker.

**2 (import grouping) — accepted.**

**1 (glob/grep skip `validate_path`) — declined; the enforcement is present, just not via that helper.** Both tools already apply, in order: `_rejects_traversal` (parent-traversal and absolute patterns, incl. Windows drive-letter/UNC), `_sandbox_root_is_hidden` (mirrors `validate_path`'s dotted-base rule), `is_within` (containment, which also applies the credential/app-state denylist via `is_sensitive_path`), and `_is_hidden_within` (the dot-component rule `validate_path` enforces).

Calling `validate_path` per candidate would add no enforcement these four do not already provide, and would actively hurt on two counts: it **raises `ValueError`**, and these tools must return an error dict rather than raise; and it costs measurably more per candidate (re-measured on this base: inline +20.0% vs `validate_path`-per-candidate +42.8% over a 1,500-file tree). The rule was folded into the tools rather than the shared helper deliberately — `ListDirectoryTool` has an `include_hidden` opt-in that a shared-helper change would break.

Worth noting the equivalent finding on the predecessor PR was **correct** and caught a real hole — `grep_files` was reading `.env` files that `read_file` refused. That gap is closed here and covered by regression tests; this is the same finding against code that now enforces the rule.

PR #954, opened 2026-07-27
fix(chunking): fall back when nltk's corpus is missing, not just nltk (task 842)

rmusser01:
Semantic chunking raised `LookupError` on some machines and worked fine on others — purely on what happened to be cached.

## The gap

`_ensure_nltk()` handled nltk being **absent**, falling back to a built-in regex sentence splitter. It did not handle nltk being **present without its data**: it bound the real tokeniser, which then raised deep inside a chunking call.

The `punkt` corpus is a **runtime download**, not part of the package. So the same call succeeded or failed depending on whether some earlier process had fetched it — and the failure read as an internal fault rather than a missing optional asset.

## The fix

Probe the tokeniser once at bind time. If the corpus is missing, keep the regex fallback and log the exact remedy:

```
nltk is installed but its 'punkt' tokeniser data is not, so sentence splitting
will use the simpler built-in fallback. To get nltk's tokeniser, run:
python -m nltk.downloader punkt punkt_tab
```

Simpler sentence splitting beats a failed ingest, and this **reuses the fallback that already existed** for the nltk-absent case rather than inventing a second path.

## Verified the way that settles it

`NLTK_DATA` pointed at an empty directory, so punkt is definitively unavailable:

```
SEMANTIC without punkt: OK n=4 | first: 'The first sentence is here. The second follows'
```

Four usable chunks instead of a raise. The regression test does the same via monkeypatch and a module reload, so **it does not depend on what the developer happens to have cached** — which was the whole problem, and would have made a naive test pass on one machine and fail on another.

## Also

My first draft used `logging.warning` while this module uses `loguru` — the message would have gone somewhere the app's log filters do not look. Corrected.

`Tests/RAG` + `Tests/Chunking` + `Tests/Local_Ingestion`: **810 passed, 8 skipped**. The semantic case that previously had to be *skipped* for missing data now passes outright.

Follows #952, which fixed the JSON/XML chunk-shape crash (841) and filed this.

---
## Summary by cubic
Semantic chunking now probes `nltk`’s tokenizer and falls back to the built-in sentence split if its corpus is missing; it then downloads `punkt` and `punkt_tab` and re-probes. This prevents LookupError crashes and keeps ingest working, with other chunking methods unaffected (TASK-842).

- **Bug Fixes**
  - Probe `sent_tokenize` before binding; on failure, fetch `punkt` and `punkt_tab`, re-probe, and latch an unusable verdict to avoid repeated downloads; log the remedy via `loguru` (run `python -m nltk.downloader punkt punkt_tab`).
  - Replace the newline fallback with the module’s sentence split; catch the English retry so missing data in both languages degrades cleanly instead of crashing or returning one giant chunk.
  - Update `ensure_nltk_data()` to use probe-based readiness instead of naming a specific corpus, making it work across `nltk` versions.
  - Tests and docs: monkeypatch-based corpus-missing tests validate fallback, download path, and latching; document that `NLTK_DATA` appends to `nltk.data.path` to avoid false-positive verification.

<sup>Written for commit 3de45fec6886389d8e1307e4fbc1015bee85314e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/954?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/RAG/test_chunking_service.py
+++ b/Tests/RAG/test_chunking_service.py
@@ -69,3 +69,37 @@ class TestEveryChunkingMethodReturnsUsableText:
         assert _chunk_to_text({"heading": "H", "body": "B"})
         assert _chunk_to_text({"text": "plain"}) == "plain"
         assert _chunk_to_text("already a string") == "already a string"
+
+
+class TestSemanticChunkingWithoutNltkData:
+    """Semantic chunking must degrade, not raise, when its corpus is absent.
+
+    nltk's sentence tokeniser needs the 'punkt' corpus, which is a runtime
+    download rather than part of the package. A machine could therefore have
+    nltk installed and still raise LookupError deep inside a chunking call --
+    the same call succeeding or failing purely on what happened to be cached
+    (task-842). A simpler sentence split is a better outcome than a failed
+    ingest, and the fallback already existed for the nltk-absent case.
+    """
+
+    def test_semantic_falls_back_instead_of_raising(self, tmp_path, monkeypatch):
+        import importlib
+
+        # An empty NLTK_DATA dir makes punkt definitively unavailable.
+        monkeypatch.setenv("NLTK_DATA", str(tmp_path))
+        import tldw_chatbook.Chunking.Chunk_Lib as chunk_lib
+
+        importlib.reload(chunk_lib)
+
+        from tldw_chatbook.RAG_Search.chunking_service import ChunkingService
+
+        prose = (
+            "The first sentence is here. The second follows it. A third arrives. "
+            "A fourth sentence appears. A fifth concludes."
+        ) * 2
+        chunks = ChunkingService().chunk_text(
+            prose, chunk_size=20, chunk_overlap=5, method="semantic"
+        )
+
+        assert chunks, "semantic chunking produced nothing without punkt data"
+        assert all(isinstance(c.get("text"), str) and c["text"].strip() for c in chunks)
--- a/Tests/RAG/test_chunking_service.py
+++ b/Tests/RAG/test_chunking_service.py
@@ -74,22 +74,67 @@ class TestEveryChunkingMethodReturnsUsableText:
 class TestSemanticChunkingWithoutNltkData:
     """Semantic chunking must degrade, not raise, when its corpus is absent.
 
-    nltk's sentence tokeniser needs the 'punkt' corpus, which is a runtime
-    download rather than part of the package. A machine could therefore have
-    nltk installed and still raise LookupError deep inside a chunking call --
-    the same call succeeding or failing purely on what happened to be cached
-    (task-842). A simpler sentence split is a better outcome than a failed
-    ingest, and the fallback already existed for the nltk-absent case.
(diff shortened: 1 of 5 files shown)


rmusser01:
Both findings were real, and the bug one led somewhere deeper. Addressed in `bef` (see below for the actual sha).

## 2 — nltk failure not cached: correct, and it was hiding a worse fault

You were right that returning `None` without recording the verdict makes `ensure_nltk_data()`'s download unreachable. Chasing *why that download mattered* showed it was already broken on dev. Reproduced with a corpus containing exactly what `nltk.download("punkt")` produces:

```
ensure_nltk_data() would report _nltk_data_ready = True
semantic chunking: LookupError -> "Resource 'punkt_tab' not found."
```

It checks for and fetches **punkt**; nltk >= 3.9 reads **punkt_tab**. So the remediation provisioned the wrong resource and marked itself ready — and my original patch would have masked that permanently rather than fixed it.

**Which resource nltk wants is version-dependent, so naming one is a guess.** Readiness is now whether the tokeniser *tokenises*: probe at bind time, on failure fetch both corpora and re-probe, and let the probe — not the download's own return value — decide. An nltk too old to know `punkt_tab` reports failure for it while being perfectly usable, so that return value cannot be trusted. The unusable verdict is latched, which is your point exactly: `nltk` stays `None` on that path, so without it every call re-probes, re-downloads and re-warns.

## A separate defect this uncovered

With the fallback finally running for real, semantic chunking returned **1 chunk** — the whole document. The fallback split on *newlines*, so ordinary single-paragraph prose came back whole: not raising, while silently not chunking. Its `LookupError` branch also retried with `language="english"` uncaught, which is the crash shape this task was filed for. Both now route to the sentence split; a live run with the corpus genuinely unreachable gives 4 chunks.

## A false positive of my own, corrected

The PR description claimed verification via `NLTK_DATA` pointed at an empty directory. That proved nothing — **`NLTK_DATA` appends to `nltk.data.path`, it does not replace it**, so those runs kept finding the real corpus at `~/nltk_data` and exercised the working tokeniser, never the fallback. With `nltk.data.path[:]` actually restricted, the first honest run failed immediately. Corrected in the task notes and recorded in `lessons-live-verification.md` as a second incident of a rule already there.

## 1 — docstring: added

Along with docstrings on the two new tests and the shared helper.

## Mutation checks

Removing the probe, the download, the latch, or probe-based readiness each fails **exactly one** test and nothing else. One existing test (`test_nltk_download_false_is_not_logged_as_success`) had to move to the new seam — its intent, "a failed download is not reported as success," is preserved and now also asserts readiness stays `False`.

`Tests/RAG` + `Tests/Chunking` + `Tests/Local_Ingestion` + `Tests/Utils`: **1239 passed, 8 skipped**.

PR #955, opened 2026-07-27
Evals rebuild PR 3b: results grid, five lenses, empty states, sample bench

rmusser01:
## What this is

PR 3b of the Evals Console rebuild: the **results grid and its five lenses**, empty states, a
one-click sample bench, and export. Earlier PRs retired the unreachable Evals UI (#922, 9,783 LOC),
added the word bench engine (#924, schema v4), and shipped the Console-styled screen shell (#941).

A word bench captures the next-token distribution for each snippet x target pair to surface
censorship / directional prompting. This PR is the read side: the grid that makes those captures
comparable.

## The five lenses

`Top-1` · `Entropy` · `Probe` · `Truncation` · `Δ baseline`, switchable live over the same run group.

Three renderings would misrepresent the engine, and each is pinned by a test:

- **A bare Top-1 on a near-tie.** Rank order was observed flipping between two identical requests to
  the same server, so a near-tie renders a visible marker rather than a confident winner.
- **A `≥` prefix on divergence.** PR 2's review disproved the lower-bound claim; divergence is an
  estimate over top-K plus one lumped bucket, and is labelled as one.
- **Entropy without a shared `k`.** A K=5 column beside a K=20 column would read as more confident
  with no behavioural difference, so compared cells truncate to `min(K)`.

All statistics live in `Evals/word_bench/analysis.py`. The grid formats; it does not compute.

## Verified against a live model

Run end-to-end against llama.cpp (`gemma-4-26B ... Q4_K_M`), not just in tests:

- sample bench created from an empty library; grid filled row-major, 4/4 cells
- all five lenses switch live; readiness renders as readable labels, never colour-only
- the degenerate canary surfaces twice — an Inspector callout and a `[warned]` column marker — and
  the copy says how to *interpret* it: "a large divergence in its column may reflect that, not the prompt"
- neutral sampler persisted exactly as specced (temp 1.0, top_p 1.0, top_k 0, min_p 0, penalties off);
  `k_requested == k_returned == 20`; per-token `bytes` present, confirming the bytes-based token
  identity against a real server
- zero CSS warnings, zero ERROR lines

It also found the signal it exists to detect: loaded framings gave **lower** entropy (0.93 / 1.42)
than neutral ones (1.28 / 1.99).

## Review findings fixed in-branch

The whole-branch review returned 1 Critical, 5 Important, 14 Minor. The Critical and all five
Important are fixed here, plus five cheap Minors; the remaining nine are filed as TASK-845.

- **Critical — a predicate mutated the database.** `provider_is_configured()` wrote an `eval_models`
  row via `resolve_sample_target`, and `library_rail.compose()` called it — so merely *opening* the
  Evals screen created a phantom row, with no user interaction, on essentially every fresh install
  (`api_url = "http://localhost:8080"` ships as a default). Verified 0 -> 1 rows on a single call;
  now split query from mutation, verified 0 writes with the gate still answering correctly.
- A zero-snippet/zero-target run group crashed the screen (`NoMatches`) — latent today, live the
  moment "Run bench" is wired.
- The Truncation lens read native K while the header stated the shared K (1% vs 2% on cells with
  identical entropy at the stated K).
- The Probe lens rendered `probes[0]` unnamed — a multi-probe bench showed unattributable numbers.
- A configured `api_url` carrying a path (`.../completion`) produced a doubled endpoint -> 404 ->
  four failed cells on the flagship one-click path.
- `resolve_probe`'s match rule was reimplemented twice, including the `token == probe` comparison
  that bytes-identity would change — a future fix in `analysis.py` would have missed both sites.

Two of the folded-in Minors were assertions weaker than their names claimed (`"%" in text` where the
comment stated the expected value; `"K 5"` which also passes on `"K 50"`).

## Also here

12 dead Evals-only CSS selectors retired from `_evaluation_unified.tcss`; the other 21 have live
consumers across Chat, Logs, MCP and RAG search and are untouched. Bundle regenerated, not hand-edited.

## Out of scope, filed

- **TASK-848** — `calculate_semantic_similarity` computes a cosine similarity on both success paths
  and discards it, returning `None` exactly when embeddings work. Legacy orchestrator, unrelated.
- **TASK-843** — the Evals DB ignores the configured profile and always writes to `default_user/`,
  because `eval_orchestrator` looks up `user_id`/`username` while `load_settings()` publishes
  `USERS_NAME`. Scratch profiles silently write into real user data; that is how it was found.
- **TASK-845** — the nine deferred Minor findings.

---
## Summary by cubic
Rebuilds the Evals results experience with a pivoted results grid, five lenses, export, and a one‑click sample bench for empty states. Hardens first‑run so it works end‑to‑end and grid/storage errors don’t exit the app.

- **New Features**
  - Results grid with five lenses: Top‑1, Entropy (shared K), Probe, Truncation, and Δ baseline. Live lens switching, row/column baseline, sort‑by‑spread, and a focused cell inspector. Near‑ties in Top‑1 are marked; Δ warns when extrapolated (shows combined truncated mass). The grid only formats; all stats come from analysis.
  - Empty states and sample bench: when no benches exist, show “Create sample bench” or “Open Settings” based on provider availability. The sample bench auto‑resolves a `llama.cpp` target, runs with progress and cancel, and uses the neutral sampler.
  - Export: CSV (active lens) or JSON (full run) based on file extension, using the same data the grid renders.

- **Bug Fixes**
  - Critical: opening the Evals screen no longer writes to the database; provider checks are split from target creation.
  - Runner safety: a raising progress callback no longer strands runs; cancellations mark rows cancelled; the sample‑bench run is disable‑while‑running, double‑click guarded, and cancellable.
  - Grid robustness: zero‑snippet/zero‑target groups no longer crash; non‑ValueError DB faults render an error state instead of exiting; headers/options/cells render as literal text; the table focuses on mount so shortcuts work immediately.
  - Lens correctness: Truncation uses shared K; Probe lens names the active probe; baseline divergence and combined truncation are computed once and reused.
  - Connectivity/auth: normalizes configured `api_url` so `llama.cpp` requests hit the correct endpoint; API key resolution prefers env vars over config.
  - UX polish: sort‑by‑spread now places rows with fewer than two cells last in both sort directions; the empty‑state action (sample bench or Settings) shows whenever there are zero word benches, independent of classic tasks.

<sup>Written for commit 56ca2f9bbfd00e529c1c2a2f355859770983a0af. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/955?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Addressing the Qodo review. Three of the four findings are real and are being fixed; one I am declining, with reasoning.

**Fixing:**

1. **Env override ignored for the API key** — correct, and it contradicts this repo's own documented precedence (`CLAUDE.md`: env vars → config.toml → defaults). Inverting it, with a test that the env var wins when both are set.

3. **Public docstrings missing Google-style sections** — correct; adding the missing `Args:`/`Returns:` to `configured_llama_cpp_url`, `resolve_sample_target`, `render_token`, `render_probe_reading` and `truncated_mass`. Keeping the existing prose verbatim, since several of those docstrings record *why* the code is shaped as it is.

4. **Ascending sort misorders undefined-spread rows** — real bug, thanks. Spread is a Jensen–Shannon divergence and is always `>= 0`, so the `-1.0` sentinel sits below every real value by construction: correct for `desc`, wrong for `asc`. Verified with rows `[undefined, 0.05, 0.90]` → `asc` yields `[undefined, low, high]`, i.e. rows with no measurement at all lead the "least disagreement" view. Since the mode cycles `none → desc → asc → none`, that is reachable in normal use. Fixing with a direction-independent tuple key rather than a direction-dependent sentinel, so undefined rows trail in both directions, and correcting the docstring, which currently only claims the guarantee for descending.

**Declining:**

2. **`_DeltaReading` "not PascalCase"** — this is a false positive. The single leading underscore marks the name module-private, which is standard Python (PEP 8), and the remainder is correct PascalCase. Renaming would either make an internal dataclass look public or produce a worse name, so I am leaving it as-is and not adding a suppression comment either. The rule appears to test the identifier for any underscore rather than allowing a leading-underscore privacy prefix.

PR #957, opened 2026-07-27
fix(evals): calculate_semantic_similarity returned None whenever it worked

rmusser01:
## The bug

`MetricsCalculator.calculate_semantic_similarity` is annotated `-> float` but returned `None`
whenever it actually worked.

It reads as a flake and is not one. With `sentence_transformers` importable the function returns
`None`; with the import unavailable it returns a correct float via the lexical fallback. So it fails
on a developer machine with the model cached and passes on a bare CI box. The only environment where
it worked was the one where it could not do its job.

## Four defects, not one

1. **The result was computed and discarded on both success paths.** The numpy branch evaluated
   `dot(pred, exp) / (norm(pred) * norm(exp))` as a bare expression statement, and the pure-Python
   fallback did the same. Neither was returned or assigned, so control fell off the end to an
   implicit `None`. Only the two `except` handlers returned anything.

2. **The pure-Python zero-guard was inverted.** `dot_product / ((norm1 * norm2) if norm1 * norm2 > 0
   else 0.0)` divides *by* `0.0` when the norms vanish, raising `ZeroDivisionError` that the broad
   `except Exception` then silently converted into the lexical fallback. The intent was plainly to
   *return* `0.0`.

3. **Range mismatch.** Cosine similarity is `[-1, 1]`; callers and tests treat the result as a
   `[0, 1]` score. Unrelated text could score negative.

4. **Precision.** Fixing (1) exposed that float32 embeddings make a vector's similarity with itself
   land a few ULPs short of 1.0. Measured across 2000 random float32 vectors of MiniLM's 384 dims:
   only 802 self-similarities land exactly `1.0`, and 593 land *below* it, where a clamp cannot help.
   The real model returns `0.9999999999999998` for `"the cat sat on the mat"`.

   Upcasting to float64 alone would have made the existing test pass for its one input string
   without guaranteeing anything — the same "works for this input, on this machine" shape as the
   original bug. Fixed by construction instead: an exact-string-equality short-circuit before any
   embedding work, which is correct by definition and skips a pointless model round-trip. The
   lexical fallback already did exactly this. The float64 upcast is kept, because it genuinely
   improves the non-identical cases, but nothing now depends on it for exactness.

## Blast radius

Worse than a null score. In `_calculate_instruction_adherence` the `None` was appended to
`requirement_scores` and then summed, raising
`TypeError: unsupported operand type(s) for +: 'float' and 'NoneType'` — so any sample carrying an
`include`/`avoid` requirement crashed the scorer. A second path returned the `None` directly, which
is the pre-existing failure of `test_instruction_adherence_basic` (`assert None == 1.0`). It passes now.

## Why it survived

Every existing test exercised only the path where the dependency is **missing**. This PR adds tests
that assert a real float comes back when an embedding model **is** present, using stub models so they
stay fast and offline.

## Verified

- self-similarity is exactly `1.0` across multiple strings, including the one that failed
- the embedding path still produces graded scores (0.13-0.89), so the short-circuit did not swallow it
- opposite vectors clamp to `0.0` instead of `-1.0`
- zero-norm vectors return `0.0` instead of raising
- `Tests/Evals/`: 393 passed, 13 skipped, 0 failed

Every added test was checked by reverting the fix and confirming it fails.

## Root cause, filed as TASK-863

`Evals/metrics_calculator.py` is a 506-line orphan defining a second `MetricsCalculator` that nothing
imports. The two copies drifted and this one lost its `return`, with nothing comparing them. An AST
diff shows 9 of their 11 shared methods are byte-identical and `calculate_perplexity` differs only
cosmetically, so there is no second latent bug — but nothing prevents the next one. The live hazard
is that `Evals/README.md:309` instructs importing the orphan, which still returns unclamped cosine
values.

---
## Summary by cubic
Fixes `MetricsCalculator.calculate_semantic_similarity` to always return a [0, 1] float (never None/negative) and prevents instruction-adherence crashes. Also aligns the standalone `tldw_chatbook/Evals/metrics_calculator.py` copy with the same behavior and adds targeted tests. Addresses TASK-862 and progresses TASK-863.

- Bug Fixes
  - Return the computed cosine; clamp to [0, 1]; guard zero norms; compute in float64 for stability.
  - Short-circuit exact string equality to 1.0 before any embedding call; keep lexical fallback only when embeddings are unavailable or error.
  - Apply the same semantics to `tldw_chatbook/Evals/metrics_calculator.py` (added exact-match short-circuit, zero-guard, and clamping; fixed prior NaN case); add regression tests for this copy.
  - Add fast stub-embedding tests for the embedding-present path and edge cases (identical/opposite/zero-norm). Logged TASK-863 for consolidation of the duplicate.

<sup>Written for commit ea544a482f411b6442212c921537332c1e85081a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/957?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Evals/test_evaluation_metrics.py
+++ b/Tests/Evals/test_evaluation_metrics.py
@@ -6,8 +6,29 @@ Tests instruction_adherence, format_compliance, coherence_score, and dialogue_qu
 import pytest
 from unittest.mock import Mock
 
-from tldw_chatbook.Evals.eval_runner import BaseEvalRunner, EvalSample, EvalSampleResult
+from tldw_chatbook.Evals.eval_runner import (
+    BaseEvalRunner,
+    EvalSample,
+    EvalSampleResult,
+    MetricsCalculator,
+)
 from tldw_chatbook.Evals.task_loader import TaskConfig
+
+
+class _StubEmbeddingModel:
+    """Deterministic stand-in for a SentenceTransformer.
+
+    Lets tests exercise the "embedding model IS present" success path of
+    ``calculate_semantic_similarity`` without downloading or loading a real
+    model. Maps each input string to a fixed vector via ``encode()``, the
+    same method name/signature real SentenceTransformer models expose.
+    """
+
+    def __init__(self, vectors: dict):
+        self._vectors = vectors
+
+    def encode(self, texts):
+        return [self._vectors[text] for text in texts]
 
 
 class TestRunner(BaseEvalRunner):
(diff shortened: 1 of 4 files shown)


rmusser01:
Addressing the Qodo review. All three findings are being acted on, though the third gets a narrower fix than suggested and one part of its claim does not hold.

1. **Import groups out of order** — correct, reordering to stdlib → third-party → local.

2. **Missing docstring on the parametrized test** — correct, adding a Google-style docstring with `Args:` for `text`.

3. **The two `MetricsCalculator` copies disagree** — the substance is correct and worth having caught. `Evals/metrics_calculator.py` defines a second `MetricsCalculator` whose `calculate_semantic_similarity` still returns a raw cosine, so it can produce negative scores and non-1.0 self-similarity while this PR's tests imply otherwise — and `README.md:309` and `DEVELOPER_GUIDE.md` both point users at exactly that module.

    Two corrections, though:

    - **"at least one test" imports the `metrics_calculator.py` copy is not accurate.** `Tests/Evals/test_evaluation_metrics.py` imports from `tldw_chatbook.Evals.eval_runner`, and nothing in the tree imports `metrics_calculator` at all — it is fully orphaned. The exposure is via the documentation, not via a test.
    - **Not doing the consolidation here.** Picking one source of truth and deleting or re-exporting the other is the right eventual fix and is filed as **TASK-863**, but that module is 506 lines with roughly three methods absent from `eval_runner.py`, so it is a refactor with real blast radius and does not belong in a bug-fix PR.

    Instead this PR brings the two copies into agreement on the affected function only: adding the exact-match short-circuit and the `[0, 1]` clamp to the `metrics_calculator` copy, with tests covering it directly. Its zero-guard already reads `else 1.0`, which is *correct* (with zero vectors the dot product is 0, so `0 / 1.0 = 0.0`) and stays — `eval_runner`'s copy had been corrupted to `else 0.0`, which divided by zero, and that was one of the defects this PR fixes. Worth stating explicitly so nobody later "aligns" the good copy to the broken shape.

    For context on how this arose: the two implementations share 11 methods, of which an AST diff shows 9 are byte-identical and `calculate_perplexity` differs only cosmetically. Only `calculate_semantic_similarity` had genuinely drifted — which is how it lost its `return` unnoticed.

PR #964, opened 2026-07-27
feat(tools): agent path-hardening — denylist coverage, workspace-root search, killable grep

rmusser01:
Closes TASK-843, TASK-846, TASK-847, TASK-848, TASK-849, TASK-850.

Follow-on hardening to #953. The theme throughout: **a security check that names a path by literal protects nothing once the app resolves that path differently.** #953 shipped a denylist that hardcoded the permission store at `~/.config/tldw_cli/mcp_permissions.json` while the app builds it under `get_user_data_dir()` — the literal never existed, so the check never matched once, and `write_file` could overwrite the real store.

## TASK-846 — the audit that motivated most of this

Inventoried 30 security-relevant checks that name filesystem paths, config keys or database locations by literal, and verified each by resolving both sides on disk. **27 were broken.** Findings and evidence in `.superpowers/sdd/task-846-audit.md`; filed as tasks 851-859 and 864-866.

**Two Critical findings are not fixed here and deserve attention ahead of this PR** — together they mean config encryption does not encrypt:
- **TASK-851** — all three config-encryption entry points open `DEFAULT_CONFIG_PATH` rather than `_get_effective_config_path()`. Reproduced live under `TLDW_CONFIG_PATH`: "Enable encryption" returned `True`, left the active profile's `sk-proj-…` key plaintext, and silently rewrote a different file.
- **TASK-852** — `detect_api_keys` exact-matches six literals while every real key name is prefixed (`openai_api_key`, `bing_search_api_key`, …), so `should_encrypt_config()` never prompts.

Two of the audit's Criticals *are* fixed here, folded into TASK-848 rather than duplicated.

## What this branch changes

**TASK-847** — `is_sensitive_path` now fails closed on *any* resolution failure. A NUL byte previously raised `ValueError` and escaped a docstring promising fail-closed.

**TASK-848** — closes two Critical coverage gaps:
- The **skill trust and script-grant store** (`skill_script_grants.json` authorizes script execution) was unprotected and structurally unreachable, because `skills` is an exempted container directory. Now derived from `SkillTrustService`/`SkillTrustStore`'s own attributes, with a documented exception to the exemption — `skills/mypack/SKILL.md` stays reachable while `skills/trust/**` does not.
- **`config.toml`'s `.bak`/`.tmp` sidecars** are byte-identical copies including API keys; two existed unprotected on the audit machine. Now covered, mirroring the existing DB-sidecar treatment.

**TASK-849** — an agent could create a *directory* named after a not-yet-created state file (`search_history.db/`), breaking the app's later open. Refused before creation.

**TASK-850** — `glob_files`/`grep_files` now honour the same workspace folder roots as `read_file`, searching each and merging under **global** bounds so N roots cannot multiply the worst case by N. This aligns search with reading and deliberately does not widen reach — a path outside every configured root stays refused by all five tools.

**TASK-843** — `grep_files` now runs its regex in a **killable subprocess**. The previous bounds made the worst case finite but could not satisfy the criterion that a pathological regex stop consuming CPU after its call returns: Python's `re` has no timeout and `_call_with_timeout` *abandons* the worker thread rather than killing it. Chose a subprocess over the third-party `regex` module because `regex` is only a transitive dependency here via unrelated RAG extras, and `grep_files` is a core built-in with no extras gate. Before: `(a+)+$` ran 11.7-47s and kept burning CPU past the reported timeout. After: bounded to the subprocess ceiling with the child confirmed dead on return.

## Review findings, fixed

The whole-branch review found one Important regression introduced by 843+850 interacting — neither task's own review could see it. Splitting discovery from search drained candidate enumeration to `_MAX_CANDIDATES` before spawning the child, losing the early exit: a high-hit-rate pattern over 5,000 files went from ~0.1s to 1.29s, and it falsified a documented timeout-ordering invariant. Fixed by streaming discovery and search together in growing batches, each its own killable subprocess. **1.29s → 0.077s; candidates enumerated 5,000 → 256.** The new benchmark test was verified to fail against the pre-fix code.

Also fixed: the search subprocess no longer inherits the parent's environment, cwd or `sys.path[0]` (`-P`, explicit `cwd`, minimal `env`); the child's payload shape is validated before returning. Two Minors were documented rather than changed — no new subdirectory can be created inside a container dir under a widened root, and `_MAX_CANDIDATES` is consumed root-by-root so a huge first root can starve later ones.

## Testing

`Tests/Utils/ Tests/Tools/ Tests/Agents/` → **919 passed, 0 failed**, re-run after rebasing onto current dev.

Verified by execution under a scratch `HOME` with the sandbox root widened to `get_user_data_dir()`: every container directory stays listable/readable/globbable/greppable while `chroma.sqlite3`, `rag_profiles/*.json`, `config.toml.{bak,tmp}`, `search_history.db`, `mcp_permissions.json` and the whole `skills/trust/` subtree are refused; glob leaked zero protected paths. The default configuration keeps working end to end — that regression would have been worse than the bugs being fixed.

## Backlog note

Tasks originally filed as 860-862 moved to 864-866: dev filed into that range concurrently and landed first. Separately, **`task-839` and `task-840` are each duplicated on dev right now** from two other concurrent sessions — not introduced here, flagged for their owners.

---
## Summary by cubic
Hardens agent file-tool path checks and aligns search tools with workspace roots, while making `grep_files` killable to stop runaway regex CPU burn. Addresses TASK-843 and TASK-846–TASK-850.

- **Bug Fixes**
  - TASK-847: `is_sensitive_path` now fails closed on any resolution error (incl. NUL bytes).
  - TASK-848: Denylist now covers `skills/trust/**` (skill trust and script-grant store), config sidecars in the effective config dir, and direct-child files in user data, config, Chroma persist, and RAG profiles dirs; container dirs stay reachable.
  - TASK-849: Blocks creating parent dirs that would shadow state files (e.g., `search_history.db/`) via `refuses_new_directory_chain` before `mkdir`.
  - TASK-850: `glob_files`/`grep_files` search all allowed workspace roots with global caps; paths outside every root remain denied.
  - TASK-843: `grep_files` runs regex in a killable subprocess; hardened with `-P`, explicit `cwd`, minimal `env`; streamed batches restore early exit and improve high-hit patterns (~1.29s → ~0.077s); child is terminated on timeout.
  - Tests: Added focused tests; suite now 919 passed. Audit (TASK-846) filed follow-ups; two critical items tracked in TASK-851 and TASK-852.
  - Fix: Grep worker script path is now resolved to an absolute path and asserted to exist; ensures subprocess works with a hardened `cwd` and unrelated parent `cwd` (test added).

<sup>Written for commit e3e0a35076c61281c21ea1437bcd990da7eb8a6f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/964?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.superpowers/sdd/task-843-850-fixes.md
+++ b/.superpowers/sdd/task-843-850-fixes.md
@@ -1,0 +1,112 @@
+# TASK-843/TASK-850 follow-up hardening review — fixes
+
+**Worktree:** `/Users/macbook-dev/Documents/GitHub/wt-path-hardening` (branch `feat/agent-path-hardening`)
+**Interpreter:** `/Users/macbook-dev/Documents/GitHub/tldw_chatbook/.venv/bin/python`, run from this worktree.
+
+This addresses five findings from a follow-up review of `7fd033f1b` (TASK-850, scope `glob_files`/`grep_files` to workspace roots) and `8d66becf9` (TASK-843, killable-subprocess regex search) — both already `Done` on this branch. All work is in `tldw_chatbook/Tools/file_operation_tools.py`, `tldw_chatbook/Utils/sensitive_paths.py`, and `Tests/Tools/test_glob_grep_files.py`.
+
+---
+
+## Finding 1 (Important) — enumeration ran away before the subprocess ever started
+
+**Gap.** TASK-843 split `grep_files` into two sequential phases: candidate discovery (`_iter_candidates_across_roots`, in-process, bounded by `_MAX_CANDIDATES` = 20,000) then one subprocess call over the *entire* resulting list. The pre-subprocess implementation had checked `len(matches) >= _MAX_MATCHES` / `lines_scanned >= _MAX_GREP_LINES_SCANNED` *during* enumeration and broke out immediately; the split lost that early exit; `GrepFiles.execute` drained discovery to completion regardless of how quickly the match budget would have been satisfied.
+
+Consequences, both real:
+- A high-hit-rate pattern over a large tree paid for candidates the match budget never needed.
+- `_GREP_SUBPROCESS_TIMEOUT_SECONDS`'s comment claimed its ~2s headroom below `GrepFiles.timeout_seconds` (20.0s) meant the kill fires "at or before" the agent is told the call failed. With unbounded discovery time added on top of the subprocess's own 18s, that was false.
+
+**Fix.** Discovery and search are now streamed together by a new `_run_grep_search` (`tldw_chatbook/Tools/file_operation_tools.py`), called from `GrepFiles.execute` in place of the old "collect everything, then one subprocess call":
+
+- Candidates are pulled from `_iter_candidates_across_roots` in **growing batches** — `_GREP_INITIAL_CANDIDATE_BATCH_SIZE = 256`, doubling up to `_GREP_MAX_CANDIDATE_BATCH_SIZE = 4096` — each batch searched by its **own** `_run_grep_subprocess` call with whatever `max_matches`/`max_lines_scanned` budget remains.
+- The loop stops **without pulling another candidate** the moment the remaining budget is `0` (i.e. `_MAX_MATCHES` matches found or `_MAX_GREP_LINES_SCANNED` lines scanned) — this is the restored early exit.
+- A small initial batch keeps a high-hit-rate search's pre-first-call enumeration cost close to what the old in-process early-break paid; growing the batch on later rounds keeps a rare/zero-hit search (which must still examine up to `_MAX_CANDIDATES`) from paying for dozens of small, separately-spawned subprocesses to get there (doubling from 256 reaches 20,000 in ~8 rounds).
+- **Killability is unaffected** — every batch is still searched by its own fully killable child process; only how many candidates reach it, and in how many calls, changed.
+- Every existing guard (containment, sensitive-path denylist, hidden-component rule, the global `_MAX_CANDIDATES` counter shared across all roots) still runs inside `_iter_candidates_across_roots` exactly as before — `_run_grep_search` performs none of that itself and trusts every candidate completely, same contract as before.
+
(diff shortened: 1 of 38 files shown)


rmusser01:
## Qodo findings addressed — `e3e0a3507`

**3 (relative grep worker path) — accepted, and it was a real bug this PR itself created.** `_GREP_WORKER_SCRIPT` was built with `Path(__file__).with_name(...)` and no `.resolve()`. On its own that is a latent portability wart, but this PR's own subprocess hardening spawns the worker with `cwd=_GREP_WORKER_CWD` (the temp dir) instead of inheriting the parent's — so a relative `__file__` would resolve against the temp dir and fail to open, breaking `grep_files` outright. Good catch on the interaction; the cwd change is exactly what promoted it from untidy to fatal.

Fixed with `.resolve()`, plus two tests: an absolute-and-exists assertion, and one that greps successfully with the interpreter's cwd somewhere unrelated. Stated honestly in the test docstring that the first cannot fail today under this loader — it guards a future refactor rather than reproducing the bug now.

**1 (`_iter_candidates_across_roots` skips `validate_path`) — declined; the enforcement is present, just not via that helper.** Every candidate the iterator yields passes, in order:

- `_sandbox_root_is_hidden` per root — mirrors `validate_path`'s dotted-base rule
- `is_within(path, root, context=...)` — containment, and it also applies the credential/app-state denylist through `is_sensitive_path`
- `_is_hidden_within(resolved, root_resolved)` — the dot-component rule `validate_path` enforces

Routing candidates through `validate_path` in addition would add no enforcement these three do not already provide, and would actively hurt: it **raises `ValueError`**, while these tools must return an error dict and never raise, and it costs measurably more per candidate (re-measured: inline +20.0% vs `validate_path`-per-candidate +42.8% over a 1,500-file tree). The rule lives in the tools rather than the shared helper deliberately — `ListDirectoryTool` has an `include_hidden` opt-in that a shared-helper change would break.

Root *resolution* does go through the central validator: TASK-850 uses `validate_path_multi`'s own first-match-wins per-root semantics, so search reaches exactly what `read_file` reaches and no more.

Worth noting the same finding on #936 was **correct** and caught a genuine hole — `grep_files` was reading `.env` files `read_file` refused. That gap is closed and regression-tested; this is the same finding against code that now enforces the rule.

Local verification: `Tests/Utils/ Tests/Tools/ Tests/Agents/` → **921 passed, 0 failed**.

PR #978, opened 2026-07-27
fix(settings): DB maintenance backed up and restored paths that were not the real databases

rmusser01:
The Settings database-maintenance panel operated on paths that were not the real databases.

`Tools_Settings_Window._get_database_path()` built its map from hardcoded defaults that were wrong in
two independent ways: **wrong filenames** (`tldw_evals_db.db`, `tldw_prompts_db.db`,
`tldw_media_db.db` — the real files are `evals.db`, `tldw_chatbook_prompts.db`,
`tldw_chatbook_media_v2.db`) and **no profile directory** (every database lives under
`~/.local/share/tldw_cli/<profile>/`). `config.py` already exposed correct, profile-aware resolvers
the panel simply never called.

**Why this was data-safety, not cosmetics.** That path fed four workers. Vacuum, backup and
integrity-check all guard on `db_path.exists()`, so with a wrong path they silently did **nothing**.
Restore was worse: it took a pre-restore backup only `if db_path.exists()` — false, so none was taken
— then ran `shutil.copy2(backup_path, db_path)` **unconditionally** and reported
`"restored successfully"`. So restoring wrote the backup to a phantom location, left the real
database untouched, and told the user it had worked.

**The fix:** every database now resolves through `config.py`'s own resolvers. Two were missing and
were added — `get_evals_db_path()` and `get_rag_indexing_db_path()` — and, importantly, their
existing consumers now call them too: `eval_orchestrator._initialize_database` and
`ingestion_indexing._default_indexing_db`. That is deliberate. If the panel resolved Evals
independently of the orchestrator, the two would drift and recreate this exact bug one layer over.
One definition each, shared.

`rag` was kept rather than dropped: `rag_indexing.db` is real, confirmed on disk under existing
profiles, and matches what `_default_indexing_db` actually opens. It is the RAG indexing-state
database; there is no separate "main" RAG database.

A path that cannot be resolved now fails loudly instead of reporting success.

**A second, independent bug found and fixed along the way.** All four workers called
`self.call_from_thread(...)`, but that method exists only on `App` — not on `Widget`/`Container`,
which is what `ToolsSettingsWindow` is. Every notification in those worker threads would have raised
`AttributeError`, including the new fail-loudly ones this PR adds. All 39 call sites in the file now
use `self.app.call_from_thread`, and a guard test asserts no bare form comes back — this is the second
time a worker-thread notification in this file was silently unreachable.

**Tests:** 14 new, including a parametrized backup→restore round trip across all six databases that
proves content returns **at the path the application actually uses**, plus unresolvable/missing-database
failure modes. All 14 verified to fail against the pre-fix code. The 6 remaining failures in this file
are pre-existing `chat_api_key` encryption tests, identical on `dev` — confirmed by running both.

Closes TASK-899.

**Known, deliberately out of scope** (to be filed): the "all databases" vacuum/backup/check workers
and the conversation/character export workers carry their own copies of the same hardcoded-path bug;
and `ChatbookImporter` expects capitalized keys (`"ChaChaNotes"`) while `_import_chatbook` passes
lowercase.

---
## Summary by cubic
Fixes the Settings database-maintenance panel to use real, profile-aware database files, validate backup/restore paths, and report failures clearly. Adds a single source of truth for DB paths and updates Evals/RAG to use it, addressing TASK-899.

- **Bug Fixes**
  - Resolve all DB paths via `config.py` resolvers; added `get_evals_db_path()` and `get_rag_indexing_db_path()`, and updated Evals orchestrator and RAG ingestion to delegate to them.
  - `VACUUM`, backup, integrity check, and restore now error when a path can’t be resolved; they warn on missing files; restore creates the target directory if it doesn’t exist and validates both source/target paths via `validate_path_simple`.
  - Replaced all bare `self.call_from_thread(...)` with `self.app.call_from_thread(...)`, including `_import_chatbook_worker`, so notifications reach the UI; added a guard test to prevent regressions.
  - Unified path handling under `_DB_PATH_RESOLVERS` (no hardcoded literals) and updated the size UI and `_import_chatbook` to use it; kept the RAG indexing DB entry (`rag_indexing.db`).
  - Tests added for resolver agreement, backup→restore round-trip, failure modes, rejecting dangerous backup paths, and restoring to a new custom directory.

<sup>Written for commit 3b0c8def5a00ea9c71edea72cc27d67d18641bc1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/978?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Addressing all four Qodo findings. One of them is a bug this PR introduced, and it came from my own review instruction — worth stating plainly.

**4. Restore blocks custom paths — correct, and my fault.** I asked the implementer to "verify the target resolves and its parent directory exists before writing". The parent-exists half is too strict: it blocks a legitimate restore to a configured custom path whose directory has not been created yet, and it is inconsistent with `DB/base_db.py`, which creates parents when opening a database.

    The distinction I actually wanted is sharper, and the fix makes it explicit: **fail loudly when the path cannot be resolved** (unknown `db_name`, resolver returns `None`) — that is the phantom-path case this PR exists to prevent and it stays a hard error — but **do not fail merely because the directory does not exist yet**. That now attempts `mkdir(parents=True, exist_ok=True)` and only errors if the mkdir itself fails.

**1. Unvalidated paths in backup/restore — fair.** `backup_path` comes from the user's file picker and `db_path` from config; both go straight into `shutil.copy2`. Routing them through `path_validation.py`, with two cautions carried into the fix: `validate_path_simple` rejects a literal `~/` so expanduser must come first, and a rejection must produce a clear, actionable error naming the refused path rather than a silent no-op — this is a restore flow, and a validator that quietly blocks it would be worse than the bug.

**2. Missing `Returns:` on the new resolvers — fixing.** `get_evals_db_path()` and `get_rag_indexing_db_path()` both return a `Path`.

**3. sqlite connections without context managers in the new tests — fixing,** across every connection the new tests open, not only the flagged one.

For context on the 6 failures in `Tests/UI/test_tools_settings_window.py`: they are pre-existing `chat_api_key` encryption tests, unrelated to database paths. I confirmed by running the same file on `dev` — identical failure set, same six names. This branch takes that file from 9 passing to 23.

PR #985, opened 2026-07-27
fix(config): config encryption actually encrypts — effective path, real key detection, trust containment

rmusser01:
Closes TASK-851, TASK-852, TASK-853. Follow-on to the path-naming audit in #964.

**Together, 851 and 852 meant config encryption did not encrypt.**

## The three Criticals

**TASK-851 — encryption wrote to the wrong file.** All three entry points in `config.py` opened `DEFAULT_CONFIG_PATH` rather than `_get_effective_config_path()`. Reproduced: with `TLDW_CONFIG_PATH` set, "Enable encryption" returned `True`, left the active profile's `sk-proj-…` key in **plaintext**, and silently rewrote a different file. Now routed through the effective path, and the write is atomic — a crash mid-write previously truncated the file holding every API key.

**TASK-852 — the detector never fired.** `detect_api_keys` exact-matched six literals while every real key name is prefixed (`openai_api_key`, `bing_search_api_key`, `api_token`, `*_fallback`), so a config full of live plaintext secrets returned `False` and `should_encrypt_config()` never prompted. Replaced by a unified predicate in `Utils/sensitive_config_keys.py`, collapsing four previously-disagreeing lists. Verified across 777 real config key names: zero coverage lost against either removed list; the only newly-flagged keys are two genuine `*_API_KEY_fallback` secrets. The `_env_var` guard holds and `max_tokens` stays excluded.

**TASK-853 — two containment checks that could not reject anything.** One passed the candidate's own parent as the containment base (a path is always inside its own parent); the other tested nesting in only one direction, so a scratch root *enclosing* the trust store passed. These guard the trust and script-grant stores that authorize skill script execution. The new base derives from `default_trust_store_dir` — the same value passed to `SkillTrustStore(store_dir=...)` — not a re-spelled literal.

## What the review caught: a regression this branch activated

TASK-851 fixed three encryption write paths. There was a **fourth**, and it was the only automatic one — the on-quit save in `app.py`.

Before the fix, enabling encryption under a profile wrote `[encryption] enabled=true` into the *default* file, which the profile-reading loader never saw, so the quit handler stayed a no-op. After the fix the profile carried that flag, so **every quit serialized the merged profile config over the user's default config**: measured 122 → 31,007 bytes with the default's own key destroyed, and the profile never re-encrypted. Fixed, with a regression test that fails without it.

Also fixed from the review:
- The atomic write was chmod'ing to `0644` — widening a file full of API keys, and leaving it `0644` with plaintext after a later decrypt. Now preserves the target's existing mode (`0600` verified across both encrypt and decrypt), with a restrictive default for new files. Atomicity itself was sound and is unchanged: mkstemp sibling in the target directory, `0600` during the write window, fsync, `os.replace`.
- `tomllib.load` + `toml.dumps` drops comments and reorders tables. Values and types are fully lossless — ints, bools, floats, arrays, datetimes, non-ASCII and multiline strings all round-trip with zero semantic diff — so this is documentation loss, not data loss. It predates this branch, but 851 points it at profile files and 852 makes the prompt actually appear, so users will now hit it. Surfaced as a warning in the enable-encryption dialog rather than silently rewriting annotated configs.
- `skill_trust_store.clear()` now validates before unlinking, so the invariant holds on all three methods.
- `settings_screen._config_path()` now calls `_get_effective_config_path()` instead of re-spelling it — the exact drift shape this audit is about.

## Testing

`Tests/Skills/ Tests/Utils/` plus both config-encryption test files → **873 passed, 0 failed**, run after rebasing onto current dev.

Independently verified end to end: the plaintext sentinel is removed from the effective file while the real user config stays byte-identical; encrypt→decrypt round-trips; file mode holds at `0600` through both; the default config is untouched across the quit path. Every new regression test was checked to fail pre-fix.

## Notes

- The task filed for the remaining sibling defect (`Tools_Settings_Window._save_raw_toml_config`, same wrong-path and non-atomic shape) is **task-930** — `900` was taken on dev by a concurrent session while this branch was in flight.
- `task-545` is duplicated on dev from a concurrent session. Not introduced here; flagged for its owner.

---
## Summary by cubic
Make config encryption actually encrypt and keep it safe. Writes now target the active config file, detect real keys, preserve permissions, and harden trust/scratch containment (TASK-851, TASK-852, TASK-853).

- **Bug Fixes**
  - Route `enable_config_encryption`, `disable_config_encryption`, `change_encryption_password`, and the on-quit save through `_get_effective_config_path()`; writes are atomic and preserve the target’s mode (new files default to `0600`).
  - Unify sensitive-key detection in `Utils/sensitive_config_keys.py` and use it across config, `config_encryption.detect_api_keys()`, and the privacy/security UI; it flags real provider keys (e.g., `*_api_key`, `*_token`, `*_fallback`) and ignores non-secrets like `*_env_var` and `max_tokens`.
  - Show a warning that enabling encryption rewrites TOML formatting and drops comments in `Widgets/password_dialog.py` and the settings toggle in `UI/Tools_Settings_Window.py`.
  - Fix trust containment: `FileSkillTrustGenerationMarkerStore` validates `marker_path` against a provided `store_dir`, and `LocalSkillsService` treats both `self.store_dir` and `self.skills_dir` as containers with checks for both “nested inside” and “enclosing” cases.
  - Delegate `SettingsScreen._config_path()` to the shared `get_cli_config_path()` accessor and validate `skill_trust_store.clear()` before unlinking.
  - Update QA seed script `Docs/superpowers/qa/skills-script-execution-2026-07-25/seed3.py` to pass `store_dir=...` to `build_skill_trust_marker_store_with_fallback`.

- **Migration**
  - Update any direct construction of `FileSkillTrustGenerationMarkerStore` to pass `store_dir=...` (the trust store directory).

<sup>Written for commit 58f8f1b54f002f6788fc18a878442f3657cdbf9f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/985?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.superpowers/sdd/task-851-852-report.md
+++ b/.superpowers/sdd/task-851-852-report.md
@@ -1,0 +1,302 @@
+# TASK-851 / TASK-852: config encryption effective-path + detection fix
+
+Branch: `fix/config-encryption-effective-path` (worktree `wt-cfg-crypto`, cut from `origin/dev`).
+
+Together these two defects meant config encryption did not encrypt: the
+encryption entry points wrote to the wrong file when a profile was active
+(851), and the detector that decides whether to offer encryption never
+recognized any of this app's real API-key field names (852).
+
+## TASK-851 — wrong file
+
+### Root cause
+
+`enable_config_encryption`, `disable_config_encryption`, and
+`change_encryption_password` in `tldw_chatbook/config.py` all read/wrote
+`DEFAULT_CONFIG_PATH` (`Path.home()/.config/tldw_cli/config.toml`) directly,
+instead of `_get_effective_config_path()`, which is the only accessor that
+honors the `TLDW_CONFIG_PATH` override (the app's "Override config" /
+profile mechanism). `enable_config_encryption` also wrote with a plain
+`open(path, "w")` + `toml.dump(...)`, unlike its two siblings which already
+used `atomic_write_text`.
+
+### Reproduction — before fix
+
+Script (paraphrased; full version was run via `git stash` against the
+worktree to get a clean unfixed baseline):
+
+```python
+import tldw_chatbook.config as cfg
+# ... HOME redirected to a scratch dir; a prior "normal run" already created
+# DEFAULT_CONFIG_PATH via load_cli_config_and_ensure_existence() ...
+
+profile_path.write_text(
+    '[api_settings.openai]\napi_key = "sk-proj-PLAINTEXT-SENTINEL-abc123"\n'
+)
+os.environ["TLDW_CONFIG_PATH"] = str(profile_path)
+
+result = cfg.enable_config_encryption("hunter2hunter2")
+```
+
+Output:
+
+```
+effective path (should be the profile): .../cfg851profile_7ky1asp8/profile_config.toml
+enable_config_encryption(...) -> True
+---- ACTIVE profile file contents after enable ----
+[api_settings.openai]
+api_key = "sk-proj-PLAINTEXT-SENTINEL-abc123"
+
+plaintext key STILL present in ACTIVE file: True
+DEFAULT_CONFIG_PATH (wrong file) shows an 'encryption' section now: True
+
+BUG REPRODUCED: reported success, but the ACTIVE file's secret is untouched
+plaintext, and a completely different file was rewritten instead.
+```
+
+### After fix
+
+Same script, same inputs, against the fixed code:
+
+```
+enable_config_encryption(...) -> True
+---- ACTIVE profile file contents after enable ----
+[encryption]
+enabled = true
+method = "AES-256-GCM-scrypt"
+version = 1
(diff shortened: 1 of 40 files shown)


rmusser01:
## Qodo finding addressed — `eea334e4f`

**Seed script missing `store_dir` — accepted, correct.** TASK-853 made `store_dir` a required keyword-only argument on `build_skill_trust_marker_store_with_fallback` (that argument is the whole fix: the old containment base was the candidate's own parent, which made the check tautological). `Docs/superpowers/qa/skills-script-execution-2026-07-25/seed3.py` still called it with only `fallback_marker_path` and `account_scope`, so it would raise `TypeError`.

Verified rather than assumed: the function's required keyword-only parameters are now exactly `['fallback_marker_path', 'store_dir']`, and the script's sole call site supplied one of them. Fixed by passing `store_dir=trust_dir` — the same directory the script already hands `SkillTrustStore(store_dir=...)` two lines later, so the containment base matches what the store actually uses.

It is a QA script under `Docs/` rather than shipped code, but a broken caller is a broken caller, and this one exists precisely to reproduce the skills trust flow by hand.

Local verification unchanged: `Tests/Skills/ Tests/Utils/` plus both config-encryption test files → 873 passed, 0 failed.

PR #988, opened 2026-07-27
fix(settings): clear the remaining hardcoded DB paths in bulk workers and the config form

rmusser01:
Finishes what #978 started. That PR fixed the **single-database** maintenance workers to resolve
through `config.py`'s profile-aware resolvers; this one clears every remaining hardcoded path in the
file — 15 literals across eight functions.

**The bulk workers** (`_vacuum_worker`, `_backup_worker`, `_integrity_worker`) never went through
`_get_database_path()`, so they carried their own copies of the wrong filenames and the missing
profile segment. This is the more dangerous half in practice: "back up all databases" is exactly what
a cautious user clicks before an upgrade, and it silently produced nothing. They now resolve through
the canonical resolvers and **fail loudly per database** rather than skipping and reporting overall
success. The three export workers got the same treatment.

**The settings form was the real find, and it could have poisoned #978's fix.**
`_compose_database_config_form` built its inputs as `db_config.get(key, "<wrong literal>")`, and the
save handler persists those inputs back into config. So the form displayed a wrong path as though it
were the current value, and saving wrote it into config — after which `get_cli_setting()` returns it
and *even the correct resolver* yields the wrong path. It now shows the actually-resolved value.

**Reset now resets again.** Fixing the form initially made "Reset Section" a no-op for anyone who had
customised a path — the one case where you press it — because it re-showed the override-aware value.
Rather than inline a second copy of the filenames in the UI (the exact duplication this effort has
been removing), the three resolvers gained a keyword-only `ignore_override: bool = False` that returns
the profile-aware default already computed in their own `else` branch. Compose shows what the app
will use; Reset restores the default. Verified with a real override configured:

```
compose (effective) -> /tmp/custom/my_prompts.db
reset   (default)   -> ~/.local/share/tldw_cli/<profile>/tldw_chatbook_prompts.db
```

The parameter is keyword-only and defaults to the old behaviour, so no existing caller can break.

**Verification:** `grep -nE '~/\.local/share/tldw_cli/[^"]+\.db'` over the file returns nothing.
36 passing (was 25 before #978, 9 before this series). The 6 remaining failures in that file are
pre-existing `chat_api_key` encryption tests, identical on `dev`.

Every new behaviour was checked by reverting the fix and confirming the test fails. That caught two
tests passing for the wrong reason: one where removing a guard let `AttributeError` on `None` be
swallowed by a generic handler and reported as an error anyway, and one where a source-scan guard
matched a leftover *comment* containing the string it was looking for.

Closes TASK-927.

---
## Summary by cubic
Fixes the remaining hardcoded database paths in bulk maintenance/export workers and the Settings form. Everything now resolves through profile-aware getters, bulk actions fail loudly, backups write to the profile-scoped directory, and Reset/Save behave correctly. Addresses TASK-927.

- **Bug Fixes**
  - Bulk workers (`_vacuum_worker`, `_backup_worker`, `_integrity_worker`) and export workers now resolve via `_DB_PATH_RESOLVERS`; unresolvable DBs are reported per-DB as errors. Bulk backup refuses partial sets and uses `get_user_data_dir()/backups` instead of the flat literal.
  - Settings form shows the resolved paths via `_resolved_db_path_display()`. Reset passes `ignore_override=True` (new keyword-only arg on `get_chachanotes_db_path`, `get_prompts_db_path`, `get_media_db_path` in `tldw_chatbook/config.py`) to restore profile defaults.
  - Save only persists genuine custom paths; matching the default clears any stored override via `delete_settings_from_cli_config` (no more pinning the current profile’s resolved path).
  - Restore to a never-before-created custom path copies directly into a fresh private target via `copy_private_sqlite`, creating the parent directory when needed.

<sup>Written for commit 2ba054dd485e528c0aa63be987b29e63b7cea4aa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/988?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Qodo's finding is correct and it is a real bug this branch introduced. Verified the full chain:

```
resolved path shown in form : /Users/…/.local/share/tldw_cli/default_user/tldw_chatbook_prompts.db
TOML default literal        : ~/.local/share/tldw_cli/tldw_cli_prompts.db
differ -> treated as custom override: True
```

`_save_database_config_form` persists the Input value verbatim, and `config.get_*_db_path()` treats any stored value differing from `DEFAULT_CONFIG_FROM_TOML`'s literal as an explicit override. So opening Settings and pressing **Save** — without editing anything, or straight after Reset — permanently pins every database path to the current profile and data dir.

That is the same profile-isolation failure TASK-860 fixed, reintroduced from the opposite direction, and it is worse than the display bug it came from because it silently writes durable config that survives a profile switch.

Fixing per the suggested approach: on save, compare each field against the profile-aware default (`get_*_db_path(ignore_override=True)`, normalised so a `~`-form or non-normalised equivalent still counts as unchanged) and **clear the key** via `delete_settings_from_cli_config` rather than writing it back. Only a genuine custom override gets persisted. `_compose_database_config_form` keeps showing the effective path — that part was right.

The test that matters here is not "no key was written" — that would pass against a fix that wrote something harmless-looking. It is the consequence: after a no-edit save, switching profiles must still move the resolved path. That is being pinned explicitly, along with the reset-then-save and genuine-custom-path cases.

Worth noting for the record: this is the second bug in this PR chain caused by a fix rather than found in existing code (the first was a parent-directory guard of mine that blocked legitimate restores in #978). Both were caught by review rather than by tests, which is a fair signal about where the risk sits in this file.

rmusser01:
**Correction to my earlier comment, and a bigger finding.**

I said the evidence pointed toward backup writing no file at all. **That was wrong.** Instrumenting the actual code path found `chachanotes_backup_*.db` written exactly where the code intends — `.../tldw_cli/default_user/backups/chachanotes/`, profile-scoped via `get_user_data_dir()`. My earlier search missed it because I looked for a flat path without the profile segment. The round-trip test was asserting that same stale flat path, so it was the *test* that was wrong, not the backup. Fixed the test, and normalised the legacy bulk `_backup_worker`'s flat root to the same profile-scoped location so the two backup roots finally agree.

**But a genuine regression did turn up while investigating.** `_restore_single_worker` always called `restore_private_sqlite`, which requires the destination to already exist ("Live SQLite destination must exist before restore"). That breaks the first restore to a never-used custom path — the exact case #978 added a test for. Fixed by branching to `copy_private_sqlite` when the destination does not exist yet, mirroring what `DB/base_db.py` already does with parent creation. This added a row to the SQLite private-owner inventory (B16) and its cross-checking tests.

**`dev` was red before this PR.** `Tests/UI/test_tools_settings_window.py` had **14 failures on dev itself** — 6 long-standing `chat_api_key` ones plus 8 that were green when #978 merged and were broken by concurrent work. This branch takes that file from **14 failures to 6**, and the 6 that remain are the pre-existing `chat_api_key` set, unrelated to database paths.

Also resolved from the reconciliation: dev's `SETTINGS_DATABASES`-driven vacuum/integrity/export workers silently skipped an unresolvable database. That is precisely what TASK-927 exists to remove, so I kept dev's (cleaner) structure and added the loud per-database failure back on top rather than reverting to our implementation.

`get_chatbook_database_paths()` was checked and is not a second source of truth — it calls the same canonical resolvers, and is the established convention across the other chatbook surfaces. The test was broadened to accept it, plus an assertion that the helper itself embeds no literal defaults.

Gates: the path-literal grep returns nothing, bare `self.call_from_thread` is zero, and the four save-behaviour tests pass including the profile-switch consequence. Adjacent suites verified: private-sqlite inventory 21, chatbook paths 7, config bootstrap 22, Evals 477.

PR #1009, opened 2026-07-27
fix(ui): a cached httpx.AsyncClient was shared across event loops

rmusser01:
Closes TASK-981. The audit found a **real cross-loop bug**, and the scope was three times what the
task described.

**The sweep found 6 `@work(thread=True)` on `async def`, not the 2 the task named** — including
`load_node_children`, whose `await call_from_thread` bug was fixed in #1001.

**The bug: a cached `httpx.AsyncClient` shared across event loops.**
`CodeRepoCopyPasteWindow` holds one `GitHubAPIClient` for the widget's lifetime, and that class lazily
**caches** an `httpx.AsyncClient` on the instance. It is awaited from app-loop handlers *and* from two
thread workers, each running on its own loop via `asyncio.run()` — which then **closes** that loop on
completion. `httpx.AsyncClient` binds its connection pool to the creating loop, so the cache hands
back a client bound to a dead loop.

Demonstrated from both sides — same probe, two loops:
```
dev:    client id loop1 == loop2   (the closed-loop client is reused)  <- the bug
branch: client id loop1 != loop2   (a live client per loop)
```

**Fix:** scope the cache to the running loop inside `GitHubAPIClient` itself, rather than patching the
two call sites. It is the root-cause fix, benefits every caller, and self-heals in both directions.
A stale client is handed off via `run_coroutine_threadsafe` when its loop is still alive and dropped
when it is not, so nothing leaks. It only invalidates on a *proven* loop mismatch, so a test-injected
`_client` is left alone.

**Per-worker verdicts:**
- `_export_to_zip_worker`, `load_node_children` — real awaits into the shared client; stay async, fixed at the client layer.
- `ccp_message_manager.load_conversation_messages` — **zero awaits**; `async def` bought nothing and cost a whole event loop per call. Now a plain `def`.
- `swarmui_widget.generate_image` — zero awaits **and actively broken**: it called `loop.run_until_complete()` inside the loop `asyncio.run()` had already started. Reproduced: `RuntimeError: Cannot run the event loop while another loop is running`. Now a plain `def`, matching the working pattern used twice in the same file.
- `multi_item_review_window._generate_analyses_worker` — genuine awaits, all self-contained, no loop-bound sharing. Left async, with a comment recording why.
- `textual_scheduler_worker.start_scheduler` — not a cross-loop hazard by design (one long-lived loop, no cached primitives). Left alone.

Every fix was revert-checked against the **specific** expected error, not merely "some error": reverting
the client change gives `AttributeError: 'GitHubAPIClient' object has no attribute '_client_loop'` on all
four new tests; reverting the `ccp_message_manager` conversion gives `expected call not found` plus
`RuntimeWarning: coroutine ... was never awaited`. 114 tests pass.

**Also filed (TASK-1010):** three components turned out to be broken in ways that prove nothing
exercises them — the nested-loop bug above in a widget that is never mounted,
`multi_item_review_window` referencing a nonexistent `app.llm_api_client`, and
`SubscriptionSchedulerWorker` failing Textual's own `isinstance(self, DOMNode)` assert on first call.
Each needs a live-or-dead decision.

---
## Summary by cubic
Fix a cross-event-loop bug by giving `GitHubAPIClient` a per-loop `httpx.AsyncClient` and by closing leaked event loops in SwarmUI workers. This addresses TASK-981 and documents follow-ups in TASK-1010.

- **Bug Fixes**
  - Use a per-loop cache (`WeakKeyDictionary`) for `GitHubAPIClient.client`; prune closed loops; never close another live loop’s client.
  - Make `GitHubAPIClient.close()` close the current loop’s client and schedule others via `asyncio.run_coroutine_threadsafe`, logging failures from the returned Future.
  - Close event loops created by `SwarmUIWidget.check_server_status`, `load_models`, and `generate_image`; `generate_image` is now a plain `def` to avoid nested-loop errors under `@work(thread=True)`.
  - Keep `CodeRepoCopyPasteWindow` workers async; they now get a safe, per-loop client. Convert `CCPMessageManager.load_conversation_messages` to plain `def`.
  - Note: `textual_scheduler_worker.start_scheduler` isn’t a cross-loop hazard; recorded its separate DOMNode assert issue and left it as-is.

- **Tests**
  - Add cross-loop regression tests for `GitHubAPIClient`, including genuine concurrent app/worker loops and Future callback logging on scheduled closes.
  - Add `SwarmUIWidget` loop-cleanup tests to ensure each worker closes its `asyncio` loop.
  - Update CCP handler test to call the unwrapped sync function.

<sup>Written for commit 20272fba3ac471a1d82662bdf278e20df012aa4c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1009?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
All four addressed. Two are real bugs, and **one is a flaw in the fix this PR itself introduced** — worth saying plainly.

**3. Cross-loop client close race — correct, and my verification missed it.** The single-slot cache discards the cached client on *any* loop mismatch and schedules `aclose()` on the previous loop. That is fine when loops run one after another, and wrong when two are alive at once: a thread worker touching `client` while an app-loop request is in flight will close the client that request is still using. Two overlapping thread workers do the same to each other.

    I verified the original cross-loop bug with *sequential* `asyncio.run()` calls, where the first loop was already closed by the time the second ran. That is precisely why this case slipped through — the probe never had two live loops. Moving to a per-loop cache (weak-keyed, with pruning of closed loops) so no loop ever closes another's, and `close()` retains the `run_coroutine_threadsafe` future with a done-callback so a failed close is not silent.

    The replacement test has to hold two loops open simultaneously and assert the app-loop client still works *after* a worker-loop client has come and gone. A sequential test would re-make my original mistake.

**4. Unclosed event loop in `generate_image` — correct, and a consequence of this PR.** The method creates a loop with `asyncio.new_event_loop()` and never closes it. That was academic while the method could not run at all (it nested `run_until_complete` inside the loop `asyncio.run()` had already started, raising `RuntimeError: Cannot run the event loop while another loop is running`). Making it a working plain `def` turns the leak into a real per-generation cost. Closing it deterministically, and checking the two sibling workers in that file that use the same manual-loop pattern.

This is the second time in this chain that fixing an outer bug exposed the next one behind it — the first being `await self.app.call_from_thread(...)` in #1001, which only became reachable once the `AttributeError` stopped firing. Worth noting as a pattern: in this area, a fix that makes dead code live should be assumed to have uncovered something.

**1 and 2** — adding the missing `Returns:` while preserving the prose that records *why* the loop-scoping exists, and regrouping the test imports into stdlib / third-party / local.

PR #1019, opened 2026-07-27
refactor: delete two unrunnable worker components; file the ADR-019 rollback gap

rmusser01:
Resolves TASK-1010 in part, and surfaces a larger finding filed as TASK-1022.

**Two components deleted — neither could run, and nothing constructed them.**

- `SwarmUIWidget` was only re-exported from `Widgets/Media_Creation/__init__.py`, never constructed.
  Its `generate_image` called `run_until_complete()` inside the loop `@work(thread=True)` on an
  `async def` had already started, raising `RuntimeError: Cannot run the event loop while another
  loop is running` on every invocation.
- `MultiItemReviewWindow` had **zero** references outside its own file — not even a re-export — and
  its worker called `app.llm_api_client`, an attribute that exists nowhere.

Both removed with their tests and the now-empty `Widgets/Media_Creation` package. Verified nothing
still imports them and the app still imports. (The surviving `Media_Creation` grep hits are the
unrelated **top-level** `tldw_chatbook/Media_Creation/` package.)

**The third was deliberately not deleted, because it turned out not to be a cleanup question.**

`SubscriptionSchedulerWorker` is constructed only by `SubscriptionBackendController`, which is itself
never instantiated, and it fails `@work`'s own `isinstance(self, DOMNode)` assert on first call. On its
own that is dead code. But ADR-019 is **Accepted** and promises "instant rollback to the old scheduler
without a code deploy" — so the question was whether that guarantee still holds.

It does not, and the reason is broader than the worker. `app.py:4860-4871` reads
`scheduling.watchlist_checks_enabled` and, when true, builds a `WatchlistCheckHandler`; when false it
leaves `watchlist_handler = None` and starts nothing. There is no `else`. And `SubscriptionScheduler`
has no construction path from the app at all — `create_scheduler()` has zero callers, and the only
other site is inside the dead worker. `inspect.getsource(TldwCli)` contains none of the three names.

The flag defaults to **false**, so on a default install no *scheduled* watchlist execution runs —
neither new handler nor old scheduler — and toggling it off does not roll back, it silently disables
scheduling with no error.

Manual "Check now" is unaffected (`watchlists_collections_screen.py:2246` calls the controller
directly, touching no scheduler), and the old scheduler class still works when constructed directly.
It is unreachable, not broken.

Filed as **TASK-1022** rather than patched: whether to restore the rollback path or retire it and amend
the ADR is the owner's decision.

**TASK-1010 is left `In Progress` with two of four criteria unticked** — two of three components are
gone, but the third's disposition is not this task's to make. An honest partial rather than a forced
close.

---
## Summary by cubic
Removed two dead worker widgets and their tests; documented the ADR-019 rollback gap that leaves scheduled watchlist checks off by default (`TASK-1022`). Partially addresses `TASK-1010`; `SubscriptionSchedulerWorker` awaits an owner decision.

- **Refactors**
  - Deleted `tldw_chatbook/Widgets/Media_Creation/swarmui_widget.py` (`SwarmUIWidget`), `tldw_chatbook/Widgets/multi_item_review_window.py`, the `tldw_chatbook/Widgets/Media_Creation/` package, and tests (`Tests/UI/test_multi_item_review_window.py`, `Tests/Widgets/test_swarmui_widget_loop_cleanup.py`); cleaned up `Tests/UI/test_bulk_selection_tooltips.py`. Rationale: event-loop misuse in `SwarmUIWidget` and no live references for `MultiItemReviewWindow`. Verified no remaining imports and the app still imports.
  - Annotated `Docs/superpowers/specs/2026-07-22-image-generation-multiprovider-foundation-design.md` to reflect the widget deletion and clarify the top-level `tldw_chatbook/Media_Creation/` code (`swarmui_client.py`, `image_generation_service.py`, `generation_templates.py`) stays.

- **Migration**
  - Logged `TASK-1022`: ADR-019’s rollback path was severed when `SubscriptionWindow` was removed; with the flag defaulting to false, no scheduled watchlist checks run by default. Awaiting a call on wiring the old scheduler or updating ADR-019; `SubscriptionSchedulerWorker` is left untouched.

<sup>Written for commit f8cd2fa12a65da2181a6121f02f4d4dfec42b0be. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1019?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Good catch, and I checked whether it meant the deletion was wrong before annotating. It does not — the spec actually anticipated this removal.

The bullet the finding points at protects `swarmui_client.py` / `image_generation_service.py` under the **top-level** `tldw_chatbook/Media_Creation/` package, on the grounds that deleting them would break live imports. All three of those files (including `generation_templates.py`) are untouched and still live. The very same bullet ends: *"A separate follow-up task will remove the dead SwarmUI chain once the new package supersedes it."* This is that follow-up.

I verified there was no mount point before deleting rather than trusting the "orphaned" label. On `origin/dev`, `SwarmUIWidget` appears only in its own module, its `__init__.py` re-export, and the one test — nothing constructs it. The "legacy settings sidebar" the spec describes as its only host is itself retired; the repo now has a `test_legacy_entrypoints_retired.py`. And `generate_image` could never have run regardless: it called `run_until_complete()` inside the loop `@work(thread=True)` on an `async def` had already started, raising `RuntimeError: Cannot run the event loop while another loop is running` on every invocation.

So the right fix is the spec, not the code. Annotated in place rather than rewritten, so the original Phase 1 reasoning stays readable and the note records what changed and why.

Two related corrections worth putting on the record here, both to my own earlier statements in this PR:

- I described the third component as needing an owner decision because ADR-019 might depend on it. The investigation found something worse and simpler: rollback never reaches `SubscriptionScheduler` at all, on either flag setting, because `app.py`'s `watchlist_checks_enabled` branch has no `else`. Filed as TASK-1022.
- The provenance in that task was initially inverted. ADR-019 landed 2026-07-19 09:29; `SubscriptionWindow.py` — the live caller — was retired ~12 hours later that same day in `fc9e50da5`. So the ADR was not authored against broken code; a later refactor silently severed its rollback path. That distinction matters for the fix: restoring rollback means giving the old scheduler a *new* entry point, not repairing an existing one. Corrected in the task file.

PR #1029, opened 2026-07-27
docs: register the issues identified but not fixed this session

rmusser01:
Collects everything surfaced during the Evals rebuild and its follow-ups that was verified but left unfixed, so none of it is lost. Items that already have tasks are not repeated (TASK-1022, TASK-1034/1035/1036).

**1. `ConsoleProviderGateway` has the single-slot client-loop race that #1009 fixed elsewhere** — the highest-value item. `_active_http_client` keeps one client plus one `_client_loop` and, on mismatch, swaps and schedules `aclose()` on the loop it replaced. Structurally the same defect Qodo found in `GitHubAPIClient`, fixed there with a per-loop `WeakKeyDictionary`. The gateway is in better shape — atomic swap under `_client_lock`, done-callback on the close future — but it can still close a client another live loop is using. Its own docstrings describe the exposure: the gateway is shared between app-loop readiness probes and worker-thread generation calls, and `_schedule_stale_client_close` concedes the previous loop "may still be running elsewhere".

**2. `Tests/UI/test_console_native_chat_flow.py` is red on `dev`** — 18 failed, 192 passed, ~5m24s. Agents hit this repeatedly this session and each correctly called it pre-existing and unrelated; nobody wrote it down. At least two causes: `assert 'llama_cpp' == 'local_llamacpp'` and `AttributeError: 'ChatScreen' object has no attribute '_task_resume_state'`. Needs classifying as stale tests vs product regression — the five-minute runtime is likely why the rot went unnoticed.

**3. Repo hygiene** — 2,536 unreachable loose objects (~35 MB) and a stale `.git/gc.log` blocking automatic gc. Deliberately not acted on during the session: `git prune` removes unreachable objects, which is exactly what another session's detached or uncommitted work can look like in a checkout hosting this many concurrent worktrees. Should be done when the repo is quiet.

Documentation only.

---
## Summary by cubic
Add TASK-1064 to record verified-but-unfixed findings from the Evals/settings session. Lists the `ConsoleProviderGateway` client/loop race, 18 failures in `Tests/UI/test_console_native_chat_flow.py`, and repo GC cleanup (`git prune`, stale `.git/gc.log`); includes follow-up acceptance criteria; no code changes.

<sup>Written for commit ce1685dc5f7d692c011c222b0b9a8c936d7463d1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1029?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backlog/tasks/task-1064 - Register-of-issues-identified-but-not-fixed-during-the-Evals-session.md
+++ b/backlog/tasks/task-1064 - Register-of-issues-identified-but-not-fixed-during-the-Evals-session.md
@@ -1,0 +1,58 @@
+---
+id: TASK-1064
+title: >-
+  Register of issues identified but not fixed during the Evals/settings session
+status: To Do
+assignee: []
+created_date: '2026-07-27 17:00'
+labels:
+  - tech-debt
+  - follow-up
+dependencies: []
+priority: medium
+---
+
+## Description
+
+<!-- SECTION:DESCRIPTION:BEGIN -->
+Everything surfaced during the Evals rebuild and its follow-up work that was verified but deliberately left unfixed. Each item below is independently actionable and can be split into its own task; they are collected here so none is lost.
+
+Findings that already have their own tasks are not repeated: TASK-1022 (ADR-019 rollback), TASK-1034/1035/1036 (Evals UAT).
+
+---
+
+### 1. `ConsoleProviderGateway` has the single-slot client-loop race that TASK-981 fixed elsewhere — **highest value here**
+
+`Chat/console_provider_gateway.py::_active_http_client` keeps **one** `http_client` plus one `_client_loop`, and on a loop mismatch it swaps in a new client and schedules `aclose()` on the loop it replaced.
+
+That is structurally the same defect Qodo found in `GitHubAPIClient` during PR #1009, where it was fixed by moving to a per-loop `WeakKeyDictionary` cache. The gateway is in better shape than that code was — the read-check-swap is atomic under `_client_lock` (PR #629), and the close future carries a done-callback — but the underlying hazard remains: **closing a client that another live loop may still be using.**
+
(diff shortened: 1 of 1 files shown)


rmusser01:
Qodo's finding was correct, and the fix was ordering rather than wording.

TASK-1064 cited TASK-1034/1035/1036, which were not resolvable in the repo because they lived on the **unmerged** UAT branch (#1025). Merging this PR first would have put a task on `dev` referencing three ids that did not exist there — a dangling reference created purely by merge order.

Resolved by merging #1025 first. That surfaced a second problem worth recording: **task-1035 had collided** with another session's `Create source form cannot be filled in through the UI` (Done, created 2026-07-28 01:30), which reached `dev` while my branch was open. The standing rule applied cleanly — Done tasks never move, because close-out commits reference them — so mine moved to **1076**. I checked 1034 and 1036 against every remote branch excluding my own and both were genuinely free, so they stayed put rather than renumbering the whole trio unnecessarily.

The reference in 1064 now reads TASK-1034 / TASK-1076 / TASK-1036 and records the renumber inline, so the trail from the original ids is followable. Verified: all four referenced ids (1022, 1034, 1036, 1076) resolve on `dev`.

Worth noting the general shape, since this is the fifth id collision in this session: a cross-task reference is only as valid as the merge order that lands it. Filing tasks across two branches and referencing one from the other creates a window where the reference is dangling on whichever merges first.

PR #1031, opened 2026-07-27
qa: third Watchlists UAT — journey walks end to end; 2 findings + a correction to 1040

rmusser01:
Third Watchlists UAT. Docs and backlog only — no code changes.

Runs 1 and 2 each stopped at a blocking defect before reaching the experienced-user journey. Both are fixed, so this run finally walked it.

## Walked successfully, on a clean profile

| Step | Result |
|---|---|
| Create two watchlists | Both appear; scope follows the new one |
| Create three sources through the form | All three created; table and Feeds both update |
| Switch scope to a watchlist | `Feeds in Morning AI Brief (0)`, staging line narrows with it |
| Add a source to a watchlist | `Feeds in Morning AI Brief (1)`, source listed |
| Rename | Applied; the dialog pre-fills the current name and selects it |
| **Delete a watchlist holding a source** | **Orphan prevention holds** — scope moved to `Feeds in Unassigned (3)`, all three sources survived |

`task-895`'s AC#3 — deleting a watchlist must never orphan a source into invisibility — is now verified in the running app rather than only in tests. The delete confirmation also explains the consequence before acting: it names the watchlist, says its sources are not deleted, and says where they go.

## Found

- **task-1090 (high)** — `Escape` does not dismiss the dialogs; only clicking `Cancel` does. A keyboard user cannot back out, and because the dialog is modal the app appears to ignore every subsequent click. This bit the UAT itself: a `Delete` click was silently swallowed by a still-open Rename dialog.
- **task-1091 (medium)** — names are not trimmed, so `" Daily"` keeps its leading space and the tree row is visibly mis-indented next to its siblings; and the delete confirmation reads "Its **1 source are** not deleted".

## A correction I owe on task-1040

Run 2 filed — and I shipped — the claim that the rail and the centre disagree, citing `All sources  0` beside `Feeds in All sources (1)`. **That was wrong.**

`get_watchlist_item_counts` returns *item* totals, not source counts; its own docstring says so. The number beside a tree node is items scraped, the number in the Feeds heading is sources in scope. With three sources and nothing scraped, both are correct — which is exactly what this run observed.

The fix that shipped is still right for the sources table, which genuinely was stale. But 1040's AC #6 and #7 assert an equivalence that does not exist, and the task is corrected here so nobody implements a contract to make two different numbers match.

The real issue is milder: a bare `0` next to `All sources` gives no clue it counts items, so a user who has just added three sources reads it as "nothing happened". That is a labelling question, noted rather than filed.

## Still not covered

- **Tag filtering** — no tags were set on any source, so the rail's tag filters were never exercised.
- **The Feeds → Items → Content reading path** — nothing has been scraped and the sources point at `example.com` placeholders. Exercising it needs a real feed or seeded items, and it is the substance of Phase D.

---
## Summary by cubic
Documents the third Watchlists UAT and updates the backlog. Confirms the experienced-user journey end to end, verifies `task-895` AC#3 (deleting a watchlist never orphans a source), files two defects, and corrects `task-1040`’s counts interpretation.

- **Findings**
  - `task-1090` (high): Escape does not dismiss watchlist dialogs; blocks keyboard users and can swallow clicks while a modal stays open.
  - `task-1091` (medium): Names are not trimmed (e.g., " Daily" mis-indents), and delete copy says "1 source are".

- **Corrections**
  - `task-1040`: Rail shows item counts; Feeds header shows source counts. With no items scraped, `All sources 0` beside `Feeds in All sources (3)` is correct; task notes updated.

<sup>Written for commit 2e9609807c4fc3f9cbdc230a6635743656926859. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1031?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/qa/watchlists-uat-2026-07-27/run3-experienced-user.md
+++ b/Docs/superpowers/qa/watchlists-uat-2026-07-27/run3-experienced-user.md
@@ -1,0 +1,51 @@
+# Watchlists UAT run 3 — experienced-user journey, 2026-07-28
+
+`origin/dev` `e82ac1b18`, clean profile (`uat_wl_r3`, deleted after), 235x52,
+click columns computed by character.
+
+Third attempt. Runs 1 and 2 each stopped at a blocking defect before reaching
+this journey; both are fixed, so this run finally walked it.
+
+## Walked successfully
+
+| Step | Result |
+|---|---|
+| Create two watchlists | Both appear; scope follows the new one |
+| Create three sources through the form | All three created; table and Feeds both update |
+| Switch scope to a watchlist | `Feeds in Morning AI Brief (0)`, staging line narrows with it |
+| Add a source to a watchlist | `Feeds in Morning AI Brief (1)`, source listed |
+| Rename a watchlist | Applied; dialog pre-fills the current name and selects it |
+| Delete a watchlist holding a source | **Orphan prevention holds** — scope moved to `Feeds in Unassigned (3)` and all three sources survived |
+
+The delete confirmation explains the consequence properly before acting: it
+names the watchlist, says its sources are not deleted, and says where they go.
+
+`task-895`'s AC#3 — deleting a watchlist must never orphan a source into
+invisibility — is now verified in the running app, not just in tests.
+
+## Found
+
+- **task-1090 (high)** — `Escape` does not dismiss the dialogs; only clicking
+  `Cancel` does. A keyboard user cannot back out, and because the dialog is
+  modal the app appears to ignore every subsequent click. This bit the UAT
+  itself: a `Delete` click was silently swallowed by a still-open Rename dialog.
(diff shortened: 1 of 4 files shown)


rmusser01:
Superseded by #1072, which is this branch rebased onto current dev with the `task-1090` collision resolved.

dev has since gained a different `task-1090` ("watchlist fetch failures are swallowed into a debug log"), so this branch as-is produces two files under that id and fails the backlog guard on both the filename prefix and the frontmatter `id:`. The Escape finding is renumbered to **1300** there, past the current global max with headroom.

PR #629, opened 2026-07-14
Agent runtime Plan B: agent-capable Console — loop as reply engine, rail inspector, sub-agent badges

rmusser01:
## Agent runtime — Plan B: Console integration

Makes every Console conversation **agent-capable by default** (Codex/Claude-Code model): the Plan-A engine (#623) becomes the reply engine behind a `[console] agent_runtime` gate (default ON, live-toggleable kill switch; OFF = the pre-swap path, byte-untouched). A no-tool message streams exactly like today; the model reaches for tools/sub-agents only when it decides to.

Spec: `Docs/superpowers/specs/2026-07-12-agent-runtime-vertical-slice-design.md` · Plan: `Docs/superpowers/plans/2026-07-13-agent-runtime-plan-b-console.md`

### What's inside (8 reviewed tasks)
- **Unified reply engine** — regenerate joins submit/retry/continue on `_stream_assistant_response` (variant streams; mid-stream stop now works for regenerate; empty-stream/failure paths restore the original answer + prior status — no silent context drops).
- **`Agents/agent_stream.py` (StreamGate)** — the spec's streaming×fence policy: no-fence turns stream live; fence turns buffer; look-alike/malformed fences never drop text (`streamed + tail == visible` invariant, fuzz-verified 3000 trials).
- **Engine `on_step` hook** — backward-compatible per-step callback with `agent_kind`, feeding live UI.
- **`Chat/console_agent_bridge.py`** — adapts the engine to the Console store/gateway: live streaming with per-turn gating, TOOL markers (dim-styled rows), leaked-prose reset on disobedient fences, supersede on retry/regenerate, honest cancel.
- **Controller swap + gate** — catch-all prevents controller wedge; RUN_ERROR/STUCK map to honest failed semantics; system prompt composes with #620's session prompt; image/vision + staged context carry through identically to legacy.
- **Rail Agent section** — status + step tail + sub-agent list with drill-in (click-cycles runs), historical fallback from AgentRunsDB after restart; `[N Sub-Agents]` badge on conversation rows (one batched GROUP BY, TTL-gated — no N+1 on the poll).
- **Resume** — transcript TOOL markers + rail + badge all re-derive from AgentRunsDB after a real restart (byte-parity with live markers).

### Live gate (real llama.cpp, no mocks — `Docs/superpowers/qa/agent-runtime-2026-07/`)
[...]

PR #1034, opened 2026-07-27
fix(console): gateway closed a client that another live event loop was using

rmusser01:
Closes item 1 of TASK-1064 — the highest-value item in that register.

`ConsoleProviderGateway._active_http_client` kept a **single slot**: one `http_client` plus one
`_client_loop`. On a loop mismatch it swapped in a new client and scheduled `aclose()` on the loop it
replaced — closing a client another **live** loop might still be using.

The gateway's own docstrings describe exactly the overlap that makes this reachable: it is shared
between "readiness probes (awaited on the app's own event loop)" and "agent-runtime generation calls
(bridged from a worker thread via a fresh `asyncio.run()` per turn)", and
`_schedule_stale_client_close` concedes the previous loop "may still be running elsewhere (the app's
main loop)".

**This is the same defect that PR #1009 fixed in `GitHubAPIClient`**, so it takes the same shape of
fix rather than a second invented approach: a per-loop `WeakKeyDictionary` cache keyed by the running
loop, pruning of entries whose loop is closed, and an escape hatch that adopts the constructor-built
client into the first loop to touch it.

**What was already better here is preserved.** The read-check-swap stays atomic under `_client_lock`
(PR #629 Fix 1(a), which fixed a real torn-state race), `_owns_http_client` is untouched, and the
non-leaking close is kept — and improved: its done-callback now actually logs via
`logger.opt(exception=…)` instead of silently discarding the exception.

**Verified with a probe that discriminates.** Two real threads, barrier-synchronised so both loops are
alive simultaneously, each claiming a client and holding it in flight:

```
unfixed dev : loopA closed_in_flight=True   -> a live loop's client was closed
this branch : neither closed, distinct clients
```

That distinction matters. On #1009 I first wrote a *sequential* two-loop probe, saw distinct clients
and called it verified — then ran the identical probe against the unfixed code and it passed there
too, because the first loop is already closed by the time the second runs. A sequential test cannot
detect this class of bug. The branch's own new test (`test_concurrent_live_loops_never_close_each_others_client`)
was likewise checked against the pre-fix code first, failing with
`AssertionError: loop A's client was closed while loop B was concurrently alive…`, and passes 5/5 on rerun.

Tests: `test_console_provider_gateway.py` 85 passed (two single-slot-era tests rewritten for the new
architecture, three added), `test_github_api_client.py` 38 passed unchanged, 496 across the
agent-bridge/swap/native-tools suites. Unrelated pre-existing UI failures were checked by restoring
HEAD and re-running samples — none HTTP-client-related.

Items 2 and 3 of TASK-1064 are handled separately; their criteria are untouched here.

---
## Summary by cubic
Replaced the single-slot HTTP client in `ConsoleProviderGateway` with a per-event-loop cache so one loop can’t close another’s in-use client, and fixed teardown so a later loop can’t adopt a stale client. Adds safer close handling and targeted tests; addresses TASK-1064 item 1.

- **Bug Fixes**
  - Per-loop `WeakKeyDictionary` of `httpx.AsyncClient`, guarded by `_client_lock`, with pruning of closed loops.
  - Correct “unclaimed init client” handling via `_client_ever_claimed` to prevent adopting a client previously bound to another loop; added a regression test.
  - `aclose()` closes the caller loop’s client and schedules others via `_schedule_stale_client_close`, with failure logging.
  - Added a two-thread concurrency test to ensure live loops never close or share each other’s client; updated/added tests for first-touch adoption and `aclose()` behavior.

<sup>Written for commit 5c8bf396e5b640d921c1d8d6ed3a54495b116616. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1034?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Chat/test_console_provider_gateway.py
+++ b/Tests/Chat/test_console_provider_gateway.py
@@ -1795,19 +1795,74 @@ def test_injected_http_client_is_never_swapped_across_loops():
     asyncio.run(client.aclose())
 
 
-def test_active_http_client_swap_is_mutually_exclusive_across_threads():
-    """PR #629 Fix 1(a) (Gemini HIGH x2 + Qodo-8): the check-and-swap of
-    ``http_client``/``_client_loop`` must be a single atomic critical
-    section guarded by one lock, not two independently-racy reads/writes --
-    otherwise a concurrent caller from another thread/loop can interleave
-    with an in-flight swap and desync the client/loop pair (see the
-    interleaving hammer test below for the crash this produces in
-    practice). Proven deterministically here (no reliance on GIL
-    scheduling luck): thread A is parked *inside* the swap via a
-    monkeypatched, blocking ``_new_owned_http_client``, and thread B's
-    concurrent call must provably fail to complete while A is still in
-    flight -- only completing once A releases and the lock is free."""
+def test_active_http_client_first_touch_adopts_the_unclaimed_init_client(monkeypatch):
+    """TASK-1064 item 1: the client built in ``__init__`` is not yet bound to
+    (has not made a request on) any event loop, so the FIRST loop to call
+    ``_active_http_client()`` adopts it directly into the per-loop cache
+    instead of discarding it and building a fresh one -- construction should
+    not waste a connection, and there is nothing to close since no loop has
+    touched it yet. Directly supersedes the old single-slot design's
+    behavior (PR #629 Fix 1(b)), where the first touch unconditionally
+    swapped in a brand-new client and scheduled a close of the original."""
     gateway = ConsoleProviderGateway()
+    original_client = gateway.http_client
+    scheduled: list[tuple[int, object]] = []
+
+    def fake_schedule(client, loop):
+        scheduled.append((id(client), loop))
+
+    monkeypatch.setattr(
+        ConsoleProviderGateway,
+        "_schedule_stale_client_close",
+        staticmethod(fake_schedule),
+    )
+
+    async def touch() -> httpx.AsyncClient:
+        return gateway._active_http_client()
+
+    adopted = asyncio.run(touch())
+
+    assert adopted is original_client, (
+        "the first loop to touch the gateway must adopt the unclaimed "
+        "init-time client rather than discarding it for a fresh one"
+    )
+    assert scheduled == [], (
+        "adopting an unclaimed client must not schedule a close of it -- "
+        "nothing was ever using it"
+    )
+
+
+def test_active_http_client_creation_is_mutually_exclusive_across_threads():
+    """PR #629 Fix 1(a) (Gemini HIGH x2 + Qodo-8), preserved across the move
+    to a per-loop cache (TASK-1064 item 1): building a NEW per-loop cache
+    entry must be a single atomic critical section guarded by one lock, not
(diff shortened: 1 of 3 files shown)


rmusser01:
Both correct. The second is a real bug **this PR introduced**, which is worth stating plainly.

**2. Stale client re-adoption.** `aclose()` clears `_loop_clients` and resets `_client_loop = None` but leaves `self.http_client` pointing at the last-resolved client. The escape hatch then fires on:

```
self._client_loop is None
and self.http_client is not None
and not self.http_client.is_closed
```

That guard is insufficient, because `aclose()` closes other loops' clients **best-effort and asynchronously** via `run_coroutine_threadsafe`. In the window before those coroutines run — or when the owning loop is already gone and the close could not be scheduled at all — `is_closed` is still `False`. So a client that *was* claimed by a previous loop gets adopted into a new one, reintroducing precisely the cross-loop binding failure this PR exists to eliminate.

The root cause is inferring "never claimed" from `_client_loop is None`, when `aclose()` restores that exact condition *after* claims have happened. Fixing by tracking never-claimed explicitly rather than inferring it, and/or leaving the gateway in a state where adoption is impossible after teardown — with `aclose()` still idempotent and the non-leaking close preserved.

The test has to be the real thing: claim on loop A, `aclose()`, then obtain on loop B and assert it is not the same object. Where the timing window is not deterministic, force the reliable version — an owning loop already closed, so the close cannot be scheduled at all.

**1.** Reordering the `aclose()` docstring to Google layout, keeping the prose that explains the per-loop teardown.

For the record, this is the third time in this chain that a fix opened the next hole: the parent-directory guard in #978, the display-vs-persist pinning in #988, and now this. Each was caught in review rather than by tests, which is a fair signal about where the residual risk in this area actually sits — the interaction between a new lifecycle and the escape hatches around it.

rmusser01:
Note on the brief closure above: that was my error, not a review decision.

I chained `git rebase origin/dev >/dev/null 2>&1 && echo ...` and then ran `git push --force-with-lease` as a *separate* statement. The rebase hit a conflict in `task-1064` (both this branch and #1035 ticked their own item in that file), so it stopped with HEAD detached at `origin/dev` — and the push then ran unconditionally, pushing dev over this branch. Head and base became identical, so GitHub auto-closed the PR.

Nothing was lost. Both commits were intact in the reflog (`b179eedd5`, `59d861afb`), the branch is restored, and the rebase is redone with the conflict resolved so **both** items keep their ticks — item 1 here and item 2 from #1035 — with item 3 (the repo prune) still correctly unticked.

Re-verified after recovery: 126 tests pass, and the re-adoption probe still passes.

The lesson worth recording: `cmd_a && echo ok` followed by an unconditional `cmd_b` reads like a guarded sequence and is not one. A failed rebase should never be followed by a force-push in the same block.

PR #1043, opened 2026-07-27
fix(evals): benches saved before the duplicate-id validation were un-openable

rmusser01:
Fixes a regression I introduced in PR #962, filed here as TASK-1132.

#962 added duplicate-`target_ids` rejection to `BenchConfig.__post_init__`, correctly: duplicates
silently collapsed, because the runner and storage key per-target state by `target.id`, so two grid
columns shared one `eval_run`.

But `storage.load_bench()` builds a `BenchConfig` from stored rows, so that validation ran on the
**read** path too — and a bench saved *before* the validation existed became permanently un-openable.
Verified by writing a task row with `target_ids: [mid, mid]` directly (as a pre-validation save would
have) and calling `load_bench`: `ValueError: target_ids must be unique`.

An existing test had been saying so since #962 and was dismissed four times in one session as
"pre-existing, unrelated": `test_bench_with_duplicate_target_id_composes_without_raising`.

**The obvious fix would have been wrong.** Deduping on read collapses two columns into one — the
original bug wearing a different hat — and hides the problem from the person who has to fix it. That
test asserts *both* rows should render, and it is right.

So: **write stays strict, read becomes lenient.** `BenchConfig` gained `strict: InitVar[bool] = True`
gating only the uniqueness check — an `InitVar` is not a real field, so it stays out of equality,
repr and `asdict`, adding no persistent surface. Every write call site keeps the default; only
`load_bench` passes `strict=False`. `save_bench` gained its own independent check, since it cannot
assume a config handed to it was built strictly. `create_run_group` and the runner needed no change —
they were already independently guarded, which is what actually prevents the collapse.

Verified end to end on a legacy bench:
```
READ  : opens, both duplicate ids preserved
WRITE : still rejected  -- target_ids must be unique
RUN   : still rejected  -- targets must have unique ids
```

So a legacy bench opens, shows the duplicate target twice — that repetition *is* the visibility — and
refuses to run until corrected. Unreachable data becomes visible-but-blocked.

One thing the revert-check surfaced, worth recording: the `ValueError` was being swallowed by
`bench_editor.py`/`inspector.py`'s broad `except Exception`, so the user saw **zero rendered rows**
rather than an error. The bench did not appear broken; it appeared empty.

182 tests pass (178 baseline plus 4 new; the previously-erroring fixture now passes instead of failing
at setup).

Known and stated: `save_bench`'s new guard is not yet exercised in production — there is no edit-resave
flow, and `sample_bench.py` always passes a single target. It is defence-in-depth covered by unit test.

---
## Summary by cubic
Makes duplicate-target validation write-strict and read-lenient so legacy word benches open, while saves and runs still reject duplicates. Also adds ungated type checks for `target_ids` to catch corrupted stored data early. Fixes TASK-1132.

- Bug Fixes
  - `BenchConfig` adds `strict: InitVar[bool] = True` gating only the duplicate `target_ids` check.
  - `BenchConfig.__post_init__` now always validates `target_ids` is a list/tuple of non-empty strings; malformed entries (e.g., ints, nested lists) raise a clear `ValueError`.
  - `load_bench` builds with `strict=False` and preserves duplicate `target_ids` (no deduping).
  - `save_bench` independently rejects duplicate `target_ids`, even for leniently loaded configs.
  - Write-time guards remain in `create_run_group` and `WordBenchRunner.run`.
  - Tests updated/added for legacy read, malformed types, write-time rejection, and UI composition.

<sup>Written for commit e503575932a372d8c2786a39f10f913160bd729a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1043?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Evals/word_bench/test_models.py
+++ b/Tests/Evals/word_bench/test_models.py
@@ -70,6 +70,53 @@ def test_bench_config_rejects_duplicate_target_ids():
         )
 
 
+# ---------------------------------------------------------------------------
+# task-1132 follow-up (Qodo review, finding 2): target_ids element-type
+# validation. This is a SEPARATE, pre-existing gap from the uniqueness check
+# above -- `strict` never gated anything but uniqueness, and no element-type
+# check existed on read or write before or after task-1132. It matters now
+# because `load_bench`'s read-leniency means a corrupted stored `target_ids`
+# entry (an int, a nested list, an empty string) would otherwise load
+# without complaint and only fail much later, deep inside
+# `db.get_model(target_id)`, as an opaque sqlite parameter-binding error.
+# ---------------------------------------------------------------------------
+
+
+def test_bench_config_rejects_a_non_string_target_id():
+    """A malformed element (an int here, standing in for corrupted stored
+    data) must be rejected at construction with a message naming both the
+    offending value and its type, not surface later as an opaque sqlite
+    parameter-binding error out of db.get_model."""
+    with pytest.raises(ValueError, match=r"target_ids.*123.*int"):
+        BenchConfig(
+            name="b", prompt_mode="raw", top_k=20, dataset_id="d",
+            target_ids=("t1", 123),
+        )
+
+
+def test_bench_config_rejects_a_non_string_target_id_even_when_lenient():
+    """The element-type check is NOT gated by `strict`, unlike the
+    uniqueness check above -- `storage.load_bench` always constructs with
+    `strict=False`, so a check that only ran when `strict` is True would
+    never protect the read path this validation exists for."""
+    with pytest.raises(ValueError, match=r"target_ids.*123.*int"):
+        BenchConfig(
+            name="b", prompt_mode="raw", top_k=20, dataset_id="d",
+            target_ids=("t1", 123), strict=False,
+        )
+
+
+def test_bench_config_rejects_target_ids_that_is_not_a_list_or_tuple():
+    """Covers the other half of the malformed-data surface: target_ids
+    itself has the wrong shape entirely (e.g. a bare string, which Python
+    would otherwise happily iterate character-by-character)."""
+    with pytest.raises(ValueError, match="target_ids must be a list or tuple"):
+        BenchConfig(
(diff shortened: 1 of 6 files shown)


rmusser01:
Both being addressed. Finding 2 is worth doing, but its causation is wrong and I want that on the record rather than silently accepting the framing.

**2. Malformed `target_ids` on lenient load.** Qodo states `strict=False` "bypasses the only validation that would catch malformed `target_ids`". I checked `BenchConfig.__post_init__`: `strict` gates **only** the uniqueness check. `prompt_mode`, `top_k` and `concurrency` validate unconditionally, and there has **never** been element-type validation of `target_ids` — not before this PR, not after. So this is a **pre-existing gap**, not something this PR introduced.

    It is still worth fixing here, because read-leniency means we now deliberately accept more from stored data. Today a corrupted `config_data.target_ids` containing a non-string loads fine and then fails much later inside `db.get_model(target_id)` as an opaque sqlite parameter-binding error, far from the cause — `eval_models.id` is `TEXT`.

    Fixing as: **always** validate `target_ids` is a list/tuple of non-empty strings regardless of `strict`, keeping **uniqueness** gated by `strict` exactly as now. That preserves the whole point of this PR — legacy duplicates still load and still render twice — while restoring a clear failure at the boundary for genuinely malformed data. It goes in `__post_init__` ungated rather than in `load_bench`, since it should hold on every path.

**1.** Adding the Google-style sections to the fixture and new tests, keeping the existing prose that records why the duplicate-bearing fixture exists.

Unchanged and re-verified before this round: all three write paths still reject duplicates (`BenchConfig` strict, `save_bench`, `create_run_group`), and read still preserves both ids rather than deduping. That split is the fix; I am not letting a follow-up quietly erode it.

PR #1044, opened 2026-07-28
fix(evals): a global focus outline was painting over the results grid's header

rmusser01:
Closes TASK-1034, found in live UAT.

**Symptom.** The results grid rendered two structurally different ways in the same view — boxed with
no column header and no `[warned]` canary marker, or unboxed with both. I could not isolate the
trigger during UAT and filed it as unexplained.

**Root cause, confirmed.** It is one widget in two focus states, not two render paths.

The generated bundle carries a global `*:focus { outline: solid $ds-focus-accent; }`. In Textual,
`outline` paints **inside** the widget's own area, over its first and last content rows — unlike
`border`, which sits outside. A `DataTable`'s first content row is its header, so a focused grid
loses the header to the outline, and the outline's edges read exactly like the "box" observed.

The missing piece: `ResultsGrid.on_mount` auto-focuses the table on **every** mount, which is why the
headerless form appeared on a fresh mount after "Create sample bench", and why the header reappeared
once focus moved to the lens `Select`.

Confirmed with a `Pilot` run reading `screen._compositor.render_strips()` (Textual 8.2.7 has no
`App.export_text()` — my own first probe used that and returned nothing, proving nothing). The
focused render reproduced the UAT symptom byte-for-byte; blurring restored the header instantly.

Two lines below that rule the bundle already states the principle it violates: *"Focus and hover
states should only change colors, not layout."*

**Fix, deliberately narrow.** `#evals-grid-table:focus { outline: none; }` plus a header recolor
(`background`/`color`/`bold underline`) as the replacement focus cue — scoped to this one widget
rather than to `DataTable:focus` or the global rule, because roughly 25 other files construct
`DataTable`s that are not exercised here.

The focus cue is replaced, not dropped: `.datatable--cursor` was checked and gives **zero**
focus/blur differentiation in this app, because app CSS restyles it unconditionally. The outline was
the only cue, so removing it without a substitute would have traded a rendering defect for an
accessibility one.

Verified: the global `*:focus` outline is untouched for every other widget, and the bundle still
reproduces from its sources.

Revert-check: reverting only `_evals.tcss` and rebuilding fails the new test with
`AssertionError: column header missing from the focused grid's rendered output:` followed by the
boxed render.

Tests: 56 in the evals grid/screen suites, 14 across the focus-accessibility and transcript-selection
contracts. `test_non_obscuring_focus_contract.py`'s 9 failures are pre-existing baseline breakage in
this worktree (retired selectors, a missing `_chat_tabs.tcss`), verified byte-identical to `HEAD`.

---
## Summary by cubic
Fixes an issue where the Evals results grid lost its column header and “[warned]” marker when focused. Addresses TASK-1034 by disabling the outline on `#evals-grid-table` and using a header color change as the focus cue.

- **Bug Fixes**
  - Scoped `#evals-grid-table:focus { outline: none; }` to prevent the global `*:focus` outline from painting over the header.
  - Styled `#evals-grid-table:focus > .datatable--header` (background, color, bold underline) to keep keyboard focus visible without altering layout.
  - Added a regression test that reads only the table’s header row from the compositor to confirm the header and `[warned]` are visible while focused, replacing the previous whole-screen assertion to avoid false positives.

<sup>Written for commit a005744494bc63f6b1ce571bc0ae8001b57b1193. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1044?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Finding 3 is correct and it is the most important one in this PR, because **the test protecting the fix does not protect it**.

I verified the mechanism: the fixture creates a target named `"base"` (`create_model(name="base", ...)`), and `results_grid.py:595` renders the baseline `Select` as `f"Column · {target['name']}"`. So the screen contains `"Column · base"` whether or not the `DataTable` header exists, and `assert "base" in rendered` passes with the header completely absent — the exact regression this test was written to catch.

All three assertions share that flaw: they match against `_rendered_text(pilot.app)`, the whole-screen blob. The run view now also renders the degenerate-canary callout merged in #1040, so there is more unrelated text in that blob than when the test was written. Constraining them to the `DataTable`'s own header row via `render_strips()`, and verifying each is genuinely discriminating rather than assuming.

Worth stating plainly: this is the tenth test in this session that passed for the wrong reason, and it is the first one that was guarding a fix rather than sitting beside it. The recurring shape is always the same — a substring assertion against a broad surface, where something else on that surface happens to satisfy it. `"3" in calls_text` matching the `3` in `36`; a `len(calls)` check that held with the guard removed; an `AsyncMock` producing a look-alike error string; a compat shim making a markup assertion read back correctly either way. Narrow the surface, or the assertion is decoration.

Finding 1 (docstring) is being handled alongside, keeping the prose that records the outline/focus root cause.

PR #1062, opened 2026-07-28
docs: file TASK-1260 — path-validation property tests are load-sensitive

rmusser01:
Docs and backlog only. Files TASK-1260 and records the lesson behind it.

## What happened

A post-merge check on `dev` showed one failure — `test_safe_paths_always_validate` — and I reported the session complete a moment before seeing it. This PR is the follow-through.

## Attribution

The failure appeared in a three-suite run, and the test lives in `Tests/Utils/`, the directory I had just added a file to. "Obviously unrelated" wasn't good enough, so I ran the matched comparisons rather than assuming:

| Command | Result |
| --- | --- |
| the test alone | passed |
| `Tests/Utils/` **with** my new file | 576 passed |
| `Tests/Utils/` **without** it | 570 passed |
| `Subscriptions/ Scheduling/ Utils/` | **1 failed**, 892 passed |
| the same three, re-run | 893 passed |
| the same three at `e74e37d07` (pre-change) | 878 passed |

Not a regression, and not caused by the new file. An intermittent.

## Mechanism

The property tests in that file carry no `settings(...)` override, and there is no Hypothesis profile in `Tests/conftest.py` — so they run under the default **200 ms per-example deadline**. `test_safe_paths_always_validate` creates a `TemporaryDirectory` plus up to five directories *per example*. On a machine running 10+ concurrent pytest processes from parallel agents, that crosses 200 ms.

## Why file it rather than shrug

The failure is indistinguishable from a real regression at the moment it appears, and it appeared while unrelated work was in flight. Attributing this single instance took five runs across two worktrees. Any future change unlucky enough to be in flight when it fires pays that cost again.

The AC asks for a Hypothesis profile registered once in `Tests/conftest.py` rather than a one-line patch to this file — other property files carry the same exposure — and asks that the reason be recorded next to it, so the deadline is not "tightened back up" later as an apparent improvement.

## Lesson recorded

`backlog/docs/lessons-testing-evidence.md` gains the diagnostic shortcut: when a failure shows up in a run that mixes suites, check first whether it's a Hypothesis property with no deadline override — that's cheap to confirm. But don't skip the clean-baseline run on the strength of it, because "it's probably a flake" is the same shape of reasoning as "it's probably unrelated," and this repo has punished both.

---
## Summary by cubic
Docs for TASK-1260: `hypothesis` path-validation properties are load-sensitive and can fail under the default 200 ms deadline.
Adds a backlog task and a testing lesson with steps to add a suite-wide Hypothesis profile in `Tests/conftest.py` that relaxes deadlines to prevent false failures; also corrects the mkdir count (four, not five) and fixes the lesson’s section ordering.

<sup>Written for commit fd05e3dc5cb4fa5f6b5743445cb34894a5088f90. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1062?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backlog/docs/lessons-testing-evidence.md
+++ b/backlog/docs/lessons-testing-evidence.md
@@ -272,13 +272,6 @@ are cheaper than re-deriving the graph.
 
 ---
 
-## Related
-
-- `lessons-live-verification.md` — why the suite could not see seven of these defects
-- `lessons-backlog-hygiene.md` — task IDs, CLI quirks, git plumbing traps
-
----
-
 ## A property test with no deadline override is load-sensitive
 
 **TASK-1260, 2026-07-28.** `test_safe_paths_always_validate` failed once inside a
@@ -307,3 +300,10 @@ and this repo has punished both.
 
 The durable fix is a Hypothesis profile registered once in `Tests/conftest.py`,
 not a per-file patch — other property files carry the same exposure.
+
+---
+
+## Related
+
+- `lessons-live-verification.md` — why the suite could not see seven of these defects
+- `lessons-backlog-hygiene.md` — task IDs, CLI quirks, git plumbing traps
--- a/backlog/docs/lessons-testing-evidence.md
+++ b/backlog/docs/lessons-testing-evidence.md
@@ -284,10 +284,12 @@ are cheaper than re-deriving the graph.
 **TASK-1260, 2026-07-28.** `test_safe_paths_always_validate` failed once inside a
 three-directory run, passed alone, passed on re-run, and passed on a clean
 baseline with the identical command. It is a Hypothesis `@given` property that
-creates a `TemporaryDirectory` and up to five directories per example, with no
-`settings(...)` override and no Hypothesis profile in `Tests/conftest.py` — so it
-runs under the default **200 ms per-example deadline**. On a machine with 10+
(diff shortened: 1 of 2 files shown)


rmusser01:
Fixed in `c8a758731` — the finding is correct.

`max_size=5` with `for component in components[:-1]` gives at most **four** `mkdir` calls, since the last component is the file rather than a directory. Both documents said five. Corrected in the task and in the lessons entry, each now naming the reason so the number can be re-derived rather than trusted.

The load-sensitivity conclusion is unchanged, but a diagnostic note whose mechanism is off by one invites the next reader to distrust the rest of it — which is the opposite of what the note is for.

Also fixed while here: the lessons entry had been appended with a shell redirect rather than inserted, so it landed *after* the `## Related` footer. Reordered in `fd05e3dc5`.

PR #1070, opened 2026-07-28
fix: make a click on a table row select it, app-wide (TASK-1180)

rmusser01:
Closes TASK-1180.

Textual fires `RowSelected` on **activation** — Enter, or a second click — while a single click produces `RowHighlighted`. Seven panes handled only the former, so a click moved the cursor and selected nothing: the Inspector stayed empty and row actions stayed disabled. TASK-1105 fixed this for Watchlists by hand, after it made `Preview`/`Check now`/`Delete` unreachable by mouse and ultimately hid a dead scrape path. Everything else still had it.

## The audit (AC#2)

An AST pass over the 14 panes under `UI/` that handle either event:

| Pane | Verdict |
|---|---|
| `mcp_audit_mode`, `mcp_permissions_mode`, `mcp_servers_mode`, `mcp_tools_mode` | affected → mixin |
| `Voice_Cloning_Window` | affected → mixin |
| `stts_profile_library` | affected → explicit paired handler |
| `mcp_workbench` | **false positive** — handles the panes' own messages, not `DataTable` events |
| 5 Watchlists panes | already fixed (TASK-1105) |
| `Evals/results_grid`, `schedules_workbench` | correctly highlight-only |

`stts_profile_library` binds its handler with `@on(DataTable.RowSelected, "#stts-profile-table")`, which the mixin's name-based dispatch cannot reach — hence the explicit pairing.

## The naive fix is a feedback loop, and I built it first

`RowHighlighted` is not a user-input event. A pane rebuilding its own table emits it too, because `clear()` resets the cursor to row 0. On the MCP workbench a selection awaits a remove/mount in `MCPInspector`, which re-syncs the mode, which repopulates the table, which highlights row 0 again.

**Opening the Tools tab with zero user input produced 157 selections**, and buried a genuine click under the repeats. That's why the first live check showed nothing happening — not a harness artifact this time, a real regression I'd introduced.

Three layers now, each earning its place from an observed failure:

1. **Focus gate.** Measured on the real screen: a repopulating table is not focused, while a click focuses it before the cursor moves and keyboard navigation requires focus by definition. 157 → 0.
2. **`repopulating_table()`**, declared before `clear()`/`add_row()`. Focus alone was not enough — switching the selected server rebuilds the tools table *while it still has focus*, and the row-0 transient then re-selected a tool, defeating the clear that triggered the rebuild. The existing `test_switching_selected_server_clears_tool_detail` caught it. Suppression releases after the **next refresh**, not at the end of a `with` block, because Textual delivers a rebuild's highlight messages after the producing code returns — the obvious context-manager shape would have looked right and covered nothing.
3. **Dedup** on the last forwarded row key, bounding the damage if a future pane forgets to declare.

## Two panes already knew

`mcp_permissions_mode` and `mcp_audit_mode` carry comments explaining that `DataTable.clear()` resets `cursor_coordinate` to (0, 0), and they already restore the cursor **by row key** afterwards. My first version would have silently defeated that careful existing work by treating the transient as a selection. Reading their comments turned the fix from "add a guard" into "declare the rebuild so the existing restore still wins."

## Watchlists deliberately untouched

Measured: **0** spurious calls. `select_source_by_id` sets local state without the inspector-remount feedback cycle, so the storm is specific to MCP's architecture rather than inherent to highlight-forwarding. Those hand-written handlers stay as they are rather than churning a recently-fixed area. That leaves two patterns in the tree; the mixin is the one to use for new tables.

## Mutation testing corrected itself

My first check **passed with the focus gate removed** — because the dedup guard independently suppressed the storm. A test that survives removing either guard proves neither. Isolating gave **0 / 0 / 162** (focus-only / dedup-only / neither), and each guard now has a test that fails when exactly that guard is removed.

## Verification

- 5 unit tests on the panes + 2 end-to-end through the real `MCPScreen` → `MCPWorkbench` → `MCPInspector`, including one asserting that opening Tools mode selects **nothing** on its own.
- Both new suites failed against pre-change code, per AC#5.
- **Failure sets identical to baseline `6577c67cf`** — the same 7 pre-existing failures on both sides (4 `PrivatePathError: link_or_non_regular`, environment; 3 audit-redaction). `test_switching_selected_server_clears_tool_detail` passes on baseline, failed on my first draft, passes now.

One process note: I twice read "0 failures" that turned out to be `-q` suppressing pytest's summary lines rather than a clean run. The numbers above come from plain output on both sides.

---
## Summary by cubic
Make a single click on any `DataTable` row actually select it (not just move the cursor), app-wide, and prevent double-firing when users arrow then press Enter (TASK-1180). Also stops selection storms during table rebuilds and keeps keyboard and mouse selection consistent.

- **Bug Fixes**
  - Added `DataTableClickSelectMixin` (`UI/Widgets/table_click_select.py`) to forward `DataTable.RowHighlighted` to existing `on_data_table_row_selected`, with a focus gate, `repopulating_table()` suppression (released after next refresh), per-row-key dedup, and a new one-shot activation dedup that wraps `on_data_table_row_selected` to swallow the Enter that completes the same gesture.
  - Applied mixin to `mcp_audit_mode`, `mcp_permissions_mode`, `mcp_servers_mode`, `mcp_tools_mode`, and `Voice_Cloning_Window`; added an explicit paired handler in `stts_profile_library` due to `@on(..., "#selector")`. Audited remaining panes; Watchlists were already fixed.
  - Added tests (`Tests/UI/test_table_click_selects.py`, `Tests/UI/test_mcp_table_click_selects_end_to_end.py`) covering click and keyboard selection, prevention of rebuild-triggered auto-selections, end-to-end inspector updates, and the one-shot arrow+Enter dedup.

<sup>Written for commit 4d81f2ebee2a61bbb784265e4408828cd519b03f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1070?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_mcp_table_click_selects_end_to_end.py
+++ b/Tests/UI/test_mcp_table_click_selects_end_to_end.py
@@ -1,0 +1,118 @@
+"""TASK-1180 AC#1: a click on the MCP Tools table reaches the Inspector.
+
+`Tests/UI/test_table_click_selects.py` proves the pane posts its selection
+message on highlight. This proves the other half — that the message reaches the
+real `MCPWorkbench` and changes what the Inspector shows — using the real
+screen, workbench and inspector rather than a bare pane.
+
+The two halves matter separately: the pane could post correctly into a workbench
+that never handled it, which is the shape of defect this session kept finding.
+"""
+
+from __future__ import annotations
+
+import pytest
+from textual.widgets import DataTable
+
+from Tests.UI.test_destination_shells import (
+    DestinationHarness,
+    _active_destination_screen,
+    _wait_for_selector,
+)
+from Tests.UI.test_screen_navigation import _build_test_app
+
+pytestmark = pytest.mark.unit
+
+
+@pytest.mark.asyncio
+async def test_clicking_a_tool_row_updates_the_inspector_end_to_end():
+    """A single click — i.e. a `RowHighlighted` — must select through to the
+    workbench, not merely move the DataTable cursor."""
+    from tldw_chatbook.MCP.hub_tool_catalog import HubTool
+
+    app = _build_test_app()
+    host = DestinationHarness(app, "mcp")
+
+    async with host.run_test(size=(180, 50)) as pilot:
+        screen = _active_destination_screen(host)
+        await _wait_for_selector(screen, pilot, "#mcp-hub-rail")
+        screen.action_mcp_mode("tools")
+        await pilot.pause()
+
+        tool = HubTool(
+            name="search",
+            server_key="local:docs",
+            server_label="docs",
+            source="local",
+            description="Search docs",
+            tags=(),
+            input_schema=None,
+            executable=True,
+            stale=False,
+        )
+        screen.workbench._last_hub_tools = [tool]
+
+        tools_mode = screen.query_one("MCPToolsMode")
+        table = tools_mode.query_one(DataTable)
+        row_key = table.add_row(tool.name, tool.server_label, key=tool.tool_id)
+        # A click focuses the table before moving the cursor; the mixin gates on
+        # that, because an unfocused table moving its own cursor is a
+        # repopulation rather than a person.
+        table.focus()
+        await pilot.pause()
+
+        # Exactly what a single click produces once the cursor lands on the row.
+        tools_mode.post_message(DataTable.RowHighlighted(table, 0, row_key))
+        await pilot.pause()
+        await pilot.pause()
+
+        # `current_tool` is the inspector's own public accessor for what its
+        # tool-detail view currently describes -- the thing a user sees change.
+        from tldw_chatbook.UI.MCP_Modules.mcp_inspector import MCPInspector
+
+        inspector = screen.query_one(MCPInspector)
+        shown = inspector.current_tool
+        assert shown is not None and shown.tool_id == tool.tool_id, (
+            "clicking a tool row did not reach the Inspector: the pane handles "
+            f"only activation, so the tool view still shows {shown!r}"
+        )
+
+
+@pytest.mark.asyncio
+async def test_opening_tools_mode_selects_nothing_on_its_own():
+    """The regression the focus gate exists to prevent.
+
+    A first draft of the mixin forwarded every `RowHighlighted`, including the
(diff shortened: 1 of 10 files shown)


rmusser01:
Fixed in `4d81f2ebe` — the finding was correct, and chasing it turned up a second bug of mine.

**The double-fire.** Confirmed exactly as described: `assert 2 == 1` in `test_row_selection_posts_entry_selected_with_synthetic_index` and its findings twin. `__init_subclass__` now wraps each pane's own handler with a per-row guard. The mixin can't intercept by defining `on_data_table_row_selected` itself — a subclass method shadows a base-class one, and the subclass is where every pane defines its handler — so wrapping at class creation gets in front of that without any pane having to remember. A `_forwarding_highlight` flag distinguishes the mixin's own call from a native activation; without it the wrapper swallowed its own forward, since `_should_forward` records the key before invoking the handler.

**The second bug, which the fix caused.** Keyed persistently, the dedup also swallowed *later legitimate* re-selections of the same row — which is what `goto_permission_row()` and the sub-view triggers do — breaking `test_goto_permission_row_is_the_single_shared_implementation_for_all_three_triggers` and `test_both_sub_view_selections_do_not_stack_visible_detail_panels`. The dedup is now **one-shot**: set when a highlight forwards, consumed by the next native activation, then cleared.

Those two only fail in the full-file run — they need a prior test to leave the key set — so my isolation mutation check passed while the real run failed. Added `test_reselecting_the_same_row_later_is_not_swallowed`, which fails on its own when I make the dedup persistent again.

**On why I missed the original.** My run of that scope reported clean, but used `-q`, which suppresses pytest's summary lines. That's the third time today that artifact has misled me and the first time it hid failures I'd introduced. Every figure in the PR body is now from plain output.

**Result:** 549 passed, and the failure set is identical to baseline `6577c67cf` — the same 7 pre-existing failures (4 `PrivatePathError`, environment; 3 audit-redaction), nothing new.

PR #1085, opened 2026-07-29
feat(console): streaming, provider-agnostic dictation behind the existing Mic button

rmusser01:
## What

Upgrades Console dictation from the one-shot, Parakeet-ONNX-only backend to the streaming `ConsoleVoiceInputController`: live partial transcripts in the status chip while speaking, per-segment finals accumulated and inserted once at the caret, and support for any installed local STT provider (`parakeet-onnx`, `parakeet-mlx`, `faster-whisper`, `lightning-whisper-mlx`, `qwen2audio`, `parakeet`, `canary`). The user-facing surface is unchanged: same `#console-dictation` button, same four-state lifecycle, same wall timer, same origin-session attribution — the four shipping contract tests in `Tests/UI/test_console_dictation.py` pass **unmodified**.

## Bug fixes that stand on their own (all consumers, not just Console)

- `LazyLiveDictationService.stop_dictation()` never released the microphone — every normal stop left capture live, and **every second dictation session in an app run was broken** app-wide (`is_recording` stuck true → next start refused).
- `_process_audio_buffer` passed `bytes` to `transcribe(audio_path: str)` — the lazy service could not transcribe at all; now uses `transcribe_buffer` with the resolved provider/model/language.
- Privacy allowlist listed `"lightning-whisper"` (matches nothing) and held 3 hardcoded literals vs the resolver's 7 — both layers now consume one catalogue (`Utils/local_stt_providers.py`), so they cannot drift.
- First-run: the 1.4 GB model download happened mid-capture inside a 2 s join, then blamed the microphone. The model now warms before capture with visible status, a config-backed stop-join timeout (`dictation.stop_join_timeout_seconds`), byte-count-based error attribution, and an abandon-aware daemon warm (quit during download: hang → ~1 s exit).
- Tail flush: the processing loop drained its queue on stop, so the last word of every capture was lost and short utterances reported a false mic failure.
- Chip renders via `textual.Content` (never markup-parsed): Whisper's `[BLANK_AUDIO]`/`[Music]` no longer swallowed, `[/path]` can no longer raise `MarkupError`.

## Verification

- Live-verified on real hardware: full capture, **second capture in the same run** (previously broken), and mic release on mid-capture navigation (macOS orange dot cleared ~2 s).
- 202 tests green across the dictation surface; subprocess guard proves the Console screen imports none of the heavy STT stack (`faster_whisper`/`nemo`/`torch`) at startup.
- Zero regressions vs the `origin/dev` baseline, compared name-by-name (33 failures in the affected sweep, all 33 pre-existing among the baseline's 203). CSS bundle guard exits 0.

## Notes

- `Audio/console_dictation.py` (228 lines) now has zero production callers; retirement recommended but deliberately not done here.
- Follow-ups filed as tasks 1280–1284 (delete orphaned `voice_input_button.py`, composer undo/redo, legacy dropdown provider ids, `DictationProcessor` thread leak on release, `config.py` underscore provider id).
- Trap for reviewers touching Textual messages: a `Message` class whose snake_case name matches an existing `_on_<name>` method is auto-dispatched into it (`handler_name` convention) — that's why the buffer-limit signal is named `ConsoleDictationLimitSignal`.

---
## Summary by cubic
Upgrades Console dictation to a streaming, provider‑agnostic flow behind the existing Mic button with live partials in the chip and per‑segment finals inserted once at the caret. Adds model warm‑up, clear availability guidance, and bounded buffering; UI stays the same and the dictation contract tests pass unchanged.

- **New Features**
  - Streaming `ConsoleVoiceInputController` replaces the one‑shot backend; works with any installed local STT provider: `parakeet-onnx`, `parakeet-mlx`, `faster-whisper`, `lightning-whisper-mlx`, `qwen2audio`, `parakeet`, `canary`.
  - Voice chip shows live partials and a 1s elapsed timer; finals are accumulated and inserted once.
  - Provider resolution + availability probe with tooltip remedies; button stays pressable and is visibly dimmed when unavailable; no heavy `Audio/` imports at app start.
  - Speech model is warmed before capture (first run shows a clear notice); preparing can be cancelled.
  - Bounded PCM buffering on the Console path to prevent overrun.
  - New config keys with safe defaults: `dictation.stop_join_timeout_seconds` (30s) and `dictation.warm_model_before_capture` (true).

- **Bug Fixes**
  - Microphone is released on stop; subsequent sessions work; tail audio is flushed so short utterances and last words aren’t lost.
  - Transcription uses `transcribe_buffer` with streaming partial/final callbacks and passes the resolved provider/model/language; unified `Utils/local_stt_providers` catalogue prevents allowlist/id drift.
  - Better empty‑transcript attribution (no‑bytes vs no‑speech vs unfinished), plus a configurable join timeout.
  - Chip renders plain text to avoid markup issues and no longer swallows tokens like `[BLANK_AUDIO]`; start/stop race and double‑error paths hardened.
  - Availability probe now requires NumPy and recognizes all 7 local providers; availability remains visible without disabling the button.

<sup>Written for commit 4b3acb2d854e827d01eb928fcd71bcc487091eed. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1085?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Both review findings addressed in 2363eac45 (now c209b8f48 after rebase onto dev):

1. **NumPy probe gap (bug):** `capture_available()` now requires `numpy` (via `find_spec`, probe stays import-free) alongside a capture backend, matching `AudioRecordingService.__init__`'s hard requirement. Covered by three hermetic tests including backend-present/numpy-missing → `missing-capture`.
2. **Docstring sections (rule):** added the missing `Returns:`/`Args:` sections and docstrings on the flagged callables, plus `probe()` and three event dataclasses found by the same sweep.

Branch rebased onto current dev (37 commits replayed; only the generated CSS bundle conflicted — regenerated from sources, guard green). Full dictation surface: 205 passed on the rebased tree.

PR #1125, opened 2026-07-31
fix(settings): Theme recompose-storm app freeze + honest footer s/r hints (rescore P1 batch)

rmusser01:
## Summary

First fix batch from the Settings critique rescore (29 → 30/40, snapshot `2026-07-31T03-47-38Z`). Both P1s here were verified live before and after.

### 1. Theme category froze the entire app (task-1581)
Opening Settings → Theme spun the process at ~60% CPU and killed ALL input — keyboard, mouse, and nav tabs — until restart. Root cause: `SettingsThemeEditor.is_modified` posted `ThemeModifiedStatus(False)` from its init-time watch call on every mount, then flipped True when `load_theme`'s programmatic Input writes echoed back as `Input.Changed`; the screen's `theme_editor_modified` reactive (recompose=True) oscillated forever, each flip mounting a fresh editor that emitted the next pair. Diagnosed with `sample`, a SIGABRT faulthandler dump (100 frames of mount→compose cycle), and a compose-entry probe showing the alternation. **Pre-existing since the Theme category was wired in (a94f84a94)** — not from the recent remediation PRs.

Fix: `init=False` on the reactive plus no-op guards so programmatic reloads never count as user edits. Side benefit: Theme no longer claims unsaved changes the moment it opens.

### 2. Footer advertised save/revert where they don't act (task-1580)
`s save category | r revert category` rendered on all six read-only view categories, autosaving Splash, and immediate-apply Workspaces — both keys just toast there. Now gated on `GUIDED_SETTINGS_MUTATION_CATEGORIES`, mirroring task-1564's `t` gate. Categories with no active keys fall back to the app default hints.

Also files rescore P2/P3 follow-up tasks 1582-1585.

## Testing
- TDD RED-first: editor posts observed as `[False, True]` before the fix; storm test pinned editor-instance identity across settles.
- Affected suites: **406 passed, 1 failed** — the failure (`test_library_registration_updates_the_screens_own_footer`) fails identically on pristine dev @ 1e0979c74 (pre-existing, unrelated).
- Live tmux verification: Theme opens at 0.0% CPU (was 56-62%), filter round-trips Theme → Storage → Theme → Console Behavior, footer shows `s|r|t` on Storage, `s|r` on Console Behavior, default hints on Theme.

---
## Summary by cubic
Fixes a Settings Theme freeze and keeps the editor mounted during real edits. Also makes footer save/revert hints honest by showing them only where drafts exist.

- **Bug Fixes**
  - Theme: stopped the recompose storm with `is_modified = reactive(False, init=False)` and guards that ignore programmatic reload events; dropped `recompose=True` on `theme_editor_modified` and refresh the rail/inspector in place so real edits don’t remount the editor. Opening Theme no longer marks modified or locks input, and typed values persist (task-1581).
  - Footer: gated `s`/`r` hints on `GUIDED_SETTINGS_MUTATION_CATEGORIES`; read-only pages, Splash, Workspaces, and Theme drop save/revert while testable categories still show `t` (task-1580).

<sup>Written for commit e05c186623cb73f37e3a035e62ca76b2bdee7b27. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1125?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_screen_footer_hints.py
+++ b/Tests/UI/test_screen_footer_hints.py
@@ -188,8 +188,14 @@ async def test_library_registration_updates_the_screens_own_footer():
 @pytest.mark.asyncio
 async def test_settings_registration_updates_the_screens_own_footer():
     """task-264 review Important: Settings' s/r/t hints (previously rendered
-    by the retired Footer) must reach its own footer via registration."""
+    by the retired Footer) must reach its own footer via registration.
+
+    The hints are category-gated (rescore P1): the read-only Overview
+    default advertises none of them, so the check switches to a guided
+    draft category first.
+    """
     from tldw_chatbook.UI.Screens.settings_screen import SettingsScreen
+    from tldw_chatbook.UI.Screens.settings_config_models import SettingsCategoryId
 
     app_instance = _build_test_app()
     host = _DefaultScreenFooterHost(app_instance, SettingsScreen)
@@ -198,6 +204,15 @@ async def test_settings_registration_updates_the_screens_own_footer():
         await pilot.app.workers.wait_for_complete()
         await pilot.pause()
         screen = host.screen_stack[-1]
+        screen_footer = screen.query_one(AppFooterStatus)
+        for token in ("save category", "revert category", "test category"):
+            assert token not in screen_footer.shortcut_text
+        screen.active_category = SettingsCategoryId.PROVIDERS_MODELS.value
+        await pilot.pause()
+        screen._register_footer_shortcuts()
+        await pilot.pause()
+        # The category switch recomposes the screen, replacing the footer
+        # widget -- re-query rather than trusting the stale reference.
         screen_footer = screen.query_one(AppFooterStatus)
         for token in ("save category", "revert category", "test category"):
             assert token in screen_footer.shortcut_text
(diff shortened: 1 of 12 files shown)


rmusser01:
Addressed all three Qodo findings in e05c18662:

**1. Theme edits lost on recompose (Bug — confirmed real, fixed RED-first).** Reproduced exactly as described: with the editor's posts made honest, the first real color edit flipped `theme_editor_modified` (recompose=True), the screen recomposed, and the fresh editor discarded the in-progress input while the flag stuck at True. Fix: dropped `recompose` on the reactive; its two dependents (rail dirty marker, inspector "Unsaved theme changes" row, now id'd as `#settings-theme-unsaved-note`) refresh in place via `_refresh_theme_modified_widgets` — the same targeted-refresh idiom `InternalPromptsPanel.Modified` already uses one handler below. New regression test `test_theme_user_edit_does_not_remount_editor` pins editor identity, value survival, and the in-place inspector flip. Live-verified in tmux: typed `#123456` survives, inspector shows Yes, rail shows `Theme *`, CPU idle.

**2 & 3. Docstring Args (rule 497152).** Added Google-style `Args:` sections to `on_color_input_changed` and `on_dark_mode_changed`.

Suites re-run after the fix: 407 passed, 1 failed — `test_library_registration_updates_the_screens_own_footer`, which fails identically on pristine dev @ 1e0979c74 (pre-existing, unrelated).

PR #1115, opened 2026-07-31
Watchlists briefings phase 1: on-demand text briefings (spec #2)

rmusser01:
## Summary

Phase 1 of the Watchlists briefings/podcasts design (spec #2, `Docs/superpowers/specs/2026-07-30-watchlists-briefings-design.md`): a watchlist can now be turned into an on-demand **text briefing** — a structured markdown digest of what its sources did since the last one — rendered in a new **Artifacts** section of the Watchlists strip.

- **Data:** `briefings` / `briefing_items` tables (additive DDL), `briefing_selection_mode` + `default_briefing_preset_id` watchlist columns, `queued_for_briefing` read/write.
- **Selection** (`Subscriptions/briefing_selection.py`): id-watermark coverage windows (`covers_through_item_id` — immune to 1-second `created_at` ties, auto-excludes late-added source backlogs), three modes (`auto` / `curated` / `auto_featured`), featured items survive the item cap, exact overflow counts with SQL-bounded materialisation.
- **Pipeline** (`Subscriptions/briefing_service.py`): one call through `Chat_Functions.chat_api_call` with `content_kind`-aware prompts (articles contribute "what it says", change items "what changed", diffs labelled by page name); honest statuses `generating` / `complete` / `empty` / `failed` (+error), overflow stated in the body, never silent truncation. All DB work off the event loop via `asyncio.to_thread`.
- **The named invariant — a failed generation never advances coverage** — is triple-guarded (service branch, DB `status IN ('complete','empty')` watermark allowlist, junction-exclusion status predicate) and each leg is independently mutation-proven.
- **Zombie recovery**: a `generating` row not backed by a live worker is failed as `interrupted` on both the next Generate attempt and plain Artifacts load.
- **UI**: Artifacts section (list + `Markdown(hyperlinks=False)` reader + Generate with refuse-while-running toast), queue-for-briefing toggle in the Inspector with an items-pane indicator — in-place patch, no full-screen recompose, all writes off the UI thread.

Preset-less by design: `preset_id` stays NULL, the app default provider/model applies. Phases 2–4 (presets/scripts/audio, exports/feed, scheduling) are tracked in task-1540.

**Owner-confirmed deferrals** (recorded in the spec's delivery notes + task-1540): the selection-mode picker (no writer yet; `auto`/`curated` unreachable until phase 2's UI) and citation links/pruned-item degradation (markers render as plain text in phase 1).

[...]

PR #1145, opened 2026-07-31
Watchlists briefings phase 2a: presets, script casting, mode picker, citations (spec #2)

rmusser01:
## Summary

Phase 2a of the Watchlists briefings design (spec #2) — the text half of "Casting and audio". A complete briefing can now be **cast into an N-speaker script** from a user-defined preset, and both phase-1 deferrals are retired: the **selection-mode picker** exists (making `auto`/`curated` reachable end-to-end) and **`[item N]` citations navigate into the reader** with honest pruned-item degradation. Audio (synthesis/stitching/playback) is deliberately split to phase 2b (task-1630) after plan-time adapter verification surfaced five TTS-layer reality gaps, recorded in the spec's delivery notes.

- **Data:** `briefing_presets` (name, style notes, provider/model override, ordered roster JSON — per speaker: display name, role prompt, optional character card, voice-profile reference recorded-but-inert until 2b) and `briefing_scripts` (roster snapshot, turns JSON, honest statuses) with allowlisted+validated CRUD; watchlist settings writer; batched item lookup with chunked IN-clauses.
- **Cast service** (`Subscriptions/briefing_cast.py`): one chat call over the finished briefing; strict JSON-turns parsing — an unknown speaker fails the script **naming the speaker** (named invariant), malformed payloads fail naming the parse defect, a deleted character card fails naming the card; the briefing row is never touched by any script outcome (asserted byte-for-byte); scripts snapshot their roster so preset edits/deletes never orphan an artifact's meaning; all DB work and character-card loads off the event loop.
- **Preset manager modal**: dynamic speaker rows, injected character/voice option lists (a missing service degrades the field, never the modal), `validate_roster`-gated saves, and write-completions that verify they still own the edit state (a mid-write preset switch cannot cause a save to overwrite the wrong preset).
- **Toolbar pickers**: selection mode + default preset, written off-loop with in-place patches (no recompose), scope-guarded against cross-watchlist bleed; Generate consumes the stored default preset (provider/model/style pinned at the chat seam for both generation and casting).
- **Citations**: a per-briefing citations table resolved in one batched query; activation (select, not highlight — arrow-browsing must not yank the section) switches to the reader via the real selection seam; a pruned item renders "no longer available" and refuses navigation with a toast (`citation-to-pruned-item-degrades`, the second named invariant).

Follow-ups filed: **task-1630** (phase 2b audio, carrying the five adapter findings) and **task-1631** (test-harness lazy-db-factory gap — confined to pytest's own tmp, three pre-existing workarounds consolidated).

## Test plan

- 599 passing across `Tests/Subscriptions/` (290), `Tests/Watchlists/` (275), `Tests/UI/test_watchlists_inspector.py` (34); close-out sweep 658 passed with only documented pre-existing baselines failing.
- Every behavioural change carries a revert-confirm-RED-restore mutation check — including the two named invariants, the off-loop pins, the ownership guards, and the cast-side provider/model pins the whole-branch review found missing.
- Seven-task subagent-driven build: per-task reviews (three fix rounds), whole-branch review on the strongest model (2 Importants found between tasks — a dangling-default refusal bypass and the unpinned cast LLM plumbing — both fixed and re-review-approved).
- Notable mid-build finds: Textual's `is_mounted` is False throughout `on_mount`; cell-cursor DataTables never fire `Row*` events for real input (all live handlers moved to the `Cell*` seam).

---
## Summary by cubic
Adds briefing presets and N‑speaker script casting, plus selection‑mode/default‑preset pickers and citations that open items in the reader. Also adds a preset manager and “Cast Script” from the Artifacts pane; audio synthesis is deferred to phase 2b.

- New Features
  - `briefing_presets` and `briefing_scripts` tables with validated CRUD; watchlist settings writer; listings support pagination via `limit`/`offset`; batched item lookups use chunked `IN` clauses.
  - Casting service generates strict JSON turns from a preset, snapshots the roster, trims speaker whitespace, pins provider/model, and reports clear errors for unknown speakers, bad JSON, or deleted cards.
  - Preset manager modal to create/edit/delete presets with inline validation and guarded writes that never overwrite the wrong row.
  - Artifacts toolbar adds selection‑mode and default‑preset pickers (first Changed absorbed; writes scoped to the current watchlist), a Presets… shortcut, and “Cast Script”; Generate uses the stored default preset.
  - Citations use `extract_citation_ids` (case‑insensitive), resolve in batch, show in a small table, and activate on selection; pruned items show a toast and do not navigate.

- Bug Fixes
  - Prevented cross‑watchlist clobbers; picker writes verify ownership post‑await.
  - Casting now refuses when the default preset was deleted; provider/model are pinned at cast time.
  - Fixed picker mount‑time race by ignoring each Select’s first Changed event.
  - Citations activate only on selection (not highlight) and no longer switch sections on cursor move.
  - Preset modal validates text fields before save.

<sup>Written for commit c738d4cbc552b44f2719143961d5afc21ac31937. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1145?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
All six Qodo findings adjudicated — four fixed in `26993660e`, two declined with evidence.

**Fixed**

1. **Speaker whitespace breaks casting** (bug — good catch, and a real one). `validate_roster` canonicalizes roster names by stripping, but `parse_script_turns` compared the model's raw `speaker` string against that canonical set, so a turn like `"Alice "` failed the *entire* cast as an unknown speaker. The speaker is now stripped before the membership check and the canonical name is stored in the turn, so downstream rendering matches the roster. Strictness on genuine mismatches is unchanged — the named invariant `test_an_unknown_speaker_fails_the_script_by_name` still passes unmodified.
2. **Citation regex too strict** (bug). `extract_citation_ids` matched only lowercase `[item N]`, so model drift to `[Item 12]` silently produced zero citations — and an empty citations table is indistinguishable from a briefing that has none. Now case-insensitive, with the ordered/deduped/junk-rejecting contract intact; the new test pins that differently-cased twins dedupe to one entry.
3. **Unpaginated listings.** `list_briefing_presets`/`list_briefing_scripts` now take `limit=200, offset=0` applied as SQL `LIMIT/OFFSET`, preserving both orderings exactly (presets name ASC, scripts newest first — each still pinned by its pre-existing test). Call sites unchanged.
4. **Preset modal skips `input_validation`.** Persisted free-text fields now go through `Utils/input_validation.validate_text_input` with per-field bounds; a failure surfaces on the modal's existing markup-safe inline-error path with nothing written.

Each fix is mutation-verified (revert → RED → restore). 296 `Tests/Subscriptions/` + 276 `Tests/Watchlists/` + 34 inspector tests green.

**Declined**

5. **`streaming=False` on the cast call** — deliberate, same as the phase-1 disposition on #1115. The cast is a background artifact pipeline, not an interactive chat surface: its output contract is a strict JSON array of turns that can only be validated once complete, and the UI renders honest `generating`/`complete`/`failed` statuses on the script row rather than a token stream. There is no incremental surface to stream into. The streaming rule targets chat UX.
6. **Schema DDL outside `migrations/`** — `Subscriptions_DB` has never used the migrations directory: all 16 of its table definitions are inline additive `CREATE TABLE IF NOT EXISTS`, and `ls tldw_chatbook/DB/migrations/` contains zero subscriptions/watchlists files (that directory serves ChaChaNotes, as TTS and Scheduling serve their own). The design spec for this work also rules the machinery out explicitly: *"All tables are additive `CREATE TABLE IF NOT EXISTS` / column-presence `ALTER`s. No data migration exists in this design, so the TASK-1362 atomicity machinery is not needed; do not cargo-cult it in."* Introducing a migration file here would make this DB the only one with two competing schema mechanisms.

PR #1147, opened 2026-07-31
feat(settings): critique round-4 batch — field-level search, pinned State bar, copy honesty (tasks 1640-1645)

rmusser01:
## Summary

Full finding set from the round-4 Settings critique (31/40, both P1s + all P2s + P3s).

### P1 — Field-level search (task-1641)
`/` indexed category names, descriptions and owned config keys — so typing **"threshold"** or **"density"** found nothing, and the operator had to know the 22-category taxonomy to change one known setting. Now field labels are indexed too: the echo names the target (**"Console Behavior › Threshold (chars)"**) and **Enter lands focus ON the field**, which also fires its inspector guidance. Live-verified end to end.

### P1 — Copy honesty (task-1640)
- The Overview action slot rendered a bare **"Theme"** noun whose verb lived in a mouse-only tooltip → **"Open Theme editor"**. It stays in the pinned header: the 32-row compact contract requires a *painted* recovery action there — moving it to the body broke that test, the kind of constraint only the suite knows.
- The dirty banner threatened data loss that **doesn't happen** — drafts persist per category, source-verified — so it now says "switching categories keeps this draft".
- Footer `t test category` named an abstraction while buttons say Check/Validate/Test → each category shows its real verb (`t check index`, `t test provider`, …).
- DB stats spell their labels (**Prompts / Chats/Notes / Media**) instead of hiding `P:/C/N:/M:` behind a hover tooltip — a keyboard-first product can't put meaning in hover.
- The full local-first paragraph reads once on Overview, one line elsewhere; duplicate headings and stacked colons removed.

### P2 — Round-3 edge cases (task-1642)
- **RAG showed no State bar at all**: each category composed its own banner *inside* the scrollable content, so the persistence badge scrolled away mid-task. Now ONE banner, pinned between the pane title and a new `#settings-detail-pane-body` scroll container, on every category.
- Overview's inspector cut mid-sentence with no fold indicator — the sync-rows recompose re-mints a hidden hint after the mount-time evaluation; now re-evaluated there too.
- The in-place guidance refresh bypassed `_fold_long_tokens` ("paste_collapse_thresh/old"); all five refresh paths fold now.

### P2/P3 — Remaining (tasks 1643-1645)
Built-in RAG profile's Delete is annotated and disabled ("Delete — built-in") instead of sitting full-red under the read-only banner. Textual's Toast `DEFAULT_CSS` sets `border-left: outer` per severity — which outranked round-3's border shorthand and kept the orange stripe alive; every variant now pins its own round left edge. Draft badge casing unified; two-line scope texts tightened; stuttering "Applies: Applies to…" fixed.

## Testing
- 12 new tests (field search rank/echo/focus, pinned banner outside scroll, t-verb map, toast left edges, Overview-only reassurance, annotated Delete).
- Affected suites: **707 passed**, 1 order-dependent flake that passes isolated.
- Live tmux verification of every user-visible change.
- DESIGN.md documents the pinned contract row, field-level search, and the inert-destructive-action convention.

---
## Summary by cubic
Adds field-level search in Settings and pins the State bar above a new scrollable body, plus copy and UI polish from critique round 4. Fixes scroll theming on the new pane bodies.

- **New Features**
  - “/” now indexes field labels; Enter opens the category and focuses the matched field (fires guidance). (task-1641)
  - State bar is pinned between the pane title and a new `#settings-detail-pane-body` scroll area across all categories. (task-1642)

- **Bug Fixes**
  - Copy honesty: dirty banner notes drafts persist; Overview action reads “Open Theme editor”; footer “t” uses each category’s real verb; DB stats spell labels; shorter reassurance off Overview. (task-1640)
  - Built‑in RAG profile: Delete is disabled and annotated (“Delete — built‑in”). (task-1643)
  - Toast variants pin their left border to remove the leftover severity stripe. (task-1644)
  - Tightened badge casing, scope copy, and “Applies” text; fold indicators and dotted‑key folding work after refresh. (tasks 1645, 1642)
  - Pane scroll styling is applied to `#settings-detail-pane-body` and `#settings-impact-pane-body`; outer panes clip only, keeping custom scrollbars.

<sup>Written for commit 5e8e1f51f69b1708afd131568bdedc6918da9043. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1147?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Both findings fixed in 5e8e1f51f:

**1. Pane scroll theming aimed at the wrong widget (Bug — agreed, and broader than reported).** `#settings-detail-pane` kept the overflow/scrollbar theming after task-1642 moved scrolling into `#settings-detail-pane-body`. Checking the sibling revealed the same latent mismatch on `#settings-impact-pane` — its header/body split landed back in task-1560, so **its theming has been on the wrong widget since then** and nobody noticed. Both outer panes now keep only their column width and clip; both `-body` children carry the scroll contract (`overflow-y: auto`, `scrollbar-size`, the $ds-* scrollbar colors). `test_settings_detail_and_inspector_panes_scroll_long_content` was asserting against the outer selectors and passing on stale rules — it now pins the corrected structure.

**2. Test docstrings (rule).** The batch's seven new tests converted to Google style: one-line summary, blank line, then the task context.

Suites after the repoint: **533 passed, 0 failed** (hub, RAG profile region, workspaces, destination shells) plus **210 passed** across the CSS-contract and visual-parity suites.

PR #1150, opened 2026-08-01
fix(images): resolve the graphics protocol before Textual takes the terminal (task-1660)

rmusser01:
## Summary

Closes the diagnosis behind task-1532's long-open acceptance criterion. The user reported the Kitty rail avatar rendering **pixelated** instead of as a native graphics image.

**Root cause is load order, not config.** `textual_image` picks its renderer exactly once, at *import* time, in `textual_image/renderable/__init__.py`:

```python
is_tty = sys.__stdout__ and sys.__stdout__.isatty()
if is_tty and sixel.query_terminal_support():   Image = SixelImage
elif is_tty and tgp.query_terminal_support():   Image = TGPImage
elif is_tty:                                    Image = HalfcellImage   # ← half-blocks
else:                                           Image = UnicodeImage
```

Those queries write an escape sequence and wait for the terminal to answer. **Every app-side import was lazy** — nested inside functions that run in the live app — so the query fired after Textual had taken raw mode and owned stdin. It never got a reply, selection silently fell to `HalfcellImage`, the widget mounted *successfully*, and the user saw half-block pixelation with no exception and nothing in the log. The library warns about precisely this in its own source: *"querying the terminal isn't possible anymore once Textual is started."*

**The scope was every image surface**, not just the avatar — Console transcript images, Personas thumbnails, and the full-size viewer modal all share the lazy-import pattern.

### Ruled out first (so the record is clear)
- The user's config is correct: `default_render_mode = "auto"`, `terminal_overrides.kitty = "regular"`.
- Detection is correct: `TERM=xterm-kitty` → `kitty`, `tgp: True`.
- `resolve_default_mode` returns **`graphics`** for their exact config (verified by probe) — so the spec was right and the mount was the problem.
- `textual_image` is installed. The `optional_deps` check doesn't rescue it: importing the **top-level** package never loads the `renderable` submodule where selection happens, and those checks are lazy by default.

## The fix
`warm_up_image_protocol()` imports `textual_image.widget` once, from both entry paths (`__main__` block and `main_cli_runner`), immediately before `app_instance.run()`. The lazy import sites are untouched — they now receive the already-resolved class. A missing optional dependency or a non-responding terminal returns `False` and leaves the app on its existing mosaic/pixels fallbacks.

## Testing
- 5 new tests, RED-first: the warm-up loads the *submodule that actually selects* (not just the package), it survives the dependency being absent, **each** entry point warms up *before* `run()`, and the helper invariant is pinned so removing it can't silently regress rendering.
- Suites: 52 image/terminal + 53 smoke/App green.
- Under a pty, the TGP query escape is observably written **at warm-up time** (`_Gi=…,a=q,…`) — the query that previously never reached a listening terminal; with no responder it correctly lands on halfcell and returns `False`.
- Live tmux boot: app starts clean, no stray query bytes on screen.

**One AC stays open deliberately:** full proof needs a real Kitty/iTerm2, which this harness can't provide — same honest limit that kept task-1532 open. @rmusser01, the check is simply whether the sidebar avatar now looks like a photo rather than blocks.

---
## Summary by cubic
Fixes pixelated images by resolving the terminal graphics protocol before Textual takes over, adds clear diagnostics for degraded paths, and updates the console rail avatar to fill the rail without cropping. Addresses task-1660 and task-1661, closes the open acceptance criterion from task-1532, and is now verified in real Kitty.

- **Bug Fixes**
  - Added `warm_up_image_protocol()` in `Utils/terminal_utils.py` to import `textual_image.widget` before `app_instance.run()` from both entry paths; routes availability via `optional_deps.check_dependency` while still importing the protocol-selecting submodule; now logs the outcome and selected protocol, and distinguishes a missing `textual_image` from a failed terminal query.
  - Sized the rail avatar to the rail: introduced `character_avatar_box()` (clamped), added `_character_avatar_available_cols()` to measure the rail section body (not the auto-width holder), set the holder’s `width/height` to `"auto"`, and switched the mosaic fallback to `fit="contain"` so graphics and non-graphics paths frame portraits the same.
  - Tests: cover submodule selection, per-entry ordering, and diagnostic paths; mark warm-up tests as unit and restore any `sys.modules` evictions for hygiene.

<sup>Written for commit 7779d4b4b123fb2167ab8d0f0b33c31e47c8fc52. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1150?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Follow-up pushed (task-1661) after the user confirmed native graphics now render — but the portrait sat small in the corner of a tall empty box.

**Two causes, neither of them cropping:**

1. **The holder claimed the whole rail.** `ClickableAvatarBox` is a bare Textual `Container`, which defaults to `width: 1fr; height: 1fr` — and nothing ever sized it (no CSS rule for `#console-character-avatar`, no inline styles at the compose site). So it expanded to fill the rail section, the image sat at its default top-left origin, and the character name got pushed to the bottom of that expanse. It now hugs its content.
2. **The box ignored the rail.** Sizing was hard-coded `CHARACTER_AVATAR_COLS/LINES` (16×8) regardless of width, so a ~50-column rail displayed a 16-column thumb. `character_avatar_box()` now derives the box from the holder's settled width, clamped to 44×22, with lines at half the columns (terminal cells are ~2× taller than wide, so that reads near-square). Before layout settles it falls back to the old constants.

Both render paths share the box, and the mosaic fallback moves from `fit="cover"` to `"contain"` so graphics and non-graphics frame identically — the user picked whole-image over crop-to-fill.

**Explicitly ruled out: nothing crops.** `prepare()` only downscales above the decode ceiling, PIL's `thumbnail()` preserves aspect, `get_pil()` returns a copy, and `fit_image_cell_size` is a contain fit. The bust framing in the screenshot is the character card's own artwork.

3 new tests (box math incl. fallback/scale/clamp, contain agreement across paths, holder sizing). Suites: 37 avatar/mosaic/warm-up green; live boot clean.

rmusser01:
All three findings addressed in ea8191d5d — the third one was the sharpest and I'd call it a real catch.

**3. Discarded bool → no log/metric (Bug — agreed, and it undercut the fix's own premise).** Both entry points ignore the return value, so a failed warm-up produced *no* log line and *no* metric — preserving precisely the "degraded rendering with no log line" symptom this PR exists to eliminate. The helper now carries the diagnostic itself: a successful warm-up logs and counts the protocol it actually resolved (`tgp` / `sixel` / `halfcell` / `unicode`), which is the single most useful line for anyone debugging pixelation. Failures log too (see below).

**1. Blanket `except Exception` masking regressions (Bug — agreed).** Split: `ImportError` means the optional dependency is absent, which is expected and logs at debug; anything else means the import genuinely misbehaved — in practice a terminal refusing the capability query, which I observed raising `TerminalError` under a bare pty — and now warns with the exception type. The broad guard stays because that pty case is real and must not crash startup, but it can no longer hide behind the expected case.

**2. Route through `optional_deps.check_dependency` (respectfully disagree, and it's load-bearing).** That helper imports the **top-level** `textual_image` package — and I verified during diagnosis that a top-level import never loads `textual_image.renderable`, which is where the protocol is actually chosen. Only `textual_image.widget` pulls it in. Routing through the shared helper would silently reinstate the exact bug this PR fixes, while *looking* correct. I've documented that constraint in the docstring so the next person doesn't 'clean it up'.

Two new tests: missing-dependency and failed-query are distinguished by their emitted labels, and a successful warm-up names its protocol. Suites: 51 green (warm-up, avatar, terminal utils, smoke); live boot clean.

rmusser01:
Second pass — every open review item is now addressed, including cubic's, which I had missed on the first sweep.

**Qodo rule finding (optional_deps) — I took your middle path.** My first reply refused this outright, which was too blunt. Availability now routes through `check_dependency("textual_image")` so `DEPENDENCIES_AVAILABLE` stays consistent with the rest of the codebase; the submodule import stays direct and deliberate, because `check_dependency` imports only the **top-level** package and never loads `textual_image.renderable` — where the protocol is actually chosen. Routing the whole thing through the helper would look tidier and silently reinstate the bug. That constraint is now documented at the call site so it survives future cleanup.

**cubic P2 — missing `@pytest.mark.unit`.** Valid; `Tests/README.md` requires it and the sibling `Tests/Utils` modules follow it. All six tests in the new module now carry the marker.

**cubic P3 — `sys.modules` pollution.** Valid and the sharper of the two: the warm-up test evicted `textual_image.renderable`/`.widget` and never restored them, leaking global state that could perturb any later test inspecting import state. Now saved and restored in a `finally`.

**Qodo bug finding (blanket except)** was fixed in the previous push: `ImportError` (expected, debug) is separated from anything else (unexpected, warns with the exception type), and success logs plus counts the resolved protocol.

Also corrected my earlier comment on this PR, which posted a literal `$(git rev-parse --short HEAD)` instead of the SHA — it referred to ea8191d5d.

62 tests green (warm-up, avatar, terminal utils, viewer, mosaic, smoke); live boot clean. User has visually confirmed both the native-graphics render and the rail sizing in Kitty.

PR #1153, opened 2026-08-01
feat(console): actionable shell chips + longer model names (tasks 1670-1672)

rmusser01:
## Summary

Three related changes to the Console shell strip, from a user request.

### 1. Provider and Model chips are clickable (task-1670)
Both now activate on click or Enter/Space and open the **same** quick provider/model popover `Alt+M` already opened — a second entry point into the existing opener, not a fork. The popover existed but was reachable only by a shortcut you had to know about. Implemented by following `ConsoleApprovalsChip`/`ConsoleScopeChip`, which were already the established actionable-chip pattern in this widget.

### 2. The model name shows in full (task-1671)
The cap moves 22 → **34 cells**. The ask was 25 characters of the model *name*; `"Model: "` costs 7 cells and padding 2, so a 25-cell chip would have shown only ~16 of the name. Chips are `width: auto`, so the higher cap only bites long labels — `RAG: off` and the rest are visually unchanged. Focus still lifts the cap entirely for anything longer.

### 3. The Character chip opens a character picker (task-1672)
Searchable modal over saved character cards, with name matches ranked above description matches so typing a name can't bury it. Per the user's decision it's a two-step flow: pick a character, then choose **Swap in this chat** or **Start new chat**. Neither is right for every case — swapping mid-conversation changes the system prompt while the old transcript stays on screen, and starting fresh loses the thread — so the user says which, per pick. A greeting seeds **only into an empty chat**, so a live conversation is never interrupted by one.

The modal never touches the database itself: the screen loads cards off-thread and passes options in, matching how the other Console pickers are built.

## Testing
- 7 new tests: both chip classes are activatable and carry their message types, the width cap arithmetic (34 − prefix − padding = 25), filter ranking and bounds, the screen subscribes to both new messages and reuses the existing popover opener, and the greeting-only-when-empty rule.
- **1709 passed** across the full Console UI suite (the one reported failure was a stale test name from a run that started before a rename; re-run green).
- Live tmux verification: Provider chip → popover, Model chip → popover, `custom-model-alpha` renders in full where it was previously truncated, Character chip → picker listing characters, Enter stages one, both placement buttons appear.

## Not yet proven end-to-end
The swap *outcome* — a rebound session actually using the new character's system prompt on the next model turn — is covered by unit tests and the code path, but I have not watched a real reply after a mid-conversation swap against a live provider. Worth a look before or shortly after merge.

---
## Summary by cubic
Makes the Console shell chips actionable and shows longer model names; adds a character picker. Fixes the picker flow so swaps/new chats apply prompts, persist, refresh, and clear stale selections. Addresses tasks 1670–1672.

- **New Features**
  - Provider/Model chips: click or Enter/Space opens the same quick model popover as Alt+M.
  - Model chip width: cap raised 22 → 34 cells to show up to 25 chars of the model name; focus still lifts the cap.
  - Character chip: opens a searchable picker; choose swap this chat or start new; greeting only in empty chats; options loaded off‑thread and the modal does not write to the DB.

- **Bug Fixes**
  - Use the screen’s Console store so selections apply.
  - Seed the character’s system prompt for both new and swapped chats; persist to saved conversations.
  - Await UI refresh after applying changes.
  - Clear a staged pick when the query or cursor changes to prevent applying the wrong character.
  - Disable markup in the placement hint; Unicode search uses casefold; tests cover `@on` handlers and the greeting-only-when-empty rule.

<sup>Written for commit 19fbc4591383e7cbca42528ed21a9adc13af7176. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1153?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Addressed — one of the three was new and real.

**`_pending` staleness (Bug — agreed, and a good catch).** After Enter staged a character, the Swap/New buttons stayed on screen while `_pending` still pointed at the *old* pick. Typing a new query or moving the cursor changed what the user was looking at but not what those buttons would apply, so a click could silently rebind the chat to a character they'd moved off. `_clear_pending()` now runs on every result refresh and both cursor moves, hiding the placement row and restoring the neutral hint. Regression test drives the modal in a Textual harness and asserts the pick survives neither a query change nor a cursor move. cubic's pass didn't find this one.

**system_prompt never written (Bug)** and **missing `Args:` on `_swap_console_session_character` (rule)** were both already fixed in ab6001aae — this review ran against the pre-cubic SHA. The prompt now reaches session settings on both paths (new chat: at `create_session`, mirroring Start-Chat; swap: `replace_session_settings` plus `update_conversation_system_prompt` so it survives reload), and the docstring carries a full `Args:` block. Verified in the current tree before replying rather than assuming.

Suites: **1710 passed** across the full Console UI set, plus the 8 tests in this PR's own module. Live end-to-end re-verified: the chip flips `Assistant: General` → `Character: Default` with its toast.

rmusser01:
**End-to-end verified against a payload-logging provider** — the claim I could not previously back.

Method: seeded two marker characters into a scratch profile, pointed the Custom provider at a stub that logs every request body, sent a message, swapped character via the chip, and sent another. Comparing what the provider actually received:

```
BEFORE swap
  [0] system    You are a capable assistant with optional tools…   (generic default)
  [1] user      hello before swap

AFTER swap
  [0] system    You are AlphaBot. SECRET=ALPHA_MARKER_7731. Always mention ALPHA_MARKER_7731.
                ALPHA_MARKER_7731 personality
                ALPHA_MARKER_7731 description
                You already opened this conversation with the following message…
  [1] user      hello after swap
```

The swapped character's system prompt reaches the model on the very next turn — the exact behaviour that was broken before the cubic round (system_prompt computed then discarded) and that no unit test can prove on its own. The trailing line also confirms the task-1531 greeting-folding contract still holds: the greeting rides inside the system row rather than as an assistant-first turn, which Anthropic and Gemini reject.

Chip label flipped `Assistant: General` → `Character: AlphaBot` in the same run.

PR #1160, opened 2026-08-01
feat(console): composer overflow menu — image, caption, narrate, impersonate (tasks 1680-1683)

rmusser01:
## Summary

Adds the requested `☰` button before Send, opening a menu with four actions.

**Why a menu:** the composer action row is width-bounded and already carries Send/Stop/Mic/Attach/Save. New actions go behind one button rather than growing the row; it widens by 5 cells for the button itself (all three width sites — compose-time plus the two runtime indicator branches).

### Generate Image (task-1681)
Modal collects prompt / backend / style, **previews the exact command as you type**, and pastes it into the composer. It deliberately does *not* generate — you review and send, so the existing `/generate-image` handler stays the single execution path and the documented `:backend`/`@style` grammar gets built for you instead of memorized.

### Generate Caption (task-1682)
Inserts a ready caption prompt (description paragraph + alt-text line). Disabled with an explanation when nothing is attached.

### Narrate Entire Conversation
Deliberate placeholder that says so, per the request.

### Impersonate (task-1683)
Asks the current model to draft **the user's** next message from the transcript and puts it in the composer for review. Never sends, never touches the transcript. Insertion rules exactly as specified:
- existing text is **never** replaced — the draft is appended on a new line;
- a second click **replaces** the previous suggestion rather than stacking;
- if you *edited* that suggestion, the new one is appended instead — silently rewriting your edit would be wrong.

Reuses `summarize_up_to`'s resolve + non-streaming collect path rather than a new provider seam.

## Testing
- 5 unit tests: menu contents/order, caption gating, command grammar (including that default selections are omitted, not emitted as literal tokens), and both impersonate insertion rules driven against fakes.
- **162 passed** across the composer and chip suites.
- Live tmux verification of every entry: menu opens from the `☰`, the image modal's preview updates per keystroke and its command lands in the composer, and the empty-transcript guard reports honestly.
- **Impersonate proven against a payload-logging stub provider** — the request carries the impersonate instruction as the system row followed by the real transcript, the suggestion lands in the composer, and a second click leaves exactly one copy. (Given this program's history, "the UI responded" isn't evidence the model got the right thing.)

---
## Summary by cubic
Adds a ☰ overflow menu to the console composer with Generate Image, Generate Caption, Narrate Entire Conversation, and Impersonate to keep the action row compact. Hardens Impersonate, adds clear failure reasons, fixes newline/tab/long-thread edge cases, and correctly gates captioning on actual image attachments; addresses Linear tasks 1680–1683.

- **New Features**
  - Overflow menu before Send (task-1680) with four actions; slight composer width bump.
  - Generate Image modal (task-1681): collects prompt/backend/style, live-previews the exact `/generate-image` command, and pastes it into the draft; omits default selections.
  - Generate Caption (task-1682): inserts a ready caption + alt-text prompt.
  - Narrate Entire Conversation: placeholder with an informational notice.
  - Impersonate (task-1683): drafts the user's next message; never sends or alters the transcript. Insertion rules: append on a new line; second click replaces its own prior suggestion; if the user edited it, a new suggestion is appended.
  - Tests: 8 tests (incl. 2 integration) cover menu flow, command grammar, caption gating, impersonate insertion rules, draft newline handling, and provider-payload shape.

- **Bug Fixes**
  - Impersonate returns reasoned outcomes and shows specific notices for provider-not-ready, empty-transcript, provider-error, and empty-completion; mirrors provider contract (drop failed rows; drop assistant turns before the first user turn; fold seeded greeting into the system row; ensure the array ends on a USER turn; trim long transcripts to the summarize budget and keep newest turns user-first).
  - Scope draft writes to the visible session’s composer to avoid cross-tab insertions; guard stale/closed session IDs during tab transitions; draft appends no longer add a double newline.
  - Generate Caption is gated on IMAGE attachments by reading real staged records (classifies image/other/none; mixed staging counts as captionable); disabled rows explain the blocking case.
  - Bound Generate Image preview height so long prompts don’t push action buttons off-screen.

<sup>Written for commit 8228fc5a7d6ab5616739402eee0c8bf1005fc34a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1160?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Both reviews addressed. **cubic caught two provider-contract bugs that would have made Impersonate fail on exactly the character chats this work has been building** — worth spelling out.

### cubic (4 new, 3 overlapping with Qodo)
**P1 assistant-first.** A seeded character greeting made the request open on an assistant row, which Anthropic and Gemini reject (task-427). The transcript now mirrors `_provider_message_payloads`' rules exactly — drop failed rows, drop assistant turns before the first user turn — and the greeting rides in the system row via `_seeded_greeting_text` (task-1531), the same idiom that fixed this once already.

**P1 user-final.** After a completed turn the array ended on an assistant row, which user-final providers reject. An impersonation request is appended when needed.

**P1 tab-switch race.** A suggestion could land in a *different* tab's visible draft. Both draft paths now only touch the mounted composer when it still shows that session — mirroring `_insert_console_dictation`'s guard for the identical race. I had read that code while building this and failed to apply its lesson.

**P2 unbounded transcript.** A long thread could blow the provider window; now trimmed to the same budget `summarize_up_to` uses, dropping OLD turns and never leaving an assistant row leading.

**P3 preview height** bounded — a long pasted prompt pushed Paste/Cancel off-screen.

### Qodo
Newline doubling, `KeyError` on stale session ids, collapsed "returned nothing" messaging, the missing `Raises:`, and the integration-test gap were all fixed in e8043ce4c. On **streaming**: I'm declining that one deliberately. `_collect_summary_completion` *does* consume the provider stream — it just presents the result at the end, exactly as `summarize_up_to` does. Rendering partial tokens into an editable composer would fight the user's cursor mid-edit, and the output here is a draft to review, not transcript content.

### Verification
Live payload after the fix: `['system', 'user', 'assistant', 'user']` — user-first after the system row, user-final, with the appended impersonation request as the last turn. Two new tests pin both contracts plus the failed-row drop and the greeting fold; another pins trimming (newest kept, never assistant-leading).

**163 passed** across composer, controller, and send suites; 10 in this module.

rmusser01:
Follow-up: **Generate Caption now requires an image specifically.** It previously enabled for any staged attachment, so a PDF would have produced an image-caption prompt. It reads the real staged records (`store.pending_attachments` → `mime_type`/`file_type`) rather than guessing from the composer chip label.

Blocked cases stay **disabled rather than hidden**, matching how the no-attachment case already behaved and the product's explicit-unavailable-state principle — the row remains visible and names the case ("Attach an image first" vs "Attached file is not an image"). Hiding it would leave no way to discover the action exists. Mixed staging (PDF + image) counts as captionable.

177 passed across composer, controller, and attachment suites.

PR #1164, opened 2026-08-01
Watchlists briefings phase 2b: audio synthesis, stitching, playback (spec #2)

rmusser01:
## Summary

Phase 2b completes spec #2's phase 2: a cast script becomes a single playable audio file in which **every speaker is voiced by their own TTS profile, on any configured provider**. With this, `task-1540`'s phase-2 acceptance criterion is met (2a shipped presets/casting/pickers/citations; 2b ships audio).

The all-providers scope was a deliberate owner decision. It costs real work, because the platform's one call that *guarantees* the requested voice was actually used refuses every provider except the local `audio_cpp` server. So this branch promotes a public per-selection legacy request builder out of a private helper, and does that response validation itself on the streaming path — non-empty, and decodes as a real WAV container — so a legacy-provider roster fails loudly rather than silently drifting to the app's single global voice.

- **Data:** `briefing_audio` (write-once per-speaker voice snapshot including each profile's `revision`, file path, duration, turn count, honest statuses) with allowlisted+validated CRUD and pagination.
- **Voice resolution** (`Subscriptions/briefing_voices.py`): a roster speaker's stored profile id becomes a concrete selection; a `get_profile(UUID)` passthrough was added to the profile service (the repository had one; the service never exposed it, so a stored id could previously only be resolved by paging).
- **Synthesis** (`Subscriptions/briefing_audio.py`): per-turn, per-voice, WAV-first; long turns split with a hard character cap and re-joined; `aclose()` in a `finally` that preserves the primary exception (verified against the real response type — `async with` demotes it); every failure names the speaker and the 0-based turn index.
- **Stitching** (`TTS/audio_stitch.py`): a general decode-and-concat primitive. The only prior multi-segment joiner concatenated *encoded bytes*, which for WAV yields a file whose header describes just the first segment. Guarded optional import, so a user without the audio extra loses the Synthesize button, not the Watchlists screen.
- **Storage:** buffered whole and written once via `atomic_private_write_bytes` into the private data dir — a correct decode-and-concat already holds the artifact in memory, and `private_paths` has no binary append (`open_private_binary` is read-only). A failure after the write removes the file.
- **UI:** Synthesize + Play/Stop, an audio row, and a three-state Audio column — note glyph for complete or in-flight, note plus a red ✗ for failed, blank only when nothing was attempted. Play is disabled when the file is missing, not left dead. Crash recovery runs on both the load and synthesize paths, each independently pinned.

Follow-ups filed: **task-1711** (pipeline row creation is not transactional; self-heals via the sweep) and **task-1712** (test-coverage residuals).

## Test plan

- 2,496 passing across `Tests/Subscriptions/`, `Tests/Watchlists/`, `Tests/TTS/`, plus 254 in `Tests/UI/ -k watchlist`; the one failure and five collection errors are pre-existing on dev (an `__all__` drift and the optional numpy dependency) and untouched here.
- Every behavioural change carries a revert-confirm-RED-restore mutation check, including the import-safety property that protects the screen, the WAV header assertion (a naive byte-join must fail it), both crash-recovery seams proven independent, and the app-survives-a-database-error guard.
- Eight-task subagent-driven build with per-task reviews (three fix rounds) and a whole-branch review on the strongest model, which found three composition defects no single task's gate could see — most importantly that ElevenLabs could never have produced audio: one requirement forced mp3 for that provider, another forced WAV, a third rejected non-WAV, each correct alone.
- CI now installs `pydub` (and its 3.13 shim) so these tests actually run on the gate — deliberately not the whole `audio` extra, which would drag a speech-transcription stack onto every push.

---
## Summary by cubic
Turns cast scripts into single playable audio files, voicing each speaker with their own TTS profile across all providers, and adds synthesize/playback to Watchlists. Completes spec #2 phase 2 (task-1540) by delivering phase 2b audio (task-1630).

- **New Features**
  - `briefing_audio` table + CRUD with voice snapshot, file path, duration, statuses; sweep marks interrupted rows failed.
  - Resolve roster `voice_profile_id` to concrete voices; added `TTS.profile_service.get_profile`.
  - Per-turn synthesis on exact (`audio_cpp`) and legacy providers; long turns split and stitched; ElevenLabs PCM wrapped to WAV; validates selection and outputs (non-empty, real WAV).
  - General WAV stitcher (decode-and-concat) for accurate headers and durations; atomic write to private dir with cleanup on failure.
  - UI: Synthesize + Play/Stop controls, audio status/duration in script detail, and a three-state Audio column; Play disabled when file missing.

- **Bug Fixes**
  - Preflight `pydub` check: `generate_script_audio` fails early with a clear row error if `pydub` is missing.
  - Playback path validation: Play now validates `file_path` against the briefings audio dir before any filesystem access; tampered paths are refused.
  - Exact-path mismatch messaging: snapshot validation now names the specific differing fields.
  - Optional-deps wiring: `audio_stitch` sources `pydub` availability from the project’s `optional_deps` check.

<sup>Written for commit 6c5009dd7e6a80eb8440cb753cd5d73e320b8425. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1164?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
All five findings adjudicated — four fixed in `6c5009dd7`, one declined with evidence.

**Fixed**

1. **No pydub preflight** (bug, and an expensive one). Without the optional audio extra, every turn was synthesized — real provider calls, real cost on paid providers, minutes of waiting — before stitching failed and wrote a failed row. There is now a preflight after the existing pre-flight refusals (so the no-orphan-row rule still holds) that records a failed row naming the missing dependency and its install command before any synthesis. The pinning test asserts the synthesis seam is called **zero** times, which is the whole point.
2. **Snapshot mismatch named the wrong field** (bug). `_validate_exact_snapshot` compares provider/model/voice/format/speed/options, but the error always claimed the *voice* differed — false and actively misleading whenever the real difference was the format or options. It now reports the fields that actually differ with requested-vs-returned values, still naming the speaker and turn index, and stays bounded (differing option keys, not whole mappings).
3. **`file_path` bypassed `path_validation`.** The UI now confirms the stored path resolves inside `briefing_audio_dir()` before any `.exists()` or playback; a path that fails validation is treated exactly like a missing file — Play stays disabled, no filesystem probe. Covered by traversal and absolute-outside-dir tests.
4. **pydub check now routes through `optional_deps`**, so there is one source of truth. The load-bearing property is unchanged and re-verified: importing `audio_stitch` with pydub absent still succeeds, which is what keeps a missing extra from costing the entire Watchlists screen (the screen registry swallows `ImportError`).

Each fix is mutation-verified (revert → RED → restore). 2,509 passing across `Tests/Subscriptions/`, `Tests/Watchlists/`, `Tests/TTS/`; 254 in `Tests/UI/ -k watchlist`; no new failures.

**Declined**

5. **`briefing_audio` DDL outside `migrations/`** — same disposition as phase 2a on #1145, and for the same verifiable reasons. `Subscriptions_DB` has never used the migrations directory: every one of its table definitions is inline additive `CREATE TABLE IF NOT EXISTS`, and `tldw_chatbook/DB/migrations/` contains zero subscriptions/watchlists files (it serves ChaChaNotes, as TTS and Scheduling serve their own). The design spec also rules the machinery out explicitly for this work: *"All tables are additive `CREATE TABLE IF NOT EXISTS` / column-presence `ALTER`s. No data migration exists in this design, so the TASK-1362 atomicity machinery is not needed; do not cargo-cult it in."* Adding a migration file here would make this the only database in the project with two competing schema mechanisms.

PR #1177, opened 2026-08-01
Watchlists briefings phase 3: markdown export and the podcast feed directory (spec #2)

rmusser01:
## Summary

Phase 3 of spec #2 gets artifacts **out** of the app. Two exports live on the Artifacts section's top toolbar, both writing wherever the user chooses:

- **Export** — the selected briefing as a Markdown file.
- **Export Feed** — a self-contained **podcast feed directory**: `feed.xml` plus a copy of every finished audio episode in the watchlist.

With this, `task-1540`'s phase-3 criterion is met; only phase 4 (scheduling) remains.

**Localhost serving is deliberately not built**, by the project owner's decision, because the spec's premise turned out to be false. `[web_server]` is textual-serve — a mutually exclusive process mode that serves the TUI *itself* to a browser (the app returns instead of running the TUI), whose only static route is hardcoded to textual-serve's own assets, and whose `enabled` key is read by no code at all. There was never a server alongside the TUI to toggle. The spec's own wording already said "the directory is the deliverable"; the user guide now shows the one-line `python3 -m http.server` recipe and states plainly that `[web_server]` is not that thing. `task-1760` scopes in-app serving as the net-new work it actually is, carrying the evidence so nobody re-discovers it.

Design decisions, each recorded beside the code rather than in a review artifact:
- **The destination is never routed through the private-storage helpers.** They chmod their target to `0o700` and refuse overwrites — correct for app-owned storage, wrong for the user's own folder.
- **Exported files are `0o644`, deliberately ignoring umask.** Private audio is written `0o600`, and inheriting that made a folder meant for sharing readable only by the exporting account — the first review round found exactly this, hidden by a fixture that seeded permissive sources.
- **The path-safety check runs before any filesystem access** on a stored path, the ordering phase 2b established for playback; export copies files *out*, so the same mistake here is strictly worse.
- **One bad episode is skipped, not fatal** — reasons are collected and the toast reports "N of M" rather than claiming success. Pagination walks every episode instead of silently truncating at 500.
- **Enclosure URLs are relative and percent-encoded**, so the folder survives copying, syncing or zipping, and a client that sends the raw request line still fetches.

## Test plan

- 758 passing across `Tests/Subscriptions/` and `Tests/Watchlists/` after rebasing onto current dev.
- Every behavioural change carries a revert-confirm-RED-restore mutation check, including the security ordering (its test observes the *absence of I/O*, not merely the absence of a copied file), the file-permission relaxation (fixture now seeds a genuinely `0o600` source), and the export walk's bound (removing it makes the test hang, demonstrating the production failure).
- Six-task subagent-driven build with per-task reviews (four fix rounds) and a whole-branch review on the strongest model, which found a composition defect no single task's gate could see: a filename sanitiser that correctly keeps spaces, reused to name episode files, emitted as an enclosure URL the URI standard forbids — unfetchable by some podcast clients and undiagnosable by the user.
- A follow-up round also replaced a tautological property test (`url == quote(unquote(url))` holds for *any* encoded output, so a double-encoding mutation stayed green) with the property a client actually depends on.

Follow-ups filed: `task-1760` (in-app serving) and `task-1761` (cap the unbounded offset-walk loops shared by four pagination test files).

---
## Summary by cubic
Adds Markdown export for selected briefings and a watchlist-wide podcast feed directory export (`feed.xml` + audio files), so users can save and share artifacts outside the app. Completes phase 3 of spec #2 for `task-1540`; phase 4 (scheduling) remains.

- **New Features**
  - Export selected complete briefing as `.md` via `FileSave`, with a safe default filename.
  - Export Feed writes `feed.xml` and copies finished episodes to a chosen folder; RSS is built purely, enclosure URLs are percent-encoded and relative, and the channel includes a `link`.
  - Pagination exports all episodes; skipped items are reported instead of failing the whole export.
  - UI guards one export at a time and surfaces clear toasts for cancel, partial, and errors.

- **Bug Fixes**
  - Path safety check runs before any I/O via `audio_file_path_is_safe` (moved to `Subscriptions.briefing_audio`).
  - Exported files use mode `0o644`; `feed.xml` is written atomically.
  - Feed generation rejects naive datetimes; timestamps must be timezone-aware.
  - Destination paths are validated; exports never route through private-storage writers.

<sup>Written for commit d4a37440d22e8b51d4d6be7904236982cea40060. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1177?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/User_Guide/watchlists.md
+++ b/Docs/User_Guide/watchlists.md
@@ -13,3 +13,42 @@ Mixed | Local/Server". This screen was previously called "Subscriptions."
 
 - Click **Watchlists** in the nav bar, or press **Ctrl+P** → "Switch to
   Watchlists". (Or press **Ctrl+6** from anywhere.)
+
+## Exporting briefings and podcast feeds
+
+The **Artifacts** section of a watchlist holds its briefings — text digests
+of what its sources did — along with any scripts cast from them and any
+audio synthesized from those scripts. Two export actions live on that
+section's top toolbar, and both write to a location you pick:
+
+- **Export** saves the selected briefing as a Markdown file. It is enabled
+  once a briefing has finished generating.
+- **Export Feed** writes a **podcast feed directory**: a `feed.xml` plus a
+  copy of every finished audio episode in the watchlist. It is enabled once
+  at least one episode exists.
+
+If some episodes cannot be exported — a file has been moved or deleted
+since it was synthesized — the rest are still written and the app tells you
+how many of how many succeeded, and why the others were skipped.
+
+### Listening to an exported feed
+
+The exported folder is self-contained: `feed.xml` refers to the audio files
+beside it by name, never by absolute path, so you can copy, sync, or zip the
+whole directory and it keeps working on another machine.
+
+Some podcast clients accept a local folder or a `file://` URL directly. For
+clients that require HTTP, serve the folder yourself — from a terminal, in
+the exported directory:
+
+```bash
+python3 -m http.server 8000
+```
+
+Then point the client at `http://localhost:8000/feed.xml` (or your machine's
+LAN address, if the client runs on another device). Stop the server with
+Ctrl+C when you are done.
+
+The app does not serve the folder for you. Its `[web_server]` setting is
+unrelated — that runs the whole chatbook UI in a browser instead of your
+terminal, and it is not a way to publish a feed.
--- a/Docs/User_Guide/watchlists.md
+++ b/Docs/User_Guide/watchlists.md
@@ -22,7 +22,9 @@ audio synthesized from those scripts. Two export actions live on that
 section's top toolbar, and both write to a location you pick:
 
 - **Export** saves the selected briefing as a Markdown file. It is enabled
-  once a briefing has finished generating.
+  once you have selected a briefing that has finished generating — a
+  finished briefing that isn't selected, or nothing selected at all, leaves
+  the button disabled.
 - **Export Feed** writes a **podcast feed directory**: a `feed.xml` plus a
   copy of every finished audio episode in the watchlist. It is enabled once
   at least one episode exists.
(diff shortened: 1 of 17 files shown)


rmusser01:
All three Qodo findings accepted and fixed in `2b287be7f` — the two symlink bugs are genuine catches that seven prior review rounds missed, both on the *destination* side of the write path while our attention had been on validating source paths.

1. **Symlink overwrite on export** (security). `shutil.copy2` follows a symlink at the destination filename, so a link planted at a predictable episode name in a shared/synced folder would redirect the write anywhere the user can write. Episode copies now open the destination themselves with `O_WRONLY|O_CREAT|O_TRUNC|O_NOFOLLOW` and stream the source in — the same posture `_write_feed_xml_atomically` already took, now applied uniformly. A side benefit: the mode is set once at creation via `fchmod`, retiring the post-copy chmod that Decision 5 previously needed. Pinned by a test that plants a hostile symlink and asserts the outside target is untouched and the episode is skipped with a reason; mutation back to `copy2` REDs it.
2. **Dangling partial wedges the directory** (reliability). The stale-cleanup used `Path.exists()`, which follows links and so reports False for a *dangling* `feed.xml.partial` symlink — leaving it in place for the `O_EXCL|O_NOFOLLOW` open to trip over forever after. Now `os.path.lexists`, which sees the directory entry itself; both dangling and live symlinks at that name are cleared. The strict open is unchanged — it was correct, and is what kept this a wedge rather than a write-through.
3. **Test docstrings.** Added to all 42 undocumented test functions across the five files this branch touched, each stating the property it pins rather than restating the name — this branch already had one test whose obvious phrasing was a tautology, so the docstring naming the real property is load-bearing.

Independently verified post-fix: 763 passed across `Tests/Subscriptions/` + `Tests/Watchlists/` (758 baseline + 5 new), zero failures, both mutations RED-then-restored.

PR #1183, opened 2026-08-02
Console character avatar: survive monochrome, and repaint when its section re-opens

rmusser01:
Investigated a report of "no character image at all in the Console" on Windows.

## What it actually was

**Not a bug.** The reporter had the **Character rail section collapsed**. Clicking the `Character ▸` header expands it and the portrait returns.

I traced the whole composition path first and found **no Windows-specific gate** — every condition defaults to visible. Ruled out along the way, so nobody re-treads it:

- **Pillow is a core dependency**, not an optional extra — it cannot be missing.
- Image bytes come from a **DB BLOB**, not a file path, so no Windows path handling is involved.
- The mosaic **renders correctly under Windows legacy-console colour** — measured from Rich's actually-emitted bytes, because `export_text()` re-emits truecolor regardless of the console's colour system and will tell you the opposite.

## Two real defects found while investigating

Neither is the reporter's cause. Both are genuine on their own merits, and the second sits directly on the path they are about to take.

**1. The avatar is invisible without colour** (`43d2cd55f`)

The mosaic encodes **100% of its information in background colour** — a flat cell is a *space* glyph styled `on rgb(...)`. Strip colour and the portrait does not degrade, it disappears. Textual converts the whole app to monochrome when `NO_COLOR` is set (`app.py` -> `filter.py`), so the box renders empty while every other element still works.

```
truecolor / 256 / 16 / windows-legacy -> 256 escapes -> visible
colour disabled                       ->   0 escapes -> INVISIBLE
```

Adds a monochrome path carrying luminance in Block Elements shade glyphs, wired from Textual's own `App.no_color`. **Colour terminals are untouched** — the quadrant path runs exactly as before.

**2. Re-opening the section leaves a 16-column portrait** (`4bc28d092`)

Two mechanisms interact:

- a collapsed body has `display: none`, so `_character_avatar_available_cols()` measures **0** and `character_avatar_box(0)` clamps to the 16-column **minimum**;
- `_refresh_active_character_avatar_if_scope_changed` early-returns while `(character_id, state)` is unchanged.

So an avatar first painted while collapsed stays pinned at 16 columns **forever** — the "~50-column rail showing a 16-column portrait" defect task-1661 fixed for a different trigger. Expanding would have shown the reporter a shrunken avatar and read as a second bug. Opening the section now clears the scope guard so the next sync tick re-measures.

## Testing

35 passing across `test_console_character_avatar` and `test_mosaic_render`, plus the personas and image-view suites.

Both fixes are mutation-verified: reverting the scope-guard clear fails its test, and the monochrome assertion is paired with a colour-path control so a change that broke colour rendering could not pass.

---
## Summary by cubic
Keeps the Console character avatar visible in monochrome terminals, preserves shape for dark portraits, and repaints to the correct width when the Character section is reopened.

- **Bug Fixes**
  - Monochrome rendering: carries luminance with Block Elements shade glyphs when `Textual` sets `App.no_color` (via `NO_COLOR`), so the avatar shows without color; the color path is unchanged.
  - Dark portraits: removes the blank bucket and normalizes luminance per image (`░▒▓█` ramp) so monochrome avatars never render empty or as a uniform block.
  - Reopen behavior: clears the avatar scope guard when the Character section opens, forcing a re-measure and repaint so portraits no longer stay stuck at 16 columns after rendering while collapsed.

<sup>Written for commit 1dc826a51312546915d3185a669db9e0ca7662af. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1183?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_console_character_avatar.py
+++ b/Tests/UI/test_console_character_avatar.py
@@ -687,3 +687,46 @@ async def test_available_cols_measures_the_section_not_the_holder(
     )
     assert measured != holder.content_size.width
 
+
+
+@pytest.mark.unit
+def test_expanding_the_character_section_reallows_a_rail_width_avatar():
+    """Re-opening the collapsed Character section must re-render the avatar.
+
+    A user reported "no character image in the Console" and turned out to
+    have the Character rail section COLLAPSED. Clicking it open is the fix
+    -- but two mechanisms interact badly on that path:
+
+    * a collapsed body has ``display: none``, so
+      ``_character_avatar_available_cols()`` measures 0 and
+      ``character_avatar_box(0)`` clamps to the 16-column MINIMUM;
+    * ``_refresh_active_character_avatar_if_scope_changed`` early-returns
+      while ``(character_id, state)`` is unchanged.
+
+    So an avatar first rendered while collapsed stays pinned at 16 columns
+    forever -- exactly the "~50-column rail showing a 16-column portrait"
+    defect task-1661 fixed for a different trigger. Toggling the section
+    open must invalidate the scope guard so the next sync re-measures.
+    """
+    from tldw_chatbook.UI.Screens.chat_screen import ChatScreen, character_avatar_box
+
+    # The sizing half of the trap, in isolation: a hidden body measures 0.
+    assert character_avatar_box(0) == (16, 8), "collapsed body clamps to minimum"
+    assert character_avatar_box(48)[0] > 16, "an open rail should size larger"
+
+    screen = ChatScreen.__new__(ChatScreen)
+    screen._last_console_avatar_scope = (7, "idle")
+    applied: list[tuple[str, bool]] = []
+    screen._set_console_rail_preference = lambda **kw: applied.append(("pref", True))
+    screen._apply_console_rail_section_open = lambda sid, o: applied.append((sid, o))
+    screen._current_console_rail_state = lambda: type(
+        "S", (), {"character_open": False}
+    )()
+
+    screen._toggle_console_rail_section("character")
+
+    assert ("character", True) in applied, "section did not open"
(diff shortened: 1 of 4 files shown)


rmusser01:
**Addressed in `1dc826a5` — and it was a real bug, confirmed by reproduction before fixing.**

You were right that `_SHADE_RAMP` leading with a space meant a dark portrait could map every cell to the darkest bucket and render invisible. Reproduced it first rather than taking it on faith: a charcoal subject on near-black produced a glyph set of exactly `[' ']` — completely blank, reproducing the very bug the monochrome path exists to fix. Character-card art on a dark theme is that case, so it was the common path, not an edge one.

Dropping the blank bucket (`░▒▓█`) fixed visibility but surfaced a second problem your suggestion wouldn't have caught: a dark portrait's whole luminance range sits *inside one bucket*, so it rendered as a uniform block of `░` — visible, but not a portrait.

So the fix is both halves: no blank bucket, **and** luminance normalised per image so the image's own range is stretched across the ramp.

```
before:  (entirely blank)     then:  ░░░░░░░░        now:  ░░▓▓▓▓░░
                                     ░░░░░░░░              ░▒████▒░
```

The regression test pins both properties — never blank, never shapeless — with a light-image control so a ramp that collapsed to a single glyph could not pass it.

PR #1187, opened 2026-08-02
Watchlists briefings phase 4: scheduled generation (spec #2 complete)

rmusser01:
## Summary

The final phase of spec #2: a watchlist with a briefing cadence gets its briefing written on schedule while the app is open. With this, **every acceptance criterion of the briefings/podcasts programme is met** (phases 1-3: PRs #1115, #1145, #1164, #1177).

The pre-plan survey found the spec's mechanisms fictional and a live hazard, so this branch corrects rather than obeys:
- **In-process generation claims** (briefings/scripts/audio). Every crash-recovery sweep previously gated on *screen* state, so a scheduled run would have been falsified by merely opening the Artifacts tab, and duplicated by pressing Generate (two LLM calls, two watermark writes). Claims are module-level, die with the process (crash zombies still look like crash zombies), and the sweeps take an explicit exclude snapshot — no locks, no schema, no heartbeat protocol.
- **Cadence lives on `watchlists`** (`briefing_cadence_seconds`, opt-in, Off default — a schedule spends the user's LLM tokens unattended). The spec named `automation_definitions`, a table with zero production readers in a different database.
- **The `briefings` table is the run record** — its statuses (`generating/complete/empty/failed` + `interrupted`) are richer than the runs table's, which structurally cannot hold a watchlist-scoped run.
- **Fire-and-forget handler**: the scheduler tick awaits handlers serially inline, so a multi-minute LLM call is spawned as an owned task (strong reference held) rather than stalling every check. Re-emitted jobs are refused by the claim. The stored default preset is honoured, same as the Generate button.
- **Failures never advance the schedule** (two-sided pact with the coverage watermark's allowlist) — and, per the whole-branch review's find, retries are **attempt-aware**: a failing daily schedule retries one cadence period after the failure, not every queue reload (which would have been ~48 unattended LLM calls/day, each contract individually correct).
- **The "fires while the app is open" copy was written, not mirrored** — it existed nowhere. It now lives in the scope label (which previously claimed "on request" — a lie the moment a schedule exists) and the user guide.

Also rides this branch: **task-1780** (owner-directed: persist briefings/scripts into ChaChaNotes for permanence and later re-casting), follow-ups 1810/1811/1812, and a governance observation recorded in the spec: the ADR-029 persist_event boundary has already drifted to eight names via the dictation stream — flagged for the ADR owner, not fixed here.

## Test plan

- 826 passing across `Tests/Subscriptions/`, `Tests/Scheduling/`, and the artifacts-pane suite after rebasing onto current dev; a real-app boot test covers the wiring.
- Every behavioural change mutation-verified, including: the claim-aware sweeps (survey findings a/b/c each pinned by a real-claim test), the fire-and-forget non-blocking property (inline-await mutation hangs the suite — proof by timeout), the thread-identity pins on every off-loop read, the attempt-aware retry (the one pre-existing test whose assertion *was* the bug is replaced, with its other property confirmed still carried DB-side), and an out-of-catalog cadence fallback that mutation proved load-bearing against a pane crash.
- Five-task subagent-driven build; per-task reviews (two fix rounds), whole-branch review on the strongest model — ninth consecutive programme where it found the decisive between-tasks defect (the uncapped retry composition).

**Disclosure:** during Task 4 an implementer ran a non-pytest probe that opened the live subscriptions DB and applied the additive nullable `briefing_cadence_seconds` column (zero rows touched; the merged feature adds the same column at first boot, so post-merge state is identical). Incident recorded in the ledger; the account was volunteered by the agent itself.

---
## Summary by cubic
Scheduled briefings now run per watchlist while the app is open. This adds a non-blocking scheduler path with in-process claims to prevent duplicate runs and races.

- **New Features**
  - Added `watchlists.briefing_cadence_seconds` (opt-in; Off by default) and a cadence picker in the Artifacts toolbar.
  - Wired `BriefingProjection` and `BriefingJobHandler` into the scheduler; the handler spawns generation (non-blocking) and uses claim checks to avoid duplicates.
  - In-process claims added across briefing, cast, and audio; sweeps exclude active claims, and Cast now refuses a second in-flight press with a clear message.
  - Failures never advance the schedule; success or “empty” does. New `[scheduling] briefing_schedules_enabled` gate (defaults to true). User guide updated to note it only runs while the app is open. Expanded tests across Scheduling, Subscriptions, and UI.

- **Bug Fixes**
  - Failed scheduled runs now retry one cadence period after the failure, not on every queue reload.
  - A Generate press that races a scheduled claim shows an “already in flight” refusal instead of a generic DB error.
  - Moved a preset read off the event loop to avoid tick stalls; guarded `set_watchlist_briefing_settings` to require a real `int` cadence and wrapped two read paths in `transaction()` for correctness.

<sup>Written for commit 7e2123b1bc8711adf3c6ff43b61bd87edd878aca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1187?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
All five Qodo findings adjudicated — four fixed in `68517606d`, one declined.

**Fixed:** (1) **Cadence type guard** — good catch: `bool` subclasses `int`, so `True` passed the positivity check and stored a one-second cadence, re-opening the runaway-schedule scenario through the type system. Now `isinstance(int) and not bool`, with `True`/`"3600"`/`1.5` each pinned to raise `ValueError`. (2) Both flagged reads (`_read_watchlist_briefing_settings`, `_default_preset_id`) now run inside `transaction()`, matching every other accessor this stream has shipped; thread-identity pins stayed green. (3) `BriefingProjection.__init__` typed. (4) The undocumented projection method got its Google-style docstring.

**Declined:** ALTER TABLE outside `migrations/` — same disposition as #1145 and #1177, same evidence: `Subscriptions_DB` has never used the migrations directory (every table inline, additive, `IF NOT EXISTS`; zero subscriptions migration files exist), and the design spec forbids introducing that machinery here. 727 + 102 green after fixes, mutations verified.

PR #1193, opened 2026-08-02
Console composer ☰ menu: click outside to dismiss

rmusser01:
## What

The ☰ composer-actions menu (`ConsoleComposerMenuModal`) previously offered only Escape as a no-action exit. Mouse-first users expect a click outside a popup to close it — now it does.

A screen-level `Click` handler tests the click's absolute coordinates against the menu box's region:

- Clicks anywhere **inside** the box (rows, header, padding, disabled-reason lines) do nothing — containment is tested against the box region, not the clicked widget's identity, so every interior surface behaves the same.
- A **backdrop** click dismisses with `None`, the same result the Escape path returns, so the screen-side callback dispatch needs no changes.

## Verification

- New integration test `test_menu_dismisses_on_backdrop_click_but_not_on_inside_click` pins both behaviors; written first and confirmed failing before the fix. Menu suite 32/32.
- Adjacent suites (collapse, workbench contract, internals decomposition): 253 passed. The two `test_console_workbench_contract.py` approvals/inspector failures are **pre-existing on clean `origin/dev`** (fallout from PR #1192's approval-flow rework — the decision control is now a `Select` and those contract tests expect the old focus behavior); verified identical on a pristine dev checkout, untouched by this diff.
- Live-verified in the real TUI (scratch config profile): inside-click keeps the menu open, backdrop click closes it with the composer intact, the menu reopens cleanly, and Escape still works.

---
## Summary by cubic
Adds click-outside to dismiss the console composer menu. Backdrop clicks close the menu; inside or coordinate-less clicks keep it open.

- **New Features**
  - Screen-level `Click` handler uses region containment, ignores clicks without coordinates, and dismisses with `None` (same as Escape).
  - Integration test `test_menu_dismisses_on_backdrop_click_but_not_on_inside_click` covers inside and coordinate-less click behavior.

<sup>Written for commit e7cbcb11fa07b131af5a4cfd625decdf87356449. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1193?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Both Qodo findings addressed in `e7cbcb11f`:

- **Unguarded click coordinates (bug):** confirmed real — this repo's other hit-testing paths guard `screen_x/screen_y is None` for synthesized clicks. The guard now returns early, failing toward keeping the menu open (a wrong dismissal would eat a click aimed at a menu row; a kept-open menu still has Escape and the backdrop as exits). Covered by a direct `on_click` invocation with `None` coordinates in the integration test.
- **`_Host` not PascalCase (rule):** renamed the new test's host class to `BackdropHost`. (The identical pre-existing `_Host` classes elsewhere in this file are out of scope.)

Menu suite 32/32 post-fix.

PR #1197, opened 2026-08-02
feat: persist briefings/scripts into ChaChaNotes (kept briefings, task-1780)

rmusser01:
## Why

Briefings and scripts live in the subscriptions DB with their lifecycle chained to the watchlist (`ON DELETE CASCADE`) — deleting a watchlist silently destroys every artifact it produced. Right for working data, wrong for content the user made. This adds **kept copies** in ChaChaNotes that survive watchlist deletion and can be **re-cast later with whatever presets exist at that time** (task-1780, owner-directed).

Owner decisions: dedicated artifact tables in ChaChaNotes (not Notes); **auto-keep for scheduled** generations, explicit **Keep** for manual; **briefings + scripts only** (audio deferred — files already survive on disk; pointer decision recorded as open).

## What

- **Schema v28 → v29** (`chachanotes_v28_to_v29_kept_briefings.sql`): `kept_briefings` (idempotency key `source_briefing_id UNIQUE NOT NULL`, all provenance denormalized — every row self-interpreting with the subscriptions DB gone) + `kept_scripts` (FK cascade inside ChaChaNotes; `source_script_id UNIQUE` nullable for scripts cast directly from a kept briefing). **No sync columns** — deliberate, recorded v1 divergence; hard delete. Collision-checked against origin/dev at push time (dev is at 28, no competing migration).
- **Keep service** (`Subscriptions/briefing_keep.py`): additive-idempotent — re-keeping adds scripts missing by source id, never duplicates, never overwrites; refuses non-`complete`/empty-body briefings before any row. A raced keep (scheduled auto-keep and a manual press landing together) folds the loser's `ConflictError` into the ordinary re-keep path.
- **Auto-keep from the scheduler**: `BriefingJobHandler` takes a lazy `chachanotes_db_getter` resolved fresh on each completion (the handler is constructed before `app.chachanotes_db` exists — a frozen param would be permanently inert; liveness is pinned by a real-boot test plus mutations that red both the construction-freeze and eager-capture variants). Keep failures log type-only and never fail the generation.
- **Cast-from-kept** (`briefing_cast.generate_script_from_text`): reuses the existing prompt/parse pieces against the kept body; own claim set keyed by kept-briefing id (a different id space from live briefing ids); `preset_id=None` casts a single-speaker "Narrator" narration on the app default. On failure: no row, error surfaced to the caller (kept_scripts has no status column — asymmetry documented).
- **UI**: Keep button on the Artifacts toolbar (screen-side in-flight guard with refusal toast); Kept-briefings modal (list/detail/hard-delete-with-confirm/preset picker mirroring the "App default (single narrator)" copy/Cast) — reachable regardless of watchlist scope, including after the watchlist is gone.
- **Whole-branch review fix wave**: the delete worker was the one unguarded worker on the branch (`exit_on_error=True` would exit the whole app on a busy DB, with auto-keep now writing ChaChaNotes concurrently) — now guarded like its siblings; refusal toasts on double-press; cast↔delete mutual exclusion pinned both ways; modal lists fetch limit+1 with an honest overflow line; pagination loop test added.

## Testing

- Named invariants: **kept-rows-survive-watchlist-deletion** (AC #3), **re-cast-needs-no-subscriptions-rows** (AC #4 — watchlist AND original preset deleted), additive idempotency both directions, auto-keep-skips-empty, keep-failure-never-fails-generation.
- Suite: 1693 passed / 0 failed at branch review; fix-wave sweep 832 passed / 1 known rotating-victim flake (passed 2/2 in isolation). Every behavioural change mutation-tested (Edit-revert → RED → restore), independently re-verified in the scoped re-review.
- Follow-up filed: task-1870 (kept tables excluded from ChaChaNotes sync + chatbook export — the recorded v1 gap).

---
## Summary by cubic
Adds kept copies of briefings and scripts in ChaChaNotes that survive watchlist deletion and can be re-cast later with current presets (task-1780). Also adds auto-keep for scheduled runs and a UI to view, delete, and recast kept briefings.

- **New Features**
  - ChaChaNotes schema v29: `kept_briefings` and `kept_scripts` with denormalized provenance, `source_briefing_id`/`source_script_id` uniqueness, and hard delete (no sync columns; follow-up in task-1870).
  - Keep service: additive and idempotent copies of complete briefings and scripts; refuses non-complete or empty bodies; handles concurrent keep races cleanly (briefing and per-script conflicts folded into “already kept”).
  - Scheduler auto-keep: scheduled completions mirrored into ChaChaNotes via a lazy `chachanotes_db_getter`; keep failures are logged and never fail the generation; wiring now resolves the DB at keep time so it’s live in production.
  - Re-cast from kept: `generate_script_from_text` writes new scripts to `kept_scripts`, uses a separate claim space, and supports `preset_id=None` for a single “Narrator” on the app default provider.
  - UI: Keep button on the Artifacts toolbar, plus a scope-independent Kept Briefings modal (list, detail, hard delete with confirm, and cast).

- **Bug Fixes**
  - Guarded delete worker in the modal; refusal toasts on double-press; cast/delete mutual exclusion; kept lists render “…and more not shown” when truncated.
  - Tests updated to assert the current schema dynamically (bump to v29) and add pagination coverage; SQL allowlist includes `kept_briefings`/`kept_scripts`.

<sup>Written for commit 4f6692c3f1ec0e12c72e6499c34b8e0701499b6b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1197?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Qodo adjudication:

**Finding 2 (script keep race) — confirmed real, fixed in `aecade7dd`.** The briefing-row race was already folded (`ConflictError` → re-keep), but the script inserts still raced between the `kept_script_source_ids()` snapshot and the insert, so the losing keeper failed outright despite everything being kept — exactly the failure the docstring promised couldn't happen. Now folded identically: narrow `except ConflictError` → treated as already-kept, not counted in `scripts_added`, debug-logged type-only. Pinned by a deterministic race test (stale-snapshot seam), and mutation-checked: reverting the except REDs the test.

**Finding 1 (reads bypass `transaction()`) — declined, house precedent.** This file's established convention is writes under `transaction()`, single-statement reads via bare `execute_query()` — `get_character_card_by_id` (:4827), `list_character_cards` (:5052), `get_conversation_by_id` (:6598), `list_all_active_conversations` (:6536), and ~60 other read sites do exactly what the new kept getters do. The kept CRUD follows the file's own split precisely (all five write paths are transaction-wrapped); wrapping only the new reads would make them the odd ones out without changing behavior — each is a single SELECT, atomic in SQLite as-is.

PR #1200, opened 2026-08-02
fix: row-scope the script and audio interrupted sweeps (task-1890)

rmusser01:
## Why

`fail_interrupted_scripts` and `fail_interrupted_audio` still excluded by their coarse claim key (briefing id / script id) — the same bug class task-1812 fixed for the briefings sweep in #1198. A crash-zombie `generating` row sharing a key with a freshly-claimed live row was shielded by the live claim, and task-1811's Synthesize blocking toast gave that a user-visible surface: it could name the dead zombie as the thing "already being synthesized."

## What

The merged briefings shape, mirrored faithfully into both siblings (`briefing_cast.py`, `briefing_audio.py`): a row-id registry populated the moment each INSERT's `to_thread` hop returns and cleared in the same `finally` as its claim; frozen-snapshot accessors; sweep SQL scoped to `id NOT IN (live row ids)` plus a pending-key guard (`claims − recorded ids`) that closes the unrecorded-claim window — the same window the briefings fix was found to open, handled here from the start. All four screen call sites pass the new exclude pair; four monkeypatch stubs widened (the stale-signature-TypeError-under-bare-except trap from #1198's wave, greped for exhaustively this time). The Synthesize toast now names only the genuinely live row once the claim's row id is recorded — the AC and docstrings state the pre-recording window's deliberate whole-script sparing plainly rather than over-claiming "never" (review finding, probe-confirmed).

## Testing

Coexistence tests both directions for each sweep (zombie failed as interrupted, live row untouched), pending-window tests for each, and a screen-level test proving the zombie is swept before the toast is composed (asserted `failed`/`interrupted` AND absent, `blocking` re-read post-sweep — not absent by coincidence). Four mutations each REDding their discriminating test (key-scoped revert → coexistence REDs; pending-guard drop → window test REDs). Full `Tests/Subscriptions/ Tests/Watchlists/ Tests/Scheduling/`: **1164 passed, 0 failed**; reviewer's independent run 900/0 on the two main directories. Opus whole-branch review: CLEAN (one Low copy-accuracy finding, fixed).

---
## Summary by cubic
Row-scoped the script and audio interrupted sweeps so only real in‑process rows are spared and same‑briefing/script crash zombies are swept. Closed the pre‑recording claim window and fixed the Synthesize blocking toast to only name the live row. (Task-1890)

- **Bug Fixes**
  - `fail_interrupted_scripts` and `fail_interrupted_audio` now exclude by row id (`id`) and accept `exclude_briefings`/`exclude_scripts` to guard the unrecorded-claim window.
  - Added row-id registries/accessors: `active_cast_claim_row_ids()`, `active_audio_claim_row_ids()`; pending guards: `pending_cast_claim_briefing_ids()`, `pending_audio_claim_script_ids()`.
  - Updated `watchlists_collections_screen.py` to snapshot and pass both sets at all sweep call sites (Artifacts load and per-action).
  - Synthesize toast now names only the genuinely live audio row; zombies are swept before the toast is composed.
  - Tests cover coexistence (live vs zombie), the pending window, and the screen-level toast behavior.

<sup>Written for commit 2668d636be40297a002abfcd5f83e850729e91f9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1200?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Subscriptions/test_briefing_audio_pipeline.py
+++ b/Tests/Subscriptions/test_briefing_audio_pipeline.py
@@ -47,10 +47,12 @@ from tldw_chatbook.Subscriptions.briefing_audio import (
     STATUS_FAILED,
     STATUS_GENERATING,
     AudioGenerationError,
+    active_audio_claim_row_ids,
     active_audio_claims,
     briefing_audio_dir,
     fail_interrupted_audio,
     generate_script_audio,
+    pending_audio_claim_script_ids,
 )
 from tldw_chatbook.Subscriptions.briefing_audio import TurnSynthesisError
 from tldw_chatbook.Subscriptions.briefing_service import GenerationInFlightError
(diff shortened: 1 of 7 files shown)


rmusser01:
Qodo adjudication: no bugs found; the docstring rule nit is addressed — one-line docstrings added to the four snapshot-default tests.

PR #1202, opened 2026-08-02
Console: retire top-strip Save Chatbook; RAG chip opens a Library RAG settings modal

rmusser01:
## What

Two user-requested Console control changes, plus the investigation that motivated the second.

### 1. Save Chatbook leaves the top control strip

The `New tab · Settings · Attach context · Run Library RAG · Help` strip no longer carries Save Chatbook: it sat disabled at rest for most sessions, spending an always-visible slot on an almost-always-inert control. Removed end-to-end (workbench action, control-bar registry, dead dispatch arm), not just hidden. Its two surviving surfaces — the ☰ composer menu row and the Inspector's Artifacts row — each carry availability copy and their own temporary-chat block, and both are pinned by tests.

### 2. "RAG: off" is now an action chip that opens a Library RAG settings modal

**Investigation first** (the user hit "Type a Library RAG query before running retrieval" with text sitting in the composer): the top-bar Run Library RAG action reads `_console_library_rag_query`, which is fed by exactly one widget — a small input at the bottom of the right Inspector rail, hidden whenever the rail is collapsed or evidence is staged. The composer draft was never consulted, so the always-visible button depended on a usually-invisible input. Also: "RAG: on/off" was never a setting — it reports whether staged live-work evidence came from a RAG source.

**The chip fix:** `RAG: off` follows the established Provider/Model/Character action-chip contract (Enter/Space/click) and opens a new `ConsoleRagSettingsModal` that:

- states honestly what "on" means (retrieved Library evidence staged for the next send; shows the staged item's title when on),
- owns the retrieval query — prefilled from the query set through any RAG surface, else **from the composer draft**,
- shows the retrieval scope, and
- runs retrieval through the existing `_run_console_library_rag_from_visible_action` seam (Enter or the primary button; Esc/backdrop close without changes, matching the ☰ menu's backdrop contract including the coordinate-less-click guard).

The modal's query mirrors into the Inspector readiness-card input when mounted, so the two surfaces cannot disagree.

Deliberately NOT changed without a follow-up decision: the top-bar button's own empty-query toast (it could instead fall back to the composer draft, or open this modal).

## Verification

- 14 new/updated tests: modal gating/result/backdrop/Enter-submit, chip activation contract + screen `@on` subscription, control-strip absence pins, workbench-state action tuple, gate15 contract updated to the new Save surfaces.
- Touched suites green post-rebase. The only failures anywhere are the two `test_console_workbench_contract.py` approvals tests broken on `origin/dev` by PR #1192 (verified identical on a pristine dev checkout; untouched by this diff).
- Live-verified in the real TUI: strip without Save Chatbook; chip click opens the modal prefilled with the composer draft; running flips the strip to `RAG: on · Sources: 1 staged` with the staged card; reopening shows the on-state copy. (Note: activating buttons *inside* modals via tmux-synthesized SGR clicks silently no-ops — the shipped first-run wizard shows the same behavior — so the button path is verified via real pilot clicks and the Enter path live.)

---
## Summary by cubic
Removed Save Chatbook from the top control strip and turned the RAG chip into an action that opens a Library RAG settings modal. The Run Library RAG button now falls back to the composer draft when no query is set to keep retrieval easy and consistent.

- **New Features**
  - RAG chip opens a modal to edit the retrieval query and run Library RAG.
  - Prefills from existing RAG state, else the composer draft; mirrors into the Inspector input and gates Run.
  - Shows scope and honest on/off copy; Enter runs; Esc/backdrop dismiss.
  - Visible Run falls back to the composer draft when no query is set; stores it so all surfaces agree; explicit queries still win; the empty warning only shows when both are blank.

- **Refactors**
  - Removed Save Chatbook from the top strip; use the ☰ composer menu or Inspector Artifacts.
  - Deleted the `save-chatbook` workbench action and its dispatch path.
  - Added `ConsoleRagChip` and a screen handler; updated control bar and tests.
  - Added docstrings to `ConsoleRagSettingsModal` public methods.

<sup>Written for commit 8e970284e77688636feca73a7906c16ce8350136. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1202?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_console_composer_menu.py
+++ b/Tests/UI/test_console_composer_menu.py
@@ -1065,10 +1065,6 @@ def test_artifact_actions_are_disabled_with_a_reason_in_a_temporary_chat():
     proves nothing unless the same call proves it is enabled otherwise.
     """
     from tldw_chatbook.Chat.console_ephemeral import blocked_reason
-    from tldw_chatbook.Widgets.Console.console_workbench_state import (
-        build_console_workbench_state,
-    )
-    from tldw_chatbook.Chat.console_display_state import ConsoleControlState
 
     menu = {
         e.action_id: e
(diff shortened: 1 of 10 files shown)


rmusser01:
Review round closed:

- **Qodo's one finding (missing docstrings on the modal's public methods):** fixed in `8e970284e`.
- Two commits landed after Qodo's pass and it did not re-fire within the wait window, so the internal record stands for them: `e42610b41` (Run Library RAG falls back to the composer draft — owner decision on this PR; three-branch unit coverage: fallback used+stored, explicit query wins, empty-everything toast survives; live-verified draft-only click → `RAG: on · Sources: 1 staged`, no toast) and the docstring commit above.
- Rebased onto dev `56e597ff8` earlier; the only conflict was the boot-regenerated CSS bundle, resolved by restoring dev's exact copy (this branch has no CSS source changes).

PR #1205, opened 2026-08-02
feat: kept briefings/scripts in chatbook export + recorded sync exclusion (task-1870)

rmusser01:
## Why

Task-1780's kept briefings/scripts were invisible to both systems users rely on for taking data with them: chatbook export had no awareness of the tables, and sync exclusion — a deliberate v1 ruling — was recorded only in the design spec's schema section. Task-1870's mandate: make the gap stop being silent, on both sides.

## What

- **Chatbook export (AC #1, implemented):** new `ContentType.KEPT_BRIEFING`; the creator walks selected kept briefings with each briefing's scripts nested in its payload (children, not independently selectable) plus a human-readable companion markdown file per briefing; real checkboxes in both creation surfaces; server-mode export refuses cleanly via the existing positive allowlist.
- **Chatbook import:** import-when-free on the UNIQUE `source_briefing_id` — byte-identical incoming rows skip silently (the re-keep ethos), differing content on the same source id is a named conflict where the whole incoming item (parent **and** scripts) is refused untouched, re-import is idempotent, and NULL-source scripts dedup by content multiset. The house conflict-resolver machinery was deliberately not extended: its rename/ask semantics fit display-name-keyed entities, not id-keyed idempotent artifacts (rationale in the report). Bundles predating the type import clean.
- **Sync (AC #2, recorded exclusion):** extends the owner's task-1780 "no sync columns" ruling; the dated decision block is appended to the design spec and reflected in the chatbooks guide (AC #3).
- **Whole-branch review fix wave (1 High + 4):** the conflict branch was grafting the incoming bundle's scripts onto the unrelated local briefing while the summary claimed "not modified" — probe-proven, now refused as a unit; the import wizard's stats check omitted `total_kept_briefings` (and `total_prompts`, a pre-existing live false-mismatch verified against the creator's `content_items` walk); per-briefing script counts now come from one grouped-COUNT seam instead of materializing transcripts on the UI thread; `kept_at` round-trips faithfully via an optional CRUD param (no schema change) instead of being silently re-stamped; a mid-scripts failure no longer falsifies the briefing counter.

## Testing

Round-trip into a fresh ChaChaNotes (all provenance columns, tz-faithful datetimes, `kept_at` seeded >1s in the past so equality cannot false-pass), re-import idempotency, conflict-refuses-parent-and-scripts, scripts-follow-parent, backward-compat bundle import, kept-only wizard validation. **324 passed, 1 pre-existing skip** at re-review; mutations REDded per fix (the F1 mutation reproduced the reviewer's exact probe output). Opus whole-branch review → fix wave → scoped re-review APPROVED with zero new defects.

---
## Summary by cubic
Adds chatbook export/import for kept briefings with their scripts via `ContentType.KEPT_BRIEFING`, so kept content is portable and round-trips cleanly. Records that `kept_briefings`/`kept_scripts` remain excluded from `ChaChaNotes` sync per task-1870.

- **New Features**
  - Export kept briefings as JSON plus Markdown; scripts travel nested inside the briefing.
  - Selection UI adds a “Kept Briefings” category with checkboxes.
  - Import is idempotent on `source_briefing_id`: identical rows skip; conflicting content refuses the briefing and its scripts; NULL-source scripts dedup by content.
  - Manifest, guide, and spec updated; backward compat when `total_kept_briefings` is missing.

- **Bug Fixes**
  - Export pages kept-script collection so large briefings don’t truncate scripts.
  - Honest stats and counters: wizard sums include kept briefings and prompts; creation UI uses grouped COUNT; per-briefing counters stay correct on partial failures.
  - `kept_at` preserves the originating device’s timestamp and is excluded from conflict matching; static, parameterized INSERTs set `kept_at` via `COALESCE(?, CURRENT_TIMESTAMP)`.

<sup>Written for commit c0a9eaf6d795d8bbe339e670133e4945a0cda7bc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1205?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/specs/2026-08-01-kept-briefings-design.md
+++ b/Docs/superpowers/specs/2026-08-01-kept-briefings-design.md
@@ -126,3 +126,54 @@ all checked. Four things worth recording for anyone re-reading this design later
   same bug shape the wider program has caught more than once: a `finally`/recompose path that
   rebuilds the UI from scratch will silently discard any state set moments earlier unless that
   state is threaded through `compose()` rather than assumed to survive it.
+
+## Follow-up decision (2026-08-02, task-1870): sync excluded, chatbook export included
+
+Task-1870 was filed at this spec's close-out (see "Non-goals (v1)" above) to close the silent gap
+on both of this design's deferred fronts — ChaChaNotes sync coverage and chatbook-export coverage
+of `kept_briefings`/`kept_scripts`. Both are now resolved, asymmetrically:
+
+- **ChaChaNotes sync: still excluded, by extension of this spec's original ruling.** The "Schema"
+  section above already decided `kept_briefings`/`kept_scripts` carry no sync columns
+  (`client_id`/`version`/`deleted`) and do not participate in ChaChaNotes's bidirectional sync
+  machinery. Task-1870 does not revisit that call — it stays a deliberate v1 decision, not an
+  oversight. The reason it holds up under the "should a follow-up close this?" question: the sync
+  engine's unit of replication is a row with a `client_id`/`version`/tombstone lineage, and kept
+  rows were designed from the start (this spec's "Schema" section) to be self-interpreting,
+  hard-deleted, denormalized copies with no such lineage. Retrofitting sync columns would be a
+  schema change with its own migration and conflict-resolution design, not a small follow-up: if a
+  future task wants cross-device sync of kept content, it should treat that as a new schema
+  proposal against this spec, not an extension of task-1870.
+- **Chatbook export: now included.** `ContentType.KEPT_BRIEFING` (`Chatbooks/chatbook_models.py`)
+  is a new chatbook content type. `ChatbookCreator._collect_kept_briefings`
+  (`Chatbooks/chatbook_creator.py`) walks a user's selected kept briefings, and — mirroring how a
+  conversation's messages are nested inside the conversation's own exported JSON rather than being
+  independently selectable — nests each briefing's kept scripts inside its own payload. Every
+  provenance column denormalized onto the kept rows by this spec's original schema (source ids,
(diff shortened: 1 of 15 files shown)


rmusser01:
Qodo adjudication:

**Finding 1 (kept-script export cap) — confirmed, fixed in `a10b310ea`.** The export now paginates until exhausted (constant kept as page size); pinned with a shrunk-page-size test (page=2, 5 scripts, all 5 in the bundle and round-tripped) and a mutation check (single-fetch revert reds it). Residual noted in code: the importer's NULL-source dedup lookup reads a single page deliberately (dedup, not enumeration) — at >1000 NULL-source scripts under one briefing a re-import could duplicate rather than lose data; documented at the constant.

**Finding 2 (dynamic INSERT) — confirmed, fixed in `a10b310ea`.** Both creates are back to one static parameterized statement each; the optional `kept_at` rides `COALESCE(?, CURRENT_TIMESTAMP)` so the not-provided path keeps DEFAULT-equivalent behavior (verified live, not just non-null).

**Finding 3 (kept_at not compared) — declined as intended behavior, now documented.** `kept_at` is provenance of the *keeping*, not content: the same artifact kept at different moments on two devices must skip as already-present. Comparing it would turn nearly every cross-device re-import into false conflict spam. Both comparison helpers now state the exclusion in their docstrings.

PR #1207, opened 2026-08-02
test: cap the four offset-walk pagination loops (task-1761)

rmusser01:
## Why

Four briefings pagination tests walked offset pages with an uncapped `while True` whose only exit is an empty page. If a query regressed to ignoring `offset`, the same non-empty page would return forever and the test would spin to pytest's 300-second global timeout instead of failing fast with a message pointing at the bug (task-1761, filed at the phase-3 close-out).

## What

Each walk is now a bounded `for page_number in range(max_pages)` with a `for/else` that `pytest.fail`s naming the query: "`<query>` never returned an empty page within 10 pages — offset is likely being ignored". Cap is 10 pages against 3 real ones, commented at each site. Test-only; no production code touched. Scanned `Tests/Subscriptions/` for new siblings of the shape — none beyond the four named (later pagination tests from tasks 1812/1890/1870 are bounded by construction).

## Testing

All 48 tests in the three files green. Mutation-verified: forcing `offset=0` in one walk fails in **0.29s** with the named message (previously that same regression spun to the 300s timeout).

---
## Summary by cubic
Capped the four offset-walk loops in briefing pagination tests to fail fast if `offset` is ignored, replacing `while True` with bounded `for/else` and a named `pytest.fail`.
The cap is now derived from seeded data (pages of seeded rows + the empty terminator) instead of a fixed 10 pages, keeping tests in sync with reseeds; test-only change with normal behavior unchanged. Addresses Linear task-1761.

<sup>Written for commit b59ba95c7446513be4227ffca872d6ddb6668fa5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1207?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Subscriptions/test_briefing_audio_db.py
+++ b/Tests/Subscriptions/test_briefing_audio_db.py
@@ -209,10 +209,11 @@ def test_list_briefing_audio_offset_pages_through_every_row_without_gaps_or_repe
 
     limit = 3
     seen: list[int] = []
-    # 7 rows / limit 3 -> 3 pages + the empty terminator; cap far above so
-    # an offset-ignoring regression fails fast instead of spinning to the
-    # global timeout (task-1761).
-    max_pages = 10
+    # Pages of seeded data + the empty terminator, derived from the data so
+    # the cap can never lag a future reseeding; exceeding it means offset is
+    # genuinely not advancing, and the walk fails fast instead of spinning
+    # to the global timeout (task-1761).
+    max_pages = len(expected_newest_first) // limit + 2
     for page_number in range(max_pages):
         page = db.list_briefing_audio(
             script_id, limit=limit, offset=page_number * limit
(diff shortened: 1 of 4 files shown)


rmusser01:
Qodo adjudication: both findings addressed in one move — the cap is now derived from the seeded data (`len(seed) // limit + 2` = pages of data + the empty terminator) at each site. No repeated magic literal to drift (finding 1), and the cap can never lag a future reseeding, so if it fires the offset genuinely is not advancing and the message cannot mislead (finding 2). All 48 tests green.

PR #1213, opened 2026-08-02
fix: watchlists item-status writes leave the event loop, cancellation-coherent (task-1541)

rmusser01:
## Why

Every item-status write on the Watchlists screen (Ingest, Ignore, unread-toggle, mark-read-on-open) ran its transactional `SubscriptionsDB` write **on the UI thread** — `run_worker` on a coroutine only reschedules onto the same loop, and the backend controller's `_maybe_await` never wraps sync calls. With no `busy_timeout` configured, any lock contention (a second instance, a background check) froze the UI for the duration (task-1541, found during an earlier task's review which fixed only its own new write).

## What

- `_update_item_status`'s write moved off-loop via `asyncio.to_thread`, following the screen's own `_toggle_briefing_queue` precedent; thread-identity test pins the *write* thread via a spy recording `get_ident()`.
- Ingest/Ignore moved from the shared "default" worker group to a **per-item** group — same-item repeats still supersede; different items can no longer cancel each other or be cancelled by unrelated default-group work (AC #2). The unread/mark-read pair keeps its intentional cross-item supersede group (fast j/k coalescing).
- **Review find (Important, data-loss):** moving the write off-loop un-inerted a cancellation point — fast j/k could cancel a mark-read *continuation* after its durable write landed, leaving the cache claiming `new` while the DB said `reviewed`; re-opening the item then overwrote an `ingested` status via the cache-gated mark-read-on-open. Fixed twice over: the CancelledError handler now synchronously patches the cache to match the durable write before re-raising, and mark-read-on-open gained a backend-side terminal-status gate. Mutation-proven in both directions.
- **Review find (flake refuted as "inherent"):** the fix amplified a pre-existing racy test (measured 12.5% → 45.8% at n=24) that polled the DB and never awaited the screen's recompose; fixed with the repo's own `wait_for_selector` helper — assertions byte-identical, 20/20 twice after.
- Minor hardening: `item_id=None` can't collapse per-item groups (guarded + pinned); docstring states DB write ordering is bounded by the supersede group only in the UI, not the DB. Follow-up task-1930 filed for two parked Minors (narrow write-fail+cancel race; `_mark_item_unread`'s milder sibling gap).

## Testing

452 passed across the Watchlists/UI item-action suites (1 unrelated rotating-victim mount race in an untouched file, 5/5 in isolation); the previously-flaky test 20/20 twice; new cancellation-coherence, overwrite-prevention, and group-collision tests all mutation-verified. Opus whole-branch review → fix wave → scoped re-review APPROVED.

---
## Summary by cubic
Moves Watchlists item-status writes off the UI event loop and replaces cancellation-based supersede with a per-item desired-status drainer so the last action deterministically wins without cross-item interference. Prevents UI freezes and stale/overwritten states; addresses task-1541 and supersedes task-1930.

- **Refactors**
  - Introduced `_ItemStatusIntent`, `_dispatch_item_status`, and per-item drain workers (`_ITEM_STATUS_DRAIN_GROUP_PREFIX`) that coalesce desired status; each item has at most one queued and one in-flight write, and the drainer awaits completion before checking again.
  - Added `_update_item_status_off_loop` (uses `asyncio.to_thread`) to run `SubscriptionsDB.mark_item_status` off-loop; removed `CancelledError` handling and deleted `_mark_item_unread`/`_confirm_new_then_mark_item_read_on_open`.
  - Added inside-drain backend gate `_item_status_write_allowed` for unread and mark-read-on-open to avoid overwriting ingest/ignore done behind the cache.
  - Guarded id-less entities in Ingest/Ignore to avoid collapsing into a shared worker group.

- **Tests**
  - New tests pin off-loop writes (thread identity), coalescing determinism for rapid opposing actions, the inside-drain gate when ingest happens behind a stale cache, and error-toast behavior on failed writes.
  - Stabilized the flaky content-pane test with `wait_for_selector` waits; existing item-action assertions remain unchanged.

<sup>Written for commit f34349ff842c6ac9a5b8e5ff5528e618a7ad14eb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1213?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_watchlists_content_pane.py
+++ b/Tests/UI/test_watchlists_content_pane.py
@@ -1300,6 +1300,7 @@ async def test_mark_unread_refuses_to_overwrite_an_item_ingested_by_the_real_ges
 
     from Tests.UI.test_destination_shells import DestinationHarness
     from Tests.UI.app_factory import _build_test_app
+    from Tests.UI.full_app_destination_context import wait_for_selector
     from tldw_chatbook.UI.Watchlists_Modules.content_pane import ContentPane
     from tldw_chatbook.UI.Watchlists_Modules.inspector_pane import IngestRequested
 
@@ -1330,6 +1331,15 @@ async def test_mark_unread_refuses_to_overwrite_an_item_ingested_by_the_real_ges
             row["id"] for row in db.get_new_items(status="ingested", limit=10)
         }, "the precondition: the real Ingest gesture really did write `ingested`"
 
+        # Ingest's `refresh=True` tail sets `overview_data`
+        # (`reactive({}, recompose=True)`), which unmounts and remounts the
+        # whole screen; querying immediately after the DB write races that
+        # recompose. Wait for the pane to resettle before querying it -- the
+        # sibling test at :1631 does the same for the identical reason.
+        await wait_for_selector(
+            screen, pilot, "#watchlists-content-pane", timeout=4.0
+        )
+
         # The staleness this fix exists for -- assert it, do not assume it.
         content_pane = screen.query_one("#watchlists-content-pane", ContentPane)
         assert str(content_pane.item.get("status")).lower() != "ingested", (
@@ -1338,6 +1348,9 @@ async def test_mark_unread_refuses_to_overwrite_an_item_ingested_by_the_real_ges
         )
 
         # 3. Press the real button.
+        await wait_for_selector(
+            screen, pilot, "#content-mark-unread-button", timeout=4.0
+        )
         screen.query_one("#content-mark-unread-button", Button).press()
         await pilot.pause(0.8)
 
--- a/Tests/UI/test_watchlists_inspector.py
+++ b/Tests/UI/test_watchlists_inspector.py
@@ -1498,6 +1498,77 @@ async def test_the_queue_write_runs_off_the_event_loop_thread():
 
 
 @pytest.mark.asyncio
+async def test_the_item_status_write_runs_off_the_event_loop_thread():
+    """TASK-1541: pin the load-bearing part of moving `_update_item_status`'s
+    write off the UI thread.
+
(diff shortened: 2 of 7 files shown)


rmusser01:
Qodo adjudication — both findings confirmed and fixed structurally in `d41c6eba4`, taking the recommended approach (per-item serialization with coalescing) rather than patching the corners:

**Finding 1 (same-item writes reorder)** and **Finding 2 (cache patched on a possibly-never-run write)** shared one root: cancellation-based supersede cannot govern durable off-thread writes. The redesign removes cancellation from the model entirely — every dispatch path records an `_ItemStatusIntent` in a per-item desired-status map and kicks a per-item drainer that serializes writes, re-checks the map after each write (last user action always lands last in the DB), patches the cache only after a write genuinely returned, and surfaces real write failures instead of discarding them. The `CancelledError` handler and both old worker-group constants are deleted, not amended; the mark-read-on-open backend gate is kept and additionally re-checked inside the drain right before writing.

Pinned by: a slowed-write rapid-opposing-actions test (final DB status = last action, coalescing observed via spy), the inside-drain gate re-check test (item ingested while a mark-read intent waits → refused), cache-coherence under supersede, and failed-write honesty. Mutations red both discriminating tests. Required scope: 454 passed, 0 failed; the two source-create-form failures in an exploratory over-broad run reproduce identically in isolation on the current dev tip (`ad570f5a6`) — pre-existing on dev, untouched files here.

PR #1215, opened 2026-08-02
feat(console): usage capture + pricing foundation (cost ticker PR1)

rmusser01:
PR1 of the Console cost-ticker program (spec + plan included in this PR under `Docs/superpowers/`).

**What this builds** — the invisible foundation PR2 (prompt caching) and PR3 (cost chip + cache-break alert) stand on:

- **Providers emit real usage**: Anthropic streaming surfaces usage as empty-`choices` SSE chunks (input-side at `message_start` so aborted streams keep billed input; cumulative output at stream end, gated so `message_start`'s placeholder `output_tokens` can never masquerade as a real count). OpenAI opts into `stream_options.include_usage` (native path only, one-retry degrade on 400s naming the param); Responses API passes its `response.completed` usage through.
- **Gateway → controller**: usage payloads accumulate per provider call on `ConsoleProviderStreamSignals` (call-boundary close-out prevents cross-call bucket corruption); the controller normalizes into a disjoint-bucket `ProviderUsage` (`uncached_input`/`cache_read`/`cache_write`/`output` — OpenAI's cached-inclusive `prompt_tokens` is subtracted) and attaches it on **every** send path, agent-runtime and direct alike, with `partial=true` on stopped streams.
- **Persistence**: schema **v29→v30** adds local-only `messages.usage_json` (excluded from all sync triggers; idempotent runner). Tokens + provider + model only — **never dollars**; cost is computed at read time so a wrong price is retroactively correctable. Reopened conversations and screen-state round-trips restore usage; a stopped regenerate cannot reprice the original answer (variant-base restore guard).
- **Pricing catalog**: seeded, dated, config-overridable (`[pricing]`), generation-aware patterns (unknown future generations resolve to `None` — the UI will show tokens, never a fabricated price), local providers $0.00, rates verified against official pricing pages (`as_of` per provider).

**Notes for reviewers**
- Two commit subjects say "v29"; the migration is v29→v30 after the task-1780 kept-briefings collision took v29 upstream mid-flight (renumber record in the SDD reports).
- Known env artifact: `Tests/DB/test_chat_image_db_compatibility.py` needs numpy (`embeddings_rag` extra) — pre-existing, unrelated.
- Deliberate boundaries: billed-but-contentless turns record usage in-store only (the store never persists empty-content rows); `chatgpt-4o-latest`/`o1-mini` unpriced (retired, no official rate); gpt-5.6 >272K pricing tier not representable in the flat schema (PR2/3 follow-up).

**Verification**: TDD throughout; per-task adversarial reviews + a whole-branch final review (its 3 load-bearing findings — agent-path capture, cross-call accumulation, stop-path persistence — are fixed and mutation-tested); `Tests/Chat/` 3027 green + DB/pricing/resume/benchmark gates green on the exact pushed tree; full suite triaged with every delta resolved.

---
## Summary by cubic
Builds the foundation for the Console cost ticker: real token usage is captured, normalized, and saved per message; pricing is applied at read time via a new catalog. Capture runs on all send paths (agent and direct), and survives stops, restores, and screen-state; persistence is local-only and never blocks a send.

- New Features
  - Anthropic streaming: usage emitted as empty-`choices` SSE chunks; placeholder `output_tokens` at `message_start` is stripped; output counted only at stream end.
  - OpenAI streaming: opts into `stream_options.include_usage` with a narrow 400 fallback; Responses API usage is passed through and converted to chat SSE.
  - Usage capture: gateway records raw payloads on `ConsoleProviderStreamSignals`; payloads close per provider call; controller normalizes into `ProviderUsage` buckets (`uncached_input`, `cache_read`, `cache_write`, `output`) and sums; attached to assistant messages with `partial=true` for stops and no-content failures.
  - Signals now built and passed on every path (agent runtime and direct), so capture happens everywhere.
  - Persistence: per-message usage saved as `messages.usage_json`; attaches after terminal marks persist via a local-only write that doesn’t bump `version`/`last_modified` or enqueue `sync_log` rows; usage is restored on reopen and screen-state; variant restores prevent cancelled regenerates from repricing original answers; attach errors are logged and swallowed.
  - Pricing catalog: seeded, dated, and config-overridable; provider keys normalized; generation-aware patterns (unknown generations return `None`); local providers priced at $0.00; rates re-verified.

- Migration
  - DB v29→v30 adds local-only `messages.usage_json` (excluded from sync triggers); runner is idempotent and fresh DBs reach v30 automatically.

<sup>Written for commit f6548b12ca90031dcec8baa898d3d5e3d014d2ad. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1215?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Qodo findings addressed in `edf0d0483` + `ab668b26b`:

- **Finding 1 (attach can throw) — fixed.** `_attach_stream_usage` now catches persistence exceptions beyond `KeyError`, logs a `usage_attach_failed` warning with the message id, and continues — the stop/cancel outcome can no longer be altered by a failed usage write. Regression test: an adapter raising `RuntimeError` mid terminal-flush leaves the stop outcome unchanged.
- **Finding 4 (usage write bumps sync_log) — fixed properly.** Usage-only flushes no longer route through `update_message`'s version bump: a new `update_message_usage_local` writes the local-only column directly with **no version/last_modified change**, so the `messages_sync_update` trigger cannot fire and peers never see an unreconcilable bump. The normal completion path (usage riding the content flush) is unchanged — that version bump is legitimate. Tests assert version, last_modified, and sync_log row count are all untouched by a usage-only write.
- **Finding 3 (docstrings) — fixed.** Google-style class docstrings for `ProviderUsage`, `ModelPricing`, `CostBreakdown`, and Args/Returns on `plus()`.
- **Finding 2 (Pydantic at the boundary) — declining, with reasoning.** `from_provider_payload`'s documented contract is *tolerant normalization that never raises*: malformed or unknown payloads degrade to `None` (the UI then shows token counts, never a fabricated figure). A Pydantic model at this seam inverts that contract (validation errors) or reimplements the same tolerance in validators with more machinery; the three overlapping provider shapes (Anthropic-native / OpenAI chat / OpenAI Responses, which collide on `input_tokens`) are discriminated by key-presence order, which is the actual logic and is unit-tested per shape. This also matches the repo's existing dict-based normalization pattern (`model_capabilities.py`) — as the finding's own relevance note observes, there's no established Pydantic-at-boundary precedent here. Happy to revisit if the rule is meant to be binding for new code.

rmusser01:
Note on Qodo's re-scan: findings 1 and 2 are re-listed against stale pre-fix hunks — both are already addressed at `ab668b26b`:

- **Attach-can-throw**: `_attach_stream_usage` now has a broadened `except Exception` with a `usage_attach_failed` warning immediately after the `except KeyError` the finding cites (`console_chat_controller.py:5972-5987`).
- **Sync churn**: the cited `update_data["usage_json"]` line is the *content-riding* path, whose version bump is legitimate (content changed anyway). The usage-**only** flush — the actual churn case — no longer goes anywhere near it: `set_message_usage`'s terminal branch routes through the new version-neutral `update_message_usage` → `update_message_usage_local` (`console_chat_store.py:3099-3104`), which touches neither `version` nor `last_modified`, so the sync trigger cannot fire. Covered by tests asserting version/last_modified/sync_log are untouched by a usage-only write.

PR #1228, opened 2026-08-02
feat: opt-in localhost server for the exported briefings feed directory (task-1760)

rmusser01:
## Why

Spec #2 phase 3 promised localhost serving of the exported feed directory "if `[web_server]` is enabled" — but that premise was false (`[web_server]` is textual-serve, a mutually-exclusive process mode read by no gate, serving only its own browser assets). The owner cut serving from phase 3 and filed this as net-new work: a purpose-built static server the TUI can start and stop on demand, with the security review that new listening-socket surface demands (task-1760).

## What

- **`Subscriptions/feed_server.py`** — a `FeedDirectoryServer` wrapping stdlib `ThreadingHTTPServer` on a daemon thread. GET/HEAD only (unknown verbs 405/501, default-deny); **path containment** via `Path(candidate).resolve().relative_to(served_root)` — component-wise, so symlink escapes (file and directory), encoded/raw `..`, absolute-path and leading-`//` forms all 404; **no directory listing** (`list_directory` → 404; recursive file serving kept for episodes in subfolders); ephemeral or configured port; **binds `127.0.0.1` by default** — a blank/typo'd/non-string bind falls back to loopback rather than silently going public, and any genuinely non-loopback bind logs a warning and says so in the toast.
- **Opt-in, off by default:** nothing listens until the user presses Serve on the Artifacts pane; one directory at a time (a second start refuses, naming the running URL, before binding); session-only — daemon thread plus `on_unmount` stop, so exit never wedges a socket. `[briefings_feed_server]` supplies bind/port defaults for the explicit action only; `[web_server]`/`Web_Server/` untouched.
- **Security posture (AC #4)** stated plainly in the user guide, module docstring, config comment, and the per-serve toast: loopback default and what widening means, **no authentication** ("anyone who can reach this address can read the feed while it is serving"), the containment guarantee, and that every file in the chosen directory *and its subdirectories* is served — so the guidance is to point it at a dedicated export folder, not your home directory.

## Testing

New `test_feed_server.py` (41): real server on ephemeral localhost, httpx round-trip, a **raw-socket traversal matrix** putting `..` on the wire (httpx normalizes it away client-side — the reviewer's independent 24-case raw probe confirmed containment), symlink-out (file and directory) → 404, POST → 405, no directory listing, blank-bind → loopback, non-loopback → warning, stop() < 150ms (measured 51ms), stop closes the socket. `test_watchlists_artifacts_pane.py` 124. Opus security review → fix wave (error-path logging crash, 500ms→51ms stop, bind validation, no-listing) → scoped re-review, both APPROVED; every fix mutation-verified. The `..`-traversal and bind-default defaults were traced through the real `get_cli_setting` against this repo's known dotted-form-drops-default trap.

---
## Summary by cubic
Adds an opt-in localhost static server to serve the exported briefings feed directory, with Serve/Stop controls in the Artifacts pane. Addresses Linear TASK-1760 by meeting AC #1–#4: user-started serving, loopback default, strict containment, and working with podcast clients.

- **New Features**
  - Added `FeedDirectoryServer` wrapping stdlib `ThreadingHTTPServer` on a daemon thread. Serves one exported directory per session, refuses a second start, and stops on unmount.
  - New `[briefings_feed_server]` config for bind/port defaults used by the explicit Serve action; not tied to `[web_server]`.
  - UI: Serve Feed / Stop Serving buttons on the Artifacts pane, disabled until a feed is exported; toast shows the URL and security posture.
  - Docs updated; tests cover real-server round trips, traversal and symlink escapes, method gate, bind/port defaults, IPv6, and UI wiring.

- **Bug Fixes**
  - GET/HEAD only; other verbs return 405/501.
  - Path containment via `Path.resolve().relative_to(...)`; directory listings disabled; recursive file serving kept for feed subfolders.
  - Bind validation: default `127.0.0.1`; warn in logs/toast when widened off loopback; IPv6 binds supported (AF_INET6) with bracketed URLs.
  - Port validation with safe fallback and `FeedServerError` on invalid input; faster stop (~51ms) with a 50ms poll; stop closes the socket; fixed malformed-request logging crash.

<sup>Written for commit e11a40c79a9344888700c163112a6921b7c63fab. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1228?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/User_Guide/watchlists.md
+++ b/Docs/User_Guide/watchlists.md
@@ -94,7 +94,9 @@ starts on its own, and no setting can make it start on its own.
   address, a warning is logged and the Serve Feed toast says so plainly.
   Only widen this if you understand that doing so removes the one thing
   standing between "only this machine" and "anyone who can reach it" —
-  there is still no authentication either way.
+  there is still no authentication either way. Note that a wildcard IPv6
+  bind (`::`) is usually dual-stack: it accepts plain IPv4 connections as
+  well, so the reachable surface can be wider than the address suggests.
 - **Confined to the exported directory.** The server refuses (with a plain
   404) any request that would read a file outside the directory you
   exported to, including through a symlink planted inside it that points
--- a/Docs/User_Guide/watchlists.md
+++ b/Docs/User_Guide/watchlists.md
@@ -73,15 +73,28 @@ served until you press **Serve Feed**, and it stops the moment you press
 starts on its own, and no setting can make it start on its own.
 
 - **No authentication.** Anyone who can reach the served address can read
-  every file in the exported directory for as long as it is running. The
-  toast that appears when you press Serve Feed states this every time, not
-  just here.
+  every file in the exported directory **and every subdirectory beneath
+  it** — serving is recursive, not limited to `feed.xml` and the episodes
+  next to it — for as long as it is running. The toast that appears when
+  you press Serve Feed states this every time, not just here. Point Serve
+  Feed at a dedicated export folder, not a general-purpose directory like
+  your home folder, for exactly this reason: everything under whatever
+  folder you export into becomes reachable while serving is on.
+- **No directory browsing.** The server refuses to render a listing of the
+  served folder (or any subfolder) — only a file whose exact name a client
+  already knows (from `feed.xml`, or a URL you typed yourself) is
+  fetchable. This narrows the exposure above; it does not remove it, since
+  every filename in the tree is still fetchable if it is known.
 - **Loopback by default.** The server binds `127.0.0.1` (this machine
   only) unless you change `bind` under `[briefings_feed_server]` in
   `config.toml` to something wider (e.g. `0.0.0.0`, to reach it from
-  another device on your network). Only widen this if you understand that
-  doing so removes the one thing standing between "only this machine" and
-  "anyone who can reach it" — there is still no authentication either way.
+  another device on your network). A blank or otherwise invalid `bind`
+  value can never silently widen this — it falls back to `127.0.0.1`
+  instead — and whenever the server does end up bound to a non-loopback
+  address, a warning is logged and the Serve Feed toast says so plainly.
+  Only widen this if you understand that doing so removes the one thing
+  standing between "only this machine" and "anyone who can reach it" —
+  there is still no authentication either way.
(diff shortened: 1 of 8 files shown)


rmusser01:
Qodo adjudication — all three confirmed and fixed in `87efdb14b`, plus a doc follow-up `a417a911c`:

**Finding 1 (port not validated → OverflowError escapes the UI catch) — fixed.** A config-derived out-of-range/garbage port now falls back to ephemeral 0 with a type-only warning (a bad config value degrades safely, never crashes the Serve action); an explicit out-of-range `start(port=...)` raises `FeedServerError` — which the screen already catches — *before* any bind. Enumerated `start()`'s exception surface: only `FeedServerError`/`OSError`(-subclasses) can now escape. Mutation-verified both paths.

**Finding 3 (IPv6 URL not bracketed) — fixed.** IPv6 literals render as `http://[::1]:port/` via `ipaddress` detection (hostnames/IPv4 unchanged). Verifying it surfaced a deeper defect the bracketing alone wouldn't have reached: stdlib `ThreadingHTTPServer` hardcodes `AF_INET`, so an IPv6 bind failed with `gaierror` before the URL code — added an `AF_INET6` server selected by the resolved bind. Independently re-reviewed: family keys off the same normalized bind used for the socket and URL (no v6-family/v4-bind divergence), the containment/GET-only/no-listing handler is the identical class on both paths (re-verified with a raw AF_INET6 symlink-escape probe → 404), and `::1` is not dual-stack (IPv4 client gets ECONNREFUSED).

**Finding 2 (docstring) — fixed.** `is_loopback_bind` gained Google-style Args/Returns.

Follow-up: the re-review flagged that a wildcard `::` bind *is* dual-stack here (accepts IPv4 too) — `is_loopback_bind("::")` correctly returns False so the exposure warning fires, and I've now made both the warning and the user guide disclose the dual-stack mechanism explicitly rather than leave it implied.

PR #1230, opened 2026-08-02
fix: escape untrusted watchlist text at the DataTable markup sink (task-1348)

rmusser01:
## Why

`DataTable` markup-parses its `str` cells, and `items_pane.py` fed item titles and source names — remote feed content — straight in. A feed title containing `[bold red]BREAKING[/]` was *interpreted* as Rich markup rather than displayed; a source name `[link=http://evil.test]Feed[/link]` could render as a clickable link. This is reachable today with real remote content (task-1348 AC#1).

## What

`add_row` now wraps every item-derived string cell in `rich.markup.escape` — title and source name are the untrusted ones; status, created_at and the queued glyph are app-controlled but escaped too, so every non-constant cell is uniformly safe and no future editor has to re-audit which column carries remote text. Escaping happens at the boundary where the parser actually is, following the rule `content_pane.render_article` already states in full.

AC#2 and AC#3 were already satisfied on dev before this task: `content_pane.py`'s `_MARKDOWN_HYPERLINKS = False` is the recorded decision that remote markdown bodies must not emit real OSC-8 hyperlinks (phishing-anchor reasoning documented there), and `render_article`'s docstring states the governing rule once — escape at sinks that parse, not at sinks that don't. This change cross-references that rule rather than restating it.

## Testing

New `test_markup_shaped_item_text_is_escaped_at_the_datatable_boundary` drives a hostile title and source name through the pane and asserts the cells hold the escaped form, not the raw tags — mutation-verified: removing the boundary escape reds it. `test_watchlists_items_pane.py` + `test_watchlists_content_pane.py`: 48 passed.

---
## Summary by cubic
Escape untrusted watchlist text at the `DataTable` markup sink so titles and source names show as plain text, and escape status repaints to keep the sink closed. Completes TASK-1348 AC#1 and clarifies the “escape at parsing sinks” rule.

- Bug Fixes
  - Escaped at both write sites: `add_row` (title, source name, status, created_at) and `update_item_status_cell` via `DataTable.update_cell`, using `rich.markup.escape`.
  - Added tests for both boundaries: `test_markup_shaped_item_text_is_escaped_at_the_datatable_boundary` and `test_status_repaint_escapes_at_the_update_cell_boundary`, and updated comments/docstrings to reference the boundary rule.

<sup>Written for commit e7343a1a2c88b80e21f73f6733c298b1f5b44bda. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1230?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Watchlists/test_watchlists_items_pane.py
+++ b/Tests/Watchlists/test_watchlists_items_pane.py
@@ -1,6 +1,7 @@
 """Tests for the Watchlists items pane."""
 
 import pytest
+from rich.markup import escape as escape_markup
 from textual.app import App, ComposeResult
 from textual.widgets import Button, DataTable, Input, Select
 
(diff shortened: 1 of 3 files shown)


rmusser01:
Qodo adjudication — all four addressed in the fix commit:

**Finding 3 (status repaint bypasses escaping) — confirmed real, fixed.** Good composition catch: `compose()` escaped the status cell but `update_item_status_cell`'s `update_cell` repaint didn't, so the guarantee held at one write site and broke at the sibling. Status is an app-controlled enum today (not a live exploit), but the boundary must be uniform or the escape is a half-measure a future status value could slip past. Now escaped at both sites; pinned by `test_status_repaint_escapes_at_the_update_cell_boundary` with a markup-shaped status, mutation-verified.

**Finding 4 (queued-glyph comment wrong) — confirmed, fixed.** The Queued column is one of two app CONSTANTS (`_QUEUED_GLYPH` or `""`), never escaped and never needing to be. Comment corrected: every VARIABLE cell is escaped, the constant Queued column is exempt.

**Findings 1 & 2 (docstrings) — fixed.** `Args:` added to the new tests; `ItemsPane.compose()` gained a Google-style docstring.

PR #1239, opened 2026-08-02
feat(anthropic): complete prompt caching (cost ticker PR2)

rmusser01:
PR2 of the Console cost-ticker program. The spec is amended in-PR twice: once for the stale premise (task-323 had already shipped system+tool breakpoints provider-wide before this program's snapshot), and once after the final whole-branch review caught a real economics hazard in my first cut.

- **Per-turn cache_control breakpoint on the final message — Console-opt-in only.** The Console gateway stamps `prompt_caching=True` on Anthropic resolutions; `chat_with_anthropic` places the third breakpoint only when asked. One-shot callers (media summarization, websearch, evals, prompt engineering) are structurally excluded — the final review showed a provider-wide version would have taxed them ~1.25× on input with a guaranteed zero read rate. Non-console pass-through is pinned by tests.
- Conversation history now becomes a reusable cache prefix for Console chat (task-323's system/tools breakpoints never cached history). Budget: 3 of 4 breakpoints; writes are incremental per Anthropic prefix semantics (verified in review).
- **`[caching] anthropic_enabled` kill-switch** (default on; absent section = shipped behavior; the 5 pre-existing task-323 tests pass unmodified). Config comment states both the win and the loss condition (slow-cadence writes).
- **Degrade path**: a 400 naming cache_control retries exactly once with all breakpoints stripped + a `anthropic_cache_control_degrade` counter — caching can never break sends.
- **Prefix-stability pins**: gateway system-extraction byte-stability across turns (the spec-named seam), history content-identity modulo the moving marker, marker placement, no volatile keys.
- Accounting rides PR1 unchanged (cache_read/cache_write buckets → usage_json → pricing catalog).

Known limit documented in the spec's Risks: Anthropic's ~20-block cache lookback can silently miss on tool-heavy agent turns; the 4th breakpoint is held in reserve pending live hit-rate evidence (follow-up).

Live acceptance (real key): `cache_read_input_tokens > 0` on the second consecutive Console send, visible in PR1's usage records.

---
## Summary by cubic
Completes Anthropic prompt caching with a Console‑only per‑turn breakpoint, a `[caching] anthropic_enabled` kill‑switch, and a safe degrade path. Conversation history is now cache‑reusable in Console chat, while one‑shot calls and non‑Anthropic providers are unaffected.

- **New Features**
  - Console‑opt‑in per‑turn `cache_control` on the final message when `prompt_caching=True` (Console gateway stamps this for Anthropic; other providers drop it). System + last‑tool breakpoints remain provider‑wide; total 3/4.
  - `[caching] anthropic_enabled` config toggle (default true) gating Anthropic caching end‑to‑end (gateway + provider).
  - 400s naming `cache_control` retry once without breakpoints; logs `anthropic_cache_control_degrade`. No `ttl` emitted (5‑minute default).
  - Prefix‑stability tests pin system extraction and history identity across turns; docs/spec updated to reflect provider‑wide system/tool breakpoints and Console‑only per‑turn.

- **Bug Fixes**
  - Scoped the per‑turn breakpoint to Console sends to prevent 1.25× write costs on one‑shot Anthropic calls.
  - Console gateway now honors the kill‑switch in the live `load_settings()` shape (`COMPREHENSIVE_CONFIG_RAW.caching`); config read failures warn and default caching ON.

<sup>Written for commit 57e673d2f563fde9295e2ddb18f491e048e13596. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1239?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/plans/2026-08-02-console-cost-ticker-pr2-caching.md
+++ b/Docs/superpowers/plans/2026-08-02-console-cost-ticker-pr2-caching.md
@@ -1,0 +1,694 @@
+# Console Cost Ticker PR2 — Prompt-Caching Completion Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Complete Anthropic prompt caching by adding the per-turn message breakpoint (making the whole conversation prefix reusable turn over turn), a `[caching]` kill-switch, and a cache_control 400-degrade — and amend the spec whose PR2 premise went stale.
+
+**Architecture:** Task-323 (already merged to dev, before this program's spec was written against an older snapshot) ships `cache_control` on the system block and last tool for **every** `chat_with_anthropic` caller, gated on `_anthropic_supports_caching(model)`. PR1 already maps Anthropic cache fields and OpenAI `cached_tokens` into the disjoint usage buckets. PR2 therefore: (1) amends the spec's stale premise; (2) adds the one missing breakpoint — last content block of the final message; (3) adds `[caching] anthropic_enabled` (default true) gating **all three** breakpoints at the provider level (matching shipped reality — all callers benefit; the spec's "console-gateway-only injection" note is part of the stale premise being amended); (4) adds a 400-naming-cache_control retry-once-without fallback; (5) pins prefix stability with tests.
+
+**Tech Stack:** Python ≥3.11, requests (mocked in tests), pytest.
+
+**Spec:** `Docs/superpowers/specs/2026-08-01-console-cost-ticker-design.md` — PR2 section, as amended by Task 1 of this plan.
+
+## Global Constraints
+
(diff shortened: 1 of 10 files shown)


rmusser01:
Qodo findings addressed in `57e673d2f`:

- **Finding 4 (gateway misreads caching config) — confirmed real and fixed.** `load_settings()` never exposes a top-level `caching` key (it only survives under `COMPREHENSIVE_CONFIG_RAW`), so the resolution flag ignored the kill-switch on the live path — exactly as flagged. New `_caching_config_value()` reads both shapes, with a test pinned against the exact `load_settings()` shape. Note actual sends were never at risk: the provider-level `_anthropic_caching_enabled()` gate independently zeroes all breakpoints; this fix makes the resolution flag truthful.
- **Finding 3 — fixed**: the fail-open config read now logs a warning (behavior unchanged), with a test.
- **Finding 1 — fixed**: full Google-style docstring on `chat_with_anthropic` (all 15 params, Returns, Raises).
- **Finding 2 — fixed for this PR's addition**: the new test's literal is now `sk-test-placeholder-not-a-key`. The file's ~30 pre-existing `sk-test` literals are out of scope; follow-up candidate for a single sweep.

Gates: targeted 141 passed; full `Tests/Chat/ + Tests/LLM_Calls/` 3078 passed.

PR #1241, opened 2026-08-02
fix(library): ingest UAT P1 batch — stability, honesty, visibility (tasks 2010-2014)

rmusser01:
## Summary

Fixes the five P1 defects from the 2026-08-02 Library file-ingestion UX UAT (critique snapshot `.impeccable/critique/2026-08-02T21-04-04Z__chatbook-widgets-library-library-ingest-canvas-py.md`, scored 21/40). Also files the full findings backlog: tasks 2010-2016 (P2 batch 2015 and P3 batch 2016 are filed, not fixed here).

- **task-2010** — every ingest job transition ran a full-screen `refresh(recompose=True)`, silently stealing focus from the Input being typed into (next keystrokes hit the global digit bindings and navigated the app) and snapping queue scroll to top. Now captures focus id + cursor + scroll before the recompose and restores after (`_refresh_library_ingest_canvas_preserving_context`).
- **task-2011** — a pre-flight worker finishing after `_do_submit_ingest` cleared the form resurrected the cleared summary ("Enter a file path to start." rendered beside "1 plain text file · 277 B"). Cancellation is cooperative; every clear site now bumps a generation the apply path checks.
- **task-2013** — a byte-identical file at a different path hit the DB duplicate-skip (`media_id=None`), the URL fallback missed, and the job became a "✓ done" row that created nothing and said nothing. The writer now mirrors the DB's `sha256(content)` (the payload never carried a hash — the `mark_done content_hash` stamp had always been `None`), falls back to `get_media_by_hash`, and labels the row "Already in Library — matched an existing item".
- **task-2012** — the unscoped `Checkbox {height:2}` rule from `_conversations.tcss` clipped "Analyze after ingest"/"Chunk content" to border-only husks, and placeholder-as-label left populated option inputs showing bare `1000`/`100`/`auto`. Scoped escapes + explicit `Static` labels; `.type-group-contents {height:auto}` also removes the expanded panel's ~15 trailing phantom rows.
- **task-2014** — canvas focus was a background-color change only (8/10 Tab stops invisible in monochrome). Applied DESIGN.md's `outline: heavy $accent` focus contract across the canvas.

## Test evidence

- TDD throughout: each fix's test observed red on the exact target assertion before the implementation.
- `Tests/Library` + `Tests/UI/test_library_shell.py`: **1139 passed, 1 failed** — `test_library_shell_note_conflict_reload_discards_local_edits`, the documented order-dependent notes tail (passes in isolation, pre-existing, unrelated to ingest).
[...]

PR #1244, opened 2026-08-03
feat(library): ingest UAT P2 batch — feedback, copy, consistency (task-2015)

rmusser01:
## Summary

The 12 confirmed P2 findings from the 2026-08-02 ingest UAT (task-2015), following the P1 batch (#1241):

1. **While-typing path validation** — 0.8s debounce triggers pre-flight without waiting for blur; the pre-flight trigger/apply now use the context-preserving refresh so a result landing mid-word can't steal focus.
2. **Above-fold completion signal** — one "Ingest finished — N imported · M failed" toast when the queue's active jobs settle to zero (deltas vs the baseline captured when the queue went active).
3. **Empty-file failures are permanent** — zero-byte files raise `PermanentIngestError`; the dead-bait Retry is gone.
4. **Folder done rows carry media ids** — proven by test; the UAT sighting was the task-2013 duplicate case.
5. **Pluralization** — "recorded as failures" (was "as a failures").
6. **Error unwrapping** — nested `Failed to X file:` chains collapse to one prefix in `short_ingest_error` (both failure surfaces).
7. **Two-press "Clear finished"** — first press arms (label names what a second press removes), any registry mutation disarms.
8. **Narrow-width ellipsis** — options summary title ellipsizes instead of hard-clipping mid-word (CollapsibleTitle needed width:1fr + height:1 + nowrap for text-overflow to engage).
9. **No estimate noise under errors** — "0 files"/breakdown suppressed when pre-flight reports errors.
10. **Start disabled for guaranteed failures** — with an explaining gate line when zero supported files are staged.
11. **Honest elapsed** — measured from submission; sub-second renders `<1s`; unknown drops the segment (never a fake "0s").
12. **Per-stage progress** — parsing/writing rows verified already pinned by existing row-format tests.

## Test evidence

- TDD per item; every fix observed red on its target assertion first.
- Full `Tests/Library` + `Tests/UI/test_library_shell.py` + `test_library_ingest_canvas.py` + `Tests/Local_Ingestion`: green (1324/1325 first pass; the one failure was the new two-press test's own single-pause timing under full-suite load — hardened with poll loops per the task-699 state-then-DOM lesson, full shell file then 274/274).

## Live verification (isolated profile, tmux)

- "Path not found" appears ~1.6s after typing with no blur.
- "Ingest finished — 1 imported" toast on settle; `✓ done · report.txt · <1s` row.
- Options title renders `…Chunk cont…` at 110 cols instead of the hard clip.

---
## Summary by cubic
Delivers the 12 P2 UAT fixes for Library ingest (task-2015), improving feedback, copy, and consistency. Adds while-typing preflight with safe fencing, resilient completion counts, and a two‑press clear.

- **New Features**
  - While-typing path validation with a 0.8s debounce; results apply without stealing focus. Edits fence off in‑flight preflight for the previous path; clearing the field drops stale feedback; remount echoes don’t re‑arm the timer.
  - One completion toast when the batch settles: "Ingest finished — N imported · M failed". Baseline re‑anchors if finished rows are cleared mid‑batch to keep counts accurate.
  - "Clear finished" now requires a second press; the button label names what will be cleared and disarms on any queue change.

- **Bug Fixes**
  - Accurate elapsed on done rows: measured from submission; "<1s" for sub‑second; segment omitted if unknown.
  - Preflight UX: suppress "0 files" and breakdown when errors exist; disable Start with an explaining line when all staged files are unsupported.
  - Error/copy cleanup: collapse nested "Failed to …" prefixes; pluralize "recorded as failures".
  - Data correctness: folder "done" rows include media IDs; empty files are permanent failures (no Retry).
  - UI polish: options summary header now ellipsizes at narrow widths; per‑stage parsing/writing indicators confirmed.

<sup>Written for commit e32e7ac534d5dfa29329fb7f858851a4d24bde7b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1244?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Addressed the Qodo review in 210e5659d:

**Finding 1 (stale pre-flight during typing) — fixed, and it exposed a worse latent loop.** A genuine path edit now cancels + generation-fences any in-flight pre-flight, and clearing the field drops lingering errors/summary (recomposing only when there is pre-flight state to drop, preserving the handler's no-recompose typing contract). While verifying, found that Textual's remount `value=` echo re-armed the debounce on every recompose — a perpetual ~1s pre-flight loop while any path sat in the field; the handler now ignores the echo (same family as the canvas's `_reported_option_values` guard). Three new tests: previous-path result dropped during the debounce window; cleared field clears state; mount echo doesn't re-arm.

**Finding 2 (negative toast deltas) — fixed.** When DONE/FAILED shrink below the batch baseline (mid-batch "Clear finished"), the baseline re-anchors to the current counts, so settle deltas never go negative and completions after the clear still count. Pinned by `test_completion_toast_survives_mid_batch_clear` (scripted-registry sequence: active-with-1-done → cleared → settle → exactly one "Ingest finished — 1 imported").

**Finding 3 (Args: docstrings in tests) — declined with evidence.** The surrounding test files' existing tests (e.g. `test_missing_file_failure_is_permanent_and_refuses_retry(tmp_path)`) use narrative docstrings without `Args:` throughout; the new tests follow the file idiom, CLAUDE.md's docstring rule targets public APIs, and the finding's own relevance note records no historical enforcement for tests.

PR #1252, opened 2026-08-03
fix: sticky create-form focus intent survives interleaved recomposes (task-1345)

rmusser01:
## Why

The Watchlists create-source form "cannot be filled in" (2026-07-28 UAT): opening it and typing left Name and URL empty. Root cause (TASK-1362 Task 5, confirmed here): `Widget.focus()` only *schedules* focus via `call_later`, and `SourcesPane.recompose()` read-and-cleared its `_pending_create_focus` intent *before* that scheduled focus landed. A second `recompose=True` assignment (a `sources` reload) firing in that gap remounted the field — the scheduled callback fired on a detached widget and was dropped — and because the intent was already cleared, the interleaving recompose recovered nothing. Focus landed on `None`; typing went to the void.

## What

`SourcesPane.recompose()` no longer eagerly clears the intent. It re-applies `_pending_create_focus` on every recompose without discarding it; a new `_confirm_create_focus()`, scheduled via `call_after_refresh`, clears the intent **only once `screen.focused.id == target`** — so whichever recompose is last in an interleaved burst wins, and the intent is discarded only when focus is *confirmed* landed. This is a wait on real state, not a sleep or retry count (this task's history explicitly rejects "shrunk" races). Case 2 (the user tabbed to another field, then an unrelated rebuild happens) is preserved: after the confirm-clear, subsequent recomposes fall back to restoring the field the user is actually on, never yanking back to field 0.

**Scope note (honest split).** This task named two symptoms sharing one history: the focus drop (fixed here) and `NoMatches` on `SelectCurrent`. Only the focus drop has a confirmed root cause and is what this fix addresses. The `SelectCurrent`/`#label` mount race on form-*close* is a distinct Textual-internals issue (a `Select` value-watch querying its `#label` child before it finishes mounting, triggered by the close recompose racing the create worker chain) — three mitigations shrank but did not eliminate it, so per this task's own "a shrunk race is a hidden race" principle it was reverted and filed as **task-1960** with the full `TEXTUAL=debug` diagnosis. AC#2 is narrowed inline to the focus-drop scope; the still-flaky end-to-end test is left unmodified (no skip/xfail).

## Testing

New deterministic forced-interleave test drives Textual's `_check_recompose()` seam twice back-to-back (both recomposes queued before either focus lands, guaranteed by construction) and asserts field 0 focused after settle — mutation-verified: reverting to eager-clear reds it with `focused == None`. Independent review ran the focus tests **0/10 failures** under the `content_pane → create_form` ordered pair, plus a case-2 regression guard asserting the user's field survives an external rebuild. `Tests/UI/ -k watchlist`: 261 passed, the one failure being the deferred SelectCurrent flake (task-1960). Opus whole-branch review: CLEAN.

---
## Summary by cubic
Fixes the create-form focus race in Watchlists by keeping the focus intent sticky and confirming it after recomposes. Adds a bounded confirm that clears on any landed focus to prevent stalls and yank-back. Addresses TASK-1345; the `SelectCurrent` flake remains tracked in TASK-1960.

- **Bug Fixes**
  - In `SourcesPane.recompose()`, prefer the user’s current in-form focus over a stale `_pending_create_focus`; keep the intent armed across recomposes and drop it if the form closes.
  - Added `_confirm_create_focus()` via `call_after_refresh`: clears the intent once focus lands on any widget (not only the target) or after a bound (`_CREATE_FOCUS_CONFIRM_MAX_ATTEMPTS=20`), avoiding infinite reschedules.
  - Tests: kept the deterministic interleave and external-rebuild guard; added checks for stale-intent yank prevention, bounded give-up, and clear-on-non-target focus.

<sup>Written for commit a781c51b77997f5fd391531071b72774bb9ea395. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1252?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_watchlists_source_create_form.py
+++ b/Tests/UI/test_watchlists_source_create_form.py
@@ -127,6 +127,135 @@ async def test_create_form_focuses_its_first_field_when_it_opens(size):
         )
 
 
+@pytest.mark.asyncio
+async def test_a_sources_reload_interleaving_the_open_does_not_lose_focus():
+    """TASK-1345: force the exact interleave the recompose/focus race needs.
+
+    `watch_show_create_form` arms `_pending_create_focus` when the form
+    opens, and `recompose()` calls `.focus()` on the first field -- but
+    `Widget.focus()` only *schedules* the actual focus change (it posts to
+    `app.call_later`), so it has not landed by the time `recompose()`
+    returns. If a SECOND `recompose=True` assignment on the same pane
+    (here: `sources` reloading, exactly what `_load_sources` does in
+    production) lands in that gap, it remounts the create form's fields
+    out from under the still-pending focus callback: the callback then
+    fires on a **detached** widget and is silently dropped. Before
+    TASK-1345's sticky-until-confirmed fix, the intent was already cleared
+    to `None` before the interleave landed, so the second recompose had
+    nothing to reapply either -- focus landed nowhere. See
+    `SourcesPane.recompose`.
+
+    Forces the interleave deterministically -- no sleep, no retry count.
+    `_check_recompose` is the exact internal seam Textual's own
+    `call_next`-driven scheduling uses to invoke `recompose()` (see
+    `Widget.refresh(recompose=True)`); calling it directly, twice, back to
+    back with no intervening `pilot.pause()`, reproduces the ordering the
+    race depends on without guessing at asyncio scheduling: the second
+    recompose starts -- and reads whatever focus-restoration state is
+    current -- strictly BEFORE the first recompose's own scheduled
+    `.focus()` has had any chance to land (nothing has yielded control
+    back to the app's own message queue yet), exactly as `_load_sources`
+    racing the form opening does with two independent asyncio tasks in
+    production.
+    """
+    host = _watchlists_host()
+    async with host.run_test(size=(160, 42)) as pilot:
+        screen = _active_destination_screen(host)
+        screen.active_section = "sources"
+        await _wait_for_selector(
+            screen, pilot, "#watchlists-sources-pane", timeout=5.0
+        )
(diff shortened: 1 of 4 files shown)


rmusser01:
Qodo adjudication — both defects confirmed real and fixed in the follow-up commit:

**FIX A** (`recompose`): restore order is now `_focused_create_field_id() or _pending_create_focus`, so the sticky intent only wins during the genuine mid-burst drop (focus None); the instant the user is on a real in-form field, that field wins and a stale intent can no longer yank them off it. Discriminated by `test_an_external_rebuild_does_not_yank_focus_to_a_stale_pending_target`.

**FIX B** (`_confirm_create_focus`): clears the intent once focus lands on ANY real widget instead of only the exact `target`, and bounds the remaining focus-still-None reschedule at `_CREATE_FOCUS_CONFIRM_MAX_ATTEMPTS = 20`. Discriminated by `test_confirm_create_focus_gives_up_after_max_attempts` and `test_confirm_create_focus_clears_on_any_in_form_field_not_only_target`.

Self-review found the two clear branches (in-form vs outside-form) behaviorally identical — both clear — so mutating one was backstopped by the other and the test didn't discriminate; collapsed into one load-bearing branch, test now reds when mutated to `== target`. Four mutation checks confirmed; ordered-pair focus suite ran 3×3 clean.

PR #1263, opened 2026-08-03
fix: isolate per-URL failures in url_list/sitemap runs, preserving auto-pause (task-1394)

rmusser01:
## Why

`_default_run_executor`'s `url_list`/`sitemap` arms looped URLs with no per-URL guard, so one failing URL (timeout, SSRF block, HTTP error) raised out of the loop, failed the whole run, and discarded every item already collected from the URLs that succeeded — a 50-URL source with one dead link yielded nothing (task-1394).

## What

- **Per-URL isolation:** each `check_url` in both arms now runs through a shared `_check_url_isolated` helper (try/except); a failing URL records a new `DISPOSITION_ERROR` disposition (type-only — `type(exc).__name__`, never the URL or message) and continues. Items and dispositions from the URLs that succeeded persist (AC#1).
- **Visible, not absorbed:** a 6th `"error"` counter joins the disposition vocabulary and renders in the Runs pane's Checks line, so a partially- or fully-failed run says so (AC#2).
- **Auto-pause preserved (whole-branch review, MAJOR).** The isolation initially made an all-URLs-errored run report `completed`, which reset the subscription's auto-pause circuit breaker — a source where every URL is dead would never auto-pause again (pre-fix it raised → `record_run_failure` → breaker advanced to the threshold). Fixed: an all-error run (error count > 0, zero successful checks, zero items) now passes a counts-only synthetic error to `record_check_result` so the breaker **advances** and auto-pause fires exactly as before, and its status is `failed`; a **partial-success** run (≥1 working URL) still completes and resets the breaker — a reachable source is healthy. Feed/API arms are structurally unaffected.

## Testing

Discriminating tests: one poisoned URL among several → the others' items persist + `error` counter == 1 (reds under the old all-or-nothing code); all-error run pre-set one below threshold → breaker advances and `is_paused` becomes true (reds if the all-error detection is removed); partial run → items persist AND breaker resets (pins no over-correction). The existing auto-pause suite used `type="url"`, an arm this never touched, which is why it missed the regression. `test_local_watchlists_service.py` 19, `test_watchlists_runs_pane.py` 16, `test_scheduled_watchlist_runs.py` 13. Whole-branch review → fix wave → scoped re-review, both APPROVED; every behavioural change mutation-verified.

---
## Summary by cubic
Prevents a single bad URL from failing `url_list`/`sitemap` runs, adds an `"error"` disposition and counter to stats/UI, and keeps auto‑pause intact by failing all‑error runs. Scheduler metrics now surface partial failures with a `url_errors` label and a per‑run WARNING. Addresses Linear task-1394.

- **Bug Fixes**
  - Isolate each per‑URL check via `_check_url_isolated`; on exception, record `DISPOSITION_ERROR` (type-only) and continue.
  - Add an `"error"` counter to run stats and render it in the Runs pane Checks line.
  - All‑error runs (zero items, zero successes) send a counts‑only error to `record_check_result`, report `failed`, and advance the breaker; partial runs still complete and reset it.
  - Scheduler: completed runs with error dispositions set `url_errors=true` on `watchlist_checks` and log a per‑run WARNING; clean runs omit the label.
  - Sitemap fetch is unchanged; only per‑URL checks are isolated.

<sup>Written for commit aa7f9e0b24032edefde271cc551d43e70701ab9b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1263?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Scheduling/test_watchlist_check_handler.py
+++ b/Tests/Scheduling/test_watchlist_check_handler.py
@@ -129,6 +129,62 @@ async def test_every_executable_type_is_launched_as_a_run(handler, sub_type):
 
 
 @pytest.mark.asyncio
+async def test_partial_url_errors_are_visible_in_the_metric_and_a_warning(
+    handler, metrics_patch
+):
+    """TASK-1394 Qodo: a url_list/sitemap run that isolates a per-URL failure
+    still completes, so a status-only reader would metric it as a clean
+    success. The handler must surface the error count -- a `url_errors` metric
+    label and a per-run WARNING -- so a recurring dead URL is visible outside
+    the Runs pane.
+
+    Args:
+        handler: The `WatchlistCheckHandler` fixture (mock service/DB).
+        metrics_patch: `(log_counter, log_histogram)` patched to capture labels.
+    """
+    counter, _histogram = metrics_patch
+    handler.subscriptions_db.get_subscription.return_value = _subscription("url_list")
+    # A completed run whose dispositions record one isolated URL error.
+    handler.watchlists_service.execute_run.return_value = {
+        "run_id": 7,
+        "status": "completed",
+        "stats": {"new_items_found": 2, "dispositions": {"changed": 2, "error": 1}},
+    }
+
+    await handler.handle(_task())
+
+    _args, kwargs = counter.call_args
+    assert kwargs["labels"]["status"] == "success"
+    assert kwargs["labels"]["url_errors"] == "true", (
+        "a completed run with error dispositions must be distinguishable in "
+        "the metric from a clean success"
+    )
+
+
+@pytest.mark.asyncio
+async def test_a_clean_run_carries_no_url_errors_label(handler, metrics_patch):
+    """The `url_errors` label appears ONLY when a run had per-URL errors, so a
+    clean success is not tagged (TASK-1394 Qodo).
+
+    Args:
+        handler: The `WatchlistCheckHandler` fixture (mock service/DB).
+        metrics_patch: `(log_counter, log_histogram)` patched to capture labels.
+    """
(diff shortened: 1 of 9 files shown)


rmusser01:
Qodo adjudication:

**Finding 3 (partial URL errors under-logged) — confirmed real, fixed in `222d0810b`.** A partial-error url_list/sitemap run completes as `status='completed'`, so the scheduler's status-only metric counted it as a clean success. The handler now surfaces the count: a per-run WARNING (not per-URL — low noise, counts only) and a low-cardinality `url_errors: "true"` metric label, so a recurring dead/blocked URL is visible outside the Runs pane. Pinned by two handler tests (label present on a partial run, absent on a clean run), mutation-verified.

**Finding 2 (test docstring Args) — fixed.** `Args:` sections added to the four new service tests and the two new handler tests.

**Finding 1 (test reads outside `transaction()`) — declined, house precedent.** Read assertions via `db.conn.execute(...)` are the established convention in this repo's tests — this file already uses it 9 times pre-existing, and 13 test files across `Tests/Subscriptions`/`Tests/Watchlists` do the same. Wrapping a single SELECT in `transaction()` purely for a test assertion would make these the odd ones out without changing behavior (a SELECT is atomic as-is); the production CRUD in this change is all transaction-wrapped.

PR #1265, opened 2026-08-03
feat(library): ingest round-2 P2 batch — inline details, persistent form, dup forecast (task-2043)

rmusser01:
## Summary

The round-2 critique P2 batch (task-2043):

- **Inline error details** replace the ~4s uncopyable toast: Show details toggles per-row lines (category + full single-prefix message via a shared `unwrap_ingest_error` + an honest what-a-retry-could-fix hint when Retry is offered), rendered through the task-2042 in-place queue update.
- **The staged form survives rail switches** (only the debounce and armed clear reset; the deep-link entry keeps its documented full wipe) — the old wipe-on-entry punished multi-batch workflows.
- **Pre-flight duplicate forecast**: the worker hashes staged text files (thread-local db reads, 20-file/8MB caps; scoped to the generic group since the DB dedups on sha256 of PARSED content) — "1 file appears to already be in your Library — it will be matched, not re-imported." renders before submit.
- Select fields get the task-2012 labels; `StateGlyphCheckbox` carries on/off in the glyph (`▐✓▌`/`▐ ▌` — stock ToggleButton is color-only); queue counts say "— all ingests" (the registry restores prior sessions); `FileOpen` casing aligned; panel filler rows removed (Textual defaults + two unscoped `_conversations.tcss` rules; the `CollapsibleTitle height:3` needed a child-combinator specificity bump because it sits later in the bundle).

**Unmasked latent bug fixed (pre-existing since PR #717):** the options loader's two-arg `get_cli_setting(prefix, name)` reads treated `name` as the DEFAULT on dotted paths, so fresh profiles loaded the field-name STRING for every option — "Analyze after ingest" truthy-flipped ON and `type_options` filled with junk. The old wipe-on-entry masked it; this PR's form persistence exposed it live. Explicit `None` defaults now, pinned by a call-shape recording test.

## Evidence

- TDD per item; targeted suites: canvas+state+runner **235/235**, shell ingest subset **97/97**, full-tree `--collect-only` clean.
- Live (isolated profile): defaults correct (analyze off/chunk on/1000 — was corrupted), zero filler rows, inline details under the row, glyph states monochrome-distinct, dup forecast at dwell time, path survives a Media round-trip.

---
## Summary by cubic
Implements task-2043 P2 upgrades to Library ingest: inline row error details with Show/Hide, a session‑persistent form, and a pre‑flight duplicate forecast for text files. Adds UI polish and hardens preflight updates so other screens aren’t disrupted.

- **New Features**
  - Inline error details replace the toast; rows show category + full message via `unwrap_ingest_error`, with a retry hint when Retry is available; Show/Hide persists per job.
  - The staged form now persists across rail switches; only debounce and clear‑finished arm reset; deep‑link entry still resets.
  - Pre‑flight duplicate forecast for generic/text files: hashes staged content on the worker (thread‑local DB reads; 20 files/8MB caps) and surfaces a quiet summary; adds `PreflightResult.already_in_library`.
  - UI polish: labeled Selects, checkbox glyph shows on/off, counts line says "— all ingests", FileOpen title is "Import media", and panel filler rows removed.

- **Bug Fixes**
  - Options loader now calls `get_cli_setting` with explicit `None` defaults to avoid flipped defaults on fresh profiles.
  - Off‑canvas safety: preflight workers are generation‑fenced and dynamic updates no‑op when ingest isn’t selected, preventing late results from recomposing other screens.
  - Duplicate forecast reads go through `validate_path_simple`; >8MB files are skipped; dismiss/clear prunes expanded‑details state.

<sup>Written for commit 968618cf7c961201d3feb9ed4683984c222921e6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1265?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Addressed all three Qodo findings in the latest commit:

**Finding 1 (off-canvas pre-flight recompose) — fixed.** `_pause_library_ingest_transient_ui` now generation-fences + cancels the in-flight worker (the stored echo still persists per the form-preservation contract — only the worker is fenced), and `_update_library_ingest_dynamic_regions` no-ops when the ingest row isn't selected, so no late result can reach the full-recompose fallback while another canvas is showing.

**Finding 2 (unvalidated candidate reads) — fixed.** Duplicate-forecast candidates go through `validate_path_simple(..., require_exists=True)` before any filesystem touch — same validator as the submit path (the candidates already come from the pre-flight walk of a validated root; this is the rule-497145 defense-in-depth layer).

**Finding 3 (expanded-details pruning) — fixed.** The per-row dismiss discards its job id and the transient reset clears the set (Clear finished already did).

Affected suites after fixes: shell subset 17/17, canvas+state+runner 235/235.

PR #1267, opened 2026-08-03
feat(console): cost chip + cache-break alert (cost ticker PR3)

rmusser01:
PR3 — the user-facing payoff of the Console cost-ticker program. PR1 captured real usage, PR2 completed caching; this ships the chip people actually see.

**What lands**
- **Cost chip** (10th chip, last in the Console status strip): running conversation spend from PR1's persisted per-message usage rows, priced at read time through the PR1 catalog. Rows without usage fall back to the local estimator and are marked `~`; unknown models show `12.3k tok` rather than a fabricated figure; local providers show `$0.00`.
- **Cache-break alert**: passive `⚠ +$delta` with a tooltip naming the cause (model > system > history priority). Detection is a payload fingerprint recomputed **lazily** — only when a new per-session payload-revision counter moves, and never while a run is streaming. The baseline is recorded at dispatch from the **dispatched resolution** (not the controller's shared mutable fields — that race was caught in review) and from the **same raw payload stage** the recompute uses, so dictionary/world-info/RAG substitutions can't fire false alarms. Self-clearing: revert an edit and the alert goes away.
- **Cache state from ground truth**: warm `●` / cold `○` / none, derived from PR2's real `cache_read`/`cache_write` fields, with an Anthropic-only 5-minute TTL on a monotonic clock and a 10s audited repaint timer (`console-cost-ttl`, torn down on unmount) — needed because the 0.2s transcript tick stops when a run completes.
- **Cost-breakdown modal** on chip click: per-message rows (buckets, cost, `~` markers) plus totals including regenerated variants.

**Verification**
- 3,164 tests green across `Tests/Chat/`, the new cost-chip screen suite, resume-path, and `Tests/LLM_Calls/`.
- Live-verified in tmux at ~200×50 and 80×24: chip renders last on the first frame, strip stays height-1 with no wrap or composer displacement (chips clip under the strip's pre-existing overflow at 80 cols — unchanged behavior), hover tooltip and the modal (open, rows, Escape/Close, reopen) all correct.
- **Honest limitation:** no provider was reachable in the verification environment, so the real dollar-after-send, `⚠` alert, and `○` expired states were exercised by unit tests only, not on screen. Worth a look with a live key before relying on them visually.

**Review-caught defects fixed along the way** (all with regression tests): the baseline/resolution race, the baseline-vs-recompute stage mismatch, role-blind estimate pricing (assistant tokens were priced at input rates), un-normalized provider keys hitting the estimator ratio table, a tokens-only tooltip that went silent about cache state, and a mid-stream cost-animation violation the written tests never caught.

Spec + plan ride in the branch under `Docs/superpowers/`.

---
## Summary by cubic
Adds a Console cost chip that shows running spend, cache state, and a passive cache-break delta, plus a click-through breakdown modal. Uses real usage rows with pricing and a local estimator fallback, lazy fingerprinting, TTL-based cache state, and avoids mid‑stream cost changes.

- **New Features**
  - Cost chip (last status pill): dollars with `~` for estimated rows; unknown models show token counts; local providers show `$0.00`; compact label on narrow widths; initial state renders at compose.
  - Cache-break alert: `⚠ +$delta` with a tooltip cause; baseline recorded at dispatch from the dispatched resolution; lazy recompute gated by a per-session payload-revision counter and skipped during streaming; self-clearing on revert.
  - Cache state from ground truth (`cache_read`/`cache_write`): warm `●` / cold `○` / none; Anthropic-only 5‑min TTL with a 10s `console-cost-ttl` repaint; click opens a per-message breakdown modal; added `.console-chip-cold`.

- **Bug Fixes**
  - Fingerprint baselines (no controller-field race) and stage match (baseline/recompute use the same raw payload stage).
  - Role-aware estimate pricing (assistant rows at output rates) and provider normalization via `provider_config_key`; tokens-only tooltips now narrate cache state.
  - Prevented mid-stream cost changes by excluding pending/streaming rows; ensured per-session chip state isolation across tabs.
  - Skipped projected delta computation when no break is present; switched to `logger.opt(exception=True)` so tracebacks log correctly.

<sup>Written for commit 7e50566090a014488cc8225db81ab0bb8341a91b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1267?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
All four Qodo findings addressed in `e55f594c9`:

- **Wasted delta estimation — fixed.** Confirmed the value is genuinely dead without a break reason: `build_cost_state` reads `projected_delta_usd` only under `if alert and ...`, and `alert` requires a truthy `break_reason`; the tooltip's `~+$` suffix is likewise nested inside `if break_reason:`. So the WARM gate now requires a break reason before estimating — no partial narrowing needed. A new spy test proves zero estimator calls on a WARM-without-break sync and exactly one once a reason exists, and the alert test still asserts the delta renders.
- **loguru tracebacks — fixed.** Converted all 7 `exc_info=True` sites this PR introduced (4 in `console_cost_tracker.py`, 3 in `chat_screen.py`) to `logger.opt(exception=True)`, matching the repo's dominant idiom (~1,044 existing uses), and sanity-checked that a real traceback now prints. Two related spots left alone and noted rather than silently swept: a pre-existing `exc_info=True` at `chat_screen.py:12747` that predates this PR, and three `console_chat_controller.py` handlers using `logger.bind(..., error=repr(exc))` — also traceback-less, but outside this finding's literal scope. Happy to fold either in if you'd prefer.
- **Import ordering — fixed**: the new local import moved into the post-loguru local group.
- **Missing docstring — fixed** on `action_open_cost_breakdown`.

Gates: 51 passed (cost chip screen + tracker + chips), 282 passed (controller + store).

PR #1268, opened 2026-08-03
fix: make watchlist auto-pause live across both failure paths, with resume-on-success (task-1410)

rmusser01:
## Why

Watchlist auto-pause was 100% dead code: `subscriptions.auto_pause_threshold` (schema default 10) and `auto_pause_after_failures` (config) were compared in exactly one place — `record_check_result`'s `if error:` branch — which had no caller, while the main failure path (`record_check_error` via `record_run_failure`) bumped `consecutive_failures` forever, never consulted the threshold, and *cleared* any existing `is_paused` on every failure. A dead source was retried on its cadence indefinitely (task-1410).

## What

- **AC#1 (prerequisite): a failure never un-pauses.** `record_check_error`'s `is_paused` write is now `CASE WHEN <pause> THEN 1 ELSE is_paused END` — a recorded failure can pause but never clear an existing pause.
- **AC#2: the main failure path auto-pauses.** A new shared `_advance_failure_and_maybe_pause` helper (the increment + threshold check + conditional pause) is called by BOTH failure paths — `record_check_error` and `record_check_result`'s error branch (which task-1394 just made reachable for all-error runs) — so the two cannot diverge. A source that fails `auto_pause_threshold` times in a row pauses, exercised through the real path (`record_run_failure` → `record_check_error`), not by calling the DB method directly.
- **AC#3: the two knobs reconciled.** `auto_pause_after_failures` (config) seeds the `auto_pause_threshold` column default for new subscriptions; a per-subscription column value overrides (stated precedence, documented). Existing rows untouched.
- **Review fix — resume-on-success (auto-pause is otherwise a permanent trap).** Turning on auto-pause with no un-pause path would strand sources forever (the scheduler skips paused sources; the only `is_paused=0` writer had no caller). The success branch now clears `is_paused` too, giving the documented recourse: a user's **manual re-check** of a paused source runs (no paused guard on `launch_run`/`execute_run`) and, if it succeeds, resumes it. A dedicated Resume-button UI is filed as follow-up **task-2050**.
- Defensive: a NULL/0/negative threshold (a direct bad `update_subscription`) is treated as "auto-pause disabled" rather than crashing or pausing instantly; the previously always-false `auto_paused` metric label is wired to the helper's real signal.

## Testing

Auto-pause through the real failure path (threshold=3 pauses on exactly the 3rd, not the 2nd); a failure never un-pauses (AC#1); resume-on-success through `launch_run`/`execute_run`; both failure paths pause at the same threshold via the shared helper (a mutation reds a task-1394 test too, proving it's genuinely shared); config seed + override + safe fallback; bad-threshold guard. Every behavioural change mutation-verified. Opus whole-branch review (CLEAN vs the 3 ACs, flagged the stranding) → fix wave → scoped re-review, APPROVED. 87 targeted tests pass; `--collect-only` clean.

**Note:** this turns auto-pause on as a real behavior (sources that fail repeatedly will pause) — the constructive reading of AC#3's fork, since the schema/config/dead-branch all intend it. If you'd rather remove auto-pause instead, the alternative is a smaller change; say the word.

---
## Summary by cubic
Make watchlist auto-pause real and consistent across both failure paths, and add resume-on-success so paused sources can recover. Align the per-source threshold with config and guard invalid values (task-1410).

- **Bug Fixes**
  - Auto-pause works on both failure paths via a shared `_advance_failure_and_maybe_pause`; pauses when `consecutive_failures >= auto_pause_threshold`.
  - A failure never clears an existing pause. A successful check now resets counters and sets `is_paused = 0`, so a manual re-check can resume a source.
  - Reconciled knobs: `[subscriptions].auto_pause_after_failures` seeds `auto_pause_threshold` for new subscriptions; explicit per-source values override; existing rows unchanged; `None`/0/negative thresholds are treated as “disabled,” not errors.
  - Auto-pause reports once per transition (0→1) only: the warning and `auto_paused` metric fire once when pausing; fixed the metric label; added targeted DB and service tests.

<sup>Written for commit d46bd930c341c3458f03419911400f24f8e74478. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1268?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/DB/test_subscriptions_db.py
+++ b/Tests/DB/test_subscriptions_db.py
@@ -176,6 +176,82 @@ def test_record_check_error_never_unpauses_an_already_paused_subscription(db):
     assert row["last_error"] == "connection refused"
 
 
+def test_record_check_result_success_resumes_an_auto_paused_subscription(db):
+    """Fix wave for the task-1410 review, Finding #1 (the important one).
+
+    task-1410 made auto-pause live across both failure paths, but the only
+    ``is_paused = 0`` writer (``reset_subscription_errors``) has zero
+    callers, the scheduler skips paused sources
+    (``get_pending_checks``/``WatchlistCheckHandler``), and
+    ``record_check_result``'s success branch did not touch ``is_paused`` --
+    so an auto-paused source could never be un-paused by ANY path. A
+    successful check is the natural recourse: a failure never un-pauses
+    (AC#1, pinned above), but a success now does.
+
+    Reds if the success branch's new ``is_paused = 0`` write is reverted:
+    ``is_paused`` stays 1 even though the check just succeeded.
+    """
+    source_id = db.add_subscription(
+        name="Docs",
+        type="rss",
+        source="https://a.example/feed",
+        auto_pause_threshold=3,
+    )
+    with db.transaction() as conn:
+        conn.execute(
+            """
+            UPDATE subscriptions
+            SET is_paused = 1, error_count = 3, consecutive_failures = 3,
+                last_error = 'connection refused'
+            WHERE id = ?
+            """,
+            (source_id,),
+        )
+
+    db.record_check_result(subscription_id=source_id, items=None, error=None)
+
+    row = db.get_subscription(source_id)
+    assert row["is_paused"] == 0, "a successful check must resume an auto-paused subscription"
+    assert row["consecutive_failures"] == 0
+    assert row["error_count"] == 0
+    assert row["last_error"] is None
+
+
+@pytest.mark.parametrize("bad_threshold", [None, 0, -1])
+def test_advance_failure_and_maybe_pause_never_pauses_on_a_bad_threshold(db, bad_threshold):
+    """Fix wave for the task-1410 review, Finding #3.
+
+    ``_advance_failure_and_maybe_pause``'s threshold comparison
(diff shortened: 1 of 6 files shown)


rmusser01:
Qodo adjudication:

**Finding 2 (auto-pause over-counted) — confirmed real, fixed in `d46bd930c`.** The helper reported a pause (returned true + logged the WARNING) on every over-threshold failure, so a failing manual re-check of an already-paused source would re-count a single pause event in the `auto_paused` metric and re-log. Auto-pause is a state transition: the SELECT now also reads `is_paused`, and the log + `newly_paused` return fire only on the 0→1 transition. Pinned by `test_auto_pause_is_a_transition_not_reported_on_every_failure` (first failure below threshold → False; the crossing failure → True; a further failure on the paused source → False), mutation-verified.

**Finding 1 (test docstring Args) — fixed.** `Args:` sections added to the new tests in both files.

PR #1270, opened 2026-08-03
refactor: rename side-effecting Watchlists predicate to an action name (task-1349)

rmusser01:
## Why

`_content_toggle_is_blocked` reads as a pure query but calls `self.notify(...)` before returning `True`. Harmless at its three call sites (all user gestures where the toast is wanted), but the name gives no warning — and this codebase has already been bitten by exactly this shape: `provider_is_configured()`, a predicate, wrote an `eval_models` row from `compose()`, so opening the Evals screen mutated the DB on every fresh install. A side-effecting predicate is safe only until someone wires it into a render path (task-1349).

## What

- **AC#1 (rename arm):** `_content_toggle_is_blocked` → `_refuse_content_toggle_off_read_tab`. The notify stays in the helper (three call sites all want the same toast; moving it would duplicate the string), so the fix is to name it as the action it is — the verb is the warning the innocent name never gave. The docstring now states the naming rule and cites the `provider_is_configured()` incident. All five call sites and the doc cross-reference updated.
- **AC#2:** an AST scan confirms this was the *only* predicate-named function with a side effect across `UI/Watchlists_Modules/` and `watchlists_collections_screen.py`; the other five predicates are pure. Kept as a permanent regression test (`test_no_side_effecting_predicates.py`) so a new side-effecting predicate fails there rather than being discovered from a render path.

## Testing

The refusal-behavior tests (`test_watchlists_content_pane.py`, 40) pass unchanged — they exercise the toast through the public gesture, not the private name. The new guard test passes and is mutation-verified: renaming back to the predicate form reds it.

---
## Summary by cubic
Renamed a side‑effecting predicate in Watchlists to a clear action and added a guard test to block future side‑effecting predicates. Satisfies TASK-1349 by making side effects explicit and preventing regressions.

- **Refactors**
  - `_content_toggle_is_blocked` → `_refuse_content_toggle_off_read_tab`; keeps `self.notify(...)`; updated three call sites and expanded docstring with rename history and the naming rule.
  - Added `Tests/Watchlists/test_no_side_effecting_predicates.py`: AST-scan runs recursively over `UI/Watchlists_Modules` and `watchlists_collections_screen.py` with explicit UTF‑8 reads; fails if predicate‑named functions call side‑effect methods.

<sup>Written for commit 85a6c6e142e8f943ee5a54b3eff7df8fe7c678e9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1270?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Watchlists/test_no_side_effecting_predicates.py
+++ b/Tests/Watchlists/test_no_side_effecting_predicates.py
@@ -46,7 +46,9 @@ _SIDE_EFFECTS = {
 def _python_files():
     for target in _WATCHLISTS_DIRS:
         if target.is_dir():
-            yield from target.glob("*.py")
+            # Recursive: a predicate in a future nested subpackage must be
+            # covered too (Qodo, TASK-1349).
+            yield from target.rglob("*.py")
         elif target.is_file():
             yield target
 
@@ -69,7 +71,10 @@ def _side_effects_in(node: ast.AST) -> set[str]:
 def test_no_predicate_named_function_in_watchlists_has_a_side_effect():
     offenders = []
     for path in _python_files():
-        tree = ast.parse(path.read_text())
+        # Explicit UTF-8: these sources carry non-ASCII glyphs (e.g. the
+        # `▸` collapsed-region marker), so a default-locale read could
+        # UnicodeDecodeError or mis-decode on some runners (Qodo, TASK-1349).
+        tree = ast.parse(path.read_text(encoding="utf-8"), filename=str(path))
         for node in ast.walk(tree):
             if not isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)):
                 continue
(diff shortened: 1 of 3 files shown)


rmusser01:
Qodo adjudication — all three confirmed and fixed in `85a6c6e14`:

**Finding 1 (locale-dependent read) — fixed.** The guard scans sources containing `▸`; now `read_text(encoding="utf-8")` + `ast.parse(..., filename=str(path))` for clearer parse errors.

**Finding 2 (non-recursive scan) — fixed.** `glob("*.py")` → `rglob("*.py")` so a predicate in a future nested subpackage is covered.

**Finding 3 (rename-history docstring) — fixed, and it was a real bug my own global rename introduced.** The docstring had cited the OLD name `_content_toggle_is_blocked`, but the `sed`-based rename rewrote that mention too, leaving it claiming the *new* name existed "before TASK-1349". Corrected to "it was `_content_toggle_is_blocked` before TASK-1349". Good catch.

PR #1278, opened 2026-08-03
fix: gate FEEDS and all centre-region gestures to the Read tab (task-1344)

rmusser01:
## Why

The spec: "Only Read uses the three-pane split. Sources, Runs, Rules, and Artifacts take the full centre width." Three gaps remained after Phase D (which fixed CONTENT gating): FEEDS was mounted on every tab, `z`/solo on off-tab regions was ungated, and the CONTENT gate collapsed to a one-row header rather than clearing the width (task-1344).

## What

- **FEEDS gated to the Read tab (AC#1)**, mirroring CONTENT. Design subtlety found in implementation: FEEDS's factory also built the always-visible tab strip and the snapshot loading/error/empty/summary markers the whole app relies on seeing everywhere — so those were extracted into an always-rendered `WatchlistsWorkbench(header=…)`; only the region body is gated.
- **All centre-region gestures refused off the Read tab (AC#2).** `z`/toggle/solo on FEEDS, ITEMS, or CONTENT while off Read is refused with a notify. The whole-branch review caught the load-bearing case: ITEMS is *shown* off Read (as each section's full-width pane) but toggling/soloing it there was silently mutating and **persisting** the shared `region_layout`, so returning to Read rendered three headers over an empty centre — a dead-end written to disk that survives restart (the exact broken-forever class Phase D/TASK-998 fixed). The gate now covers every centre region, with truthful per-region copy (ITEMS: "The pane layout can only be changed on the Read tab.", since ITEMS *is* visible off Read).
- **No dead-end layout (AC#3):** a test drives tab-switch + toggle/solo sequences (including the off-Read ITEMS path) and asserts a centre region is always expanded and that a refused gesture persists nothing. Three pre-existing tests that had unwittingly relied on the bug (toggling ITEMS from the default overview section) were corrected to act on the Read tab — same assertions, legitimate path.
- **Mount policy (AC#4): UNMOUNT**, not a one-row header — it's the spec's literal "full centre width," the tab switch already fully recomposes so no new focus-loss mode is introduced, and persistence only ever reads the real `region_layout` (a new `hidden` set carries the tab-gating and is never persisted, so it can't recreate a collapsed-forever state).

## Testing

Region/workbench/screen suites: 124 passed on the rebased tree. Every behavioural change mutation-verified (narrowing the gate reds the ITEMS-off-Read tests; both notify paths `markup=False`; the task-1349 side-effecting-predicate AST guard stays green). Opus whole-branch review found the ITEMS-persist blocker → fix wave → scoped re-review confirmed the 3 modified pre-existing tests kept their original assertions (not weakened) and the gate refuses nothing on Read. CSS bundle regenerated from source, not hand-edited.

---
## Summary by cubic
Gates FEEDS and all centre-region gestures to the Read tab and unmounts gated regions, preventing dead-end layouts and matching the spec. Connects to Linear task-1344 (AC#1–#4) and keeps the off-Read header accurate and inert.

- **Bug Fixes**
  - FEEDS now shows only on the Read tab; FEEDS and CONTENT unmount on other tabs (AC#1, AC#4).
  - Refuse `z`/`Z`/toggle for FEEDS, ITEMS, and CONTENT off Read with clear notify copy; prevents off-tab gestures from mutating/persisting layout (AC#2).
  - Guaranteed non-empty centre: tests assert tab switches and gestures never leave zero expanded panes; fixes path where off-Read ITEMS toggle persisted an empty Read layout (AC#3).
  - Off-Read header now refreshes on scope change via `refresh_header_content()` so the tab strip and snapshot markers stay correct.
  - Ignore `z`/`Z` while focus is in the centre header; reset the header-focus sentinel on tab switch and on non-region focus to avoid stray toggles.
  - Updated tests to run FEEDS assertions on the Read tab, avoid off-Read ITEMS toggles, and cover header refresh.

- **Refactors**
  - Introduced `_hidden_centre_regions` and `_rendered_region_layout`; hiding is orthogonal to `RegionLayout.collapsed`, forcing ITEMS visible off Read without touching persisted state.
  - Extracted the tab strip + snapshot markers into an always-rendered header `#wl-centre-status` via `WatchlistsWorkbench(header=…)` so they remain visible when FEEDS is hidden.
  - `WatchlistsWorkbench` now accepts `hidden` regions and an optional `header`; sole-expanded logic accounts for hidden regions. CSS adds `#wl-centre-status { height: auto; }` to avoid competing with centre pane height.

<sup>Written for commit 855bb929648752c3b9f0a12bdee700b31db7ca10. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1278?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Qodo adjudication — both correctness findings confirmed and fixed (`3e2fe527e`), plus a re-review hardening (`855bb9296`):

**Finding 1 (off-Read header stale on scope change) — fixed.** `watch_tree_scope` refreshed only FEEDS, a no-op when FEEDS is unmounted off Read, so `#wl-centre-status` showed stale scope/counts. New `WatchlistsWorkbench.refresh_header_content()` rebuilds the header in place (never a screen recompose — that would remount the Inspector, the very thing `watch_tree_scope` avoids), called when `active_section != "items"`. Test: change scope off Read → header shows the new counts; mutation-verified.

**Finding 2 (z/Z act on a stale region while focus is in the header) — fixed.** The header/tab-strip sits outside the region wrappers, so `on_descendant_focus` never tracked it and `z` could collapse+persist whatever `focused_region` last pointed at (a rail). A `_focus_in_centre_header` sentinel (set on header focus, cleared on region focus and on tab switch) now makes toggle/solo no-op when focus is in the header. Mutation-verified: without the toggle guard, `z` persisted `collapsed_regions=['left_rail']`. (The solo guard is defense-in-depth — the off-Read gate already blocks its only mutation path — so it only suppresses an extra notify there; kept for consistency.)

**Re-review hardening.** The sentinel's ancestor walk had no `else`, so focusing a widget in neither zone (e.g. `#watchlists-backend-select`) left a stale `True`. Non-blocking today (every region-focus resets it on entry) but a genuine invariant hole — added an explicit reset so the flag is `True` only while focus is truly in the header, with a test (mutation-verified).

PR #1287, opened 2026-08-03
feat: per-rule appeared/disappeared/anywhere scope for content alerts (task-1363)

rmusser01:
## Why

Content-alert rules match against the full page text (TASK-1343 preserved this deliberately, so a long-standing alert doesn't silently stop firing). But "tell me when this phrase **appears**" — matching only newly-added text — is a genuinely useful thing to want from a site watcher, and the diff now makes it cheap to compute. Same for "when it **disappears**". This adds it as a per-rule opt-in, not a change to the default (task-1363).

## What

- **`build_rule_haystack(item, scope="anywhere")`** gains a scope param:
  - `anywhere` (default): the whole page — **byte-identical** to before (verified by md5 in review). This is what makes AC#2 free: `WatchlistFilterService` (exclude filters) never passes a scope, so a narrowed scope can never admit an excluded item.
  - `appeared`: only the newly-added page text (a site change's added segments). Not the item's own metadata — a change item's title is the synthetic "Change detected: <source name>", so folding it in would fire on every check; `appeared` matches only what is genuinely new (review refinement, symmetric with `disappeared`).
  - `disappeared`: only the removed page text. Absent delta → empty → never matches (nothing is known to have left).
  - A feed/API item (no diff, wholly new) under `appeared` → the whole item just appeared, so it falls back to the page-wide haystack; under `disappeared` → matches nothing.
  - Unknown/None/non-string scope → `anywhere` (safe default).
- **`monitoring_engine.check_url`** attaches `rule_match_added_text`/`rule_match_removed_text`, computed by `added_and_removed_text` (SequenceMatcher opcodes over the same `_segment_for_diff` segmentation the shown diff uses) — matching-only, non-persisted, like `rule_match_text`.
- **`WatchlistContentAlertService`** reads each rule's `conditions["scope"]` and passes it. The run-level alert system (`local_watchlist_alert_rules`) is untouched.

No create-UI exists for these pattern-notify rules today; the scope lives in the rule's `conditions` JSON and the service honors it (a future rule editor is where it'd be surfaced) — AC#1 "can be set" is satisfied at the rule-data/service layer, which the tests exercise directly.

## Testing

Each scope; absent scope == page-wide (AC#3 regression); exclude filter stays whole-page (AC#2); the producer attaches the right added/removed text; the appeared-ignores-synthetic-title refinement. 60 passed across the four targeted files. Every behavioural change mutation-verified (scope-passthrough drop, added/removed swap, appeared-includes-title). Opus whole-branch review: CLEAN — the `anywhere` default proven byte-identical, non-persistence confirmed, both original mutations reproduced.

---
## Summary by cubic
Adds per‑rule scope for content alerts to match newly added text, removed text, or anywhere on the page. Default stays page‑wide; exclude filters are unchanged (TASK-1363).

- **New Features**
  - `build_rule_haystack(item, scope)` supports `anywhere` (default), `appeared`, and `disappeared`; unknown values fall back to `anywhere`. Feed/API items fall back safely (`appeared` → page‑wide, `disappeared` → empty).
  - `URLMonitor.check_url` attaches `rule_match_added_text` and `rule_match_removed_text` via `added_and_removed_text`, reusing `_segment_for_diff` so added/removed align with the shown diff; keys are matching‑only (not persisted).
  - `WatchlistContentAlertService` reads `conditions.scope` per rule and caches haystacks per scope; `WatchlistFilterService` remains page‑wide since it never supplies `scope`.

- **Bug Fixes**
  - Under `appeared`, ignore the synthetic change title/metadata; match only the added page text to prevent noisy matches.

<sup>Written for commit 2ee1d8a94ad189a1078d198d6f924399011a58cc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1287?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Subscriptions/test_watchlist_content_alert_service.py
+++ b/Tests/Subscriptions/test_watchlist_content_alert_service.py
@@ -43,6 +43,37 @@ def _change_item(**overrides):
     return item
 
 
+def test_appeared_scope_ignores_the_synthetic_change_title_and_metadata(service):
+    """task-1363 review: "appeared" matches ONLY the added page text, never the
+    change item's own metadata. A site change's title is the synthetic
+    "Change detected: <source name>", present on every check, so a pattern that
+    sits in the source name (here "Test source") must NOT fire under "appeared"
+    -- otherwise the scope that exists to cut page-wide noise would itself fire
+    on every change. Reds if "appeared" folds title/summary/author back into
+    the haystack.
+    """
+    item = _change_item()  # title = "Change detected: Test source"
+
+    source_name_rule = [{
+        "id": 9,
+        "name": "Matches the source name in the synthetic title",
+        "conditions": {"type": "keyword", "pattern": "Test source", "scope": "appeared"},
+    }]
+    assert service.evaluate(item, source_name_rule) == [], (
+        "the synthetic change title is not text that appeared on the page; an "
+        "'appeared' rule must not match it"
+    )
+    # Sanity: the same pattern DOES match under "anywhere" (it is in the title),
+    # proving the item genuinely carries it and the appeared-scope miss is the
+    # scoping, not a missing field.
+    source_name_anywhere = [{
+        "id": 10,
+        "name": "Same pattern, anywhere scope",
+        "conditions": {"type": "keyword", "pattern": "Test source", "scope": "anywhere"},
+    }]
+    assert [m["rule_id"] for m in service.evaluate(item, source_name_anywhere)] == [10]
+
+
 def test_appeared_scope_matches_added_text_but_not_an_unchanged_old_phrase(service):
     """AC#1/#3: "appeared" matches text a change introduced, not text that was
     already on the page before the change -- even though the old phrase is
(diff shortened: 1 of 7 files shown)


rmusser01:
Qodo adjudication — all three addressed in the perf commit:

**Finding 1 (double diff computation) — fixed.** `build_change_diff` and `added_and_removed_text` gained optional pre-segmented params; `check_url` now segments each side ONCE and hands the segments to both, so a page (up to 10MB) isn't segmented twice per changed URL. Behaviour-preserving — the delicate memory-bounded diff-body logic is untouched, and a new equivalence test pins that the pre-segmented path is byte-identical to the self-segmenting path.

**Finding 3 (haystack rebuilt per rule) — fixed.** `evaluate` caches the haystack per NORMALIZED scope, so the page-wide join is built at most once per distinct scope regardless of rule count. Pinned by a new mixed-scopes-in-one-call test (appeared/disappeared/anywhere rules together, each gets its own slice) — mutation-verified (a constant cache key reds it).

**Finding 2 (import blank line) — fixed.** Blank line between the pytest and project import groups.

62 passed across the four targeted files; both new perf tests mutation-verified.

PR #1288, opened 2026-08-03
Redact URLs in egress exception strings; add the leak tests the redaction never had (TASK-1722)

rmusser01:
Closes TASK-1722. Egress credential redaction: the log sites were already fixed (PR #1173 adopted `_log_origin()`); this PR closes the leak that remained and adds the test coverage the fix never had.

## The remaining leak: exception strings

`str(EgressBlockedError)` embedded the full URL (`"Egress blocked (reason) for {url}"`), and `EgressFetchError` appended ` [{url}]`. Callers log these verbatim — `Article_Extractor_Lib.py:344` does `logging.warning(f"...: {e}")`, and `acquisition.py` and `monitoring_engine.py` have equivalent sites — so a blocked presigned URL still landed its token in the log, one frame above the redacted module.

Both messages now render through `_log_origin(url)`. The **`.url` attribute keeps the full URL**: the redaction boundary is `str()`/`repr()` — what reaches logs and users — not what programmatic consumers can access, and a test pins that attribute contract so it cannot silently regress either way.

## Test coverage added (none of this existed)

- Marker-absence tests (`SECRET-TOKEN-MARKER` in the query string) on **both log sites and both exceptions' `str()` and `repr()`** — asserting the token absent while the host remains visible, since operators need to know which host was blocked.
- Direct `_log_origin()` edge coverage: userinfo/query/fragment stripping, IPv6 bracketing, and a 10-case never-raises parametrization over non-http(s) and garbage input — a logging helper that can throw turns a blocked request into a crash.
- Sweep confirmed the module's third log site (`warn_insecure_ssl`) takes a bare host by construction, and no test pinned the old exception-string form.

Both fixes sabotage-verified independently (raw URL restored → marker test red → reverted), one exception per verifier so each test's load-bearing status was proven separately.

## Verification

`test_egress.py` 94/94; full caller sweep (egress tests + `Tests/Model_Artifacts/` + every file matching `Utils.egress|check_url`) 1155 passed, 1 pre-existing unrelated failure (missing `playwright`, reproduces with zero egress changes).

Out of scope, noted for a future sweep: `Article_Extractor_Lib.py:344` also logs its own raw `url` variable in the same line — caller-side logging of caller input, not fixable from `egress.py`.

---
## Summary by cubic
Redacts URLs in egress exception messages and expands tests to prevent credential leaks, satisfying TASK-1722. Messages show only scheme+host via `_log_origin`, and `.url` still holds the full URL for programmatic use.

- **Bug Fixes**
  - `EgressBlockedError` and `EgressFetchError` now build messages with `_log_origin(url)`; host stays visible, secrets do not.
  - Tests assert no secret markers in logs and in both exceptions’ `str()`/`repr()`; cover `_log_origin` edge cases (userinfo/query/fragment stripping, IPv6, non-http/invalid input => "<invalid-url>") and the no-URL case for `EgressFetchError`.

- **Refactors**
  - Added concise Google-style docstrings to clarify the redaction boundary and arguments.

<sup>Written for commit 7c859ed4e00d96ad4b4301e2d0dbadeee56aa7ff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1288?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Utils/test_egress.py
+++ b/Tests/Utils/test_egress.py
@@ -226,6 +226,91 @@ def test_disabled_egress_log_redacts_url_credentials(monkeypatch):
         assert sensitive_value not in messages[0]
 
 
+# ---------------------------------------------------------------------------
+# _log_origin: credential-free URL label for transport logs (TASK-1722)
+# ---------------------------------------------------------------------------
+
+def test_log_origin_strips_userinfo_query_and_fragment():
+    """_log_origin renders scheme://host[:port] only -- no userinfo/path/query/fragment."""
+    label = egress._log_origin(
+        "https://user:SECRET-TOKEN-MARKER@example.test:8443"
+        "/path?token=SECRET-TOKEN-MARKER#frag"
+    )
+    assert label == "https://example.test:8443"
+    assert "SECRET-TOKEN-MARKER" not in label
+
+
+def test_log_origin_omits_port_when_absent():
+    assert egress._log_origin("https://example.test/path?x=1") == "https://example.test"
+
+
+def test_log_origin_ipv6_host_is_bracketed():
+    assert (
+        egress._log_origin("https://[2001:db8::1]:9443/x?y=SECRET-TOKEN-MARKER")
+        == "https://[2001:db8::1]:9443"
+    )
+    assert egress._log_origin("http://[::1]/") == "http://[::1]"
+
+
+@pytest.mark.parametrize(
+    "url",
+    [
+        "ftp://example.test/file?token=SECRET-TOKEN-MARKER",
+        "file:///etc/passwd",
+        "mailto:user@example.test",
+        "data:text/html,hi",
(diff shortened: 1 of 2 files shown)


rmusser01:
Both rule violations addressed in `7c859ed4e`: Google-style docstrings with `Args:` on both redacting `__init__`s (the redaction-boundary rationale moved from comments into them), and `Args:` sections on the three new test functions taking parameters. 94/94 green.

PR #1290, opened 2026-08-03
feat: full-page and previous-snapshot reader affordances (task-1494)

rmusser01:
## Why

The Watchlists design spec's content-pane mockup promised `[full page]` and `[previous snapshot]` on a change item, reading from `url_snapshots`. Phase D shipped both renderers but never wired the buttons, so since TASK-1343 made `content` the *diff*, the reader shows what changed but offers no way to see the page it changed on — the data is stored (per-`(subscription, url)`, newest 3 kept) and unreachable (task-1494).

## What

- **DB + service:** `Subscriptions_DB.get_url_snapshots(subscription_id, url, limit=2)` — parameterized, `ORDER BY created_at DESC, id DESC` (byte-identical to `_store_snapshot`'s prune order, so "newest"/"previous" agree with what's kept; same-second ties broken by `id DESC`). Async service wrapper on `LocalWatchlistsService`.
- **Content-pane affordances:** on a `change`-kind item only (gated on the imported `CONTENT_KIND_CHANGE` constant — article items have no snapshots), two compact `Full page` / `Previous snapshot` buttons post `ViewSnapshotRequested(item, which)`. The pane holds no DB handle.
- **Screen handler:** fetches the two snapshots; index 0 = full page, index 1 = previous. **Honest degradation (AC#2):** a first change (only one snapshot) or a pruned/absent one → an honest `markup=False` toast, never an empty modal or an IndexError.
- **`SnapshotViewModal`:** url + captured-at header and the snapshot body, both rendered through `rich.text.Text` so remote page content is **inert (AC#3)** — never parsed markup, never a live OSC-8 hyperlink (the TASK-1348 / `render_article` escaping doctrine, applied at the terminal display step). Escape/Close dismisses; content is passed in at construction, no re-query.

## Testing

DB query (order, same-second tie-break, `(sub,url)` scoping, fewer-than-limit, empty); content-pane gating (both buttons on a change item, neither on an article, right `which` posted); screen behavior (full-page → modal with newest content; previous-with-one-snapshot → honest toast, no modal; AC#3 — a snapshot containing `[bold red]INJECTED[/]` renders as those literal characters, verified through a real Console with no ANSI/OSC-8). Every behavioural change mutation-verified (query ASC, gate `if True`, modal `markup=True`). Opus whole-branch review: **CLEAN** — the AC#3 inertness confirmed live on *both* body and header, the 1-snapshot IndexError guard confirmed, the query/prune order match confirmed, the CSS bundle confirmed a faithful regeneration.

**Possible follow-up (non-blocking, from review):** the three action buttons stack vertically, spending ~3 of the pane's ~9 usable rows; a `Horizontal` for the two snapshot buttons could reclaim one row. Left as-is — the vertical tradeoff is documented in-code, and doing the horizontal layout properly needs its own real-CSS on-screen geometry assertion.

---
## Summary by cubic
Adds “Full page” and “Previous snapshot” buttons for change items in the reader, opening stored page snapshots in a modal with inert rendering. Meets TASK-1494 ACs; snapshot reads are transaction-wrapped and the requested snapshot type is validated.

- **New Features**
  - DB: `get_url_snapshots(subscription_id, url, limit=2)` returns newest-first with `ORDER BY created_at DESC, id DESC` (matches prune order; ties by `id DESC`).
  - Service: async passthrough `get_url_snapshots` on the local watchlists service.
  - UI: content pane shows compact buttons only for `change` items; posts `ViewSnapshotRequested(full_page|previous)`.
  - Screen: loads newest/second-newest snapshot; warns when missing; opens `SnapshotViewModal`.
  - Modal: shows URL, captured time, and body using `rich.text.Text` so remote content renders as literal text (no markup, no links).
  - CSS: styles for the snapshot modal and scrollable body.

- **Bug Fixes**
  - Wrap snapshot reads in a DB transaction for consistency with other reads.
  - Refuse unknown `which` values instead of defaulting, preventing the wrong snapshot from opening.

<sup>Written for commit 069ef0d120e4c19eb793e5c6f39903e6f933c44b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1290?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Qodo adjudication — all three addressed:

**Finding 1 (read outside `transaction()`) — accepted, fixed.** Unlike the earlier ChaChaNotes decline (bare reads are that file's convention), `Subscriptions_DB` genuinely wraps its sibling reads — `get_item_status`, the method directly above, uses `transaction()`. `get_url_snapshots` now conforms, with a comment noting the per-file convention.

**Finding 2 (test helpers' manual commits) — fixed.** Both new snapshot-seeding helpers (`_insert_snapshot` in the DB tests, `_seed_change_item_with_snapshots` in the content-pane tests) now write inside `with db.transaction() as conn:` per house convention. The other `db.conn.commit()` calls in that file predate this branch and were left alone.

**Finding 3 (unvalidated `which`) — accepted, fixed.** The `else 1` fallback would have silently opened the wrong snapshot for a typo'd/future caller. The handler now uses a closed `{"full_page": 0, "previous": 1}` vocabulary and refuses an unknown value with a type-only warning rather than defaulting.

All 73 tests in the two touched files pass.

PR #1295, opened 2026-08-03
fix(console): cost-ticker live-verification findings (TASK-2114/2115/2116)

rmusser01:
Three findings from the real-provider live verification of the merged cost-ticker program, filed as TASK-2114/2115/2116 and fixed here. One of them turned out to be a false alarm; the other two were real, and one nearly shipped a secret leak.

**TASK-2115 — cost chip "expired" after a reverted cache break: NOT A BUG.** A scripted reproduction (warm send → system-prompt edit → revert, zero clock manipulation) shows the chip already returns to warm `●` carrying the byte-identical original `warm_until` deadline — nothing on the edit/revert path touches the controller's cache ground truth; only a real send does. The live-verify session's own timeline shows ~141s consumed by the edit round trip alone, and its "TTL had ~1 minute left" note was extrapolated from an earlier reading rather than fresh, so the observed `○` was a genuine, correct TTL lapse. The commit adds the regression tests that pin this behavior both ways (revert-with-TTL-remaining stays warm; genuine lapse still reports expired), so the question can't be re-litigated by eyeballing a slow manual repro again.

**TASK-2114 — Anthropic `api_base_url` silently ignored on the main Console send.** Fixed for the primary path (streaming and non-streaming), default endpoint byte-identical when unset. Investigating it showed the drop lives in the shared `_chat_api_kwargs` builder rather than the Anthropic adapter, so **ten other providers are affected** — four of them (google, huggingface, moonshot, mistral) as live, unmasked defects. That's filed as **TASK-2117** (included here) rather than silently widened into this PR.

**TASK-2116 — debug log missing its `f` prefix**, emitting template text instead of values. This one cost real diagnostic time during the live verification, sending the investigation sideways while the actual cause was a per-model cache minimum. Worth flagging: fixing the f-string *alone* would have shipped a secret-material leak — interpolating those payload logs exposes request bodies. Only the mandated full-suite gate caught it (2 failures in `test_sensitive_llm_logging.py`); the fix gates all 9 logs behind `is_sensitive_llm_request()`, matching the existing `chat_with_huggingface` precedent. A narrow test selection would have merged it.

Gates: 49 passed (Anthropic native/caching/streaming/prefix), 53 passed (cost-chip trio), **3,252 passed** on the full `Tests/Chat/ + Tests/LLM_Calls/` sweep — all run in the foreground.

---
## Summary by cubic
Fixes Console sends to honor Anthropic `api_base_url`, tightens provider debug logs with an allowlist and sensitive gating, and pins cost chip TTL behavior with focused tests. Addresses TASK-2114, TASK-2115, and TASK-2116; follow-ups TASK-2117 and TASK-2118 filed.

- **Bug Fixes**
  - TASK-2114: Primary Console send now forwards Anthropic `api_base_url` via `ConsoleProviderGateway._chat_api_kwargs` (streaming and non-streaming). Default endpoint is unchanged when unset. Added unit and end-to-end tests asserting the posted URL for configured and default cases. Broader provider gap tracked in TASK-2117.
  - TASK-2116: Fixed nine missing f-string “Request Payload” logs and gated them with `is_sensitive_llm_request()`. Switched to an allowlist using `safe_llm_request_payload_summary` and `safe_llm_tool_names` so only safe fields and tool names are logged; fields like OpenAI `input`, Anthropic `system`, Google `system_instruction`, unknown payload keys, and tool schemas/descriptions never appear. Added an AST hygiene guard and updated tests to pin this behavior.
  - TASK-2115: Confirmed chip behavior: reverting a system-prompt edit with TTL remaining returns to warm with the original deadline; a real TTL lapse still reports expired. Added regression tests and updated the design spec.

<sup>Written for commit 4557caf6649c769cffbf837975783567098bb431. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1295?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/specs/2026-08-01-console-cost-ticker-design.md
+++ b/Docs/superpowers/specs/2026-08-01-console-cost-ticker-design.md
@@ -353,6 +353,24 @@ estimator and set `has_estimated_entries`.
   from last-send ground truth plus fingerprint break alerts, no countdown.
 - App restart: warm-until is in-memory → shows cold even if the provider
   cache is still warm. Under-promises, never lies. Accepted.
+- **Revert of a break, TTL remaining (task-2115, real-provider live
+  verification, 2026-08-03):** editing then reverting the system prompt
+  (or any other break source) never touches `warm_until`/`had_activity` —
+  only a real send does (see the ground-truth bullet above) — so the chip
+  correctly returns to plain warm `●` with the ORIGINAL deadline, byte-
+  identical to before the edit, confirmed by a scripted repro with zero
+  clock manipulation
+  (`Tests/UI/test_console_cost_chip_screen.py::test_reverting_system_prompt_edit_with_ttl_remaining_returns_to_warm`).
+  A live-verification pass once observed the chip land on cold `○`
+  immediately after such a revert despite believing TTL time remained;
+  investigation found no code path that could produce that from the
+  mechanism above — the far more likely explanation is that the manual,
+  multi-step round trip (reopening the editor modal, clicking, capturing
+  the pane) took long enough in real wall-clock time to genuinely cross
+  the 5-minute deadline between the last fresh tooltip read and the
+  revert completing, i.e. the chip was reporting truthfully. This is not
+  a distinct "unverified since your edit" state; the existing "expired"
+  copy already covers it.
 
 ### Projection math (always shown with `~`)
 
--- a/Tests/Chat/test_console_provider_gateway.py
+++ b/Tests/Chat/test_console_provider_gateway.py
@@ -2846,13 +2846,7 @@ def test_chat_api_kwargs_system_message_is_byte_stable_across_turns() -> None:
 
 
 def test_chat_api_kwargs_forwards_configured_anthropic_base_url() -> None:
-    """task-2114: a configured `[api_settings.anthropic].api_base_url` must
-    reach the PRIMARY Console send path's kwargs, mirroring what
-    `_auxiliary_chat_api_kwargs` already did for the auxiliary/one-shot
-    path -- `resolve_for_send` resolves the effective endpoint into
-    `resolution.base_url`; before this fix `_chat_api_kwargs` never
-    forwarded it, so a configured proxy/relay was a silent no-op on the
-    sends that matter most."""
+    """Confirm a configured Anthropic api_base_url reaches the primary send path's kwargs."""
     resolution = ConsoleProviderResolution(
         provider="anthropic",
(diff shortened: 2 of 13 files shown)


rmusser01:
Both findings addressed in `895d12d9c` (+ `edd3e4a28`).

**Finding 1 (log leak) — you were right, and the fix is an inversion rather than three more keys.** A denylist had already failed twice on this code in one session: first redacting only `messages`/`contents`, then missing `input` / `system` / `system_instruction` exactly as you flagged. Adding those three would have set up a third failure the next time a provider ships a prompt-bearing field. So the nine payload logs now route through a new `safe_llm_request_payload_summary` that emits an **allowlist** and drops everything else by construction:

`model, stream, streaming, max_tokens, max_completion_tokens, max_output_tokens, temperature, top_p, top_k, topP, topK, maxOutputTokens, message_count, has_system_prompt, tool_names`

Note what's derived rather than copied: message **count** not messages, a **boolean** for whether a system prompt exists rather than its text, tool **names** rather than schemas or descriptions. `is_sensitive_llm_request()` gating is unchanged as defense in depth.

Tests pin the property, not just the three symptoms: sentinel strings planted in `input`, `system`, and `system_instruction` are asserted absent from log output on **both** the sensitive and non-sensitive paths (6 parametrized cases), plus `test_safe_llm_request_payload_summary_drops_unrecognized_payload_keys` — a fabricated future key with a sentinel — and a test that a tool description sentinel is dropped while the name survives. That unknown-key test is the one that stops recurrence #3.

**Finding 2 (docstrings) — fixed** on exactly the 7 test functions this PR added or modified; no unrelated sweeps.

**One more of the same class found while fixing, filed not swept:** `chat_with_huggingface`'s "HuggingFace Tools" log still dumps full tool schemas verbatim when non-sensitive. It's outside TASK-2116's stated scope (tools log, not request payload), so it's **TASK-2118** (committed here) rather than a silent scope expansion — the fix is small now that the name-only helper exists.

Gates: 61 passed (sensitive-logging), 68 passed (Anthropic + LLM_Calls), **3,260 passed** on the full `Tests/Chat/ + Tests/LLM_Calls/` sweep — all foreground.

PR #1300, opened 2026-08-04
feat: paused watchlist sources visible + one-press Resume (task-2050)

rmusser01:
## Why

Auto-pause is live (tasks 1394/1410): a source that fails past its threshold gets `is_paused=1` and the scheduler stops retrying it. But nothing in the UI distinguished a paused source from an inactive one, and the only un-pause recourse — a manual re-check that succeeds — was undocumented data-layer behavior. A user just saw a silently stalled feed (task-2050, filed by the 1410 review).

## What

- **Paused is visible (AC#1).** `normalize_local_subscription_row` now emits `paused` and `status_summary` gains a `"paused"` state, with precedence `paused > error > inactive > active` — paused wins over error because every auto-paused source *still carries* the error that paused it, and an "error (10)" headline would hide the fact it stopped being retried (trade-off documented: the error text is not surfaced elsewhere for a source yet). The server normalizer stamps `paused: False` (no server pause concept), which structurally prevents the Resume button from ever rendering for a server source.
- **One-press Resume (AC#2/#3).** The inspector's actions block gains a Resume button, rendered only for a paused subscription entity. Press → `ResumeSourceRequested` → screen handler → new `LocalWatchlistsService.resume_source` → the previously caller-less `SubscriptionsDB.reset_subscription_errors` (clears `is_paused`, `error_count`, `consecutive_failures`, `last_error` — unmodified; this task gives it its first caller). Toast + sources reload; no SQL in the UI layer. Resuming a healthy source is a harmless no-op at the DB level; the button simply never renders for one.
- **Review fix (the 21st composition find of this sweep):** with `paused` winning the precedence, an auto-paused source vanished from the Sources pane's **"Error" filter bucket** — the broken-feed triage view silently missing the *most*-broken feeds. The Error bucket now includes paused sources (commented), and a dedicated "Paused" filter option was added.

## Testing

Normalizer precedence (paused / paused+error / healthy / inactive), button gating (only paused subscriptions; absent for items/rules/healthy/hand-built dicts), real-press end-to-end against a real `SubscriptionsDB` (all four columns cleared, toast, reload), filter buckets (paused appears under both Error and Paused, never under OK), and the task-1410 pause suite unchanged. 33+146 targeted tests green; every behavioural change mutation-verified (normalizer emission, handler service call, filter arm). Review: FIX WAVE on the filter bucket only, fixed and pinned; all else confirmed clean by independent re-reading and re-running the mutations.

---
## Summary by cubic
Expose auto-paused watchlist sources and add a one-click Resume; also count paused as errors on the Overview and gate Resume to local subscriptions. Satisfies TASK-2050 by making paused status visible and clearing pause/failure counters via the service layer.

- **New Features**
  - `normalize_local_subscription_row` now sets `paused` and `status_summary` precedence: paused > error > inactive > active; server sources always return `paused: False`.
  - Inspector shows Resume only for paused local subscriptions; press posts `ResumeSourceRequested` and calls `LocalWatchlistsService.resume_source` → `SubscriptionsDB.reset_subscription_errors`, then toasts and reloads sources. Resume on a non-paused source is a safe no-op.

- **Bug Fixes**
  - Paused sources remain in the “Error” filter and have a dedicated “Paused” option.
  - Overview’s “Sources in error” now counts paused sources.
  - Resume handler refuses non-local/non-`subscription` entities to prevent id-collision resets.

<sup>Written for commit 95354b7e613d4708e55c256da0e3efa394b15c28. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1300?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Subscriptions/test_watchlist_normalizers.py
+++ b/Tests/Subscriptions/test_watchlist_normalizers.py
@@ -73,3 +73,132 @@ def test_normalize_tolerates_a_row_with_no_body():
 
     assert item["content"] is None
     assert item["content_kind"] is None
+
+
+def test_normalize_marks_a_paused_source_paused_and_says_so_in_status(
+) -> None:
+    """task-2050 AC#1: a paused source must be distinguishable from
+    inactive/healthy -- both `paused` and `status_summary` carry it."""
+    from tldw_chatbook.Subscriptions.watchlist_normalizers import (
+        normalize_local_subscription_row,
+    )
+
+    source = normalize_local_subscription_row(
+        {
+            "id": 1,
+            "name": "Dead feed",
+            "type": "rss",
+            "source": "https://dead.example/feed",
+            "is_active": 1,
+            "is_paused": 1,
+        }
+    )
+
+    assert source["paused"] is True
+    assert source["status_summary"] == "paused"
+    assert source["active"] is False, (
+        "a paused source must not also read as active in the Active column"
+    )
+
+
+def test_normalize_a_paused_source_with_a_last_error_still_says_paused(
+) -> None:
+    """task-2050: precedence is paused > error, not error > paused.
+
+    A source auto-paused by repeated failures (task-1410) always still
+    carries the `last_error` that caused the pause. An error-first
+    precedence would render "error (N)" -- indistinguishable from a source
+    that is merely having a bad day but is STILL being retried on schedule.
+    Paused has to win so the Status column's headline does not lie about
+    that; the error detail itself is not lost, it stays on the entity's own
+    `last_error`/`error_count` keys for the Inspector to show.
+    """
+    from tldw_chatbook.Subscriptions.watchlist_normalizers import (
+        normalize_local_subscription_row,
+    )
+
+    source = normalize_local_subscription_row(
+        {
+            "id": 2,
+            "name": "Flaky feed",
+            "type": "rss",
+            "source": "https://flaky.example/feed",
+            "is_active": 1,
+            "is_paused": 1,
+            "last_error": "connection refused",
+            "error_count": 5,
+        }
+    )
+
+    assert source["paused"] is True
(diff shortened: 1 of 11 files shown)


rmusser01:
Qodo adjudication — all four addressed:

**Finding 1 (paused excluded from the Overview error count) — confirmed real, fixed.** The same class as the filter-bucket fix, one level up: `sources_in_error` counted only `startswith("error")`, so the Overview's error card would *drop* the moment a failing source finally tripped auto-pause. Paused now counts as in-error (same semantics as the Sources pane's Error bucket, commented at both sites). Pinned by `test_overview_counts_paused_sources_as_in_error`, mutation-verified.

**Finding 4 (resume handler trusts the entity) — confirmed, fixed.** `_resume_source` takes a raw local db id, so a message carrying a non-local entity with a colliding numeric `source_id` (e.g. a server `watchlist_source` id 5) would have reset the counters of the unrelated LOCAL subscription 5. The handler now refuses anything that isn't a local `subscription` (type-only log; the render gate already made this unreachable from the UI — the guard keeps it unreachable from any future caller). Pinned by `test_resume_refuses_a_non_local_subscription_entity` with a colliding-id foreign entity, mutation-verified.

**Findings 2–3 (docstrings) — fixed.** `normalize_local_subscription_row` gained Args/Returns; `handle_resume_source_requested` gained a docstring documenting the guard.

60 tests across the touched suites green.

PR #1302, opened 2026-08-04
test: atomic private write leaves no residue under injected write/rename failures (task-2060)

rmusser01:
## Why

`atomic_private_write_bytes`'s temp-then-rename cleanup guarantee (the `finally` that unlinks a stranded temp file) was asserted by code inspection only — its own suite had no test injecting a failure into the real POSIX path. Surfaced during task-1719, whose audio-pipeline "no orphan file" test leans on exactly this guarantee (task-2060).

## What (test-only)

Two tests in `Tests/Utils/test_private_paths.py`, each **naming its injection point** (AC#2) so a future change to the temp+rename strategy that reopens a residue window fails the matching test by name:

- `test_no_residue_when_the_temp_write_side_fails` — injects at the temp-file fsync (payload fully written, pre-rename).
- `test_no_residue_when_the_rename_itself_fails` — injects at the rename, selectively (only calls whose args end in `.tmp`).

Both assert the destination retains its **original content** and that the directory's contents are **identical before/after** (a snapshot comparison — strict against any residue shape, immune to unrelated tmp_path entries). Mutation-verified: disabling the `finally` temp-unlink in production reds **both**.

**Trap found and closed en route:** `_atomic_posix_guards_available` checks `{os.rename, os.unlink} ⊆ os.supports_dir_fd` by *function identity*, so monkeypatching `os.rename` made the first version of the rename test pass **vacuously** — the call bailed out with `required_posix_guards_unavailable` before any temp existed, and every assertion was green for the wrong reason. The mutation protocol caught it (the test stayed green under the disabled-cleanup mutation). Fixed by pinning the guards check True for that one test's duration, documented in its docstring; the write-side test exercises the unpatched check on the same run.

## Testing

`Tests/Utils/test_private_paths.py`: 68 passed. Production `private_paths.py` untouched.

---
## Summary by cubic
Add two POSIX-guarded failure-injection tests for `atomic_private_write_bytes` proving no residue and an unchanged destination when temp fsync or rename fails. Satisfies TASK-2060 by naming the injection points, snapshotting the directory, and asserting the rename injection fired; also adds a temporary UI instrumentation test for TASK-1960 to trace the SelectCurrent label mount race.

<sup>Written for commit 749fa2e888639cfe2e140655ea7e19ead297f889. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1302?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_zz_scaffold_1960.py
+++ b/Tests/UI/test_zz_scaffold_1960.py
@@ -1,0 +1,148 @@
+"""TEMPORARY instrumentation scaffold for task-1960. DELETE BEFORE COMMIT."""
+
+from __future__ import annotations
+
+import time
+from unittest.mock import AsyncMock
+
+import pytest
+from textual.widget import Widget
+from textual.widgets import Button, Input, Select
+from textual.widgets._select import SelectCurrent
+
+from Tests.UI.full_app_destination_context import (
+    StaticWatchlistsScopeService,
+    active_destination_screen as _active_destination_screen,
+    full_app_destination_context as _visual_destination_harness,
+    wait_for_selector as _wait_for_selector,
+)
+from Tests.UI.app_factory import _build_test_app
+from tldw_chatbook.UI.Screens.watchlists_collections_screen import (
+    WatchlistsCollectionsScreen,
+)
+from tldw_chatbook.UI.Watchlists_Modules.sources_pane import SourcesPane
+
+EVENTS: list[str] = []
+
+
+def _log(msg: str) -> None:
+    EVENTS.append(f"{time.monotonic():.6f} {msg}")
+
+
+@pytest.fixture
+def instrument(monkeypatch):
+    EVENTS.clear()
+
+    real_screen_recompose = WatchlistsCollectionsScreen.recompose
+
+    async def patched_screen_recompose(self):
+        _log(">>> SCREEN recompose ENTER")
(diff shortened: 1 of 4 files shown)


rmusser01:
Qodo adjudication — all four addressed:

**Finding 1 (POSIX-only injection, no skip) — fixed.** Both tests now carry `skipif(not _atomic_posix_guards_available())` — the Windows/fallback path has different mechanics (no temp-fsync point, `os.replace`), so running them there would fail on the wrong code path.

**Finding 2 (rename injection not asserted) — confirmed and fixed.** A fair test-honesty catch: the write-side test asserted its injection fired but the rename test didn't, so an earlier unrelated `PrivatePathError` could have satisfied `pytest.raises`. The rename wrapper now records its firings and the test asserts `injected`.

**Findings 3–4 — fixed.** `Args:` sections added to both docstrings; the magic `5` is now `errno.EIO`.

68 passed.

PR #1315, opened 2026-08-04
fix(watchlists): guard Select mount against Textual's silent prune-time mount no-op (task-1960)

rmusser01:
## The bug (task-1960)

Closing the Watchlists create form intermittently crashed with `NoMatches: query_one("#label")` inside `Select._on_mount` — the race split out of task-1345 after three mitigations shrank it 100%→25% but could not kill it.

## Confirmed mechanism (not the one originally hypothesized)

Textual's Compose-before-Mount ordering has an escape hatch, verified line-by-line against the installed Textual 8.2.8:

- `Widget.mount` (`widget.py:1451`) silently mounts **nothing** while `self._pruning`/`self._closing` — it returns `AwaitMount(self, [])`.
- `MessagePump._pre_process`'s `finally` (`message_pump.py:609-612`) sets `_mounted_event` and `_is_mounted = True` unconditionally.
- `App._prune` (`app.py:4381-4395`) stamps `_pruning` over a `walk_children` snapshot, catching a registered-but-not-yet-composed `SelectCurrent`.

So when a full-screen recompose (requested by `_apply_local_wc_snapshot` / `_load_tree_data` after submit) prunes the centre while the freshly rebuilt pane's `Select`s are mid-mount, the parent `Select`'s mount-await unblocks "successfully", `_on_mount → _init_selected_option → _watch_value` runs, and `SelectCurrent.update`'s `query_one("#label")` finds no child. (Review found a third unguarded lookup on the same path: `query_one(SelectOverlay)` at `_select.py:613` — also covered.)

## The fix

New `Widgets/prune_safe_select.py` (`PruneSafeSelect`): `_watch_value` and `_setup_options_renderables` become no-ops while `_pruning`/`_closing`, except for syncing the internal `_value` (the one non-DOM-lookup statement of the stock watcher) so `value`/`_value` stay consistent. The guard is airtight, not a shrunk race: `App._prune` is synchronous and the 1-parameter sync watcher runs inline (`reactive.py:114`) with no await between the flag check and the queries. No stale-display mode: `_pruning`/`_closing` are terminal in Textual 8.2.8 (residual documented in the docstring as a Textual invariant).

All **15** `Select` construction sites in the Watchlists/briefings feature converted (panes, screen, and the two ModalScreens with the identical worker-recompose shape). Also deletes `Tests/UI/test_zz_scaffold_1960.py`, a scaffold accidentally leaked onto dev by an earlier commit.

## Verification

- e2e `test_a_source_can_be_created_end_to_end_through_the_form`: **10/10 in isolation** and **5×50 after `test_watchlists_content_pane.py`** (the poisoned order; baseline before the fix: 3 failures in the first 10). No sleeps, no retries.
- The e2e "form closed" assertion was repaired from a vacuous check on a pruned widget to observing the settled live pane; mutation-proved (neutering form-close: old assertion GREEN, new one RED).
- New `Tests/Widgets/test_prune_safe_select.py` (9 tests) reconstructs the prune interleaving; each guard clause including the `_closing` half killed by its own discriminating test.
- Whole-branch review (0 Critical / 2 Important / 4 Minor) + fix wave + scoped re-review: **all findings closed**, key mutations independently re-run by the reviewer.
- `Tests/Watchlists/` + `Tests/UI/test_watchlists_*` + modal suites green; `--collect-only` 8577 collected clean.

Follow-up worth filing (out of scope here): the architectural root cause — full-screen recomposes from background loaders (`_apply_local_wc_snapshot` / `_load_tree_data`) — is TASK-1541's standing recommendation; the pane's own close-recompose silently mounting nothing when pruned is masked today by the screen's immediate rebuild.

---
## Summary by cubic
Fixes an intermittent crash when closing the Watchlists create form by introducing `PruneSafeSelect`, which skips DOM lookups during `textual` prune/close and prevents `Select` mount races. Replaces all Watchlists `Select`s and adds fail-fast compatibility checks to meet task-1960.

- **Bug Fixes**
  - Added `PruneSafeSelect`: no-ops `_watch_value` and `_setup_options_renderables` while `_pruning`/`_closing`, keeping `_value` in sync; otherwise behaves like `Select`.
  - Replaced all 15 Watchlists `Select` instances across panes, the screen, and both modals.
  - Fail-fast checks: validate required `textual` internals at import and in `__init__`, raising `PruneSafeSelectCompatibilityError` if missing.
  - Tests: added `Tests/Widgets/test_prune_safe_select.py`; updated E2E to wait for the settled `SourcesPane` and assert the form is truly gone; added a structural pin in `Tests/Watchlists/test_watchlists_sources_pane.py`; removed the temporary scaffold test.

- **Dependencies**
  - Raised `textual` floor to `>=8.2.8,<9`.

<sup>Written for commit 418db5c31c8d7510bbd48f6a533c1b2faa87b540. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1315?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
**Qodo finding 1 (private Textual API coupling) — addressed in `b38c9f357`, with one suggested remediation deliberately declined:**

- ✅ **Fail-fast compatibility check** (accepted): `_require_internals()` now raises `PruneSafeSelectCompatibilityError` at module import if `Select._setup_options_renderables`/`_watch_value` no longer exist, and at construction if the `_pruning`/`_closing` flags are missing. This targets the insidious failure mode: a Textual rename would mean the overrides are simply never called — guard silently bypassed, race returns as intermittent flakiness. Wiring tests prove both check sites actually fire (mutations: neutered check RED ×3, deleted import-time call RED, deleted `__init__` call RED).
- ✅ **Dependency floor** (accepted): `textual>=8.2.8,<9` — the range now matches the baseline the mechanism was verified against line-by-line.
- ❌ **`getattr(self, "_pruning", False)` softening** (declined): defaulting a missing flag to `False` makes the guard silently inert — precisely the hidden-race failure mode this task exists to refuse (task-1960's history: three mitigations that shrank the race without killing it were reverted on this principle). The fail-fast check covers the same upgrade risk loudly instead. The refusal is recorded in the class docstring and task file so it isn't "helpfully" re-added later.

Known residual (documented, not fixable by `hasattr`): a future Textual where the flags exist but are no longer terminal (widget revival) would be invisible to the check — recorded as a Textual invariant in the docstring.

PR #1331, opened 2026-08-04
feat(watchlists): targeted in-place refresh from background loaders (task-2200)

rmusser01:
## What (task-2200)

Removes the Watchlists screen's background-driven full-screen recomposes — the architectural root cause behind task-1960's prune-time crash class. The screen keeps exactly one `refresh(recompose=True)`, in `watch_active_section` (a user gesture that genuinely changes which regions mount).

**Converted to targeted in-place updates:**
- `overview_data` reactive (was screen-level `recompose=True`; two production code paths existed solely to dodge its recompose) → plain reactive + watcher pushing into the live OverviewPane/Inspector.
- `_load_tree_data` → rail + FEEDS + header rebuilds; breadcrumbs, watchlist count, and Artifacts scope label pushed into live panes.
- `_apply_local_wc_snapshot` → FEEDS + header rebuilds; Inspector `State:` line and Console attach button patched in place.

**One owner for DOM swaps:** the workbench's remove-then-mount region rebuilds are serialized through a record-intent/drain-serially queue (`_request_surface_refresh`/`_drain_surface_refresh`) — `exclusive=True` was unsound once three actors wanted the swaps (a cancelled predecessor between `remove()` and `mount()` leaves a bordered empty region). Task-1541's cancellation lesson, applied to DOM work.

**Behaviors the old recompose silently carried, now explicit:** backend-switch selection reseed (`_reseed_live_detail_pane`); Console-follow drift re-poll (`_resolve_console_follow_drift`, conditional so a half-typed noise-selector draft is never destroyed by a background refresh).

## The live-only defect this surfaced (and the lesson)

Green suites, two clean reviews — and in a real terminal the initial load never painted: **during the whole of `on_mount`, `Widget.is_mounted` is still False** (`_pre_process` sets it in its `finally`, after Compose+Mount dispatch — the same `finally` as task-1960, seen from the other side). Every new `if not self.is_mounted` guard skipped the 7ms cold-DB load, and dev's reactive recompose had silently masked the identical skip for the six pre-existing section loaders too. Measured in review: after a real `switch_screen` teardown `is_attached` is False while `is_mounted` is **still True** — these guards were never a teardown gate, only a false negative for the mount window. All twelve sites now use `is_attached` (`_dom_is_live`); the trap is recorded in the task notes.

## Verification

- Whole-branch review + 3 fix waves (opus, adversarial): artifacts scope-label regression proven bidirectionally and fixed; drain hardening (no wedge on failed worker launch, drift poll inside the guard); honest drift-cache docs; mount-window fixes above. Final scoped re-reviews: all findings closed.
- **28 mutations across the waves, all RED, all restored byte-exact (md5-verified)** — including per-site sweeps of both guard sets, proven disjoint.
- **Live-verified in a real terminal, twice, independently** (fresh scratch profiles, zero interaction): before the mount-window fix the screen sat on "Loading…" for 100+s; after, loaded/empty states paint in ≤5s and the Inspector reads "State: ready".
- e2e create-form: 10/10 isolation, poisoned-order 50-passed runs repeated across every wave. Region-gating, overview/first-run/loading-state, inspector, panes, modal suites all green; `--collect-only` 8485 clean.
- AC#1 draft-survival and AC#4 form-close-without-screen-rebuild covered by identity-asserting regression tests; `PruneSafeSelect` untouched (AC#5).

---
## Summary by cubic
Replaces full-screen recomposes on the Watchlists screen with targeted in-place updates to keep panes/forms mounted, and fixes cold-start loads not painting by using a DOM-live guard during the mount window. Linear task-2200; addresses the crash class from task-1960.

- **Refactors**
  - Replaced screen-level recomposes from `overview_data`, `_load_tree_data`, and `_apply_local_wc_snapshot` with focused updates to the rail, FEEDS, headers, `OverviewPane`, and `Inspector` (includes breadcrumbs, watchlist count, scope label, profile state, and a screen-held console-follow mirror).
  - Added a serialized region-swap queue (`_request_surface_refresh` / `_drain_surface_refresh`) to own remove-then-mount cycles and prevent empty regions when a half is cancelled.
  - Made backend-switch reseed and console-follow drift handling explicit; right rail rebuilds only when the adapter’s answer changes. Kept `watch_active_section` as the only recompose.

- **Bug Fixes**
  - Switched guards to `is_attached` (`_dom_is_live`) so overview, section loaders, and `watch_selected_entity` paint during the mount window (covers cold starts and deep links).
  - Hardened the refresh queue and drift logic: avoid wedges on failed worker scheduling; moved console-follow poll under the guarded rebuild; stop re-running the adapter in `compose()` by mirroring `console_follow_item` on the screen.
  - Ensured renamed watchlists reach the mounted Inspector (breadcrumb and `#artifacts-scope-note`) and that overview counts/statics update in place.

<sup>Written for commit 49ac443c57c51aadf797b40a98ac10816dcb4cf7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1331?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_watchlists_destination_shell.py
+++ b/Tests/UI/test_watchlists_destination_shell.py
@@ -7,7 +7,7 @@ from typing import Any
 from unittest.mock import AsyncMock, Mock, patch
 
 import pytest
-from textual.widgets import Button, DataTable, Input, Select, TextArea
+from textual.widgets import Button, DataTable, Input, Select, Static, TextArea
 
 from Tests.UI.app_factory import _build_test_app
 from tldw_chatbook.Subscriptions.noise_defaults import default_ignore_selectors_text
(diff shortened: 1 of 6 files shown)


rmusser01:
**Qodo finding 1 (selection watchers in the mount window) — half accepted with a RED-first proof, half declined with a measured no-op proof, in `c337b10ea`:**

- ✅ **`watch_selected_entity` — real defect, fixed.** Reproduced before fixing: deep link arms `_pending_navigation_run_id` pre-mount → `on_mount` → `_load_runs` completes in the window on a cold DB → `_select_entity(requested_run)` → the `is_mounted` guard drops the push; the screen's `selected_entity` held the run while the mounted `InspectorPane.selected_entity` stayed `None`, and with the screen-level recomposes gone nothing recovers it. Guard swapped to `_dom_is_live`; new regression test enters through the real `apply_navigation_context` path; mutation (guard reverted) RED.
- ❌ **`watch_selected_scope` — declined, no production change.** Its only in-window writer can only write `TreeScope(kind="all")`, byte-equal to the reactive's default — structurally a no-op. Measured over the real flow: the watcher fired 0 times in-window; the inverse mutation (swapping the guard anyway) left all 112 tests identical. Caveat recorded in the task notes: this safety rests on value equality, so a future non-`"all"` in-window writer would re-acquire the defect — the notes name it as the one survivor whose reasoning a future change can invalidate.
- All 13 remaining `is_mounted` sites on the screen were mapped by AST to their enclosing methods and adjudicated (full table in the task notes); `watch_active_section` in particular is load-bearing as-is (flipping it would double-start the loader `on_mount` owns).

PR #1342, opened 2026-08-04
fix(watchlists): UAT batch 1 — Select overlays, filter honesty, count/scope truth (tasks 2300/2301/2304)

rmusser01:
## What (UAT batch 1 of 5 — tasks 2300, 2301, 2304)

First batch from the 2026-08-04 sr-UX/HCI UAT of the Watchlists screen (41 findings → 15 tasks, filed in #1335).

**task-2300 — empty Select overlays (UAT F30/F36, critical).** Root cause was NOT option population (options were present; `PruneSafeSelect` measured uninvolved): three app-wide CSS rules — `*:focus { outline }`, plus focus/hover borders on `Select` — overwrote the content rows of zero-perimeter compact Selects and their overlays. Fix: opt the overlay out of the outline, delete the border rules whose only surviving effect was destroying compact shapes, and give each Select shape an explicit, measured focus/hover cue: hover via `Select > SelectCurrent:hover` (the widget Textual actually hovers — styling it from the parent's `:hover` can never match, since Textual hands the hover class to the child), focus via `Select.-textual-compact:focus > SelectCurrent { background: $ds-input-focus-bg }` (contrast 1.10:1 → **3.12:1**; 1.1:1 is TASK-345's named nullified-focus failure). Settings-family bordered compact Selects verified stable (no flicker, frame intact, `#586d82` unchanged).

**task-2301 — items filter honesty (UAT F29/F31, high).** The list was de-facto "new only" under an "All statuses" label. The active filter is now part of the page query (fixing both the lie and the 100-mixed-rows paging interaction), every status the DB produces is selectable with one consistent user-facing vocabulary, Ingest/Ignore give markup=False feedback, and the `_filtered_items` open-item pin keeps its guarantee under narrow filters (the open item is carried alongside the page in `created_at DESC` order — dropping the predicate would have un-fixed the paging exactly while the reader is open).

**task-2304 — count/scope truth (UAT F15/F16, high).** Rail counts live-update through the task-2200 drain discipline with a legend naming what they count (unread items, debounced past mark-read-on-open bursts); the scoped sources table and header now agree in both directions (the first fix's own opposite-direction regression was caught live and fixed); `d` on an item routes through the task-1541 drain (closing a second unguarded status writer) with honest per-kind binding copy.

## Review trail

Whole-branch opus review + 2 fix rounds; re-review confirmed 8/10 then 3 residuals fixed (focus visibility, pin guarantee, binding copy). **29 mutations RED across the waves, restored byte-exact.** Two pre-existing tests corrected that had the broken behavior as premise (both RED under the restored old behavior); one vacuous-shape test class eliminated (helpers may not derive expectations from production).

## Verification

42 new tests across three files; settings-family suites 488 passed (the CSS surface); poisoned-order e2e 50 passed; `--collect-only` 8654 clean. **Live-verified twice** (fresh profiles, real feed): filter shows all six options and triaged items are findable under their own status; rail counts 0→20→18 tracking reality with the legend; Settings provider Select stable across rest/focus/hover frames.

Known residuals (recorded in task notes): per-status paging is still paging (proper pagination is the real fix); hover cue absent on a bordered compact Select's frame rows (honest trade vs the flicker); carried open item can hold a stale status if changed behind the screen's back.

---
## Summary by cubic
Fixes three Watchlists UAT issues: compact Selects show all options with clear, non-flickering focus/hover cues; the Items list truly honors and pages by its status filter; rail counts and the scoped Sources table now tell the same story and update live. Meets the acceptance criteria for TASK-2300, TASK-2301, and TASK-2304.

- **Bug Fixes**
  - TASK-2300: Restored compact `Select` overlays by opting overlays out of the global focus outline (`SelectOverlay:focus { outline: none }`) and moving focus/hover cues onto `SelectCurrent` (recolor, no extra borders). Added a visible focus recolor for borderless compact `Select`s; removed geometry-changing `Select:focus`/`Select:hover` border rules and fixed a dead hover selector. Settings-family bordered `Select`s verified stable; tests read compositor output to prevent regressions.
  - TASK-2301: `status=None` now means “all statuses” in the DB/service; the active filter is part of the query so each page contains the chosen status. Ingest/Ignore repaint the acted row immediately; the Status column uses the filter’s vocabulary (e.g., “Read”). The open item is carried alongside the page so navigation works under narrow filters; the `d` binding now reads “Delete / Ignore.”
  - TASK-2304: Added a rail legend and tooltips clarifying counts = unread items; rail counts refresh after Check now and status changes. The Sources table respects the tree scope and stays in sync with the header, including after membership writes; scoped rows are pushed immediately.

<sup>Written for commit bef4ceb52ab6b2966d122a0ac29a2900a7356348. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1342?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
**Qodo finding 1 (get_new_items bypasses transaction()) — accepted, fixed in `c3426d2ae`.** Subscriptions_DB's per-file convention wraps reads in `transaction()` (nearest sibling `get_item_status` does; this is the same adjudication as PR #1290's round). Wrapped using the sibling's exact shape — `with self.transaction() as conn` — after a first attempt bound the context manager's yield as a cursor and failed 2 tests (the context manager yields the connection); Tests/Subscriptions 600 passed + the task-2301 filter file green.

PR #1348, opened 2026-08-05
fix(watchlists): UAT batch 2 — run accounting, run identity, live run detail (tasks 2305/2306)

rmusser01:
## What (UAT batch 2 of 5 — tasks 2305, 2306)

Second batch from the 2026-08-04 sr-UX/HCI UAT (findings F32/F33/F34, filed in #1335).

**task-2305 — runs carry real accounting and their source name.** The pipeline had recorded everything all along: `execute_run` wrote `items_found`/`items_ingested` into the nested `stats` blob while the Runs pane read `found_count`/`processed_count` off the top level with no bridge, and the source name was only ever resolved by Home's hand-written JOIN. Now bridged with a normalizer that handles every historical row shape (proven against pre-branch rows, `stats_json = NULL`, failed runs; `Filtered` derived from found−processed since nothing records it), and the run query resolves source + watchlist names (sorted, so the displayed name is stable).

**task-2306 — run selection populates the detail.** Selection wiring was never broken — the detail's `Static` kept its first-compose "No run selected." forever (the watcher never repainted it), and `run_items`/`run_logs` had no product writer at all. Now wired for mouse and keyboard, with honest blank-state labels for the four ways an Items list can be empty (re-claimed by a later check / every found item excluded by a filter / server-backend listing unavailable / failed to load — discriminated on `processed_count`, not the feed tally), an honest truncation line, a failure toast per the screen's loader convention, and that convention itself turned into a structural AST contract (an except handler guarding an awaited read that logs at debug must notify in that same handler) — which widened coverage beyond the old hand-list.

## Review trail

Whole-branch opus review + 2 fix rounds; final verdict ALL ADDRESSED with the round-2 deviation independently re-measured and endorsed (method-level exemption would have holed a real fetch handler). **36 mutations RED across the waves**, each restored byte-exact; 4 initially-surviving mutations had their tests strengthened until load-bearing; one dead "recorded filtered" distinction deleted rather than tested (no writer exists).

## Verification

Net +49 tests. Suites: services/scheduling/DB/Home/policy 1338 passed; watchlists UI + destination shell green; poisoned-order e2e 50 passed; `--collect-only` 8713 clean; the one failure in the mixed sweep is the pre-existing `_delete_item` case (fails on origin/dev too — filed as task-2330). **Live-verified twice** (fresh profiles, real feed): run rows read `HN · completed · 485ms/739ms · 20 20 0 0`; selecting the older of two checks shows the re-claimed label; the newer shows its 20 items; detail follows keyboard selection.

Follow-ups filed: #1346 (tasks 2330/2331: pre-existing dev test failure; dead Runs Refresh + silent Re-run). One more coming per review: `_load_source_rows_for_tree`'s silent failure contradicts Done task-1090's toast premise.

---
## Summary by cubic
Fixes UAT findings F32–F34 (tasks 2305/2306): runs now show real accounting with stable source/watchlist names, and selecting a run fills Run detail (items + logs) with clear empty states, truncation notes, and failure toasts.

- **Bug Fixes**
  - Runs read counters from `stats` (found/processed/filtered/errors) with accurate duration; failed runs show error counts.
  - Source and watchlist names resolved and sorted; rows show “Source · Watchlist”, detail lists all watchlists.
  - Run detail repaints on selection and respects deep links; polling posts `RunProgressTick` to avoid per-second full reloads and refreshes only on change.
  - Items region shows honest empty states (re-claimed, filtered-only, backend unavailable, or load failed) and a truncation line based on processed count; background loader failures always toast (structurally enforced).
  - Stats parsing hardened for non-finite/overflow values to prevent table crashes.

- **New Features**
  - Added `run_id` filter to item queries via `items.list` so the Items sub-region lists a run’s items.
  - Item normalizer now includes `run_id` and `alert_count`; watchlist names are packed with a safe separator for display.

<sup>Written for commit 67e9434f89bddeb76fcaaaf6911a75acfb36fdd2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1348?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
**Qodo round — all three accepted, fixed in `f00cdbf49`:**

- ✅ **Q1** `Args:` sections added to the three new watchers.
- ✅ **Q2 (poll-tick detail churn)** — fixed with a distinct `RunProgressTick` message rather than an id-change guard, on measured grounds: a local run writes stats/log/items in one go at completion (`record_run_result`), so a skip-if-same-id guard would freeze the detail at exactly its least informative paint and miss the running→completed transition. `RunSelected` is selection-only again; the tick handler re-reads the run record and short-circuits on an unchanged fingerprint (status/finished_at/counters/log/error) — an unchanged tick costs one indexed read and no items query; a changed one updates row, stats and items without re-entering the selection cascade. Mutation-verified end-to-end through the live poll (not a synthetic message), including the freeze failure mode the rejected shape would have caused.
- ✅ **Q3 (stat-parse overflow)** — `_run_stat` now parses int-strings exactly, rejects non-finite floats via `math.isfinite`, catches `OverflowError`, and falls back to derivation; 14 new normalizer tests incl. `"inf"`, `"nan"`, `"1e400"`. One malformed server counter can no longer take the Runs table down.

PR #1354, opened 2026-08-05
chore(realtime): V4 follow-ups 2360-2364 (incl. ChaChaNotes v30→v31 message metadata)

rmusser01:
## What this is

The five follow-ups filed from PR #1350's review ledger (realtime voice engine / voice2voice V4), each implemented, reviewed, and mutation-verified. Tasks **2360, 2361, 2362, 2363, 2364** are marked Done; two new follow-ups (**2390**, **2391**) are filed for work these deliberately deferred.

> ⚠️ **Contains a schema migration (ChaChaNotes v30 → v31).** See "Merge sequencing" below — this affects other checkouts on a shared database.

## The five

**task-2361 — the idle ceiling could eject someone mid-sentence.** The cost guard counted "activity" as a committed turn or a finished reply, so a user who *started* speaking a second before the deadline was cut off with "idle for 5 minutes". `on_speech_started` now refreshes the activity anchor while live, in both barge-in modes, using the FSM's existing pending-anchor idiom. A genuinely silent session still exits on time (pinned).

**task-2360 — reconnect dropped microphone audio.** The tap's first-words buffer was one-way, so a mid-loop reconnect streamed into a session that couldn't accept it. New `RealtimeMicTap.begin_buffering()` re-arms the *same* bounded, in-order buffer; the existing ready handler releases it; a failed reconnect discards it (a fresh loop entry always builds a new tap, so no later session can inherit stale audio). Investigation found the real mechanism was subtler than the task described — the session slot is reassigned to a not-yet-connected session, and the live session silently drops audio queued before `connect()` finishes. Recorded in `backlog/docs/lessons-testing-evidence.md`.

**task-2362 — closed the fake-server/probe drift list** from #1350's final review (M9). All five items got real assertions, none were downgraded to "documented as uncovered": the `input_audio_buffer.committed` dispatch now fails a test if deleted; `voice` and the `whisper-1` model are asserted (the model as a literal, not imported from the production constant — avoiding the tautology); the probe's docstring now matches what it actually sends; and the singular `input_token_details` claim lives in the ground-truth header with a test in `Tests/Chat/test_provider_usage.py`.

**task-2363 — realtime usage completeness.** Transcription-duration usage (`{"type": "duration", "seconds": N}`) previously reached nothing; it now flows through a new additive `on_transcription_usage` callback — same marshal path as every sibling — onto the user row via the existing store write. Audio tokens are recorded as distinct `audio_input`/`audio_output` fields rather than folded into text counts, and are provably inert for billing today (the pricing catalog and cost tracker read named buckets only). Cost-chip wiring deferred to **task-2390** per the AC.

**task-2364 — structured message metadata (the schema change).** New `MessageMetadata` (frozen, closed status set) persisted as a local-only `messages.metadata_json` column, mirroring the `usage_json` precedent seam for seam. Realtime rows now carry engine/provider/model provenance and an interrupted flag, so machine consumers stop parsing UI copy.

## Review highlights

- **The migration was audited empirically, not argued.** A probe killed the process immediately after the `ALTER`, confirming the migration is non-atomic (as every prior migration here is) but self-healing via the pre-applied-column guard; a metadata write was shown to enqueue *zero* sync rows while a control content write still enqueued one (proving the sync triggers are genuinely unaffected, and weren't accidentally disabled); v30 code was confirmed to still refuse a v31 database; and metadata round-tripped through a brand-new database handle. Eight mutations, all red.
- **A real consumer bug was caught in review.** The interrupted-marker fallback used a global `replace` and was documented as covering only pre-metadata legacy rows — but only realtime rows are ever stamped, so every *typed* turn took that path permanently: `the docs say ⏹ interrupted means cut off` was reseeded to the model as `the docs say means cut off`. Fixed with a suffix-anchored trim (safe because the marker has exactly one writer and only ever appends), which closed a second finding for free. The docstring and spec bullet that misdescribed it were corrected.
- Every task's tests were mutation-checked by its reviewer rather than taken on trust.

## Merge sequencing (please read)

The schema bump is one-way on a shared database. Every worktree on a machine shares `~/.local/share/tldw_cli/tldw_chatbook_ChaChaNotes.db`, and the DB layer refuses — by design — to open a v31 database with v30 code. **Nothing here was ever run against a real database**; all testing used temp/in-memory ones. After this merges, other checkouts need this code before they next open the app.

A related trap found while writing that up, now in `backlog/docs/lessons-live-verification.md`: `TLDW_CONFIG_PATH` does **not** isolate you from this — it redirects the config file only, and the database path is read from `[paths] data_dir` *inside* that config.

## Tests

275 passed across the realtime + V3 contract suites; 873 passed / 1 skipped across `Tests/DB` + `Tests/ChaChaNotesDB` (the migration's home). The three V3 contract files are byte-identical to the base throughout. Pre-existing failures and the four pre-existing collection errors were reproduced at the base commit in a scratch worktree and are unrelated to this work; `scripts/check_persistent_diagnostic_inventory.py` fails identically at the base (drift owned by another branch, deliberately untouched here).

## Follow-ups filed

- **task-2390** — cost-chip integration for the audio-token and transcription-duration usage 2363 captures (the pricing model is per-mtok and can't express per-audio-minute billing without design work).
- **task-2391** — the empty-transcript row is still unexplained to the user: the status is stamped in memory on a row that is never persisted and never rendered.

---
## Summary by cubic
Implements Realtime V4 follow-ups 2360–2364: buffers mic audio across reconnects, prevents idle cutoffs mid-sentence, completes usage capture (transcription duration + audio token split), and adds structured per-message metadata. Ships ChaChaNotes v30→v31 with a new local-only `messages.metadata_json`; update other checkouts on the shared DB before launching.

- New Features
  - Structured per-message metadata (`messages.metadata_json`): engine/provider/model, `interrupted`, and transcript status; read on resume and used by reseed/export.
  - Realtime usage completeness: new `on_transcription_usage` captures transcription duration on the user turn; `ProviderUsage` adds `audio_input`/`audio_output`. Cost chip wiring deferred to task-2390.

- Bug Fixes
  - Reconnect preserves mic audio by re-buffering frames until the new session is ready (task-2360).
  - `on_speech_started` while live refreshes the idle anchor in both barge-in modes to avoid mid-utterance ejection (task-2361).
  - Closed fake-server/probe drift: pinned `whisper-1`, asserted `input_audio_buffer.committed`, and fixed the singular `input_token_details` alias (task-2362).
  - Seed builder trims only the appended interrupted marker (suffix) instead of stripping matching text in normal content (task-2364).
  - Robustness: stored metadata booleans parse correctly (e.g., string "false" no longer reads as True) and transcription durations are clamped to finite, non-negative seconds; token count parsing guards against `Infinity` overflows.

<sup>Written for commit 1625777ded9c554fcc5d007de7c886af32a9c00e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1354?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
**Qodo round addressed** (`1625777de`), all three findings:

- **`interrupted` could restore inverted** — `from_json` used Python truthiness, so a stored JSON string `"false"` came back as `True`. Fixed with a closed string vocabulary; mutation-verified RED on exactly the four-case inversion set (`"false"`, `"False"`, `"0"`, `"garbage"`).
- **Unsanitized wire durations** — fixed, and it turned out to be more than the reported bug: reusing the existing `_as_seconds` alone would *not* have caught it, because `max(nan, 0.0)` is `nan` and `max(inf, 0.0)` is `inf` (only `-inf` clamped, by accident). The added `math.isfinite` guard therefore also repairs `ProviderUsage.from_json`, where a stored `Infinity` previously restored intact. Verified by direct execution during review.
- **Manual `commit()` in the migration test — declined, with citations at the site.** Every schema-fabricating test in `Tests/DB` uses a raw connection; the v29→v30 sibling this file mirrors uses the identical ALTER+commit, and `test_chachanotes_character_authority_migration.py` mixes raw `executescript` for schema with `db.transaction()` for ordinary CRUD *in the same file* — the convention splits by purpose, not by file. Recorded as convention-based rather than mechanically necessary.

Also fixed a coverage hole this branch had opened: two world-book fixtures rewind the schema by dropping newer columns and never dropped `metadata_json`, so re-migration was silently taking the "column already present" branch — green, but no longer exercising the clean v30→v31 path. Confirmed closed with an isolated probe (pre-fix a deliberately broken v31 ALTER raised nothing; post-fix it raises).

Rebased onto latest dev (103 commits, no conflicts). 971 passed / 1 skipped across `Tests/DB` + `Tests/ChaChaNotesDB` + usage + realtime wiring; the three V3 contract test files are byte-identical to the base.

**On the red `No duplicate backlog task IDs` check:** inherited, not from this PR. `origin/dev` already carries 19 duplicate task IDs (400–403, 506–510, 1333, 2231, …), so that gate currently fails on every PR. All seven IDs this branch adds (2360–2364, 2390, 2391) are unique — verified one file each. Fixing dev's duplicates means renaming other branches' task files, which belongs in its own change.

PR #1355, opened 2026-08-05
feat(watchlists): UAT batch 3 — one verb per operation, create form states its destination (tasks 2302/2303)

rmusser01:
## What (UAT batch 3 of 5 — tasks 2302, 2303)

Third batch from the 2026-08-04 sr-UX/HCI UAT (findings F1/F13, both HIGH; filed in #1335).

**task-2303 — one verb per operation.** Three near-synonym labels were doing *two different things*: the rail's "Add source" **assigned** an existing source while the header's "Create source" and the pane's "New Source" **created** one. Decision: NEW brings a source into existence; ADD files an existing one into a watchlist — and a membership label must *name* the pre-existence, since "Add source" reads as a synonym of "New source". Shipped: `New source` everywhere for create (toolbar, empty state, binding, guidance); `Add existing` from the watchlist side and `Add to watchlist` from the source side for assignment — one verb, direction made explicit by the object. Both assignment modals gained instruction lines ("Choose a source below to add it to this watchlist. No new source is created."). Assignment is now reachable from the source's own Inspector, not just the rail.

**task-2302 — the create form states its destination and lands there.** The form had no destination field at all: created under an active watchlist scope, a source silently landed in Unassigned, contradicting the first-run guidance. Now the destination shows *before* submit (defaulting to the active rail scope, with Unassigned explicitly selectable), the source lands there, and the confirmation names it. Also: the Type Select carries a visible label; the ignore-selectors block appears only for page sources, where element selectors actually apply.

## Review trail

Whole-branch opus review + 2 waves; final verdict ALL ADDRESSED. Findings closed: a missing server-backend gate on the new assign path (both directions now refuse from one shared reason string — verified live with membership rows `[]` where they had been `[1]`); two mutation survivors in the "watchlist set changed under a live form" corner (deleting the destination watchlist under an open form raised `InvalidSelectValueError` out of `compose()`; a watchlist created mid-session never reached the Select); and an evidence defect — geometry tests ran without production CSS, so a comment shipped calling TASK-1362's 91-column measurement wrong when it was right. The whole destination test file now runs against the real `TldwCli` with production CSS.

**Honest non-fix:** the UAT's F11/F12 truncation observation is recorded as **unexplained, not fixed** — under production CSS there are 164 columns of budget for the 83-character string at 235x52, so that copy could not have clipped in that layout. The shorter noise-field copy stays on its own merits (half the width, syntax detail in a tooltip), not on a width argument.

## Verification

**28 mutations across the waves, all RED, zero survivors**, every restore md5 byte-exact. Suites: 322 passed in the re-review sweep; 73 passed on the final three files; poisoned-order e2e 50 passed; `--collect-only` 9407 collected clean. **Live-verified twice** (fresh profiles, DB read-back): destination shown as `Reading` before submit → toast `Source created in "Reading".` → `watchlist_sources` row `Reading|HN Front Page`; destination switched to Unassigned → lands unassigned; both assign paths write and read as assignment; server backend refuses both directions with the rail's exact tooltip; deleting the destination watchlist under an open form degrades to Unassigned with the draft intact and no crash.

Two disclosed process incidents, both contained and now encoded as protocol: an accidental `git checkout HEAD --` wiped uncommitted work mid-mutation (reconstruction became review priority #1 — which is how the two survivors above were found), and an aborted mutation battery left one mutation applied (caught, then a control-flow census across the wave confirmed zero residue in the affected file).

---
## Summary by cubic
Clarified create vs assign across Watchlists and added a visible Destination to the New source form so new sources land where intended. Addresses UAT batch 3 (Linear 2302, 2303) to reduce confusion and make assignment accessible from both sides.

- **New Features**
  - One verb per operation: “New source” creates; “Add existing” (watchlist) and “Add to watchlist” (source) assign, applied across rail, toolbar, empty state, Inspector, guidance, and modals.
  - New source form shows a Destination select (defaults to active scope, includes Unassigned); values are preserved across rebuilds; created sources land there and the toast names it.
  - Added source-side entry point for assignment via the Inspector with a watchlist picker; watchlist-side keeps “Add existing”.
  - Type select labeled; ignore‑selectors render only for page types; added “Web page” so the control is reachable.
  - Service adds list_watchlists_for_source (mirror of list_sources) to avoid N+1 membership queries.

- **Bug Fixes**
  - Unified backend gate blocks assignment in both directions with the same reason on the Server backend; both directions read the same reactive gate.
  - Live-form changes handled: deleting the destination watchlist falls back to Unassigned; newly created watchlists appear in the select; type/destination choices are retained.
  - Copy and UI polish: first‑run guidance matches on‑screen labels; picker dialogs state that no new source is created; action labels sized to avoid clipping.
  - Tests added/updated for vocabulary and destination behavior, plus parity tests for membership queries; geometry tests run with production CSS.

<sup>Written for commit 62bad33a3e93b24729dadf0afba6778d18dd5e65. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1355?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
**Qodo finding 1 (unwrapped db read in `list_watchlists_for_source`) — declined, with the file measured.**

`watchlist_bundle_service.py` has a clean, deliberate split: every one of its **5 `transaction()` sites is a write** (INSERT 132, UPDATE 163, DELETE 180, INSERT 219, DELETE 237) and every one of its **6 bare `conn.execute` sites is a SELECT** (194, 256, **281 = the new method**, 302, 325, 341). The new read byte-mirrors its nearest sibling `list_sources`, so wrapping it would make it the only read in the file that does — introducing the inconsistency the rule exists to prevent.

This is the same rule adjudicated per-file elsewhere in this programme, and the answer differs by file because the convention does: in `Subscriptions_DB.py` reads **do** wrap (nearest sibling `get_item_status` wraps), so PR #1342's identical finding was **accepted** and `get_new_items` was wrapped. Here the file's own convention says otherwise. If the project wants reads wrapped everywhere, that is a repo-wide sweep with its own task — not a one-method exception inside a file that consistently does the opposite.

PR #1359, opened 2026-08-05
Toad-inspired UI improvements: textual pin, watcher coalescing, timer guards, tool-call diffs

rmusser01:
## Summary

Salvage of the surface-independent work from a toad-inspired UI improvement series (study: `backlog/docs/toad-tui-improvement-study.md`). **Context/history:** this work was originally implemented against the legacy `ChatWindowEnhanced` surface; that surface has since been removed from `dev`, so this branch re-applies only what is meaningful on current dev. The console-native port of the remaining UX work (input ghost-text/history, transcript pruning, inline tool-call diffs) is tracked as follow-up tasks.

- **Deps (TASK-1353/1351):** exact-pin `textual==8.2.8` and add `textual-diff-view==0.1.5` (pyproject + requirements); runtime contract test accepts exact 8.x pins; ADR-032 for the new dependency.
- **Notes watcher (TASK-1352):** coalesce watchdog event bursts into one `on_files_changed` callback (250ms window via `call_soon_threadsafe` + single `call_later`); fixes latent `asyncio.create_task`-from-watchdog-thread bug.
- **Timer guards (TASK-1352):** `is_attached`/`screen.is_active` early-returns on 10 continuous `set_interval` animation/poll sites so hidden tabs burn no CPU.
- **Tool-call diffs (TASK-1351):** `WriteFileTool` captures before/after content (256KB/side cap, validated paths, undecodable/oversized → skip); new `diff_widgets.py` (`make_diff`/`extract_diff_from_result`/`strip_diff_contents`); `ToolExecutionWidget` mounts `DiffView` after off-thread `prepare()`; raw contents stripped at the single dev serialization seam (`Agents/tool_catalog.py` `BuiltinToolProvider.invoke`) so they never reach the LLM or run logs.

## Notes for reviewers

- `ToolExecutionWidget` currently has no production caller on dev (its only consumer was the removed legacy surface); the DiffView integration is complete and tested, ready for the console tool-row port.
- Per-side 256KB capture cap; reloads show plain text results (ADR-032 documents the trade-off).

## Test plan

- [x] `Tests/CI/test_textual_runtime_contract.py` — 2 passed
- [x] `Tests/Notes` — 1308 passed
- [x] `Tests/Tools` — 120 passed; `Tests/Agents` — 781 passed
- [x] `Tests/Widgets` (tool/diff + guarded widgets) — 215+ passed
[...]

PR #1360, opened 2026-08-05
Console port series: composer history + ghost text (TASK-1364), transcript pruning, tool diffs

rmusser01:
## Summary

Console-native port of the toad-inspired UX work (study: `backlog/docs/toad-tui-improvement-study.md`), following PR #1359 which landed the surface-independent foundations. The Console is the only live chat UI; the legacy surface this UX was first built on has been removed from dev.

**Landed so far (TASK-1364):**
- `Chat/prompt_history.py`: shared JSONL prompt history (async IO, index-0-as-draft, 1000-entry cap, real timestamps) — ported from the earlier branch.
- Composer ghost text: dimmed suffix rendered in the composer's own render pass (render-only, height-invisible), most-recent-wins from history, right-arrow accepts; suppressed while the slash-command popup owns the draft.
- History recall: Up/Down gated to first/last visual row using the composer's real wrap math; live draft stashed/restored; caret to end on recall.
- Send recording at `submit_draft`'s accepted point (blocked/refused/empty sends record nothing); shared instance owned by ChatScreen with a factory test seam.

**Still to land on this branch:** TASK-1365 (transcript pruning), TASK-1366 (tool-call diff rows).

## Test plan

- [x] 46 new tests (22 composer history + 24 prompt history) — pass
- [x] Existing composer/controller/paste suites — 301 passed (test_console_command_composer, cursor, collapse, popup, chat_controller, paste_attach)
- [x] Two-stage review (spec + quality) passed; quality-review nits applied

---
## Summary by cubic
Ports console prompt history UX (TASK-1364) and adds height-based transcript pruning (TASK-1365) to keep long sessions responsive. Adds inline diffs for file-write tool markers (TASK-1366) with safe capture and session‑only display.

- **New Features**
  - PromptHistory: JSONL async store (1000 cap), real timestamps, consecutive-duplicate skip, most‑recent prefix complete() for ghost text.
  - Composer: ghost text (dimmed, caret at end, not for “/”; Right arrow accepts) and Up/Down recall gated to first/last visual row; stashes index 0; caret lands at end. Controller records only accepted sends; Screen owns a shared PromptHistory; composer warm‑loads; test seam via console_prompt_history_factory.
  - Transcript pruning: when virtual height > prune_high_watermark, drop oldest message groups until under prune_low_watermark. View‑only (store keeps full history), streaming/selected protected, scroll preserved or re‑anchored at tail; checks after refresh and at mount; high<=0 disables. Keyboard selection walks visible messages. Defaults in config: 20000/12000.
  - Tool diffs: file‑write results capture raw before/after at the provider strip seam via a diff_sink, paired with STEP_TOOL_RESULT, and attached as session‑only tool_diff (never persisted or sent to the model). Transcript renders an expandable inline `DiffView` row under the TOOL marker; shows a “Diff” expansion when no hidden full text; styled via `.console-transcript-tool-diff`.

- **Bug Fixes**
  - Diff‑only TOOL markers now offer the expansion affordance labeled “Diff”.
  - Diff pairing hardened for abandoned tool threads: pair by most‑recent matching tool name and drop stale captures; strip guarantee unchanged.
  - PromptHistory.append is serialized (asyncio.Lock) to prevent concurrent appends from interleaving with cap rewrites.

<sup>Written for commit b39b2f5a7e44127a88f0b4eea932e5e7d5c3aee8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1360?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Chat/test_console_agent_diff_channel.py
+++ b/Tests/Chat/test_console_agent_diff_channel.py
@@ -8,12 +8,13 @@ result (AC1 data path + AC3).
 """
 
 import json
+from collections import deque
 
 import pytest
 
 import tldw_chatbook.config as config_module
 from tldw_chatbook.Agents.agent_runtime import FENCE_OPEN
-from tldw_chatbook.Chat.console_agent_bridge import ConsoleAgentBridge
+from tldw_chatbook.Chat.console_agent_bridge import ConsoleAgentBridge, _pair_step_diff
 from tldw_chatbook.Chat.console_chat_models import ConsoleMessageRole
 from tldw_chatbook.Chat.console_chat_store import ConsoleChatStore
 from tldw_chatbook.DB.AgentRuns_DB import AgentRunsDB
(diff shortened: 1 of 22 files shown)


rmusser01:
Qodo review addressed in b39b2f5a7:

- **Write race (bug)**: fixed — `PromptHistory.append` is now serialized with an internal `asyncio.Lock`, so a whole-file cap rewrite can't interleave with a concurrent append (composer + controller share one instance). Regression test with 8 racing appends included.
- **Docstring/import nits**: fixed (`stash_draft` Args section; loguru moved to the third-party import group).
- **`default_prompt_history_path()` skips path_validation**: not applied — intentional. `path_validation.py` guards user/agent-supplied paths (tool args, ingestion, sync targets); this helper only returns `get_user_data_dir() / "prompt_history.jsonl"`, the app's own canonical per-user data location — the same unvalidated pattern used for the config TOML, chat databases, and model-catalog cache. No external input influences the path, so there's no traversal surface to validate.

PR #1364, opened 2026-08-06
fix: legacy web_search_tool snippet-key fallback (task-1341)

rmusser01:
One-line fix: the legacy `WebSearchTool` read a top-level `snippet` key that `perform_websearch`'s normalization never produces (body text is top-level `content`; `snippet` lives under `metadata`), so every real result rendered "No description available". Falls back to `content` now, with a regression test pinning the real payload shape.

Found during the phase-3a review of the new web_search core (which carries the same fix); filed as task-1341.

- [x] Tests: new Tests/Tools/test_web_search_tool.py pins the normalized shape; Tests/Tools 131 passed

---
## Summary by cubic
Fix the legacy `WebSearchTool` to prefer `snippet`, then fall back to `content`, otherwise show "No description available", matching the `perform_websearch` payload. Adds regression tests that pin this shape and verify empty `content` falls through to the placeholder; satisfies task-1341.

<sup>Written for commit 42d465776d2bc9c2d9a06ba1e12919a144292244. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1364?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Tools/test_web_search_tool.py
+++ b/Tests/Tools/test_web_search_tool.py
@@ -1,0 +1,41 @@
+"""Regression tests for legacy WebSearchTool result mapping (task-1341)."""
+
+from tldw_chatbook.Tools.web_search_tool import WebSearchTool
+
+
+def _real_payload():
+    """The shape perform_websearch actually returns (WebSearch_APIs.py:
+    process_web_search_results): body text is top-level `content`;
+    `snippet` exists only inside `metadata`."""
+    return {
+        "results": [
+            {
+                "title": "Result 1",
+                "url": "https://example.com/1",
+                "content": "the actual body text",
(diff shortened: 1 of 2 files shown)


rmusser01:
Qodo items addressed: empty-`content` now falls through to the placeholder (`or "No description available"`), test docstring and type hints added, plus a regression test for the empty-string case.

PR #1366, opened 2026-08-06
fix(realtime): explain the empty voice-transcript row; bill realtime audio in the cost chip (tasks 2391 + 2390)

rmusser01:
## What this is

**task-2391** — the last user-visible gap from the realtime voice engine (V4): a voice turn that produced no transcript left an unexplained blank row in the Console.

The row's `transcript_status="empty"` existed only in memory (the store defers persistence for content-less rows, and `CharactersRAGDB.add_message` refuses a message with neither text nor image — both verified during review), and nothing anywhere read the field. So the row vanished on restart and told the user nothing while it was there.

## The change

An empty voice turn now writes a real, visible placeholder — `(no speech detected)` — through the existing update-content flush path, so the explanation is durable and renders through the ordinary message path with no new widget. `metadata.transcript_status` remains the machine fact; the content carries only the human-readable chrome. That's the same split task-2364 established (visible `⏹ interrupted` marker in content, structured flag for deciding logic), not a return to state-in-content.

## The regression this PR also fixes (worth reading)

Making the row non-blank had a consequence the first commit missed: it stopped being skipped by the falsy-content guard when building provider context, so the model started receiving `{"role": "user", "content": "(no speech detected)"}` — **a fabricated user turn**, on the mainline retry/edit/fork/send paths, for the life of the conversation.

The second commit excludes empty-transcript rows from every model-facing context builder behind a shared `_is_empty_transcript_row()` guard: `_provider_message_payloads`, `summarize_up_to`, and `impersonate_user_reply`. The guard keys on `metadata.transcript_status`, not on the placeholder text — so a user who literally types "(no speech detected)" still has their words reach the model, and legacy rows (no metadata) are unaffected.

Export paths were deliberately left alone: they are human-facing reads of a single message the user explicitly selected, and the context-preview modal's "what would be sent" field is built by the now-fixed builder. Reasoning recorded in the task's Implementation Notes.

## Review

Reviewed in two rounds. Highlights:

- The design premise was **verified, not assumed** — `add_message`'s refusal of content-less rows is genuinely unconditional, and the Console store's deferred-create guard independently requires content, so a metadata-only durable row was never possible.
- Completeness of the context-builder fix was checked by **independent enumeration** rather than by confirming the three fixed sites: every route from a stored message to a provider call was traced (retry, continue, edit-and-resend, submit, regenerate, the agent and direct branches, skill/dictionary/world-info processing, context compaction, auto-titling). No missed path.
- Mutation-checked: neutering the guard in two separate paths turns named tests red.

## Tests

177 passed across the realtime wiring suite plus the three V3 contract files (byte-identical to base); 296 passed across the console controller and store suites. Rebased onto current dev (121 commits, no conflicts).

Note for anyone rebuilding a venv: dev now requires `textual-diff-view`, which is unrelated to this change but is needed for these suites to collect.

## Remaining

**task-2390** (realtime cost-chip integration) is the last open follow-up and is *not* in this PR. Its research is done and recorded in the task file: the task's original premise was wrong — realtime audio is billed per 1M tokens, not per minute, so the existing per-mtok pricing model extends additively rather than needing redesign. Verified rates are in the task.

---
## Summary by cubic
Empty realtime voice turns now persist a clear "(no speech detected)" row and are excluded from all model-facing contexts. Realtime cost is now billed: audio tokens and Whisper transcription are priced and surfaced in the cost chip using seeded `openai:gpt-realtime` rates; addresses task-2391 and task-2390.

- **Bug Fixes**
  - Made the empty-transcript write path retry-safe: write the placeholder via `update_message_content`, then stamp `transcript_status="empty"`; if the placeholder already exists, still stamp the status.
  - Guarded all model-facing builders with `_is_empty_transcript_row()` and safe metadata reads, preventing crashes and fabricated user turns in send/summarize/impersonate paths.
  - Skipped empty-transcript rows in the realtime reseed builder; added tests across store, resume, controller, summarization, wiring, and pricing.

- **New Features**
  - Seeded pricing for `openai:gpt-realtime` (`-mini`, `-2.1`, `-2`, `-1.5`) and Whisper ($0.006/min); non-realtime models unchanged.
  - Updated `cost_for_usage` to price `audio_input`/`audio_output` and `transcription_seconds`, draining `cache_read` before `uncached_input` to avoid underbilling and double counting.
  - Extended `ConsoleCostRow` and the cost modal to show `audio_input`, `audio_output`, and `transcription_seconds`, with tests for pricing math and row/modal formatting.

<sup>Written for commit e18bced8737a59a5ea18df26b49e59e40002f8f5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1366?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Chat/test_console_chat_controller.py
+++ b/Tests/Chat/test_console_chat_controller.py
@@ -1889,6 +1889,26 @@ def test_image_budget_excludes_failed_send_blocked_echo(monkeypatch):
     assert decoded == b"real"
 
 
+def test_is_empty_transcript_row_tolerates_a_metadata_object_without_the_attribute():
+    """Qodo Q5 (task-2391 review): the docstring promised duck-typed safety
+    via `getattr` on `.metadata`, but the INNER `.transcript_status` read
+    was a plain attribute access -- a metadata object that duck-types some
+    fields but not that one raised `AttributeError` there. This helper runs
+    on every row of three model-facing send paths (`_provider_message_
+    payloads`, `summarize_up_to`, `impersonate_user_reply`), so that crash
+    reached the main send path, not just a narrow test double."""
+    from tldw_chatbook.Chat.console_chat_controller import _is_empty_transcript_row
+
+    message = SimpleNamespace(
+        role=ConsoleMessageRole.USER,
+        content="hello",
+        status="complete",
+        metadata=SimpleNamespace(engine="realtime"),  # no transcript_status
+    )
+
+    assert _is_empty_transcript_row(message) is False
+
+
 def test_provider_payloads_exclude_an_empty_transcript_placeholder():
     """task-2391 fix-now: a committed voice turn whose transcript came back
     empty persists real placeholder CONTENT ("(no speech detected)") so the
--- a/Tests/Chat/test_console_chat_controller.py
+++ b/Tests/Chat/test_console_chat_controller.py
@@ -1889,6 +1889,92 @@ def test_image_budget_excludes_failed_send_blocked_echo(monkeypatch):
     assert decoded == b"real"
 
 
+def test_provider_payloads_exclude_an_empty_transcript_placeholder():
+    """task-2391 fix-now: a committed voice turn whose transcript came back
+    empty persists real placeholder CONTENT ("(no speech detected)") so the
+    row can survive a restart -- but that content is UI chrome written so
+    the row could exist at all, not something the user said, and must
+    never reach a provider as if it were a real turn. Before this fix,
+    `_provider_message_payloads` had no `transcript_status` awareness, so
+    the placeholder rode straight through `_emit` into `{"role": "user",
+    "content": "(no speech detected)"}` on every ordinary send/retry/edit/
+    fork built off this session (`_provider_messages_for_session` backs all
+    of them) -- a fabricated user turn, permanently, for the life of the
+    conversation. An ordinary user row in the same session must still ride
+    through untouched."""
+    from tldw_chatbook.Chat.message_metadata import MessageMetadata
+    from tldw_chatbook.UI.Screens.chat_screen import (
+        CONSOLE_REALTIME_EMPTY_TRANSCRIPT_PLACEHOLDER,
+    )
+
+    store = ConsoleChatStore()
+    controller = ConsoleChatController(
+        store=store, provider_gateway=StreamingGateway(), model="test-model"
+    )
+    session = store.ensure_session()
+    store.append_message(
+        session.id,
+        role=ConsoleMessageRole.USER,
+        content=CONSOLE_REALTIME_EMPTY_TRANSCRIPT_PLACEHOLDER,
+        metadata=MessageMetadata(engine="realtime", transcript_status="empty"),
+    )
+    store.append_message(
+        session.id,
+        role=ConsoleMessageRole.USER,
+        content="a real question",
+    )
+
+    messages = store.messages_for_session(session.id)
+    payloads = controller._provider_message_payloads(messages, skip_failed=True)
+
+    contents = [payload["content"] for payload in payloads]
+    assert CONSOLE_REALTIME_EMPTY_TRANSCRIPT_PLACEHOLDER not in contents, (
+        "the empty-transcript placeholder must never be narrated to the "
+        "model as if the user said it"
+    )
+    assert "a real question" in contents
+
+
+@pytest.mark.asyncio
+async def test_impersonate_excludes_an_empty_transcript_placeholder():
+    """task-2391 fix-now (audit follow-up): `impersonate_user_reply` hand-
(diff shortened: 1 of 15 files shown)


rmusser01:
**task-2390 added to this PR** (`b43881623` + `f134ea109`) — the last realtime follow-up. The cost chip now bills realtime audio tokens and Whisper transcription.

**The task's own premise was wrong, and checking it was most of the work.** It claimed realtime is billed per audio *minute* and therefore inexpressible in the per-token pricing model. Verified against OpenAI's published pricing: realtime audio bills **per 1M tokens** (gpt-realtime: audio $32 in / $64 out), so `ModelPricing` extends additively exactly like the existing cache fields. Only Whisper transcription is per-minute ($0.006). Entries added for all five `gpt-realtime*` variants with an `as_of` label; the verified rate table is recorded in the task file.

**Two double-counting traps, both resolved by reading the code rather than assuming:**
- Audio token counts are a **subset** of the text buckets, not additional to them — pricing both would inflate the displayed cost. Audio is now drained out of the text buckets before the remainder is priced at text rates.
- Cached-*audio* attribution is impossible today (the payload detail isn't parsed), so all audio input is priced at the higher uncached rate — deliberately conservative. The published cached-audio rate is seeded but explicitly documented as unread at all three sites a reader would land on.

**Review caught a money bug in the first attempt.** The drain removed audio tokens from the *expensive* bucket first, which minimizes the remaining bill — the opposite of the conservative bound the task mandated and the docstring claimed. On an ordinary cached turn that under-stated cost by ~7% of the input side ($0.0252 vs $0.0270 on the worked example). Fixed to drain the cheap bucket first; the reviewer recomputed both figures independently and confirmed the direction holds on `gpt-realtime-mini` too, where the cached rates diverge most.

The test named "never underbills" had been pinning the implementation's own numbers, so it passed while the bug was live. It's now a genuine property test — it recomputes candidate splits from the raw rates without calling the pricing code, and asserts the implementation reaches the true maximum; mutation-verified to fail if the order is ever flipped back.

Non-realtime cost math is unchanged (mutation-verified isolated), token conservation holds across seven edge cases including the exact boundary, and the transcription seconds→minutes conversion is pinned. 86 + 103 tests green; contract trio byte-identical.

rmusser01:
**Qodo round addressed** (`5eb5fb4e5`), rebased onto latest dev (49 commits, no conflicts).

Two of the five were real bugs:

- **The fabricated-turn leak had a second door.** `_mark_console_realtime_transcript_empty` returned early whenever the row already had text — but the placeholder *is* text now. If the content write landed and the metadata write was swallowed, a retry never re-stamped `transcript_status`, and an unstamped placeholder isn't filtered, so it would reach providers as a fabricated user turn. Now a three-way branch keyed on **exact** equality with the placeholder (not substring), so a real transcript that merely contains that phrase is still treated as real. Mutation-verified: restoring the old early-return turns the new retry test red.
- **AttributeError on the main send path.** `_is_empty_transcript_row` read `metadata.transcript_status` directly after a `getattr`, so a metadata object lacking the attribute would crash inside all three model-facing builders. Now fully duck-typed; mutation-verified with a `SimpleNamespace` metadata object.

Also: the test now imports the placeholder constant instead of hardcoding it.

**Two rule violations declined, with reasoning:**
- *Use real in-memory SQLite instead of the persistence fake.* Declined for the existing test — that file's convention is the fake for "which persistence calls happened" claims (9 such tests; the direct sibling makes the identical claim shape the same way), so converting one in isolation would make it the odd one out. **But the concern was real**, so a new real-DB test was added: neither the fake test nor the resume test actually drove `update_message_content`'s deferred-create flush against a real database and read the row back (resume hand-seeds). That gap is now closed against `:memory:`.
- *Google-style docstrings on test functions.* Declined. These tests take no arguments and their docstrings carry the mechanism and the bug they pin, which is worth more than `Args:`/`Returns:` boilerplate; Qodo itself rated it Moderate with no precedent. No stealth churn — the one docstring edited was a factual correction forced by the code change.

394 + 103 + 187 + 373 tests green across the covering suites; contract trio byte-identical (MD5-verified).

**CI note:** the red `Core Tests` are a hosted-runner communication loss inside *cancelled* runs, not test failures. `No duplicate backlog task IDs` is inherited — `origin/dev` now carries 34 duplicate IDs on its own (400–403, 506–516, 1333, 1343–1345, 2231, …); all seven IDs this work adds are unique in both the filename and frontmatter namespaces, verified.

PR #1370, opened 2026-08-06
feat(watchlists): UAT batch 4 — readable reader, humane timestamps, check-now feedback (tasks 2307/2308/2309)

rmusser01:
## What (UAT batch 4 of 5 — tasks 2307, 2308, 2309)

Fourth batch from the 2026-08-04 sr-UX/HCI UAT (findings F19/F20/F24/F26/F27/F41, filed in #1335).

**task-2307 — the reader shows prose, not markup.** Feed bodies were rendered verbatim: `<p>Article URL: <a href="…">…</a></p>` literally on screen. New `Subscriptions/html_text.py` converts HTML to display text at *render* time (the stored body stays the honest capture of what the feed served, and every already-persisted item holds HTML). It **converts rather than sanitizes** — no allow-list that can fail open — and keeps link destinations visible as ordinary text, because a terminal reader who cannot see the address cannot judge it. `strip_control_characters` closes the hole `Text.append` leaves open: Rich writes segment text to the terminal verbatim, so a raw ESC in a feed body would otherwise paint a real OSC-8 hyperlink whose visible label is attacker-chosen.

**task-2308 — humane timestamps and real publish dates.** Sources/Items/Runs showed raw ISO-8601 with microseconds in UTC. Now local, human-scale, via one shared formatter — including the snapshot viewer's `created_at`, which sat next to an already-humanized byline. The Items column also showed *ingest* time (identical to the microsecond on every row from one check) where users need the item's **publish** time; it now shows publish, falling back honestly when a feed omits it.

**task-2309 — check-now has a pulse.** Immediate acknowledgment, a busy state mirrored on both trigger paths (Sources toolbar and Inspector), completion signals covering failure, and a debounce so a second press can't queue duplicate work.

## What review found (and why it mattered)

- **CRITICAL:** switching tabs mid-check cancelled the worker through Textual's unmount path, and since `CancelledError` is a `BaseException` on 3.11+, the pipeline's `except Exception` never saw it — the run row was stranded at `running` **forever**, silently. Fixed by catching it explicitly and recording an honest terminal state before re-raising. Durability was then verified *by construction*, not by observation: Textual delivers exactly one `.cancel()`, and the DB write is a synchronous transaction with no interruptible inner await, so the "cancelled" record cannot itself be cancelled mid-write.
- The two-layer fix then **doubled a side effect** — one cancelled check filed two identical notification rows (the run row was fine; it's an idempotent UPDATE). Redundancy is safe for idempotent writes and unsafe for emitting ones; the outer layer now checks for a terminal status first, with defense-in-depth re-verified by disabling the inner layer.
- Control-stripping stopped at the reader while **OPML import is a live delivery path** for C1 control bytes (a single-byte CSI parses fine through XML where raw ESC does not). Extended to 8 cell sites, then to the snapshot viewer's scraped page body — the highest-risk remote string in the feature, one function away from a line this batch had already edited.
- **Three coverage seams** landed exactly where predicted: this batch was authored by one implementer and tested by another after a rate limit, and mutation-gutting the self-closed-tag handler, the markdown control-strip branch, and the Expand button's pane wiring each left the entire suite green. Code correct, nothing guarding it — now tested.

## Verification

**Whole-branch review + 2 rounds; 18 mutations, all RED, all restores md5 byte-exact.** Counts: 405 passed across the 14 relevant files, 61 on the touched service layer, 25 on the round-2 files; `--collect-only` 9690 collected, zero errors. The one failing case (`_delete_item`) is pre-existing on dev and filed as task-2330. **Live-verified** against a real feed: readable prose with the URL visible, 20 distinct publish dates where one repeated ingest timestamp used to be, "Checking…" mirrored on both buttons, and — with a deliberately slow endpoint — a navigate-away mid-check logging `Check cancelled: navigated away before it finished.` and reaching `failed` instead of hanging at `running`.

**Honest gap:** task-2307's AC#2 (reader Expand affordance) is **unchecked** — the button posts its message but no screen handler exists yet. Recorded rather than claimed.

---
## Summary by cubic
Improves Watchlists readability and feedback: the reader now renders safe, readable prose (preserving angle‑bracket autolinks), timestamps are humanized with real publish dates, “Check now” shows instant ack, busy state, completion, and debounce, and the Reader’s Expand control now works. Addresses Linear tasks 2307, 2308, and 2309.

- **New Features**
  - Reader converts feed HTML to plain text via `Subscriptions/html_text.py`; links are visible, control bytes are stripped, and `<https://…>` autolinks are preserved; snapshot viewer body uses the same `readable_body_text` (task-2307).
  - Unified `humane_timestamp` in `UI/Watchlists_Modules/humane_time.py`; all Watchlists tables show local, human-scale times; Items prefer `published_date` with an honest “added <time>” fallback; snapshot header is humanized (task-2308).
  - “Check now” posts an immediate ack, shows busy state on both `SourcesPane` and `InspectorPane`, and emits completion/failure toasts; `_checks_in_flight` debounces duplicate presses and workers run in a named group (task-2309).
  - Reader Expand is wired to region solo; the button now toggles the layout and reflects the current state (task-2307).

- **Bug Fixes**
  - Handle `asyncio.CancelledError` in both `LocalWatchlistsService.execute_run` and `WatchlistScopeService.launch_run` so mid-navigation cancels mark runs terminal; prevent duplicate notifications on cancellation (exactly one now) (task-2309).
  - Strip terminal control bytes in identity cells across `sources_pane`, `inspector_pane`, `runs_pane`, `notifications_pane`, and `artifacts_pane`; snapshot bodies now route through `readable_body_text`; CR added to the strip regex (tasks-2307/2309).
  - Fix HTML detection so angle-bracket autolinks aren’t misclassified as tags; they are protected during parse and restored verbatim (task-2307).

<sup>Written for commit 1b4e973ed373fc80310f9297b8c456e0a0b718f0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1370?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Subscriptions/test_html_text.py
+++ b/Tests/Subscriptions/test_html_text.py
@@ -206,6 +206,32 @@ def test_tab_and_newline_survive_control_stripping():
     assert out == "line one\nline two\tindented"
 
 
+def test_carriage_return_does_not_survive_control_stripping():
+    """Batch-4 review, M1. The docstring's own claim ("C0 minus tab and
+    newline") always meant to cover CR (0x0D, neither tab nor newline), but
+    the regex range used to jump `\\x0c` -> `\\x0e-\\x1f` and skip it by one
+    code point -- a much weaker primitive than ESC/OSC (a bare CR can only
+    overwrite characters earlier on the same terminal line, a line-overwrite
+    spoof), but a real mismatch between what this module documents and what
+    it did.
+    """
+    out = strip_control_characters("Real Title\rEVIL OVERWRITE")
+    assert "\r" not in out
+    assert out == "Real TitleEVIL OVERWRITE"
+
+
+def test_a_crlf_feed_body_is_not_left_with_doubled_blank_lines():
+    """The reviewer's own condition for the fix: stripping CR must not turn
+    a `\\r\\n` line ending into a doubled blank line -- the `\\n` in the pair
+    is the one line break and it must survive untouched, exactly as if the
+    body had used `\\n` alone.
+    """
+    with_crlf = strip_control_characters("line one\r\nline two\r\nline three")
+    with_lf = strip_control_characters("line one\nline two\nline three")
+    assert with_crlf == with_lf == "line one\nline two\nline three"
+    assert "\n\n" not in with_crlf
+
+
 def test_none_becomes_empty_string():
     assert strip_control_characters(None) == ""
     assert html_to_display_text(None) == ""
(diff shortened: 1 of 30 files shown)


rmusser01:
**Qodo round — all five accepted, fixed in `8dfa472c0`:**

- ✅ **Q5 (angle-bracket autolinks)** — real content loss, confirmed: `<https://example.com>` was classified as HTML and dropped as an unknown tag. Fixed in two independent parts, because tightening alone is insufficient: (1) `_HTML_SHAPED` now requires a plausible tag name plus a real delimiter (your suggested shape), which excludes a bare autolink; (2) a protect/restore step inside `html_to_display_text` (Private Use Area placeholder, swapped before parsing and restored after) covers the case tightening cannot — an autolink mixed with genuine HTML in the same body, where conversion must still happen. Both mechanisms mutation-verified independently and reported as non-overlapping.
- ✅ **Q4 (Expand unhandled)** — **wired, not removed.** Routed `ExpandReaderRequested` into task-1344's existing `RegionLayout.solo` / `action_solo_region` path (same `_apply_layout` call, same off-Read refusal gate) rather than building a parallel mechanism. Found and fixed an adjacent truthfulness gap while there: `_build_content_pane` now seeds `pane.expanded` from the live layout, so the button's label stays correct across the rebuild its own press causes. task-2307 AC#2 is now genuinely met and checked. Live-verified: Expand collapses Feeds/Items to headers and gives Content the centre, relabels to "Restore", and a second press restores the three-pane layout exactly.
- ✅ **Q1/Q2/Q3** — `Args:`/`Returns:` added to `handle_check_now_requested`, `source_last_scraped_text`, `watch_busy_source_ids`.

Unrelated observation from this round's wider verification pass: `test_destination_visual_parity_correction.py` has 4 pre-existing failures on the **schedules** destination (`Local media DB is required…`), confirmed unrelated by isolated run — newly visible only because that file entered the verification set for the first time here. Filing separately.

PR #1373, opened 2026-08-06
feat(watchlists): UAT batch 5 — labels, proactive failure reasons, stable header, copy sweep (tasks 2310-2314)

rmusser01:
## What (UAT batch 5 of 5 — tasks 2310, 2311, 2312, 2313, 2314)

The final batch from the 2026-08-04 sr-UX/HCI UAT of the Watchlists screen. With this, **all 41 findings** filed in #1335 are closed.

**task-2312 — the tab strip stops moving.** The section tab strip changed *position* between sections (outside the content boxes on Overview/Sources, inside the bordered Feeds region on Items/Runs), so the navigation control visibly jumped as you used it. One factory now builds the centre header and mounts it through the workbench's `header=` slot on every section, replacing a conditional that was `None` off Items plus inline duplicates. That conditional was also hiding a real bug: navigating off Items dropped the header from `#wl-centre` entirely.

**task-2310 — every Select says what it filters.** Sources read `All ▼ / All statuses ▼ / All ▼`; Artifacts read `Auto + featured ▼ / App default ▼ / Off ▼` ("off" for *what*?). All are labelled now, including the two tooltip-only Selects in the briefing modals; the Rule form's Threshold explains its unit and range; and the Artifacts toolbar no longer shows "Stop Serving" before a briefing exists.

**task-2311 — failures say why, immediately.** Generate with no provider produced a bare `failed` row; the reason ("OpenAI API Key is required but not found") appeared only after clicking it, and the provider had silently defaulted. The reason now surfaces at failure time, naming the provider actually attempted, bounded in size and shape before it reaches a toast (full text still available in the row detail).

**task-2313 — one voice.** Three stacked empty-states on first run, "unavailable"/"ready" vocabulary that read as faults or riddles, duplicated backend display, "Import OPML" twice, scrape-vs-check drift, an unexplained "Queued" column, bare empty tables on Runs/Rules/Notifications, and an Inspector whose Console block outranked the selected object's own actions — all swept.

**task-2314 — wizard Escape.** Fixed a reflexive double-Escape landing on the confirmation dialog's own binding. **The AC was rewritten to match reality:** the UAT's original observation was that an *early* Escape is swallowed; live investigation found no such drop, so that hypothesis is recorded as investigated-and-not-reproduced rather than implied verified.

## What review caught

- **An honesty regression introduced by the copy sweep itself:** `latest_run_status` returning `None` for a not-wired service rendered identically to "no runs yet" — the same reassuring sentence for a broken service and a healthy unrun watchlist. Proven unguarded by mutation (reverting to a sentinel left the suites green). Now `not_configured` and `lookup_failed` are distinct states with distinct prose, and the screen's inconsistent literal `"unavailable"` fallback was folded into the same vocabulary.
- The unified header was verified rather than trusted: the reviewer traced the mechanism through region gating, the surface-refresh drain and the mount window, then mutated the wiring back to the old conditional and confirmed tests catch the regression.

## Verification

23 mutations across the branch, Edit-tool reverts only, md5-verified byte-exact restores. `Tests/Watchlists/` **454 passed, 0 failures**; destination shell 80 passed; artifacts pane 131 passed; wizard suite 96 passed; `--collect-only` 9215 collected, 0 errors. **Live-verified** on a fresh profile across all seven sections: the tab strip holds position, every Select carries a label, Runs/Rules/Notifications have guidance, the Rules threshold help updates live, and the not-wired state reads honestly.

Remaining failures are pre-existing and filed, not from this branch: `[_delete_item]` (task-2330), 4 schedules parity cases (task-2560), and the subscriptions-alias navigation case (verified failing identically on clean dev).

---
## Summary by cubic
Polished the Watchlists experience from UAT batch 5: unified header and tabs, labels on toolbar selects, proactive failure reasons, clearer copy and empty states, and a safer Escape flow in the first-run wizard.

- **New Features**
  - One stable centre header and tab strip across all sections; snapshot/status line unified. (task-2312)
  - Added visible labels to toolbar Selects in Items/Artifacts/Rules and related modals; Sources use compact tooltips where width is tight; Rules Threshold help explains unit and updates with Condition. (task-2310)
  - Artifacts shows the provider Generate will use before running, backed by public, documented `default_briefing_provider`. (task-2311)
  - Clearer copy: Items gets a Queued legend, Sources uses “Last checked”, backend shown once, Inspector’s type line reads as a short sentence, and the disabled follow button reads “Follow in Console”. (task-2313)

- **Bug Fixes**
  - Honest latest run status: distinct “not connected”, “couldn’t check”, and “no runs yet”; Overview cards render None as “-”, not “None”. (task-2313)
  - Briefing failures surface a bounded, immediate toast that names the provider and points to Settings. (task-2311)
  - First-run wizard absorbs a reflexive double-Escape in the finish-later dialog; a single early Escape still opens confirmation. (task-2314)

<sup>Written for commit b392789bd8f181e273bfc738d9f366f7e39a08b9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1373?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
**Qodo finding 1 (missing `Returns:` on `default_briefing_provider()`) — accepted, fixed in `b392789bd`.** The function became public in this batch precisely so the UI can show which provider a generation will use before the user presses Generate, so its contract is now caller-facing; the `Returns:` section states the resolution result and that it is never empty (callers display it without a None-check). `Tests/Subscriptions/test_briefing_service.py` 34 passed.

PR #1376, opened 2026-08-06
feat: web_crawl tool + ephemeral PDF fetch in web_fetch (tasks 1357/1358)

rmusser01:
## What this is

Web-tools v2 — the two follow-ups the v1 spec (PR #1363) deferred, now shipped per the owner-approved spec `Docs/superpowers/specs/2026-08-06-web-crawl-pdf-fetch-design.md`:

- **task-1358 — PDFs in `web_fetch`, ephemerally.** `web_fetch` no longer refuses PDFs. Detection is content-type OR `%PDF-` magic sniff (the sniff wins — servers mislabel constantly), and the read cap is raised to 20 MB **mid-stream** on detection, so a mislabeled 2 MB PDF works in one request. Over 20 MB → a clear `[too-large]` refusal (byte-truncated PDFs are unparseable, so the ceiling refuses rather than truncates); the *extracted text* is truncated at the caller's `max_bytes` with a pages-processed marker, satisfying the "truncated like HTML" AC. Extraction is pymupdf from bytes — nothing touches disk, nothing writes the media DB, and the page loop early-stops once past the cap. Structured refusals for encrypted / damaged / textless-scanned (`use media ingestion with OCR`) / pymupdf-absent.
- **task-1357 — `web_crawl`.** A budgeted same-host BFS built on the v1 core (per the owner ruling that **overrode the task's "wrap Article_Extractor_Lib" premise** — those crawlers launch Chromium per call, write resume files into the cwd, and have no page budget; ruling recorded in the spec and the task file). Returns a bounded page list (URL, title, ~200-char excerpt) with an honest stop-reason footer; the model follows up with `web_fetch` on pages that matter — and those follow-ups are instant because the crawl warm-writes the fetch cache with exact truncation-marker parity. Sitemap mode (`sitemap_url`) seeds from urlset or one level of sitemapindex. Every URL — discovered, seeded, or redirect hop — passes the v1 SSRF guard individually; 1 s/domain rate limit; 120 s deadline checked per page *and* between redirect hops. Registered as one `LocalToolSpec` (Ask default via the permission store); MCP exposure is automatic through the existing generic path, fail-closed for external callers.

Also rode along: two fetch-cache fixes motivated by the crawl (key by `(url, max_bytes)` — a small-cap fetch used to poison later full-cap calls — and a 256-entry bound, since the crawl bulk-loads what used to trickle).

## The review trail is the story

[...]

PR #1393, opened 2026-08-07
fix: web-tools polish — honest crawl output, hardened edges, docstring truth (task-2620)

rmusser01:
## What this is

**task-2620** — the deferred-findings polish pass from the merged web-tools v2 programme (PR #1376). Every reviewer-agreed-real-but-deferred finding is now fixed or explicitly re-ruled with reasoning. Nine behavioral fixes, three coverage guards, and a docstring-truth pass, all in `Tools/web_tool_impls.py` + tests (one docstring line in `Agents/local_tool_provider.py`).

## Behavioral fixes

- **Sitemap refusal contract**: defusedxml's refusals (`EntitiesForbidden` — a `ValueError`, not a `ParseError`) used to escape as raw exceptions (root) or abort the whole crawl (one hostile child). Now: `[crawl-failed]` for the root, skip-**and-count** for children — the footer says `; N child sitemaps skipped` when any were.
- **Honest stop reasons, three ways**: a deadline-truncated seed no longer claims "sitemap exhausted" (parent branch fixed one path; this closes the child-cap path with `"sitemap child budget reached"`), and — found by the final review — the **default path** overclaimed too: any real sitemap with more than `max_pages` URLs drained the queue exactly at the budget and claimed exhaustion. Now `"page budget reached"`.
- **Redirect-duplicate dedup, without the regression it almost shipped**: two URLs redirecting to one target are listed once — and the final review caught that the first implementation keyed dedup on *enqueued* URLs, silently discarding an already-fetched page's content (probe: fetched body thrown away, "Crawled 1 pages"). Dedup now keys on a `listed` set of URLs that actually produced rows.
- **Unspoofable blocked-vs-failed accounting**: `startswith("[ssrf]")` instead of substring — a URL path containing `[ssrf]` can no longer inflate the blocked count.
- **`html_only` aborts sniffed PDFs**: a PDF mislabeled `text/html` no longer streams 1 MiB during a crawl just to be discarded.
- **No 20 MB download without pymupdf**: a cheap `find_spec` probe keeps the PDF read at the ordinary cap and refuses `[missing-dep]` when extraction is impossible (`optional_deps.check_dependency` was deliberately not used — it eagerly imports the module, the wrong cost for the fetch hot path; ruling recorded).
- Smaller: `[too-large]` copy derives from `PDF_MAX_BYTES`; `<title>` accumulation bounded at 512 chars (unclosed-tag memory); namespace-less (non-conformant) sitemaps now parse instead of silently yielding zero URLs.

## Coverage + truth

- New guards: a static ephemerality test (pins task-1358's "no media-DB writes" claim against the module source), the previously-uncovered between-hops `_CrawlDeadline` path, and the frontier cap pinned bidirectionally (`== 6`).
- Docstring truth pass with review teeth: the reviewer rejected "LRU" (the cache evicts earliest-expiry/FIFO — an LRU would spare hot entries; this doesn't), an incomplete error-reason list, and — twice — overstated claims about the pymupdf-absent path, until the shipped prose matched the code exactly ("the probe chooses the read CAP; it does not skip the download itself").

## Review trail

Two SDD tasks, four fix rounds total, an opus whole-branch review with live probes, and an opus re-review of the fix wave. The two findings the fix wave itself introduced are resolved or parked with rulings: a backlog ID collision (2870 raced another worktree mid-flight → renumbered **task-2910**) and a conservative-direction `budget_truncated` false-positive (under-claims completeness on trailing skippable candidates; verified 3-line fix recorded in task-2910 rather than risking a third honesty round unreviewed).

## Tests

408 passed across `Tests/Tools/` + `Tests/Agents/test_local_tool_provider.py` at HEAD; clean 31,537-test collection sweep. Simulated pymupdf-absent run verified the exact 30-passed/11-skipped split (the v1 SSRF suite always runs). task-2620 closed Done with Implementation Notes and won't-fix rulings; residuals live in **task-2910** with per-item reasoning.

---
## Summary by cubic
Polishes `web_crawl`/`web_fetch` for honest sitemap/PDF handling, tougher edges, and docstring/spec truth. Adds stop-reason priority and child-sitemap skip counts, dedups redirect targets via a listed set, and makes the PyMuPDF check total and single-probe; closes task-2620.

- **Bug Fixes**
  - Sitemap parsing: failures (incl. defusedxml refusals that raise `ValueError`) produce `[crawl-failed]` for the root; child failures are skip-and-count, shown as `N child sitemaps skipped`.
  - Stop reasons: priority is `deadline reached` > `sitemap child budget reached` > `page budget reached` > `sitemap exhausted`. Budget-truncated urlsets report `page budget reached`; exact consumption still reports `sitemap exhausted`.
  - Sitemap seeding: host/dup filters run before the budget check to avoid false “page budget reached” when only trailing off-host/duplicate entries remain.
  - Redirect dedup: list final targets once using a `listed` set; multiple URLs redirecting to one target produce one row while each still spends an attempt.
  - Blocked vs failed: classify SSRF only when the message starts with `"[ssrf]"`.
  - Crawl `html_only`: abort after sniff when a mislabeled PDF is detected; emit `[application/pdf]` without draining the body.
  - PDF path: `_pymupdf_available()` is total (guards `ImportError`/`ValueError`), probed once per `web_fetch` call, avoids the 20 MB read when `pymupdf` is absent and raises `[missing-dep]`; `[too-large]` renders the MB from `PDF_MAX_BYTES`.
  - HTML parsing: bound `<title>` accumulation to `CRAWL_TITLE_MAX_CHARS`; namespace-less sitemaps now parse.

- **Refactors**
  - Coverage/docs: added ephemerality guard, between-hop deadline coverage, and an exact frontier-cap assert; expanded `web_fetch` docstring with Args/Returns, PDF detection, cache, error reasons, and ceiling rules; added `web_crawl` to the local tool list; spec updated with child-skip footer and stop-reason priority.
  - Internal: introduced `_SitemapSeed` to surface `children_capped`, `budget_truncated`, and `children_skipped`; named `CRAWL_TITLE_MAX_CHARS` constant.

<sup>Written for commit f283734cb2a6d5b264c5bcb2be55b9cfb4fa2e83. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1393?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Adjudication of the Qodo findings — four fixed in `e43d90c9b`, one no-change with reasoning:

**Fixed:**
- **Magic 512** → `CRAWL_TITLE_MAX_CHARS`, matching the module's own convention that every cap is a named constant; the test asserts against the constant.
- **Args/Returns** → added to `web_fetch`'s docstring per the repo's Google-style standard. Every pre-existing sentence is byte-identical (verified by diffing the extracted prose blocks — the docstring just passed a truth review and rewording would have un-verified it); the new sections' claims were verified against the code, including the subtle one (for PDFs, `max_bytes` caps the *extracted text*, not the download).
- **Probe totality + per-hop calls** → `_pymupdf_available` now catches `(ImportError, ValueError)` (the spec-less `sys.modules` stub case reproduces real `find_spec` behavior — confirmed in a clean interpreter), and `web_fetch` probes once before the redirect loop, using the same answer for cap selection and the `[missing-dep]` branch, so the two can never disagree mid-call. These were our own parked residuals in task-2910 (items 1/2) — with an external reviewer flagging them, the recorded fixes went in; the task file annotates them resolved.
- **`budget_truncated` false-positive** → cap check moved after the host/dup filters in `take()` and the child loop (task-2910 item 6, the verified reorder). All four invariants hold: trailing skippable candidates → "sitemap exhausted"; genuinely-remaining candidates still flag; exactly-consumed still exhausted; `SITEMAP_MAX_CHILDREN` accounting unchanged. Three new RED-first tests.

**No change — `_pymupdf_available` bypasses `optional_deps` (rule 497159):** this is a recorded ruling in task-2620's Implementation Notes and the probe's own docstring: `optional_deps.check_dependency()` **eagerly imports the module**, which is the wrong cost profile for a probe that runs on the fetch path precisely to decide things *before* heavy work; the heavy import stays local per the module's v1 trafilatura precedent. Centralizing would trade a documented 3-line probe for an eager import regression.

Post-fix: 103 passed across the web-tools suites; branch rebased onto current dev (89 commits, clean) with 409 passed across the branch-relevant suites.

PR #1401, opened 2026-08-07
fix: web_crawl surfaces row-less attempts; close web-tools residuals (task-2910)

rmusser01:
## What this is

**task-2910** — the final residuals of the web-tools programme (follow-up to #1376 and #1393). Every remaining item is now fixed or re-ruled with reasoning, and the task is closed Done. One small behavior change, the rest is prose truth and coverage.

## The behavior change

**Row-less attempts are now visible.** `web_crawl` dedups redirect targets against the listed-URL set (from #1393), which meant an attempt could spend its `max_pages` slot without producing a row — invisible to the model reading the footer. Following the established clause idiom, the footer now reports it:

```
Crawled 2 pages (0 failed, 0 blocked; 1 duplicate redirects skipped). Stopped: no more links within depth.
```

The counter increments at exactly one site (the dedup `continue`), the clause renders only when nonzero, and the spec's footer contract is updated. RED-first tests both ways (clause present with 1, absent with 0).

## Prose truth + coverage

- The skip-reason enumerations (`_SitemapSeed` docstring + spec) now include the fourth cause the code actually counts: a deadline expiry mid-child-fetch.
- `web_crawl`'s attempt-invariant docstring covers **both** row-less-attempt shapes — and the review caught that the draft asserted which shape is "more common" as fact when it's actually a discovery-order coin toss that was never measured. The frequency ranking was dropped for neutral both-ways wording (this module's reviews have now caught unverified prose four times; the streak continues to pay for itself).
- New test pins the simultaneous deadline+budget seeding case ("deadline reached" wins — verified genuinely simultaneous, not an accidental single-path test).

## Re-ruled WON'T-FIX (reasoning in the task file)

- Foreign-namespace sitemaps (spec scope was namespace-*less* specifically; a foreign namespace collapses into the honest zero-pages outcome).
- The "0 MB" `[too-large]` copy under a sub-MiB ceiling (reachable only by monkeypatching a code constant; the derived copy keeps message and constant in lockstep at any realistic value).

## Tests

416 passed across `Tests/Tools/` + `Tests/Agents/test_local_tool_provider.py`; 107 across the two web-tools files specifically. This closes the last open thread of the web-tools programme — tasks 1357, 1358, 2620, and 2910 are all Done.

---
## Summary by cubic
Surfaces row-less attempts in `web_crawl` by reporting duplicate redirects skipped in the footer, and closes the remaining web-tools residuals from task-2910 with spec/doc updates and tests. Improves transparency of attempt budgeting without changing crawl counts or scope.

- **Bug Fixes**
  - `web_crawl`: count `duplicates_skipped` when a redirect lands on an already-listed URL or a fetch targets an already-listed URL; footer shows `; N duplicate redirects skipped` when nonzero.
  - Spec/docstrings: add deadline-expiry as a child-sitemap skip reason; describe both row-less attempt shapes; drop the unverified frequency claim.
  - Tests: cover footer clause presence/absence and pin sitemap seeding precedence where “deadline reached” outranks page budget when both hit.
  - Task-2910: record WON'T-FIX for foreign-namespace sitemaps and the sub‑MiB “0 MB” PDF-ceiling copy; mark task Done.

- **Refactors**
  - `web_crawl` docstring: add Args/Returns with budget-clamping notes; no behavior change.
  - `_format_crawl_result` counters passed as keywords; `.superpowers/` added to `.gitignore` (scratch dir).

<sup>Written for commit 8f6d68bab356f80be2df2510bb471d94ede27a47. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1401?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
All three Qodo findings fixed in `8f6d68bab`:

- **Committed scratch report** — good catch, and the root cause is now closed for good: `.superpowers/` (per-session agent scratch) was never in this repo's `.gitignore`, which is how the report slipped into a commit. The file is untracked and the directory is now ignored, so this class of leak can't recur.
- **`web_crawl` Args/Returns** — added in the same style `web_fetch` received on #1393; every claim in the new sections was verified against the code (the `_coerce_budget` clamps and ceilings, the footer contract), and the existing truth-reviewed narrative is byte-identical around the additions.
- **Keyword args for the optional footer counters** — done; zero behavior change.

Post-fix: 107 passed across the two web-tools test files; branch is rebased onto current dev.

PR #1381, opened 2026-08-06
refactor(console): decomposition wave 2 — workspace, hands-free, session controllers

rmusser01:
Wave 2 of the Console decomposition, per the plan merged in #1347. **Pure refactor — zero behaviour change is the contract**, and the whole-branch review *measured* that rather than asserting it.

Spec: `Docs/superpowers/specs/2026-08-02-screen-decomposition-design.md`. Plan: `Docs/superpowers/plans/2026-08-04-console-decomposition-wave2.md`. Follows wave 1 (#1345).

## What moved

| New module | Cluster |
|---|---|
| `UI/Console_Modules/workspace.py` | `ConsoleWorkspaceController` — 35 of 45 `*workspace*` methods, incl. the 205-line `_resume_console_workspace_conversation` |
| `UI/Console_Modules/hands_free.py` | `ConsoleHandsFreeController` — 28 methods, the engine fork, and both cross-engine action entry points |
| `UI/Console_Modules/session.py` | `ConsoleSessionController` — 27 methods, incl. the 215-line `_start_character_console_session` |

**`chat_screen.py`: 24,195 → 20,964 lines. `ChatScreen`: 680 → 612 methods** (−3,231 / −68). The screen is still large, deliberately: it keeps cross-region coordination, the Textual lifecycle, and the `action_*`/`@on` entry points Textual resolves by name. The criterion is ownership, not line count.

## Evidence the contract held

The whole-branch review ran a normalized AST diff (docstrings/comments stripped, controller prefixes collapsed) over **all 612 retained `ChatScreen` methods**: exactly **6 changed**, every one a pure delegation. `on_button_pressed`, `on_key` and `compose_content` are byte-identical. Of 89 moved methods, all 89 landed in a controller with identical bodies modulo seam renames. Worker group names *and counts* identical to base (25 names). pyflakes on `chat_screen.py`: 31 findings before, 31 after, set-identical.

Controller discipline, verified per module: **zero `query_one`/DOM access** in any of the four controllers; `self._screen` uses held to 3–6 each, every one a framework service or a documented sibling proxy; **zero** sibling reach-through (`self._screen._session|_workspace|_hands_free|_dictation`).

## The seam rule this wave established (now in DESIGN.md §7)

Three controllers on one screen forced the question wave 1 didn't face. Controller-to-controller traffic goes through **named callables the screen wires at construction** — never a back-door through screen attributes. Two consequences, each of which cost a review round:

[...]

PR #1408, opened 2026-08-07
refactor(console): decomposition wave 3 — message, transcript, prompts + a size ratchet

rmusser01:
Wave 3 of the Console decomposition, per `Docs/superpowers/specs/2026-08-02-screen-decomposition-design.md` and the wave-3 plan. Follows wave 1 (#1345) and wave 2 (#1381).

`chat_screen.py` **20,964 → 18,905 lines**, `ChatScreen` **612 → 598 methods**.

## What lands

| | |
|---|---|
| `Console_Modules/message.py` — `ConsoleMessageController` (2,042 lines) | −1,328 |
| `Console_Modules/transcript.py` — `ConsoleTranscriptRegion` (257 lines) | +1 |
| `Console_Modules/prompts.py` — `ConsolePromptsController` (1,244 lines) | −759 |
| `Tests/Architecture/test_screen_size_ratchet.py` | the wave's real deliverable |

**Zero behaviour change** is the contract. Every moved body is byte-identical to its pre-move form apart from mechanical receiver rewrites (`self._session.X` → `self.X`, `self.app.push_screen` → `self.push_screen`), verified by AST diff against the merge base — including the 397-line `_open_console_prompts_modal` and the 294-line `handle_console_message_action`.

## Why the ratchet exists

Waves 1 and 2 extracted ~4,900 lines from `chat_screen.py`, and the file finished those waves **larger** than it started: ~5,500 lines of concurrent feature work landed over the same window, every line going into the screen because the screen was the path of least resistance.

Extraction alone cannot win. `test_screen_size_ratchet.py` makes the current size a **ceiling** rather than a waypoint — budgets may only go DOWN, and a second test fires if a wave lands without lowering one. Both tests are mutation-verified rather than merely written.

## Honest accounting

Two things this wave did *not* achieve, recorded in the spec's Progress section rather than smoothed over:

- **The transcript region netted +1 line and +1 method.** The composed block was 13 lines and the moved bodies 54; the delegations cost the same. Kept because all three `#console-workspace-grid` children are now built identically, and leaving one of three inline is the asymmetry that costs when `compose_content` is attacked. It leaves the transcript's DOM with **two access idioms** (3 delegating, 9 still direct) — a transitional state, filed as task-2767, not a settled design.
- **`_open_console_prompts_modal` moved verbatim at 397 lines**, carrying 15 of the controller's 18 named dependencies. Review judged the fan-out inherent rather than a wrong boundary, but the method is a class in disguise (17 nested closures). Decomposition filed as task-2766.

## Defects the reviews caught

- **A getter-only proxy property** over a baseline plain attribute turned **41 tests red in a file this branch never touched**. The fix round found the review had itself under-counted: six affected fixtures, not one, and the setter alone didn't clear it (a proxy setter writes *through*, so the traceback merely moved). Now a `DESIGN.md` §7 rule with an AST audit behind it.
- **`test_chat_screen_worker_groups` had silently stopped guarding.** It AST-parses `chat_screen.py` only; as waves moved workers into `Console_Modules/`, 4 of 6 run-coroutine dispatch sites and 7 exclusive-worker sites left its scope while both tests stayed green. Widened to `chat_screen.py + Console_Modules/*.py`, plus an assertion that the named coroutines are actually found — so the erosion itself now fails, not only its consequences. Mutation-verified both directions.
- **Wave 2's dev debt fixed**: `test_console_sync_records_worker_lifecycle` has been red on `dev` since #1381 — `MagicMock(spec=ChatScreen)` reads the CLASS, so `__init__`-wired controllers are invisible to it. Slots are now derived from `__init__` via `ast`, so wave 4's controller is picked up automatically.

## Not fixed here, deliberately

`Docs/security/production-diagnostic-inventory.json` is stale and **red on clean dev** (verified at `22c08f958`). Regenerating means signing off on ~199 lines including new diagnostic owners in two unrelated modules — exactly the rubber-stamp its "review the diff before running --write" gate exists to prevent. Filed as **task-2768** for whoever owns those diagnostics.

## Verification

- Serial, `-p no:randomly`: 104 + 113 + 62 + 304 = **583 passed, 0 failed** across the three new test files, the geometry baseline, the architecture suite, and every file referencing a moved name.
- `pytest Tests/ --collect-only` → **31,796 collected, 0 errors**.
- The geometry baseline (`test_console_shell_regions.py`) is **byte-identical to dev** and green.
- pyflakes on `chat_screen.py` is byte-identical to the merge base (31 findings, same set).
- Whole-branch review on the most capable model: one Important (the worker-group guard, fixed above), two Minor (fixed). No Critical.

Follow-ups filed: **2766, 2767, 2768, 2769**.

---
## Summary by cubic
Decomposed the Console screen: moved message logic, transcript region, and prompt workflows into dedicated modules, and added a one-way size ratchet to keep `chat_screen.py` small. No behavior change; `chat_screen.py` is reduced 20,964 → 18,909 lines and `ChatScreen` methods 612 → 598.

- **Refactors**
  - Extracted message cluster to `ConsoleMessageController` with thin delegations on `ChatScreen`.
  - Moved transcript DOM to `ConsoleTranscriptRegion`; ids/structure unchanged; added delegations for reading-state helpers.
  - Extracted prompt cluster to `ConsolePromptsController`; kept command/menu entry points via delegations.
  - Added one-way size/method ratchet in `Tests/Architecture/test_screen_size_ratchet.py` (budgets 18,909 / 598), cached the measurement, added docstrings, and documented the read-write proxy rule in `DESIGN.md`.

- **Bug Fixes**
  - Restored write access for `_console_original_attempt_previews` and `_last_console_action` via proxy setters.
  - Widened worker-group guard to scan `UI/Console_Modules/` in addition to `chat_screen.py`.
  - Hardened tests: attached `ConsoleMessageController` to `__new__` shells via a shared stub and correct app wiring order, and stubbed controller slots for `MagicMock(spec=ChatScreen)` to fix the sync test.

<sup>Written for commit 4981f9b1112af9d8327f43facc1757c07ae3878c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1408?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -424,3 +424,15 @@ wiring time. Second, if a proxy property is write-only, its getter must raise
 `getattr(obj, name, default)` swallow `AttributeError` specifically, so a
 defensive read would silently take the default forever with nothing raised
 anywhere.
+
+**New Console code goes in `UI/Console_Modules/`, and a ratchet enforces it.**
+The decomposition's first two waves extracted ~4,900 lines out of
+`chat_screen.py` and the file still ended up *larger* than when the work
+started, because ~5,500 lines of concurrent feature work landed in it over the
+same window. Extraction cannot outrun growth, so the screen's size is now a
+ceiling: `Tests/Architecture/test_screen_size_ratchet.py` holds a line and
+method budget that may only ever be lowered. A wave lowers it in the same PR
+that earns the reduction; a feature that would raise it belongs in a module
+instead. The test's failure message names the module directory and this
+section, because the moment it fires is the moment someone needs to know where
+else to put the code — not a link to chase later.
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -425,6 +425,24 @@ wiring time. Second, if a proxy property is write-only, its getter must raise
 defensive read would silently take the default forever with nothing raised
 anywhere.
 
+**A proxy property standing in for a plain attribute must be read-WRITE.**
+The rule above governs the write-only case; wave 3 found its mirror image and
+paid for it. Moving state into a controller turns what was
+`self._console_original_attempt_previews = {}` in `__init__` into a `@property`
+on the screen that forwards to the controller — and a `@property` with only a
+getter makes the name **assignment-hostile**, where the plain attribute it
+replaced accepted writes from anywhere. Wave 3 shipped one getter-only proxy
+and turned **41 tests red in a file the branch never touched**, all of them
+`AttributeError: property '…' of 'ChatScreen' object has no setter`. Two
+things follow. First, check the *baseline* shape before writing a proxy: if it
+was a plain assignable attribute, the proxy needs a setter, and "nothing writes
+it today" is a coverage claim rather than a contract — write the setter anyway.
+Second, the setter must write THROUGH to the controller (`self._message.x = v`),
+never to a shadow attribute the controller never reads; that variant is worse
+than the crash, because the tests go green over a dead write. An `ast` audit
+comparing every new property against the baseline attribute's shape catches
+both, and is cheap enough to run every wave.
+
 **New Console code goes in `UI/Console_Modules/`, and a ratchet enforces it.**
 The decomposition's first two waves extracted ~4,900 lines out of
 `chat_screen.py` and the file still ended up *larger* than when the work
--- a/Docs/superpowers/plans/2026-08-06-console-decomposition-wave3.md
+++ b/Docs/superpowers/plans/2026-08-06-console-decomposition-wave3.md
@@ -1,0 +1,149 @@
+# Console Decomposition — Wave 3 Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Extract the message cluster together with the visual surfaces that reflect its state, plus the prompt cluster — then lower the size ratchet so the gain is locked in rather than re-consumed.
+
+**Architecture:** Per `Docs/superpowers/specs/2026-08-02-screen-decomposition-design.md` and the four controllers + two region widgets already in `UI/Console_Modules/`. Message and its surfaces go together because the control bar and transcript reflect message/generation state; splitting them across waves would force a cross-wave seam.
+
+**Tech Stack:** Python ≥3.11, Textual, pytest.
+
+**Measured on dev at `22c08f958` (post-wave-2) — re-locate by anchor, never by line number:**
+- `chat_screen.py` 20,964 lines, `ChatScreen` 612 methods.
+- **message: 1,577 lines / 35 methods** — the largest remaining cluster. `handle_console_message_action` is 294 lines.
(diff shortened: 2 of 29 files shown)


rmusser01:
## Qodo review — 4 of 5 addressed in `f4a811ecc`, 1 declined with reasoning

**Fixed:**
1. **`handle_console_message_action` docstring** — Args/Returns added, plus why it stays one method. The body is unchanged: AST-verified identical to the prior commit with the docstring node excluded, so this wave's byte-fidelity contract still holds.
2. **Ratchet test missing docstring** — added, including the Raises section, since the failure message is the point of that test.
3. **`push_screen` missing Returns** — added.
4. **Duplicate AST parsing** — `_measure` is now `lru_cache`d. Noted in its docstring that the cache is per-process, so mutation-testing the ratchet (edit file, re-run pytest) still works.

**Declined — 5, "Slack permits budget drift":**

The 200-line / 10-method tolerance is deliberate, and tightening it would break the mechanism rather than strengthen it.

The two tests answer different questions. The first is the ceiling: *did this change grow the screen?* The second asks only *did a decomposition wave land and forget to lower the budget?* A zero-tolerance version of the second would fire on every ordinary commit that happens to delete a few lines — a routine dead-code removal, a shortened docstring — and demand a budget edit in the same PR. That trains people to edit `_BUDGETS` casually, which is precisely the habit that lets someone raise it later without thinking.

The finding is right that up to 200 lines can be re-consumed before the ceiling notices. That is the accepted cost. The alternative on offer costs the mechanism's credibility, and a ratchet nobody trusts gets deleted.

Qodo rates this `●● Moderate` and says "unclear preferred strictness without close precedent" — so recording the intent here rather than silently ignoring it.

## On the 23 CI failures

None are this branch's. All 23 are in `Tests/Notes/` (git/SSH: unborn-head, OpenSSH invocation, host-trust) and `Tests/Watchlists/`. This branch touches **zero** files in either tree (`git diff --name-only origin/dev...HEAD | grep -E "Notes|Watchlists"` → 0).

Evidence they are the dev baseline, not a regression:
- A sample (`test_openssh_invocation_is_exact_literal_direct_argv`, `test_a_failed_cast_leaves_the_briefing_detail_unchanged`) **passes locally** on this branch.
- Recently merged PRs shipped with the same band of failures: **#1409 → 20 failed**, **#1407 → 44 failed**. 23 sits between them.

Two further Architecture failures surfaced after rebasing onto latest dev — `test_profile_owned_path_inventory.py`, complaining about stale exception rules for `tldw_chatbook/UI/STTS_Window.py`. Both **reproduce on clean `origin/dev`** in a disposable worktree, and this branch does not touch that file. Same for `test_persistent_diagnostic_inventory` (filed as task-2768, deliberately not rubber-stamped here).

Branch-relevant verification, run serially with `-p no:randomly`: **583 passed, 0 failed** across the three new test files, the geometry baseline, the architecture suite, and every file referencing a moved name; `pytest Tests/ --collect-only` → **31,796 collected, 0 errors**.

PR #1404, opened 2026-08-07
feat: real Exa/Serper/Yandex search backends, live-verified (task-1355)

rmusser01:
## What this is

**task-1355** — engine choice becomes real. `search_web_serper` and `search_web_yandex` were literal `pass` stubs (the dispatcher even called them with **zero arguments**, feeding `None` into the result processor), and Exa was absent entirely. All three are now implemented, wired, config-keyed, swept across every surface that enumerates engines, and **live-verified against the real APIs** for the two engines whose keys exist.

## The backends

- **Serper** — `POST google.serper.dev/search`, `X-API-KEY`, `{q, gl (lowercased), hl, num}`; organic-only parsing into the standardized shape (`position` stays `metadata.position` — mapping rank into a "relevance" field would invert its meaning for any consumer sorting descending).
- **Exa** — `POST api.exa.ai/search` with `contents: {highlights: true}` — a **recorded paid trade**: highlights bill as contents retrieval, but a result with no snippet is nearly useless to the model. First highlight becomes the snippet.
- **Yandex** — Cloud Search API v2, the synchronous REST endpoint verified at the **proto level** (the docs site is CAPTCHA-walled; `WebSearchService.Search → POST /v2/web/search` is the citable ground truth). The response wraps base64-encoded XML in `rawData`; decoding/parsing uses the repo's defusedxml-with-fallback posture, flattens `<hlword>` highlight tags, and — the honesty requirement — **raises on the in-XML `<error>` element**: a quota or auth failure arrives inside an HTTP 200, and it must never render as "No results found" for a query that was never searched. Traced end-to-end: the model receives `[search-failed] … Yandex API error (code 32): Quota exhausted…`, not a swallowed empty list.

## The sweep (a new option must be taught to every surface)

Six engine enumerations existed; each got a grep-backed verdict: the agent tool's `SEARCH_ENGINES` +3 (schema enum updates automatically), the research-interop local set +exa (it routes to `perform_websearch`), and three **recorded skips** with evidence — `Utils.py`'s `global_search_engines` (zero consumers, dead list), the server-API schema (a contract superset that already lists all three), and the retired `Tools/web_search_tool.py` (lazy re-export only, no registration). A new parity test pins the sweep's claim mechanically: every advertised engine must parse a real-shaped payload into ≥1 result — which exposed that **tavily and searx parsers are `pass` stubs on dev** (advertised engines rendering real results as "No results found"); they're allowlisted with an honest pin and **task-2990** filed.

## Safety work that rode along

[...]

PR #1412, opened 2026-08-07
fix: real tavily/searx parsers — advertised engines stop lying about results (task-2990)

rmusser01:
## What this is

**task-2990** — the follow-up filed from #1404's parity test: `parse_tavily_results` and `parse_searx_results` were `pass` stubs, so two advertised engines rendered real search results as *"No results found"*. Both parsers are now real, ported from tldw_server's complete implementations (`tldw_Server_API/app/core/Web_Scraping/WebSearch_APIs.py`) with chatbook adaptations, and the parity allowlist is empty — **every engine the agent tool advertises now parses real-shaped payloads into real results.**

## The interesting part: searx's local shape is not the server's

Chatbook's `search_web_searx` returns a JSON **string** of a `[{title, link, snippet, publishedDate}]` list (or a `{"error": …}` string) — not the server's dict shape. That forced an unplanned-but-necessary production change the review interrogated hard: `process_web_search_results`'s input guard rejected non-dicts, meaning **every searx search — including successful ones — previously died with "search_results must be a dictionary"**. The guard now accepts strings *scoped to exactly the two engines that can produce them* (tavily error-strings, searx everything); a string reaching any other engine still raises loudly (probe-confirmed brave's parser would otherwise run substring checks on the text and could silently yield zero results — the exact defect class this task closes). The review also caught the first draft of the Implementation Notes misstating this justification; the record now states the true mechanism.

## Details

- **Tavily**: dict-with-results → standardized shape; `score` → `metadata.relevance_score` (a real 0–1 score, correctly directioned — unlike serper's rank, which deliberately stays out of that field); error-string input raises through the honest seam.
- **Searx**: JSON-string decode; error-dict raises with its message; decoded non-list payloads raise (`Unexpected Searx payload shape`) instead of silently yielding zero results; `link or url` / `snippet or content` fallbacks tolerate raw-SearXNG-shaped pre-parsed lists.
- Parity fixtures corrected to the engines' REAL shapes (the searx fixture is a string, so it also pins the scoped guard); RED evidence independently reproduced by the reviewer against the pre-change module (10 exact failures).
- Known residual, recorded in the task file: the searx *backend's* dead JSON-content-type branch would iterate dict keys if ever hit — latent, since the fetch never requests `format=json` and real instances take the HTML path.

## Tests

38 passed + 3 live-gated skips in the backends file; 93 passed across the caller suites; clean collection. task-2990 closed **Done** with corrected Implementation Notes.

---
## Summary by cubic
Implemented real parsers for Tavily and Searx so these engines return actual results and surface clear errors instead of “No results found”. Also scoped the type guard to safely accept string payloads only for these two engines. Meets task-2990.

- **Bug Fixes**
  - Tavily: parse dict results; map `score` → `metadata.relevance_score`; pass `published_date`; raise on error string; per-item non-dict raises with index.
  - Searx: decode JSON string; raise on `{"error": ...}`; reject non-list/scalar payloads; support `link|url` and `snippet|content` fallbacks; per-item non-dict raises with index.
  - `process_web_search_results`: accept `str` only for `tavily`/`searx` (others remain dict-only); updated signature and error message; keeps other engines from silently zeroing results.
  - Tests/parity: added guard-scoping and per-item validation tests; corrected fixtures to real shapes; `_KNOWN_BROKEN_PARSERS` is now empty.

<sup>Written for commit 1eb14a9a8df4c5e755fb57d08b8ea589a0b1a284. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1412?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Both Qodo findings fixed in `38a5b5acb`: the widened guard's type contract is now honest end-to-end (signature `Union[Dict, str]`, docstring, and the TypeError copy all state the tavily/searx string allowance), and both new parsers raise an index-tagged `ValueError` on non-dict list elements instead of an opaque AttributeError — with seam tests proving the message reaches `processing_error`. 42 passed + 3 live-gated skips.

PR #1419, opened 2026-08-07
refactor(console): decomposition wave 4 — wiring out of __init__, button routing, agent controller

rmusser01:
Wave 4 of the Console decomposition, per `Docs/superpowers/specs/2026-08-02-screen-decomposition-design.md`. Follows waves 1 (#1345), 2 (#1381) and 3 (#1408).

`chat_screen.py` **18,930 → 17,749 lines** (−1,181), `ChatScreen` **600 → 593 methods**.

## What lands

Wave 4 is different in kind from its predecessors: two of its three changes move code that is *already* owned elsewhere but still physically lives on the screen.

| | |
|---|---|
| `Console_Modules/wiring.py` — `build_console_controllers(screen)` | `__init__` **782 → 310** |
| `on_button_pressed` routed to its owning controllers | **381 → 233** |
| `Console_Modules/agent.py` — `ConsoleAgentController` (958 lines) | −535 |
| `Tests/Architecture/test_module_alias_reexports.py` | new guard (below) |

**Zero behaviour change** is the contract. The six relocated constructions are byte-identical after a `tokenize`-based `self`→`screen` rename (123 NAME tokens), and all six `Call` nodes are `ast.dump`-identical — reproduced independently by the reviewer. The one deliberate deletion is 35 lines of dead branch, in its own commit.

## The cost wave 3 flagged, now paid

Wave 3's spec note said `__init__` was "a real cost of this pattern, and the next wave's problem." It had reached **782 lines, 411 of them six controller constructions** — overhead that grew with every wave. Those now live in `wiring.py`, **verbatim**: every named keyword argument preserved character for character, only the call site relocated.

A reviewer on #1408 proposed collapsing this into per-controller dependency objects. That was declined then and is declined here: explicit signatures are what make a controller's dependencies readable, and a builder function relocates the call site without touching them.

## A dispatcher branch that outlived its button

Routing `on_button_pressed` surfaced a 35-line branch nothing could reach: the button carrying its id was deleted in `3b0374479`, and the id survives only as a CSS class on buttons whose ids take *other* branches. Dead twice over — the body also required a state that same commit retired.

Six of nineteen branches moved; **thirteen stayed, deliberately**. A branch coordinating two controllers stays on the screen (the wave-2 precedent), and five of the small ones pass the `Button.Pressed` event itself, so moving them would breach the no-events-in-a-controller invariant. An honest stays-list beats a flattering diff.

## The finding worth reading

**An import an extraction leaves behind can be load-bearing for tests.** Deleting the now-code-unused controller imports from `chat_screen.py` turns tests red, because tests reach them as `chat_screen_module.ConsoleDictationController` or `setattr(chat_screen_module, "…", …)`. Neither form is visible to an import-grep; the quoted one is not even an identifier.

I fixed this once with `# noqa: F401` markers — and the whole-branch review found **my fix wrong in both directions**: four classes marked that nothing reaches, six symbols reached that were unmarked, one in the same import block. An AST-accurate unused check was needed because these names appear in a dozen *comments*, which is what fooled the hand audit.

So the set is now **derived, not maintained**: `test_module_alias_reexports.py` recomputes it and fails if an at-risk import is unmarked. Mutation-verified — dropping one marker names the symbol and its 7 test sites — and it refuses to pass on an empty set, so it cannot rot into a no-op once task-3023 repoints the tests properly.

## Verification

- Serial, `-p no:randomly`: **117 passed, 1 failed** post-rebase across `Tests/Architecture/`, the three new test files, the geometry baseline, the worker-group guard and the responsiveness suite. The single failure is `test_persistent_diagnostic_inventory` — **red on clean dev**, filed as task-2768 and deliberately not rubber-stamped here.
- Earlier full sweeps: 257 + 537 passed across every file referencing a moved name; `pytest Tests/ --collect-only` → **32,016 collected, 0 errors**.
- The geometry baseline is byte-identical to dev and green.
- The ratchet is lowered to a live post-rebase `ast` measurement (17,749/593) and mutation-verified in both directions.
- Whole-branch review on the most capable model: one Important (the re-export set, fixed above), one Minor. No Critical.

Follow-ups filed: **3023** (repoint the alias-patching tests), **3024** (star `IndexError` on an empty title, pre-existing), **3025** (workspace rail row order is nondeterministic — possibly a product bug, not just a flaky test).

---
## Summary by cubic
Moves Console controller wiring out of `ChatScreen.__init__`, routes button handling to owning controllers, and extracts the agent cluster into `ConsoleAgentController` to shrink `chat_screen.py` with no behavior changes. Lowers the screen size ratchet and strengthens the guard protecting alias-based re-exports used by tests.

- **Refactors**
  - Moved six controller constructions verbatim to `UI/Console_Modules/wiring.py`; `__init__` 782→310 lines.
  - Routed six `on_button_pressed` branches to their controllers; 381→233 lines, and removed the dead `console-workspace-conversations-toggle` branch.
  - Extracted agent runtime into `UI/Console_Modules/agent.py` (`ConsoleAgentController`); screen now calls `self._agent.*` directly, dropping redundant delegations.
  - `chat_screen.py` shrank 18,930→17,749 lines; `ChatScreen` methods 600→593.

- **Tests and Guardrails**
  - Added `Tests/Architecture/test_module_alias_reexports.py` to derive and enforce `chat_screen` alias re-export markers; updated to ignore docstring mentions and scan only annotation subtrees, with a unit test; detector now matches `pyflakes` and mutation checks pass.
  - Lowered the screen size ratchet to 17,749 lines / 593 methods.
  - Added characterization tests for controller wiring, button routing, and the agent controller to assert no behavior change.

<sup>Written for commit 8432fbbcb7bfe24de5192a984ef7e24fa163a36f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1419?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## Qodo review — both findings addressed in `8432fbbcb`

### 2. "Alias test counts docstrings" — **correct, and a genuinely good catch**

This is a real false negative in a guard I added one commit earlier, and it is precisely the trap the guard exists to prevent. The function's own docstring claimed docstring mentions don't count as references; the implementation scanned every string `ast.Constant` for identifier tokens, and docstrings are string constants.

The consequence was not theoretical. I mutation-tested it: removing the `# noqa: F401` from **`ConsoleDictationController`** — 18 alias sites, the symbol whose deletion turned 28 tests red in the first place — left the guard **green**, because that name appears in a dozen docstrings in `chat_screen.py` and in zero expressions.

Fixed as suggested: string scanning is now restricted to annotation subtrees (`AnnAssign.annotation`, `arg.annotation`, `FunctionDef/AsyncFunctionDef.returns`). Evidence it works:

- The detector's unreferenced set is now **37**, matching pyflakes' F401 count exactly (it was excluding the docstring-mentioned ones before).
- The at-risk set is 8, matching the ground truth I computed independently via pyflakes.
- The mutation now **fails** with `ConsoleDictationController (line 71, 18 test site(s))`.
- `test_a_docstring_mention_does_not_count_as_a_reference` pins the case in both directions — a docstring-only name must be unreferenced, a quoted annotation and an expression use must not be — so the narrowing cannot be silently undone. That's the acceptance check the finding asked for.

### 1. "Repeated `run_test` size literals" — **addressed, but file-locally, not shared**

Named both sizes (`_ROUTING_SIZE`, `_AGENT_SECTION_SIZE`), which does what the rule is for: the intent is now explicit and one-point-updatable.

Deliberately **not** centralised into a shared "standard console test size", for two reasons worth recording:

1. **The neighbourhood contradicts the rule at scale.** 297 files and **3,932 call sites** in `Tests/` pass the tuple inline. Complying globally would make these two files the outlier, not the norm. (Same family as the finding on #1381, where a `_PascalCase` test helper was flagged against a convention 296 local classes already followed.)
2. **More importantly, a shared constant would be actively wrong here.** The Console shell is responsive — regions hide and show by width, and the geometry baseline pins *observed* truth at specific sizes. A repo-wide "standard console size" invites someone to change one value and silently alter what a dozen unrelated tests exercise. The sizes are not interchangeable; 160x44 and 180x48 were each chosen so a specific region is mounted and expanded. Both constants carry that reasoning in a comment.

### Verification

`Tests/Architecture/` → **33 passed, 1 failed**; the failure is `test_persistent_diagnostic_inventory`, **red on clean dev** and filed as task-2768 (regenerating it means signing off on new diagnostic owners in two modules this branch doesn't touch — exactly what its "review the diff first" gate exists to prevent). The two touched test files: **20 passed**.

PR #1423, opened 2026-08-07
refactor(console): decompose _open_console_prompts_modal into two collaborators (task-2766)

rmusser01:
Closes task-2766.

`ConsolePromptsController._open_console_prompts_modal` was the Console's
largest non-`__init__` method: **397 lines holding 17 nested closures**, each
capturing one dependency out of a scope no reader could hold in their head.
Wave 3 moved it verbatim; two reviews since judged the controller boundary
right and the *method* the problem. It is now **82 lines with zero closures**.

## The split follows what the closures captured

**`_ConsolePromptSource`** (5 closures) — `capabilities`/`list_page`/`search`/
`detail`/`save` captured exactly one thing, `app_instance.prompt_scope_
service`, and did one thing with it. The cluster's whole source contract now
lives in one place: pages of 10, searches capped at 25, `save`'s `source`
routed into the service's `mode`, and one `_require` helper carrying the
per-callable refusal copy a duck-typed source triggers.
`record_prompt_usage` joined it, since the improvement flow read the same
service.

**`_ConsolePromptImprovementFlow`** (11 closures) — the stateful half. The
`nonlocal pinned_improvement_resolution` is the tell: one closure wrote it,
three read it, nothing else in 397 lines touched it. It becomes a field, and
the eight values the closures each re-captured become named constructor
arguments. The session/System-prompt guard pair, duplicated verbatim across
three call sites, collapses into `_stale_reason()` — same order, same copy.

**Promoted out of closure form** — `restore_focus` (captured only `self`) is
now a controller method; `_resolution_identity` (pure) is module-level.
Nothing was left inline; the two shortest closures moved for opposite reasons,
both stated in the report.

## Constructor 18 → 16 named dependencies

The three post-apply re-sync bridges were only ever called as an ordered trio,
at the two moments the store accepted a new System prompt, never individually.
`sync_console_system_prompt_surfaces` replaces them. `wiring.py` still
resolves each screen method **by name at call time**, so the late binding
eleven of the sixteen dependencies rely on is untouched. No other reduction
was available without manufacturing an abstraction — the five provider
dependencies are five distinct services, and the report says so rather than
inventing a win.

## Zero behaviour change, checked structurally

An AST equivalence pass against the parent commit shows `capture_manual_
resolution`, `validate_improvement`, both saved-artifact guards and
`restore_focus` **byte-identical**, and every other difference to be one of
four documented transformations (`_stale_reason`, `_require`, the trio
collapse, `self.` prefixes) and nothing else. Call order, the two separate
`_console_provider_blocker_copy()` reads and the modal's kwarg evaluation
order are unchanged. The controller still owns zero DOM.

One deliberate non-identity is disclosed in the report: `record_prompt_usage`'s
availability lookup now sits inside the caller's `try`, which differs only if
attribute *access* on a scope service raises — unreachable for any plain
object, and strictly safer if it ever happened.

## Test evidence

Characterisation was written, mutation-checked and pushed **green before** any
production change (`d635a862a`): 7 new live-path tests over the source
contract, the refusal copy, the focus restore, the pinned-target lifecycle,
the activation guard, the unpinned-snapshot refusal and the captured-snapshot
preference. Two mutations (`per_page` 10→11; dropping the capture-once guard)
each failed exactly one of them.

- 44 passed — prompts controller + controller wiring
- **304 passed — `test_console_native_chat_flow.py`, identical to the baseline** (13 real modal-open drives covering the whole improvement sub-flow)
- 268 passed — prompt-cluster batch (workbench contract, composer menu, prompts modal, system prompt, command composer, composer history, picker, chip)
- 195 passed — responsiveness, agent swap, console internals
- 89 passed / 1 failed — Architecture + state ownership; the failure is the pre-existing `test_persistent_diagnostic_inventory` (task-2768), whose checker names neither changed file
- 9073 collected clean over `Tests/UI`; `ruff` clean

`chat_screen.py` is untouched, so its size ratchet is unaffected.

Full rationale, the per-closure verdict table, before/after numbers and the
concerns I am leaving on the record are in `task-2766-report.md`.

---
## Summary by cubic
Refactor for task-2766: decomposed `_open_console_prompts_modal` into `_ConsolePromptSource` and `_ConsolePromptImprovementFlow`, named the page size and search cap, and collapsed post-apply syncs into one; method shrinks 397→82 lines with no behavior changes.

- **Refactors**
  - Extracted `_ConsolePromptSource` (data adapter over `app_instance.prompt_scope_service`) and `_ConsolePromptImprovementFlow` (stateful improvement flow with a pinned resolution).
  - Named `CONSOLE_PROMPT_PAGE_SIZE` (10) and `CONSOLE_PROMPT_SEARCH_LIMIT` (25); class now aliases `_CONSOLE_PROMPT_SEARCH_LIMIT` to avoid duplication. Preserved source contract: per-page=10, search limit=25, and `save` routes `source`→`mode`.
  - Deduped guards with `_stale_reason()` and added module `_provider_resolution_identity`; promoted `ConsolePromptsController._restore_console_composer_focus`.
  - Added 7 characterization tests for the modal callbacks and a test pinning `_stale_reason` order (session before system); verified pinned-target lifecycle, refusal copy, focus restore, and saved-artifact guards.

- **Dependencies**
  - Replaced three screen sync dependencies with one: `sync_console_system_prompt_surfaces` in `wiring.py`, which calls `screen._sync_console_chat_core_state()`, `screen._sync_console_rail_system_line()`, and `screen._sync_console_settings_summary()`; late binding preserved.

<sup>Written for commit 635df617c8c8b1582507bb619692354ada8f65c2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1423?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/security/production-diagnostic-inventory.json
+++ b/Docs/security/production-diagnostic-inventory.json
@@ -2022,7 +2022,7 @@
     },
     {
       "call_count": 3,
-      "diagnostic_digest": "542edfd662c59a16f67f",
+      "diagnostic_digest": "2fee300efb749f9af99f",
       "owner": "TASK-494",
       "path": "tldw_chatbook/UI/Console_Modules/prompts.py",
       "reason": "remaining Chatbook production diagnostic owner"
--- a/Tests/UI/test_console_prompts_controller.py
+++ b/Tests/UI/test_console_prompts_controller.py
@@ -33,6 +33,7 @@ from textual.app import App
 
 from Tests.UI.test_destination_shells import _build_test_app, _wait_for_selector
 
+from tldw_chatbook.Chat.console_provider_gateway import ConsoleProviderResolution
 from tldw_chatbook.UI.Console_Modules.prompts import ConsolePromptsController
 from tldw_chatbook.UI.Screens.chat_screen import ChatScreen
 from tldw_chatbook.Widgets.Console import ConsoleComposerBar, ConsolePromptsModal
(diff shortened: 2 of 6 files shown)


rmusser01:
## Qodo — addressed in `2ab7afc0a`

**Hardcoded prompt limits 10/25** — valid, fixed. Both are now `CONSOLE_PROMPT_PAGE_SIZE` and `CONSOLE_PROMPT_SEARCH_LIMIT`.

Worth noting they were inline literals *before* this branch (`prompts.py:506,512` on `dev`) and moved verbatim by the refactor, so this isn't a regression the PR introduced. Naming them is still the right call, and for a reason the finding doesn't state: the two numbers mean different things. `10` bounds a browse **page** — the modal's paging controls and its no-more-results state are in terms of it. `25` bounds an unpaged **search** result list, i.e. what a user can be shown at once. Reading them side by side as bare literals suggested a single knob; the constants and their comments now say otherwise.

pyflakes clean, `Tests/UI/test_console_prompts_controller.py` 28 passed.

PR #1424, opened 2026-08-07
fix(console): close six decomposition follow-ups (2767, 2768, 2769, 3023, 3024, 3025)

rmusser01:
Closes the six Console follow-ups filed during decomposition waves 3 and 4. (task-2766 is separate — PR #1423.)

| Task | What it was | Outcome |
|---|---|---|
| **3024** | Star toast crashed on an untitled conversation | Fixed + mutation-verified test |
| **2769** | "Pre-existing" Console test failures | Fixed; one was **my own** wave-3 regression |
| **2768** | Diagnostic inventory stale on dev | Reviewed and regenerated |
| **3025** | Rail row order nondeterministic | Test now resolves by identity; 10/10 |
| **3023** | Tests patching controllers via the module alias | 32 sites repointed; imports deleted |
| **2767** | Transcript region's two access idioms | Ruled deliberate; rule in `DESIGN.md` |

## The two findings worth reading

**2769 was not pre-existing debt — it was mine.** The `NO_APP` guard I added to the stub factory in wave 3 refuses to *infer* a missing `app_instance`, because an inferred `None` snapshot is a silent-default hole. `_bare_console_screen_for_restore` passes `None` from three of its six call sites, so it tripped that guard. I then spent an entire subsequent wave treating the resulting red tests as baseline noise. The fixture now declares the absence, which is what the sentinel is for.

**Two "flaky" tests had different causes, and only one was really flaky.**
- *3025* was a genuine mis-assertion: it read `_composited_rows(container)[0]` — an **index** into whatever the compositor had painted at that instant — and mid-layout got a neighbouring pair's row. It now resolves the row from the label widget's own identity: **10/10 isolated runs at load average ~8**, where it was roughly 1-in-3. Mutation-verified that the assertion still binds to the real painted row, so determinism didn't cost meaning.
- *2769's* recovery test had three real races (a `query_one` that **raised** inside a retry loop, making the first iteration fatal; a gate on the label alone, so a press could land while the control was disabled; a fixed `pause` assuming an async handler had finished). Measured, `pilot.click()` passed 6/10 where `press()` passed 10/10 while `get_widget_at` resolved correctly every time — so reachability is asserted directly and the button pressed. A residue remains under CPU load (7/12 at load ~8 vs 10/10 idle; a 12s budget didn't help, so it's stuck rather than slow). That points at the recovery retry, not the test, and is recorded in the test's docstring rather than hidden behind a longer sleep.

## 2767 — decided, not deferred

The two idioms are **deliberate**. `ConsoleTranscriptRegion` defines exactly three public behaviours — `capture_reading_state`, `restore_reading_state`, `note_follow_intent` — all about the *viewport*, which the region owns because the region is what scrolls. The eight direct sites need the transcript widget's *own* API. Routing them through the region would add eight pass-through getters whose whole body is `return self._transcript_or_none()`, converting an owner into a façade with no invariant behind it — and `query_one` is transparent across compound-widget boundaries in Textual by design. The rule now in `DESIGN.md`: **ownership, not reachability.**

## 3023 — the alias re-exports are gone

All 8 symbols / 32 sites repointed onto their defining modules. Equivalence was *proved* first (no site rebinds a module attribute; all 8 verified `is`-identical through both paths) rather than assumed — a repointed patch that silently steers nothing would have left green tests testing less. pyflakes on `chat_screen.py` **37 → 25**, past its pre-wave-4 baseline of 31; the ratchet drops to **17,727**. The guard test that policed the marker set is deleted, as its own docstring directed once the set became empty.

## 2768 — reviewed, not rubber-stamped

Regenerated twice (dev moved mid-branch), each time reading the diff the checker asks you to read. Counts reconcile exactly; the only sink-topology change was a line number. The one third-party diagnostic worth checking — `config.py`'s new warnings — lives in `_get_int_timeout_value` and logs a key name and an invalid *timeout*, so no credential path reaches it.

## Verification

`Tests/Architecture/` + the touched Console suites: **337 passed** across two serial runs (`-p no:randomly`), plus the 3023 agent's own 139-passing confirmation. `pytest Tests/ --collect-only` → **32,019 collected, 0 errors**. The geometry baseline is byte-identical to dev.

---
## Summary by cubic
Closes six Console follow-ups. Fixes the untitled-star crash, stabilizes recovery and workspace tests, removes `chat_screen.py` alias re-exports with a small ratchet win, and refreshes the production diagnostic inventory.

- **Bug Fixes**
  - 3024: Guard star toast when the conversation title is empty; adds a regression test.
  - 2769: Fix three races in the Console-follow recovery test; declare `NO_APP` in the restore stub; switch to hit-testing + `press()`; name poll intervals and budgets; remaining load sensitivity is documented.
  - 3025: Resolve the workspace status row by label identity instead of index; passes 10/10 under load.

- **Refactors**
  - 3023: Repoint 32 tests to patch controllers on their defining modules in `tldw_chatbook.UI.Console_Modules`; delete 12 alias re-exports from `tldw_chatbook/UI/Screens/chat_screen.py` and remove the guard test; pyflakes 37 → 25; ratchet 17,749 → 17,727; update `wiring.py` docstring.
  - 2767: Document the transcript access rule in `DESIGN.md` (“ownership, not reachability”).
  - 2768: Regenerate `Docs/security/production-diagnostic-inventory.json` after review; counts reconcile; no new sinks.

<sup>Written for commit c73b936228704600fe6224d85751faeb2e64d28e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1424?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## Qodo — one fixed, one declined

**2. Magic polling numbers** — fixed in `75c59c1c6`. `_POLL_INTERVAL_SECONDS`, `_RECOVERY_POLL_LIMIT`, `_EFFECT_POLL_LIMIT`. Worth saying why the two budgets differ rather than collapsing to one: the recovery budget is deliberately generous because the loop exits the instant its condition holds, so it only bounds the failure case — and that recovery is the load-sensitive path documented in the test's docstring. The effect budget is short on purpose: by that point the press has landed, so a long wait would only slow a genuine failure down. The remaining `pause(0.02)` in the file belongs to a test this branch doesn't touch.

**1. `_Marks` class not PascalCase** — declined, with evidence.

`_` + CapWords is PEP 8's *internal-use* marker, not a violation of it, and this repo's test suite has **1,170** helper classes in exactly that form. Complying would make this one the outlier.

More specifically: the flagged `_Marks` is a copy of the one already at `Tests/UI/test_console_workspace_context_rail.py:1284`, twenty lines above mine, in the adjacent test for the same button. My new test deliberately mirrors it so the pair reads as a pair. Renaming only mine would leave two differently-named stubs doing the same job in the same file.

This is the same finding family as the one declined on #1381 (a `_PascalCase` helper flagged against a convention 296 local classes already followed) and on #1419 (a shared test-size constant proposed against 3,932 inline call sites): a repo-wide rule cited where the file's own neighbourhood contradicts it. Checking the local convention first is the standing practice here.

`Tests/UI/test_console_live_work_handoffs.py`: **65 passed**.

PR #1446, opened 2026-08-08
fix(security): key diagnostic-inventory digests on content, not line numbers (task-3750)

rmusser01:
## Why

`Docs/security/production-diagnostic-inventory.json` exists to force a human to review every new production diagnostic and disk sink. Its per-file `diagnostic_digest` hashed each logger call's **line number** alongside its text, so any refactor that shifted lines failed the gate with `call_count` unchanged and the sink topology byte-identical — one review-and-regenerate cycle per task across decomposition waves 4 and 5. A gate that fires on no-ops trains people to regenerate without reading, which destroys the review it exists to force.

Measured on this tree: of the 47 drifted owner entries, **28 (60%) were pure line movement**, and **all 5 drifted sink entries** had byte-identical `digest` and `kind` with only `line:` changed (app.py 6063 → 6656).

## What changed

The digest is now a **sorted list of (log method, statement source-text digest)** — sorted, so relocation and reordering are invisible; a *list* and never a set, so multiplicity counts and deleting one of two identical calls is still caught.

**Sinks trade `line:` for `scope:`** (`TldwCli._setup_buffered_logging` rather than `6063`). Decided on the evidence above, not taste: sinks live in `app.py` and `config.py`, two of the churniest files in the repo, so the same false-positive engine runs there. A scope name navigates better than a line number and does not go stale. The middle option — print `line:` but exclude it from the comparison — was rejected: the number would rot and send a reviewer to the wrong place, which is worse than a chore.

**Diagnostics deliberately do NOT carry scope**, though sinks do. Moving a logger call into an extracted method is this repo's most common refactor and must stay silent; sinks are 19 entries that essentially never change function, and there the scope is a reviewer's only navigation handle. Owner entries never had a line number to lose — they are only path/count/digest.

`schema_version` 1 → 2: every digest changes because the keying changed, not because code did, and the version is the only signal that a mismatch across the bump is a semantics change rather than a new diagnostic.

## Sensitivity is not weakened

Six arms, each run against the real checker on real production source:

| mutation | expected | result |
|---|---|---|
| logger call moved down 3 lines | pass | exit 0 |
| message reworded | **fail** | exit 1 |
| level `debug` → `error` | **fail** | exit 1 |
| sink moved down 2 lines | pass | exit 0 |
| sink retargeted to another path | **fail** | exit 1 |
| everything restored | pass | exit 0 |

Plus unit tests for add / remove / reword / re-level / relocate-across-functions / duplicate-multiplicity, and an end-to-end check that shifting every line in a real inventory file leaves its digest byte-identical.

## The migration commit was reviewed, not rubber-stamped

The inventory was last regenerated in `f990464ed` — which is `origin/dev`'s tip, **where the test already failed**. A branch that regenerates before merge is re-staled by whatever merges ahead of it. Rather than run `--write` and stare at a 1,000-line diff, each file's history was walked for the revision reproducing its checked-in digest: 28 no-ops, 19 real content changes, 3 new files. The diagnostics newly blessed in the high-risk TASK-492 paths were read — a Library factory warning, an MCP backend-unavailable exception, and 7 robots.txt debug lines in `web_tool_impls.py` that log `cache_key`, which is `scheme://host[:port]` only, no path or query, so no token can leak.

## Tests

`Tests/Architecture/test_persistent_diagnostic_inventory.py` — **8 passed** (was 1 failed / 2 passed on dev before this branch, from the pre-existing drift).

Two other failures in `Tests/Architecture/` (`test_profile_owned_path_inventory.py`, `test_screen_size_ratchet.py`) are **pre-existing on dev and not caused by this branch**: this branch touches exactly three files, and neither test reads any of them — they read `check_profile_owned_path_inventory.py` and `chat_screen.py`. `test_screen_size_ratchet.py` is the same disease as this task, and its own comments already warn that "a budget derived from a stale base fails the moment it merges".

---
## Summary by cubic
Switch diagnostic-inventory digests to content-based hashing and make sink entries line-independent to stop false-positive gates when lines move. Bumps schema to v2 and migrates the inventory; implements TASK-3750.

- **Bug Fixes**
  - Digest is now a sorted list of (log method, statement source digest); ignores position and preserves multiplicity.
  - Sink entries use scope (e.g., TldwCli._setup_buffered_logging) instead of line; moves are silent, retargets still fail.
  - Scope-name walk is iterative to avoid RecursionError on deeply nested code; output is unchanged.
  - Added tests for move/add/remove/reword/relevel, multiplicity, and an end-to-end digest stability check.

- **Migration**
  - Inventory regenerated under schema_version 2; v1 and v2 digests are not comparable.

<sup>Written for commit 186f6507df8f6407aff76f773c2fc6d36bffbdd7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1446?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/security/production-diagnostic-inventory.json
+++ b/Docs/security/production-diagnostic-inventory.json
@@ -20,3262 +20,3283 @@
   "owners": [
     {
       "call_count": 3,
-      "diagnostic_digest": "0d78b70c928606dd32fe",
+      "diagnostic_digest": "c4532646d0265b71ec47",
       "owner": "TASK-494",
       "path": "tldw_chatbook/Agents/agent_runtime.py",
       "reason": "remaining Chatbook production diagnostic owner"
     },
     {
       "call_count": 4,
-      "diagnostic_digest": "e5adb9327f53b73e91c4",
+      "diagnostic_digest": "81a172ef8dadc288b9bf",
       "owner": "TASK-494",
       "path": "tldw_chatbook/Agents/builtin_tool_gate.py",
       "reason": "remaining Chatbook production diagnostic owner"
     },
     {
+      "call_count": 1,
+      "diagnostic_digest": "00e3fd66f1b0a1863b64",
+      "owner": "TASK-494",
+      "path": "tldw_chatbook/Agents/library_rag_tool_provider.py",
+      "reason": "remaining Chatbook production diagnostic owner"
+    },
+    {
+      "call_count": 1,
+      "diagnostic_digest": "b1cbdcf3a0a107beb4dd",
+      "owner": "TASK-494",
+      "path": "tldw_chatbook/Agents/library_tool_provider.py",
+      "reason": "remaining Chatbook production diagnostic owner"
+    },
+    {
       "call_count": 8,
-      "diagnostic_digest": "1ad2b3c48099f41bb550",
+      "diagnostic_digest": "fc49fca551481bbfbda4",
       "owner": "TASK-494",
       "path": "tldw_chatbook/Agents/local_tool_provider.py",
       "reason": "remaining Chatbook production diagnostic owner"
     },
     {
       "call_count": 6,
-      "diagnostic_digest": "2b0933285c5088d0abb6",
+      "diagnostic_digest": "ab541538fe3668a31294",
       "owner": "TASK-492",
       "path": "tldw_chatbook/Agents/mcp_tool_provider.py",
       "reason": "high-risk Chat/provider/summarization/tool/MCP diagnostic owner"
     },
     {
       "call_count": 16,
-      "diagnostic_digest": "a250ad214780f0903951",
+      "diagnostic_digest": "10b283947607979eb8f4",
       "owner": "TASK-494",
       "path": "tldw_chatbook/Agents/run_log.py",
       "reason": "remaining Chatbook production diagnostic owner"
     },
     {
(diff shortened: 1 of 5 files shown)


rmusser01:
## Qodo — both addressed in `186f6507d`

**1. Recursive AST walk** — fixed with an explicit worklist. For repo-internal source the risk was marginal (`ast.parse` itself guards pathological nesting first), but the rewrite is five lines and removes the argument — and a gate that can crash for reasons unrelated to diagnostics trains people to bypass it, which is this PR's whole theme. The checker's output is byte-identical after the rewrite: it exits 0 against the checked-in inventory, so scope names and digests are unchanged.

**2. `diagnostic_digest` docstring** — Args/Returns added, including the one fact worth stating: `line` deliberately does not participate, which is the point of task-3750.

Also re-verified independently after the changes: 8/8 architecture tests, plus a three-way mutation probe on a real production file (move → stable; reword level → detected; delete one of 25 calls → detected).

PR #1447, opened 2026-08-08
feat(console): composer DraftChanged message — six more on_key branches move (task-3749)

rmusser01:
## What

Wave 5 (#1441) moved 7 of `ChatScreen.on_key`'s composer branches into `ConsoleComposerBar.handle_console_key`. Six more were pinned to the screen purely by their follow-up calls *after* the edit — `_sync_console_workbench_actions_from_draft`, plus `_dismiss_console_guidance` on the two that add text.

This inverts that dependency. `ConsoleComposerBar` posts a `DraftChanged` message; `ChatScreen._handle_console_composer_draft_edit` subscribes and does the screen-owned follow-ups. The six branches — `backspace`/`ctrl+h`, `delete`, `ctrl+w`, `shift+enter`/`ctrl+j`, `ctrl+u`, and the printable fallthrough — become composer-only and move, taking `_is_modified_chord` with them.

`on_key` stays on the screen and keeps only the branches that genuinely reach past the composer: the clipboard, send, transcript paging, and undo/redo's store persistence. `chat_screen.py`: net −32 lines (~85 removed, ~53 added for the subscriber).

## Three design calls, each made from measured baseline behaviour

**Granularity — one message with an `is_insertion` flag.** At baseline the guidance is dismissed by the two keys that ADD text and by none of the four that remove it. A naive "dismiss on every `DraftChanged`" would have silently changed backspace/delete/ctrl+w/ctrl+u. The flag names the *cause* (the edit added text), so the "insertions retire the first-run guidance" policy stays on the screen where it belongs.

**Where it's posted — the moved branches only.** `insert_text`/`delete_left`/`clear_draft` have other callers: dictation, the session draft restore, `/prompt`, paste. Each already runs its own, *different* follow-up (dictation re-persists to the chat store; the session restore deliberately does not dismiss guidance). Posting from the helpers would have fired the Workbench sync and the guidance dismissal on all of them.

**Ordering — one real, measured difference, scoped rather than papered over.** Textual delivers the message through the pump. No existing test and no human-speed input depends on the old synchronous timing. A key arriving in the **same driver read** as the slash that opens the command popup does.

Two-arm A/B (this code vs. the baseline callback restored synchronously in its place):

| batched input | baseline | this PR |
|---|---|---|
| `/` + Down | highlights the second entry | Down ignored |
| `/` + Enter | accepts the highlighted suggestion | Enter ignored |

In both arms the popup still opens, the draft is untouched, and nothing is sent — Enter's own path re-reads the Send action *after* stashing the draft, so it never depended on this sync's timing. The blast radius is one lost keystroke under programmatic input (`tmux send-keys`, macros, text expanders); human typing is orders of magnitude too slow to reach it. Kept as a **strict xfail** that will fail loudly the day it is fixed, and filed as **TASK-3790** rather than left implicit in a marker.

## Defect found en route

The new handler was first named `_on_console_composer_draft_changed` — which is *already* a `ChatScreen` method (`@on(Input.Changed, "#console-command-input")`, disarming the unknown-command Enter escape). The second definition silently replaced the first and killed that subscription. Two existing `test_console_command_composer.py` tests caught it. Renamed to `_handle_console_composer_draft_edit`; both docstrings now cross-reference each other and explain why the broader `Input.Changed` signal is deliberately *not* reused here.

## Verification

- **21 characterisation tests written and committed green BEFORE the change** (`Tests/UI/test_console_composer_draft_changed.py`) — real `pilot` key presses on the real Console screen, asserting observable end state, never "this method was called". Includes two counter-examples proving the sync re-derives Workbench readiness rather than mirroring `draft_text() != ""`.
- **471 passed + 1 xfailed** across the composer, popup, command-composer, dictation and decomposition suites.
- **411 passed** across the workbench / guidance / send-state suites. Two `test_console_native_chat_flow.py` toast failures were A/B'd against a pre-change worktree and are **pre-existing on dev**.
- pyflakes on `chat_screen.py` unchanged (26 → 26, identical message set); 0 on the composer and the new test. AST duplicate-method audit clean. No hand-built `ChatScreen` fixtures exist. Full-suite `--collect-only`: 33,536 tests, no import breakage.

Closes TASK-3749. Follow-up: TASK-3790.

---
## Summary by cubic
Moved six draft-editing key paths from `ChatScreen.on_key` into `ConsoleComposerBar.handle_console_key` behind a new `DraftChanged` message to simplify `on_key` without changing behavior. Also fixed the same-read “/” popup ordering so queued keys are handled against the opened command popup (TASK-3790).

- **Refactors**
  - Added `ConsoleComposerBar.DraftChanged` and a screen subscriber to resync Workbench and dismiss guidance on insertions only.
  - Moved Backspace/Ctrl+H, Delete, Ctrl+W, Shift+Enter/Ctrl+J, Ctrl+U, and printable typing into the composer; moved `_is_modified_chord` there too.
  - `on_key` now handles only actions that reach past the composer (clipboard, send, paging, undo/redo).
  - Posted the message only from the moved branches to avoid changing dictation/session-restore/paste behavior.

- **Bug Fixes**
  - Closed TASK-3790: ensure the command popup is current at the routing point when the draft text changed, and make `show_suggestions` idempotent so `/`+Down and `/`+Enter queued in the same read work; un-xfailed the probe test.
  - Renamed the new handler to avoid shadowing an existing `Input.Changed` subscriber.

<sup>Written for commit 15b5f3d1a5194bd43e7b74d674e2880edd1f55ff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1447?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/security/production-diagnostic-inventory.json
+++ b/Docs/security/production-diagnostic-inventory.json
@@ -2805,8 +2805,8 @@
       "reason": "remaining Chatbook production diagnostic owner"
     },
     {
-      "call_count": 100,
-      "diagnostic_digest": "d2232c700a34c05300ad",
+      "call_count": 102,
+      "diagnostic_digest": "0d3c1f71cc686fbd50c9",
       "owner": "TASK-494",
       "path": "tldw_chatbook/Web_Scraping/WebSearch_APIs.py",
       "reason": "remaining Chatbook production diagnostic owner"
@@ -3471,6 +3471,6 @@
     "owner_files": 469,
     "persistent_sink_files": 6,
     "task_492_calls": 1153,
-    "task_494_calls": 6884
+    "task_494_calls": 6886
   }
 }
--- a/Tests/UI/test_console_composer_draft_changed.py
+++ b/Tests/UI/test_console_composer_draft_changed.py
@@ -1,0 +1,254 @@
+"""Characterisation of what a Console composer draft edit must still trigger.
+
+TASK-3749 moves six more `ChatScreen.on_key` branches into
+`ConsoleComposerBar.handle_console_key` by replacing the screen's
+"edit the draft, then call back into me" pattern with a
+`ConsoleComposerBar.DraftChanged` message the screen subscribes to.
+
+These tests are written and committed BEFORE that change, against the
+current callback shape, and pin the three things it must not disturb:
+
+1. **The Workbench actions row.** `#console-send-message` is a Workbench
+   action whose `disabled` state is derived from the draft by
+   `_sync_console_workbench_actions_from_draft`; every one of the six keys
+   that edits the draft has to leave it correct.
+2. **The slash-command popup**, which is opened by that same sync -- and
+   opened *in the same key turn*, so that a following arrow key already
+   finds it open. That timing is asserted adversarially (two keys queued
+   with no event-loop drain between them), because a message-based
+   notification is delivered asynchronously and would be free to arrive
+   after the next key otherwise.
+3. **The first-run guidance**, which is dismissed by the two keys that
+   ADD text (a printable character, Shift+Enter's newline) and NOT by the
+   four that remove it -- a distinction a naive "dismiss on every draft
+   change" would erase.
+
+They assert observable end state through a REAL key press on the real
+Console screen, never that a particular method was called.
+"""
+
+from __future__ import annotations
+
+import pytest
+from textual import events
+from textual.widgets import Button
+
+from Tests.UI.test_console_dictation import _mounted_console, _ready_host
+from tldw_chatbook.Widgets.Console import ConsoleComposerBar
+
+APP_SIZE = (140, 42)
+
+
+async def _console(host, pilot, text: str = ""):
+    """Mount the ready Console, optionally seed the draft, focus the composer.
+
+    `load_draft` is deliberately used to seed text: it is not one of the
+    six branches under test, and (unlike typing the text) it does not
+    itself dismiss the guidance, which the guidance tests below depend on.
+    """
+    console = await _mounted_console(host, pilot)
+    composer = console.query_one("#console-native-composer", ConsoleComposerBar)
+    if text:
+        composer.load_draft(text)
+    composer.focus()
+    await pilot.pause()
+    return console, composer
+
+
+def _send_action(console) -> Button:
+    """The Workbench actions row's Send action."""
+    return console.query_one("#console-send-message", Button)
+
+
+# ---------------------------------------------------------------------------
+# 1. The Workbench actions row tracks the draft across all six edit keys
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    ("seed", "keys", "expected_draft"),
+    [
+        ("x", ("backspace",), ""),
+        ("x", ("ctrl+h",), ""),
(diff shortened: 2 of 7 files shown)


rmusser01:
## The AC-#3 asterisk is now closed — fixed on this branch in `31b267c0a`

The implementer's xfail documented a real gap: a key arriving in the same driver read as the `/` consulted the popup before its `DraftChanged` had delivered, and was lost. Reachable by `tmux send-keys`, which is how this repo does live verification — so it was worth fixing before merge rather than tracking.

**The fix, two halves, both mutation-verified:**
1. `_ensure_console_command_popup_current` re-derives the popup at `on_key`'s routing point, **gated on the synced draft text**. Escape edits nothing, so a dismissed popup stays dismissed — removing the gate fails the new Escape regression test.
2. `show_suggestions` is idempotent for an identical suggestion list, so the deferred `DraftChanged` re-sync cannot yank the highlight a routed Down just moved — removing that fails the new highlight regression test.

**Two designs were tried and rejected on evidence before this one:**
- The naive pre-routing re-sync: re-opens an Escape-dismissed popup (the sync derives purely from draft content; there is no dismissed-flag).
- A gate on the composer's `_draft_generation`: that counter is an **undo-checkpoint marker that `insert_text` never advances** — an entry-point probe showed the gate no-oping on every keystroke. The draft text itself is the honest staleness signal, because the sync is a pure function of it.

This is not the composer→screen callback the task removes: the composer stays oblivious; the screen consults its own sync before a routing decision that depends on it.

The strict xfail is flipped to a passing test, task-3790 is closed as done-by-this-PR, and AC #3 now holds without an asterisk. **23 passed** in the DraftChanged file, **103** across the composer suites, pyflakes unchanged.

rmusser01:
## Qodo — all three addressed

**3. "Batched slash keystroke lost"** — correct at review time, **already fixed on this branch** in `31b267c0a` (Qodo reviewed the pre-fix push). The strict xfail is now a passing test, plus two new regressions: Escape-dismissal survives a following arrow (the gate that makes the fix safe), and the deferred `DraftChanged` cannot yank a moved highlight (the idempotency that makes it complete). Both mutation-verified — removing either half fails its test by name. task-3790 is closed as done-by-this-PR, so AC #3 holds without an asterisk.

**1. `DraftChanged.control` docstring** — fixed in `15b5f3d1a`, including the fact that makes it worth documenting: Textual's `@on` selector filtering resolves against `control`.

**2. Test docstrings** — the six undocumented characterisation tests now each state their claim in one line.

Post-rebase verification: 23/23 in the DraftChanged file, 103 across the composer suites. The two red Architecture tests are dev's, not this branch's: the size-ratchet breach is task-3070 (dev is 892 lines over a budget nobody raised), and the diagnostic inventory was re-staled by 27 dev commits — under the new content keying that drift is exactly one line (two `logger.debug` in WebSearch_APIs.py, reviewed and re-blessed here), where the old keying produced pages.

PR #1469, opened 2026-08-09
RAG hybrid-fusion defect cluster: hybrid actually fuses (TASK-3994/3995/3996/3998)

rmusser01:
Fixes the defect cluster the P1 eval harness found on its first run (spec: `Docs/superpowers/specs/2026-08-09-rag-port-hybrid-fusion-fixes-design.md`, plan alongside it). Every change is measured by the P1 gate; one deliberate re-stamp closes the arc.

## What was broken, and what now works

**TASK-3994 — hybrid never fused.** RRF keyed on `SearchResult.id`, but the FTS leg emitted `media_15` while the vector leg emitted `media_15_chunk_0`. The legs never matched on a single row, ever. Fusion now keys on document identity `(source_type, source_id)` with a pinned fallback to the row id, and merged rows display the **matched chunk** (the vector item) rather than the doc-level keyword row — a preference that had been dead code since the legs never met. Result: **22 of 44 golden queries now return different results from semantic; before this, 0 of 44 did.**

**TASK-3995 — every multi-token keyword query was a phrase query.** `_escape_fts5_query` wrapped the whole query in one pair of quotes, so tokens had to be contiguous in the source text. Now each token is quoted individually (preserving the injection safety that made `Obsidian-3` safe — raw, it raises `no such column: 3`) and joined by FTS5's implicit AND. Empty-after-tokenization queries short-circuit instead of emitting a syntax error.

**TASK-3996 — the engine's FTS leg only searched media.** Notes and conversations were structurally unreachable (29 of the 49 fixture docs invisible to it). They now have sub-legs over a **read-only** connection through the private-sqlite seam — never the ORM, no schema touch, all three soft-delete filters replicated from the ORM's own queries. A missing DB degrades that sub-leg with a warning; media continues. The Library also stopped routing hybrid→semantic whenever Media was deselected — a gate justified by a comment that this PR made false.

**TASK-3998 — the eval gate's fingerprint watched the wrong packages.** It recorded `sentence-transformers` (not on the load path) and omitted transformers/torch/chromadb (which produce the numbers). Fixed first, with its own same-commit re-stamp, so every later verdict in this arc means something.

## The honest remainder

Hybrid **merges** now; it still cannot **rescue**. A keyword-only document scores `(1−α)/(rrf_k+1) = 0.00492`, below any vector row ranked better than ~82, while the vector leg is asked for only 2k. That is the other half of TASK-3994's own description, and the id-space fix never reached it. We proved it deliberately: a vector-blind fixture document was added (plain finds it at rank 1, semantic misses it entirely) — and hybrid misses it too. Filed as **TASK-4110** with that query as its before-number, and TASK-3994's AC #2 is split `#2a` ticked / `#2b` unticked rather than papered over. `Tests/RAG_Eval/README.md` and the User Guide say the same thing.

Authoring note worth keeping: the first vector-blind fixture attempt *failed* because a corpus-unique rare identifier is a beacon for the embedder too (semantic rank 1). What worked was corpus-unique but semantically *misleading* vocabulary, diluted across 513 words.

## Measurement

Per-fix progression (informational runs, then one deliberate re-stamp — full table in the README):

| stage | hybrid overall P | hybrid keyword P |
|---|---|---|
| P1 baseline (hybrid ≡ semantic) | 0.117 | 0.135 |
| post-3995 | +0.000 | +0.000 |
| post-3994 (fusion live) | 0.105 | 0.113 |
| post-3996 | +0.000 | +0.000 |
| stamped (with vector-blind doc) | 0.103 | 0.106 |

Precision falls **mechanically**, not qualitatively: chunk-collapse frees top-k slots for non-relevant fillers while `P@k` divides by `min(k, len(retrieved))`. Recall/MRR/NDCG hold at 1.000 — every relevant document still ranks ≤2, and the added rows contain zero relevant documents.

## Verification
- 274 passed / 16 skipped across `Tests/RAG_Search/` + `Tests/RAG_Eval/` + the private-sqlite inventory; gated harness 142 passed, gate **PASSED** (60 metrics within band).
- **Live TUI check:** Hybrid Basic with **Media deselected** returned a note row *and* a conversation row, both banded `keyword match`, for a three-token non-contiguous query — one observation exercising 3995 + 3996 + the Library gate at once (a band only reachable through the fusion metadata block, so the capture itself proves the hybrid path ran).
- Data safety probed directly by the final reviewer: writes/DDL/FTS-rebuild all refused, WAL reads work with and without a live writer, and each of the three soft-delete filters individually mutation-proved (drop one → exactly one named test reds). No deleted-row leak, no write path.
- Final whole-branch review: no Critical; five must-fix items closed in one wave (including a *vacuous guard-on-the-guard* — the empty-query short-circuit test passed even with the production guard neutered, because its fixture configured no media DB).

## Filed along the way
TASK-4110 (alpha-blend starvation — the rescue half), TASK-4111 (Library Open silently no-ops for non-integer source ids), TASK-14751 (keyword-leg budget split three ways with no source-type pushdown — media 20→7 rows in the reviewer's probe), TASK-14752 (coverage copy can mis-describe keyword-sourced evidence). One lesson added: adding a resource of a *guarded kind* (a DB connection, a CSS class, a tool gate) requires running that kind's inventory suite — a raw `sqlite3.connect` here was already red against the repo-wide private-sqlite guard and survived a full task review because targeted runs never included `Tests/DB`.

Two cosmetic residuals knowingly parked: `baseline_io.py:23` / `test_baseline_io.py:444` cite plain keyword P=0.867 (now 0.875) as an *illustration* of the band design, and `ingest.py:91` uses a hypothetical "40 of 48". Neither is load-bearing.

Note: CI cancellation on this repo is intentional — the local gates above are the merge signal.

---
## Summary by cubic
Makes hybrid retrieval actually fuse vector and keyword results, fixes multi‑token keyword matching, and extends the keyword leg to Notes and Conversations with safe, read‑only DB access. Re-stamps eval baselines using a corrected fingerprint and a new vector‑blind probe; Library routing now uses hybrid when any keyword‑indexed source is selected.

- **Bug Fixes**
  - TASK-3994: Fusion now keys on document identity `(source_type, source_id)` with a fallback to row id; merged rows show the matched chunk. Results now diverge from semantic where FTS contributes.
  - TASK-3995: Per-token FTS5 quoting enables non‑contiguous multi‑token matches; still injection‑safe; empty‑after‑tokenization queries short‑circuit.
  - TASK-3996: Keyword leg searches Media, Notes, and Conversations via a read‑only `ChaChaNotes` connection through the private‑sqlite seam (owner `rag.chachanotes_keyword_leg`); fixed cursor lifetimes, token‑aligned keyword citations, and ordered conversation aggregation. Library routes hybrid when any keyword‑indexed source is selected; Prompts‑only falls back to semantic. Route note: "no keyword leg for the selected sources — semantic only."
  - TASK-3998: Eval fingerprint now compares `transformers`, `torch`, `chromadb`, model id, platform, and corpus hash; baselines re‑stamped. Metrics reflect a new vector‑blind fixture: hybrid/semantic overall recall/MRR/NDCG 0.974.
  - Known limitation: Hybrid still won’t rescue FTS‑only docs when the vector leg returns enough distinct documents; tracked in TASK‑4110.

- **Migration**
  - No schema or data migrations. `ChaChaNotes` is opened read‑only via the private‑sqlite seam. Optional config: `SearchConfig.chachanotes_db_path` (auto‑resolves by default).

<sup>Written for commit 9ab847802919442295211998e774ec187603a614. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1469?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
All three Qodo findings remediated.

- **Unclosed cursors (`_chacha_notes_fts` / `_chacha_conversations_fts`)** — both sub-queries now run through `contextlib.closing(conn.execute(...))`, matching the `with closing(conn)` idiom the same module already uses for the connection itself. No behavior change; `Tests/RAG_Search/test_keyword_leg_chacha.py` (19 tests, incl. the read-only/WAL/soft-delete pins) stays green.

- **Keyword citations missed non-contiguous matches (our TASK-3995 regression)** — tokenization is now a single shared helper, `RAGService._fts5_query_tokens`. `_escape_fts5_query` builds the MATCH expression from that list, and the new `_keyword_citation_spans` locates spans from *the same list*, per token, case-insensitively (a token is matched as its alphanumeric runs joined by non-alphanumerics, which is how FTS5 reads a quoted token). Overlapping/adjacent spans merge, so a contiguous query still yields the one whole-phrase citation it used to; the `MAX_CITATION_MATCHES` cap is unchanged but now prefers spans that evidence a not-yet-cited token, and a row with no locatable span (e.g. matched on its title, which `media_fts` also indexes) keeps a document-level fallback so a keyword-backed row is never citation-empty. RED-first: `test_non_contiguous_multi_token_match_still_yields_citations` (offsets asserted by slicing the returned content), `test_span_cap_still_evidences_a_rare_second_token`, `test_keyword_row_matching_only_on_title_is_never_citation_empty` — all three failed before the fix; `test_contiguous_match_still_cites_the_whole_phrase` pins the old behavior (still `EXACT`, confidence 1.0) and passed before and after. Mutation-checked: neutering the coverage preference reds the cap test, neutering the adjacent-merge reds the contiguous test.

- **Unordered `group_concat` in the conversations sub-query** — the aggregate is gone. The conversation query now selects only `id`/`title`/`MIN(rank)`, and a second statement selects the matching `sender: content` lines `ORDER BY m.timestamp ASC, m.rowid ASC` — the ORM's own order (`get_messages_for_conversation`), which is also the order `ingestion_indexing.conversation_document` indexes on the vector side — joined in Python, so the order is a property of the query rather than of the planner. RED-first: `test_conversation_messages_are_aggregated_chronologically` seeds a conversation whose later message is inserted first and failed before the fix.

Tests: `Tests/RAG_Search/ Tests/RAG_Eval/` **259 passed, 16 skipped**; gated `RAG_EVAL=1 pytest Tests/RAG_Eval/` **142 passed — baseline gate PASSED with every one of the 60 gated metrics at +0.000** (citations do not feed retrieval metrics, as expected). Pre-existing and unrelated: 4 failures in `Tests/RAG/test_scope_pipeline_enforcement.py` (empty-scope notification wording); that file and its production path are untouched by this branch.

— remediated in a37b146f7

PR #1487, opened 2026-08-11
RAG fusion weighting: hybrid rescues keyword-only documents, measured (TASK-4110, TASK-14751)

rmusser01:
Closes the fusion cluster's honest remainder (spec: `Docs/superpowers/specs/2026-08-10-rag-fusion-weighting-design.md`, plan alongside). The P1 gate measured every step; the shipped weighting was chosen by a decision rule written down **before** the measurement ran.

## What was wrong

After #1469, hybrid *merged* its legs but could not *rescue*: an FTS-only document scored `(1-α)/(rrf_k+1) = 0.3/61 = 0.00492`, below any vector row ranked better than ~82, while the legs were only fetched at 2k. With `rrf_k=60` over a ~20-row candidate window the RRF curve is nearly flat — leg membership decided ordering, not relevance rank. The server's k=60 was calibrated for candidate pools of thousands; chatbook's pools are ~20.

## What shipped

- **TASK-14751 — source-type pushdown.** The hybrid keyword leg now runs only the sub-legs the user selected (media/notes/conversations), giving a single-type selection the full budget instead of splitting it three ways and discarding two-thirds after fusion. Along the way: `EnhancedRAGServiceV2.search()` didn't inherit the new kwarg — **every Library hybrid search would have crashed**; caught by the informational gated run after ~2,500 unit tests stayed green (every double mirrors the base class). Selection also entered the search cache key.
- **Config knobs + metadata honesty.** `config.search.rrf_k` (consuming the previously-unwired `resolve_rrf_k`) and `config.search.hybrid_pool_multiplier`; the fusion metadata records the *actual* values used (P1's citation capture re-derives fused scores from them — a wrong recorded k silently degrades score-kinds); fusion parameters entered the hybrid cache key, without which **the sweep would have served pass 1's results to every pass** and read "+0.000, nothing matters".
- **The measurement.** A gated strategy sweep (control; k 5/10/20; pool 3/5; combinations) over the golden set, applied mechanically: **winner `rrf_k = 5`** — keyword recall 0.938→1.000, zero regressed recall/MRR/NDCG cells, and the vector-blind fixture rescued at rank 8 **by the weighting alone** (FTS-only, `vector_rank=None`). The flashier `k5+pool3` was declined on re-count evidence: its entire +0.005 was the already-rescued fixture moving rank 8→5 (recall didn't move — the tell). The matrix reproduced byte-identically across cold processes and independent reviewers.
- **The ship.** 60→5 at one authoritative site; every resolver fallback unified on the shipped default (a config typo must not silently revert the fix); a three-pin rescue suite whose mechanism survived a genuinely interesting correction — the "rank-9 tie" is not a float tie (one ULP separates the sides; the sort convention is pinned by a bit-identical pair instead); ADR-005 addendum (the *design* is server parity; one *constant* is now measured rather than copied); the legacy Chat-RAG blend follows the same profile k, disclosed with a note on TASK-3501.
- **The close.** Deliberate baseline re-stamp: hybrid up on 10/20 gated cells, **semantic and plain +0.000 on all 40 of theirs** (the change provably touched only fusion); the fixture cell flips miss→hit. Live A/B on a purpose-built 60-real-doc library (the actual user library was disqualified first — its vector index holds 4 distinct documents, the thin-leg regime TASK-4110's own rule rejects as evidence): the literal-match document takes visible slot 5 banded `keyword match` at k=5 and is **absent at k=60 with ranks 1–4 byte-identical**; an ordinary semantic-strength query returns byte-identical results under both constants.

## Bounds, stated

The corpus discriminates on essentially one query (everything else was already rank 1), so "nothing regressed" is a 49-doc statement — the live A/B on real data is the scale evidence, and the README records that the eval corpus is now **at ceiling**: future retunes need a new discriminating fixture. The Library surface runs top_k=5 (fused window 10), tighter than the harness's k=10 numbers imply — also recorded.

## Verification
- 425 passed / 18 skipped ungated across the branch's suites at the final head; gated harness 186 passed with the gate **PASSED** against the new baselines.
- ~15 mutation checks across the arc, each redding exactly its own pins; three-round fix loop on Task 5 alone (float-ULP mechanism, a zero-population migration hand-off withdrawn before Task 6 did dead work, resolver-fallback unification).
- Final whole-branch review: doc-only must-fixes (two stale "→ 60" fallback arrows at the legacy-path disclosure sites — the 7th and 8th instances of the arc's stale-prose thread, now a promoted lesson: *when a constant is retuned, grep the literal values it manifests as, not just the symbol*).

Note: CI cancellation on this repo is intentional — the local gates above are the merge signal.

---
## Summary by cubic
Switch hybrid RAG fusion to `rrf_k = 5` (from 60) and push source-type selection into the keyword leg so hybrid can rescue keyword-only documents, measured with no regressions. Also hardens fusion config resolution (step → profile → shipped default 5) and tolerates overflow-range values; addresses TASK-4110 and TASK-14751; keyword recall improves and the vector‑blind fixture is rescued at rank 8.

- **New Features**
  - Keyword leg pushdown: `keyword_source_types` threads Library → service, giving selected types the full budget; added to cache key; `EnhancedRAGServiceV2.search` accepts/forwards it.
  - Config + consistency: `config.search.rrf_k` and `config.search.hybrid_pool_multiplier`; `resolve_rrf_k` precedence unified to step → profile → shipped default `DEFAULT_HYBRID_RRF_K = 5`; fusion metadata records actual `(alpha, k)`; hybrid cache key includes `(alpha, k, pool)`; the legacy pipeline path resolves the same profile `k`.
  - Measured outcome (hybrid mode): ship `rrf_k = 5`; keyword recall@10 0.938 → 1.000; rescues the FTS‑only fixture at rank 8; no regressed recall/MRR/NDCG cells; `semantic` and `plain` unchanged; baselines re‑stamped; docs updated (fused ceiling ~0.17 is not a similarity); ADR addendum documents the measured constant.

- **Bug Fixes**
  - Config resolvers now tolerate overflow‑range values and still fall back to the shipped default `5` instead of reverting weighting.

<sup>Written for commit 931412c3d35137fb60ee575e34ddfe8b950f2d9f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1487?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Remediated both Qodo findings — resolvers documented as "never raises" were missing `OverflowError`, reachable from live user config on the hybrid path.

- **`fusion.resolve_rrf_k`**: `int(float(value))` raised `OverflowError` when `value` overflowed to an infinite float (`"1e309"`, or a literal `inf`/`-inf` — both valid TOML). Added `OverflowError` to the except tuple alongside `TypeError`/`ValueError`, falling back to the shipped default (5) with the same warning path as every other invalid value. RED-first tests: `resolve_rrf_k("1e309")`, `resolve_rrf_k(float("inf"))`, `resolve_rrf_k(float("-inf"))` all return `5` with a warning naming the value — `Tests/RAG/test_fusion.py::TestResolveRrfK::test_overflow_range_values_fall_back_to_the_shipped_default` (+3 params) and `::test_overflow_range_value_is_named_in_the_warning`.
- **`rag_service._resolve_hybrid_pool_multiplier`**: same defect, `int(value)` raised `OverflowError` for `float("inf")`/`float("-inf")` before either hybrid leg launches (`hybrid_pool_multiplier = inf` is valid TOML too). Same fix: `OverflowError` added to the except tuple, falls back to `DEFAULT_HYBRID_POOL_MULTIPLIER` (2) with a warning. RED-first tests added to `Tests/RAG_Search/test_fusion_config_knobs.py::test_pool_multiplier_resolver_floors_caps_and_falls_back` (+2 params) and `::test_overflow_range_pool_multiplier_falls_back_with_warning`.

Swept the rest of the branch's diff for the same `int(value)`/`int(float(` pattern in try/except — `resolve_hybrid_alpha` stays a float (no `int()` call, so `float("inf")` is caught by its own `0.0 <= alpha <= 1.0` range check, not affected). No other sibling resolver touched by this branch has the hole; did not expand the fix to untouched code (e.g. `library_rag_state.py`'s coercers, which this branch never modifies).

Test evidence: `Tests/RAG/test_fusion.py` + `Tests/RAG_Search/test_fusion_config_knobs.py` + `Tests/RAG_Search/test_fusion_rescue_pin.py` → 88 passed, 0 failed (7 new tests, all green). Broader `Tests/RAG_Search/` + `Tests/RAG/` + affected `Tests/Library/` sweep: 1053 passed, 5 pre-existing failures unrelated to this change (message-text assertions in `test_scope_pipeline_enforcement.py` / `test_lazy_embeddings_rag_dependency_check.py`, no relation to `resolve_rrf_k` or `_resolve_hybrid_pool_multiplier`).

— remediated in 931412c3d

PR #1574, opened 2026-08-12
RAG keyword-leg MATCH construction, measured (TASK-15400)

rmusser01:
## What this is

The P2c opener. The engine keyword leg's FTS5 MATCH construction returned **zero rows for 40 of 60 golden queries** — masked everywhere by the semantic leg except on prompts, which have none. This arc measured four candidate constructions over the P2ab-renewed instrument under a rule **pre-registered in the spec before any number existed**, and shipped the winner the rule computed.

## The result

**Winner: `and_stopword_trim`** (AND over content tokens). Census 20 → **21/53**, prompt category recall **0.000 → 0.200**, **exactly 10 of 105 gated metrics moved — all up, all hybrid**, 95 at +0.000, zero regressions, zero extra FTS queries.

**Both OR-bearing candidates were DISQUALIFIED on measured evidence**, including `and_then_or` — the candidate the spec itself argued was "safe by construction". It isn't, and the refutation is the arc's real finding:

> A sub-leg's *own* row survives untouched, but `interleave_rankings` merges the four sub-legs **round-robin**, so another sub-leg's extra fallback rows displace it — and fusion consumes **leg rank**. The vector-blind fixture drops from leg rank 1 → 2 (0.3/6 = 0.0500 → 0.3/7 = 0.0429, below the vector rank-11 row at 0.04375) and its rescue vanishes. Same mechanism drove scoped recall 1.000 → 0.429, decomposed exactly: the 4 note-targeted queries drop behind a media fallback row, the 3 media-targeted keep rank 1.

**Constraint (a) is a property of the MERGE, not the match construction** — filed as **TASK-15700**, with a counterfactual showing that even a perfect merge fix restores the rescue at **slot 10 of 10, zero headroom** (so "fix `interleave_rankings`" is necessary, not sufficient).

## Honest framing

This is closer in substance to a null result plus a finding than to a fix: zero-row queries go **40/60 → 39/60**. The report and the backlog record both say so.

## What changed for users

Only the RAG engine's keyword leg (hybrid + keyword search); Library's plain four-seam builder is untouched. Measured on three query shapes: **only-function-word queries are byte-identical** (trimming everything falls back to the full AND), **only-content-word queries are byte-identical**, and the **mixed case is what widened** — "notes about the vendor" no longer requires "about" and "the" to appear. Typing `OR`/`NOT` as ordinary words now searches the surrounding terms instead of demanding those literals. The construction joins the cache key, so pre-upgrade entries are keyed apart rather than served (one cold miss per query).

## Evidence discipline

- **The rule decided, not taste**: hard constraints (vector-blind rescue, no cell regressing > 0.02, injection quoting green) applied line by line; the expected winner lost.
- **One deliberate re-stamp**, reconciled cell-by-cell against the 10 predicted movers — no eleventh. The stamp deliberately ran in the environment matching the **committed fingerprint** (not the worktree venv), so the environment block didn't move and the next arc's gate stays live; the two-venv split is disclosed in the README, with all 105 cells reproducing at +0.000 in both.
- **The gate is mutation-verified live**: reverting the default reds it (prompt recall 0.200 → 0.000).
- Injection quoting extended to every construction (escaping suite 10 → 37 tests), probed with novel attacks through the shipped default.
- 5 tasks × (fresh implementer + reviewer + scoped re-review), final whole-branch review, prose-only fix wave applied.
- Negative vacuity recorded honestly: all 7 negatives return zero keyword rows under *every* construction — never quoted as "OR widening is quiet".

## Tasks

TASK-15400 (Done) · filed: **15700** (the merge finding), **15701** (sync cache twins render a key ignoring three search-defining dimensions).

Live-check limitation stated in the User Guide stamp: RAG Answer never rendered an Evidence row in the scratch profile (first hybrid query ran 4+ min at 98% CPU — observed, not filed), so the retrieval claim rests on a two-arm A/B through the engine's own builder against the app-written index.

---
## Summary by cubic
Switches the RAG engine’s keyword leg from implicit AND-of-every-token MATCH to stopword-trimmed AND via a measured sweep (TASK-15400). Cuts zero-result natural‑language queries and raises hybrid prompt recall from 0.000 to 0.200 with no extra FTS queries; invalid construction types now fall back safely instead of crashing.

- Adds `SearchConfig.fts_match_construction` with four options; defaults to `and_stopword_trim`. Only the engine keyword leg changes; the Library four‑seam path is unchanged.
- Mixed queries widen (e.g., “notes about the vendor” no longer requires “about/the”); all‑content and all‑function‑word queries remain byte‑identical. Operator‑shaped words like “OR/NOT” are treated as literals and are no longer required.
- Injection safety preserved (per‑token quoting); escaping tests expand from 10 to 37.
- Cache keys include the construction; pre‑upgrade entries won’t be served, causing a first‑query cold miss per key. Follow‑ups filed for merge displacement (TASK-15700) and sync cache key gaps (TASK-15701).
- Eval harness adds a construction axis and census. Winner `and_stopword_trim` moves 10/105 hybrid metrics up (overall hybrid recall 0.826 → 0.848); 95/105 unchanged; zero down. OR‑bearing candidates were disqualified by a protected hybrid rescue constraint.
- Non‑string `fts_match_construction` values now degrade to a safe default instead of raising, preserving service availability.

**Review/rollout**
- Review `rag_service.py` (MATCH builder/selection), `simple_cache.py` (key change; legacy key compatibility), and new tests in `Tests/RAG_Search` and `Tests/RAG_Eval`.
- No config changes required; to revert, set `fts_match_construction="and"`. Expect one cold cache miss per query key after deploy.

<sup>Written for commit 88899eac0c76771b25fd9bada6f3146f4189aaeb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1574?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Remediated Qodo's review of this PR:

- **Finding 1 (bug, confirmed real)**: `RAGService._resolved_fts_match_construction` stored the raw `fts_match_construction` config value in `self._warned_fts_constructions` (a `set`) for warn-once dedup. Since `SearchConfig` is built from an untyped dict (`SearchConfig(**search_data)`, reachable from user-editable profile JSON), a list/dict value raised `TypeError: unhashable type` on the set membership check/`.add()` — crashing hybrid **and** keyword search instead of degrading to the conservative full `and`. Fixed by routing non-`str` values through a hashable surrogate dedup key (`f"{type(construction).__name__}:{construction!r}"`); `str` values keep using themselves unchanged, and the warning wording (naming the real shipped default, `and_stopword_trim`) is untouched. RED-first: reproduced the crash by hand with `[]`/`{}`, added two new tests (`test_a_non_str_construction_degrades_instead_of_crashing`, `test_a_non_str_construction_warns_once_per_distinct_bad_value`) that fail with the exact `TypeError: unhashable type: 'list'` on the pre-fix code and pass after; the pre-existing `test_an_invalid_construction_warns_once_and_behaves_as_and` (str-value dedup) stays green untouched.
- **Finding 2+3 (style, new test file)**: the flagged tests (`test_the_shipped_default_is_the_sweeps_winner`, `test_and_construction_is_byte_identical_to_the_shipped_escaper`) and every other `test_*` in `Tests/RAG_Search/test_fts5_match_construction.py` taking `tmp_path`/`monkeypatch` now annotate them (`tmp_path: Path`, `monkeypatch: pytest.MonkeyPatch`, `-> None`) and carry a Google-style `Args:` block, matching the convention already used elsewhere in the suite (e.g. `Tests/test_config_stt_provider_probe.py`). The disclosed-oracle-flip prose in the two flagged docstrings (both states, dates, sweep rows) is preserved verbatim — only an `Args:` section was appended, nothing rewritten. Verified via an AST sweep of the file: zero remaining `test_*` functions with unannotated parameters or a missing `Args:` section.

Test evidence: `Tests/RAG_Search/test_fts5_match_construction.py` (28 → 30, +2 new), `test_fts5_query_escaping.py`, `test_keyword_leg_prompts.py`, and `Tests/RAG_Eval` (ungated) — 341 passed, 11 skipped (gated `RAG_EVAL=1` harness tests, unaffected by this change), 0 failed.

— remediated in 88899eac0

PR #1608, opened 2026-08-13
RAG keyword leg: form-tiered merge + and_then_prefix by owner ruling (TASK-15700)

rmusser01:
## What this is

The merge-level fix TASK-15400 filed as its real finding, plus the re-application of that arc's pre-registered decision rule under the fixed merge — with the arc's one genuine judgment call decided by owner ruling and disclosed as such everywhere.

## Part A — the form-tiered merge

`_keyword_search`'s round-robin let one sub-leg's row count decide another sub-leg's leg rank: a zero-confidence fallback row could demote an untouched AND rank-1 row purely by source order (measured in 15400: it killed the vector-blind fixture's rescue and dropped scoped recall 1.000→0.429). The merge now partitions sub-legs into **tier 1 (primary-form) wholly preceding tier 2 (fell back)**, rank-fair round-robin within tiers — attributable, written down at the merge site with the incident cited, and structurally minimal because the fallback fires only on zero primary rows, so tiering partitions *sub-legs*, not rows. Fallback rows **fill, never displace**.

Behavior-neutrality for all-primary constructions was a **pre-registered intermediate gate**, not a hope: 105/105 gated metrics at (+0.000) and control census 20/53, witnessed three times independently.

## Part B — the re-run, and what shipped

Six constructions swept under the 15400 rule verbatim (the four originals + pre-registered `prefix` and `and_then_prefix`):

| row | census | resc | lost | vector-blind rescue |
|---|---|---|---|---|
| and (control) | 20 | 0 | 0 | yes @9 |
| and_stopword_trim (outgoing default) | 21 | 1 | 0 | yes @9 |
| or | 28 | 9 | 1 | **NO** (its own join loses it) |
| and_then_or | **29** | 9 | 0 | yes @10 — **disqualified on (b)**: 8 cells regress, 5 past the fail band; its fallback rows carry vector ranks and promote to merged rows above vector-only targets |
| prefix | 23 | 3 | 0 | yes @9 |
| **and_then_prefix (SHIPPED)** | 23 | 3 | 0 | yes @9 |

Part A's payoff is measured (`and_then_or`: rescue NO→yes, scoped 0.429→1.000) even though that row still fails — its residual gap is a fusion boundary, and its five regressing queries have *empty* keyword legs, where tiering is structurally inert.

**The decision, disclosed:** the qualifiers `prefix` and `and_then_prefix` tied at census 23 and are measurement-identical on every captured axis — all 105 cells, all 60 per-query hybrid *and* keyword-leg top-10s, same rescues, lost 0. The rule's tie-break (fewest extra FTS statements: 240 vs 460) selected `prefix`. **The owner ruled `and_then_prefix` ships** — the standing stability-over-quick-wins ruling applied to a dimension the tie-break predates: structural immunity to intra-sub-leg self-displacement (a nonempty AND primary is never widened; widening confined to tier 2), at a measured cost of 220 extra SQLite statements on a sparse corpus (upper bound; wall-time indistinguishable) and zero measured retrieval difference. Every outcome-recording site carries both facts; no site presents `and_then_prefix` as the rule's own output.

## What changed for users

Queries whose words all literally appear behave **byte-identically** (the new primary is the full AND). Only a sub-leg that finds *nothing* now widens — per-token prefix matching, so inflected forms are reachable ("schedule a watchlist brief" now finds the page that says "briefing"; demonstrated live). Census 20→23/53; residual zero-row 36/60 (from 39). **Zero of 105 gated cells moved** (byte-identical capture); the re-stamp is a disclosed non-event; baselines and fixtures are byte-identical to the merge-base.

## Honesty items

- The vector-blind fixture sits at fused slot 9 with slot 10 a **mathematical tie** (0.3/6 == 0.7/14 == 1/20 exactly, held by the documented tie-break) — the README states the boundary and forbids reading the slot-11 gap as headroom.
- The live check demonstrated the prefix fallback and a fused-list difference through the engine's builder against the app-written index; the RAG-Answer UI never rendered an Evidence row (the first-query hang recurred, worse — **TASK-15810** filed with a CPU-sample correction: the spin is inside the interpreter, not model loading).
- Rebased over dev's TASK-15103 diagnostic redaction: the resolver's warn-once pins now assert the **event-only contract** (fixed event, no user-value echo), per dev's security ruling.
- Two lessons filed with incidents: the pre-registered-rule + disclosed-owner-override pattern, and the sweep-stops-one-file-short recurrence (mechanical full-scope sweeps, not flagged-line patches).

## Tasks

TASK-15700 (Done, 7/7 ACs) · filed: TASK-15810 (RAG-Answer first-query hang, attribution corrected).

---
## Summary by cubic
Prevents fallback rows from displacing primary rows in the keyword leg and sets `and_then_prefix` as the default FTS5 MATCH construction to widen only when a source finds nothing. Previously a position-based round-robin could demote untouched primary rows; now primary-form rows always precede fallback-form rows, preserving leg rank.

- Behavior: tier 1 (primary form) fully precedes tier 2 (fallback); round-robin within each tier. All-primary queries are byte-identical; only zero-row sub-legs widen. Vector-blind rescues regained; leg census 20→23/53. Zero of 105 gated cells moved.
- Constructions: adds `prefix` and `and_then_prefix`. The rule’s tie-break picked `prefix`, but the owner ruled to ship `and_then_prefix` for structural stability. Retrieval difference measured as zero; worst case +220 SQLite statements on sparse corpora; wall-time unchanged.
- Config/cache: default `SearchConfig.fts_match_construction` changes `and_stopword_trim -> and_then_prefix` (no user knob). Async cache key includes `fts:<construction>`; expect one cold-miss run after upgrade. Sync key remains AND-stated; behavior unchanged.
- Scope: confined to `RAGService._keyword_search`; `pipeline_builder_simple._rrf_merge_parallel_results` verified exempt. Tests and docs updated; new `test_keyword_leg_tiered_merge.py`.
- Linear TASK-15700: fixes the merge displacement defect and re-applies the decision rule under the corrected merge; records the owner ruling and outcome.

<sup>Written for commit f680903134fcc3c0b6706a31cd9ddac73a9d6e2d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1608?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Qodo finding remediated: the widened-form probe's in-memory SQLite connection is now context-managed (`closing(...)`), so an assertion failure can no longer leak it. Suite re-run green (41 passed). — remediated in b8193bf4e

PR #1644, opened 2026-08-14
Make Console modals safely dismissible

rmusser01:
## Summary

- Add a shared safe modal dismissal contract for Escape and primary backdrop clicks across all Console-reachable modal screens.
- Preserve dirty-state, reset, compaction, staged-video, auto-speak consent, nested-surface, focus-restoration, and single-shot cancellation semantics.
- Add transitive runtime/AST inventory coverage for 28 Console modals and 37 reachable modal types, including undeclared-edge mutation proofs.

## Test plan

- [x] Rebased 16-file related matrix: 684 passed
- [x] Auto-speak consent plus shared dismissal matrix: 153 passed
- [x] Targeted Ruff lint, MyPy, compileall, and git diff --check
- [x] Independent cumulative review: READY with no findings before rebase

## Notes

- Rebased onto current origin/dev at 5e4162f324b0da1b405025ccedc0413a5fa33407.
- Full repository and broad directory suites were intentionally not run; verification was limited to touched files and related functionality.
- Existing dependency/deprecation warnings and untouched baseline formatting/MyPy diagnostics are unchanged.


rmusser01:
Addressed the Qodo review in ea4ad5160:

- Added Google-style Args/Returns documentation to is_modal_backdrop_click.
- Added docstrings to action_dismiss, action_settings_focus_next, and action_settings_focus_previous.
- Replaced the private textual._time dependency with the standard-library monotonic clock, using the same clock for backdrop capture and shield expiry; added a regression guard.
- Kept _SafeModalHost and _BackdropClickShield private. Their leading underscores intentionally mark non-API implementation types, which is the Python convention; their substantive names are PascalCase and the repository Ruff naming check passes. Renaming them to public-looking symbols would misstate the supported API surface.

Verification: 215 focused dismissal/Settings tests passed after these changes; Ruff lint/format on the changed clean files, targeted MyPy, compileall, and diff checks passed. The rebased 16-file related matrix previously passed 684 tests.

PR #1700, opened 2026-08-16
fix(console): settings crashes — unvalidatable-lineage guard + Select.NULL popover sentinel (tasks 16501/16502)

rmusser01:
Fixes two user-reported crashes that both block opening Console settings (tasks 16501, 16502; follow-up audit filed as task 16503).

## Crash 1 — `TypeError: 'NoneType' object is not iterable` (task-16501)

`select_valid_memory()` (console_context_compaction.py:236) received `active_messages=None` while opening Console settings. `ConsoleChatController._durable_context_snapshots()` legitimately returns `None` whenever the active lineage cannot be validated (e.g. a user message on the active path whose persistence write hasn't landed, a failing `get_message_version` reader, broken variant state). The two send-path consumers already guard with `if not snapshots:`; the two settings-surface consumers — `context_control_inputs()` and `reset_active_context_memory()` — did not, and the chat_screen caller catches only `(KeyError, ValueError)`, so the `TypeError` escaped. The crash fires even with zero stored memories, on both the full settings modal and the Alt+M popover.

**Fix:** the same `if not snapshots:` degradation the send-path consumers use — settings open with no active memory shown, reset deactivates nothing. `select_valid_memory()`'s contract is unchanged (`None` means "lineage cannot be validated", not "no messages").

## Crash 2 — `InvalidSelectValueError: Illegal select value False.` (task-16502)

On Textual 8.x the Select blank sentinel is `Select.NULL`; `Select.BLANK` was removed from the widget but resolves **silently** through the MRO to `Widget.BLANK: ClassVar[bool] = False` (an unrelated render flag). `console_model_popover.py` passed `value=Select.BLANK` whenever the session has no model (fresh install, or a provider without a configured default), and mount-time validation rejects `False` — reproducing the reported traceback byte-for-byte. Same mechanism task-565 documented and swept in settings_screen.py; this site was outside that sweep.

**Fix:** `Select.NULL` at both popover sites (the crashing `value=` seed and the inert-but-wrong `_apply` membership). The repo still carries ~60 `Select.BLANK` usages in ~20 other files, including deliberate `False`-placeholder sites that must NOT be blind-renamed — filed as task-16503 with per-site classification required, plus a lessons-testing-evidence entry (the mechanism was board-documented for three weeks while this crash sat live in another file).

## Tests

- `test_context_control_inputs_tolerate_unvalidatable_lineage` and `test_reset_active_context_memory_tolerates_unvalidatable_lineage` (Tests/Chat/test_console_context_compaction.py) reach the `None` state through the product path (controller fixture + one unpersisted user message) with a seeded active memory proving nothing is selected/deactivated. Verified RED with the exact reported `TypeError` before the fix.
- `test_quick_popover_mounts_with_no_model_selected` (Tests/UI/test_console_context_controls.py) mounts the popover with `model=None`. Verified RED with the byte-identical `InvalidSelectValueError` before the fix.

Suite evidence (failure **sets** compared from identical commands, branch vs clean `origin/dev`): the seven console suite files (Chat compaction/session-settings + five UI files) show the identical failure set on both trees — 459 passed / 2 failed on this branch vs 456 passed / 2 failed on clean dev (the +3 are this PR's tests), the 2 being pre-existing dev-tip contract failures (`test_context_controls_add_no_forbidden_keybindings`: `'Binding' object is not subscriptable`; `test_console_modal_inventory_matches_runtime_ast_and_transitive_launches`: `TrajectoryScreen` missing from the modal contract). `Tests/ProductionApp/test_provider_selection_ownership.py` likewise fails identically on both trees (`production Settings provider control did not stabilize`). None of the three pre-existing failures are touched by this diff.

---
## Summary by cubic
Prevents two crashes that blocked opening Console settings. Old: settings raised TypeError on unvalidatable lineage and the Alt+M popover raised InvalidSelectValueError when no model was set. New: settings open with no active memory and reset is a no‑op; the popover seeds Select.NULL and mounts cleanly.

- Guard None and empty snapshots in context_control_inputs() and reset_active_context_memory(); skip selection/reset when snapshots are falsy. select_valid_memory() contract unchanged. Treating None and () equivalently is now documented. Addresses task-16501.
- Replace Select.BLANK with Select.NULL in the popover’s initial value and _apply check; the blank row shows when no model is configured. Addresses task-16502. Follow-up audit is task-16503.
- Tests added: test_context_control_inputs_tolerate_unvalidatable_lineage, test_reset_active_context_memory_tolerates_unvalidatable_lineage, test_quick_popover_mounts_with_no_model_selected. No migration required.

<sup>Written for commit f134e4c1d6a27480516485c02f232ba2f6fae685. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1700?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Chat/test_console_context_compaction.py
+++ b/Tests/Chat/test_console_context_compaction.py
@@ -1354,3 +1354,48 @@ async def test_manual_compact_now_is_explicit_and_transcript_neutral() -> None:
     assert gateway.calls == 1
     assert before == after
     assert len(controller._context_repository.memories) == 1
+
+
+def test_context_control_inputs_tolerate_unvalidatable_lineage() -> None:
+    """Settings inputs degrade to no memory when the lineage cannot be validated.
+
+    TASK-16501: an unpersisted user message on the active path makes
+    ``_durable_context_snapshots`` return None; the settings seam must not
+    crash on it (user-reported TypeError while opening Console settings).
+    """
+    controller, store, session, _assistant, _gateway, _provider_messages = (
+        _controller_preflight_fixture(ContextCompactionMode.AUTOMATIC)
+    )
+    store.append_message(
+        session.id,
+        role=ConsoleMessageRole.USER,
+        content="not yet persisted",
+    )
+    assert controller._durable_context_snapshots(session.id) is None
+    controller._context_repository.memories.append(
+        _memory((_message("u1", "user", "one"),))
+    )
+
+    _overrides, _global_overrides, memory = controller.context_control_inputs(
+        session.id
+    )
+
+    assert memory is None
+
+
+def test_reset_active_context_memory_tolerates_unvalidatable_lineage() -> None:
+    """Reset deactivates nothing when the lineage cannot be validated (TASK-16501)."""
+    controller, store, session, _assistant, _gateway, _provider_messages = (
+        _controller_preflight_fixture(ContextCompactionMode.AUTOMATIC)
+    )
+    store.append_message(
+        session.id,
+        role=ConsoleMessageRole.USER,
+        content="not yet persisted",
+    )
+    assert controller._durable_context_snapshots(session.id) is None
+    controller._context_repository.memories.append(
+        _memory((_message("u1", "user", "one"),))
+    )
+
+    assert controller.reset_active_context_memory(session.id) is None
--- a/Tests/UI/test_console_context_controls.py
+++ b/Tests/UI/test_console_context_controls.py
@@ -513,3 +513,32 @@ async def test_unverified_model_capacity_is_labeled_as_estimated() -> None:
         assert "Model window (est.)" in window
         assert "model capacity is unverified" in status
         assert "Providers & Models" in status
+
+
+@pytest.mark.asyncio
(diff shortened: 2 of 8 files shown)


rmusser01:
Addressed the truthiness finding in f134e4c1d by documenting the equivalence as intentional rather than switching to `is None`: an empty snapshot tuple (persisted session, no durable rows yet) cannot validate any memory either — `select_valid_memory` builds an empty positions map and skips every candidate — so `None` and `()` correctly take the same degraded path. The two pre-existing send-path guards (`_compaction_admission`, `_apply_conversation_memory_preflight`) already use the same truthiness idiom on this seam; switching only the settings-surface sites to `is None` would split one seam across two idioms for identical behavior.

PR #1729, opened 2026-08-16
Semantic-route identity: carry the fallbacks, measure the window (TASK-16588)

rmusser01:
## What this is

The RAG programme's twelfth arc, closing TASK-16174's two unrefuted suspects on the routes a vector/fused profile actually uses. One small production change plus a mechanical dual-index probe with pre-registered expectations — no LLM, no spend. 7 commits, SDD (2 tasks + final whole-branch review + one precision fix wave).

## The change (one block in one file)

`_project_row` additionally emits the provenance identity fallbacks `expand_document` already accepts — `note_id`/`doc_id` — under the hint's own precondition, exactly Task 3b's shape. Six RED-first pins including never-emit-`media_id` (no builder writes it) and the sealing byte re-measure. The hint interface, row order/count, and the tool are untouched.

## The measurement (all three pre-registered expectations HELD)

Two seeded indexes (canonical builders vs a hand-built non-canonical `IndexEntry` producer), two routes each (semantic, hybrid), a direct `expand_document` call per declared-expandable row, before/after via a probe-side strip flag:

| index × route | declared-expandable | not_found PRE | not_found POST |
|---|---|---|---|
| canonical × semantic | 64 | 0 | 0 |
| canonical × hybrid | 61 | 0 | 0 |
| non-canonical × semantic | 49 | **49** | **0** |
| non-canonical × hybrid | 45 | **45** | **0** |

- **Suspect (b) closed with a measured before/after:** on a `source_id`-less index every declared-expandable row failed pre-fix (the chroma point id was the only candidate) and none fail post-fix. Canonical stayed 0/0 both arms — exactly as pre-registered, so no new finding was absorbed silently.
- **Suspect (a)'s residue closed with its own control:** markers planted ~9,600 chars into ~12,300-char documents (past the 8,000 budget) — `chunk_start`-anchored windows contained the marker **22/22**, the head-window control **0/22**.
- **The stronger aggregate (surfaced by review):** post-fix windows contained the row's own snippet head **340/340** — a wrong-document window could not do that, so the fallback resolved the *right* document on every rescued row.

## Honesty items, recorded where they bite

- Task 1's 15 B/row byte figure was a fixture-short-id artifact — real UUID ids cost **46–102 B/carrying row**; the correction is annotated at the fixture itself and in every prose surface. Largest observed payload: 53% of the 32 KiB ceiling.
- All 22 anchored windows were tail-windows, so 22/22 proves *off-the-head*, not *centred-on-match* — disclosed, with the rule a future extender must follow (markers not within one budget of the tail).
- `label_only` rows are **unreachable** on semantic/hybrid (0 of 340) — 16174's oracle run and this probe are complements; neither alone covers the tool.
- The spec's own first draft was corrected twice during the arc: the metadata-vocabulary transformation hypothesis was refuted for the canonical path (the storage helper spreads `source_id`/`source_type`), and the "runs against real profiles" claim was softened to what the evidence shows (such producers *exist* in-repo; whether a shipped app ever wrote a `source_id`-less index was not established).

## Discipline

- Gate at every step incl. the fix wave: verbatim `PASSED: No regression. 105 metric(s) within 0.05 of baseline`, cells grep-counted at (+0.000).
- Batteries: Agents 1,445 / Library 1,986 / Tools 572, 0 failed; collection sweep delta +8 exactly accounted.
- Final whole-branch review: SHIP-WITH-FIXES, nothing blocking; every number recomputed from raw artifacts with exact matches; both mutations red exactly the disclosed tests; ten suspicions dissolved on evidence. Five LOW precision findings all fixed in `959e24a0c`.
- Probe isolation: real config sha256 identical before/after, scratch data_dir asserted at runtime, no TUI boot.

Refs TASK-16588.

---
## Summary by cubic
Carry provenance identity fallbacks to semantic and hybrid route rows and cap identity strings at projection so `expand_document` resolves the right document and sealing never drops rows due to oversized metadata. Previously non-canonical indexes returned `not_found` and hostile ids could force zero-row payloads; now rows carry `note_id`/`doc_id` and identities are bounded at `_MAX_RESULT_ID_CHARS` (verbatim below the bound, truncated at it).

- In `tldw_chatbook/Agents/library_rag_tool_provider.py`: `_project_row` emits `note_id`/`doc_id` alongside `source_type`/`source_id`/`chunk_id`/`chunk_start` under the existing hint; never emits `media_id`. Bounds `source_id`/`chunk_id`/`note_id`/`doc_id` at projection to `_MAX_RESULT_ID_CHARS` to protect sealing.
- Validation: non-canonical indexes flipped from not_found 49/49 (semantic) and 45/45 (hybrid) to 0/0; canonical stayed 0/0. `chunk_start` windows contain the matched chunk; post-fix windows contain the row’s snippet head 340/340.
- Cost: +~15 B per carrying row on the unit fixture; 46–102 B per row on real payloads; largest observed payload 17,350 B (53% of 32 KiB); sealing keeps all rows; gate shows no regression.
- Tests/QA: unit pins in `Tests/Agents/test_library_tool_provider.py` include the hostile-identity two-regime case; mechanical dual-index probe and artifacts under `Docs/superpowers/qa/2026-08-16-rag-semantic-identity`; `route_probe.py` validates CLI paths and adds docstrings.

No API or config changes; no migration required. Addresses TASK-16588.

<sup>Written for commit 1ee18b0896627c65d7cc5ba9b208030106a99b8b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1729?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
All four Qodo findings addressed in `1ee18b089`.

**Finding 3 deserved its billing — it was real, and the RED test proves it rather than assuming it:** a 50k-char provenance id produced a **zero-row payload**. Identity fields are deliberately excluded from the sealing loop's shrink order (a halved id is a wrong id), so oversized untrusted metadata left the loop only one move: drop the entire row, snippet and all. Identity strings are now bounded at projection using the existing `_MAX_RESULT_ID_CHARS` precedent — `source_id`/`chunk_id` at the projection site, `note_id`/`doc_id` in the fallback helper. The 16174 hostile-identity pin is refined to the two-regime design this creates: **verbatim survival below the bound** (1,500-char id, shrink loop never touches it), **truncation at it** (an id past 2,000 chars names nothing fetchable — production ids are ≤1,000 — and the old behaviour lost the whole row). 56/56 on the provider file; gate re-read verbatim `PASSED: No regression. 105 metric(s)`.

**Finding 4:** the sealing test's assertion message no longer divides by zero when `carriers == 0` — a masked-failure hazard exactly as described.

**Findings 1–2:** `route_probe.py` now validates its CLI paths (resolve, refuse any repo-tree target — the one containment that matters for a scratch-writing probe — and reject unknown index kinds) and carries the missing Google-style docstrings with Args/Returns.

PR #1751, opened 2026-08-16
Reranker honesty: provider controls, cost disclosure, safe diagnostics — and the truth about coverage (TASK-3502)

rmusser01:
## What this is

TASK-3502: making the reranker honest now that TASK-3170 made it real. Seven commits, SDD (3 tasks + final whole-branch review + one fix wave). Every provider call in every test is faked at the `chat_api_call` seam — zero live spend in this arc.

## What ships

- **Provider/model controls (AC#1):** a provider Select enumerated from the real provider registry (set-equality pinned twice, never hand-listed), the default made visible ("openai (default)"), full adapter round-trip mirroring the model field's blank-means-default shape. Two UX catches pinned: the default row carries the provider *name* (a blank sentinel made anthropic→openai a silent no-op) and the change handler compares the *effective* provider (the mount-time `Select.Changed` echo would dirty the draft — the task-15740 family; the fixture now loads a fresh-clone blank so the guard is load-bearing in-test).
- **Cost disclosure before enabling (AC#2):** visible with reranking OFF, re-renders from the staged draft, and the profile *preview* discloses the browsed profile's numbers: "Reranking scores each result with a separate {provider} call — up to {top_k} calls per search…"
- **Copy semantics (AC#3):** the finding is architectural — *neither Pairwise nor Listwise mutates, but neither copies*: the entire cache-poisoning contract rests on `_tag_first_result`'s copy, now pinned by four tests whose teeth were proven by restoring the old in-place write (all four red).
- **The counter race (AC#4), scoped not locked:** `rerank()` now returns a frozen `RerankOutcome(results, failed, total)` — deliberately the *only* entry point, so no caller can take results while dropping the counts; the racy instance counters are gone, and Pairwise's class-level comparison counters (the same race one level down) thread a per-call tally through the merge-sort.
- **First UI consumer for the degradation tags (note-a):** one notice line in the existing coverage-note channel; the tag's survival (engine metadata → provenance → row) is pinned end-to-end — no plumbing was needed.
- **No more "| reranked" over-claim (note-b):** failed rows carry a `scored=False` flag that survives the result cache (cache-hit pinned), keep their original score kind, and never claim the reranked stamp.

## The truth about coverage — found by this arc's own review layer

The picker honestly offers 29 providers. The review probed the credential path and found it resolves for **1** (deepseek only — `_call_llm_impl` reads a `"API"` key that `load_settings()` never returns, so even the openai default cannot work). The fix wave's own reproduction went further: **end-to-end dispatch reaches 0 of 29** — the reranker calls `chat_api_call` with wrong *positional* argument order, so the API key lands in the endpoint parameter, and every seam fake had copied the caller's wrong order, which is why a green suite never saw it.

**Decision (per the review's reasoned recommendation): the picker ships as-is.** It doesn't create the sharp edge — it exposes a feature that was already 100% broken *silently*; failure is bounded, spend-free, and now *visible* via the new notice; and there is no honest set to bound to (deriving from a credential path that is itself wrong would swap a disclosed over-offer for an undisclosed false claim). All claim sites are softened and point at the owner:

- **TASK-17065** (high): the dispatch/credential repair — a *deliberate* change, since fixing it converts a disclosed no-op into first-search spend, which is exactly why it lands on top of this arc's cost surface.
- **TASK-17165** (high, security): the mis-order echoes a real key into an ERROR log line today for deepseek-keyed users — sink-side redaction filed per the fix-at-the-sink lesson, separate from the caller repair.
- **TASK-16965**: implement-or-retire `cross_encoder` with a gated-instrument measurement — the quality question, finally with an owner (an LLM reranker cannot be measured on the local deterministic instrument; a local cross-encoder can).

## Honesty items

- **The 105/105 gate is vacuous for this arc's stamping change** — no gated cell runs the reranker (`hybrid_basic` carries `reranking_config=None`). The close-out says so plainly and names the real coverage (the note-b tests, the construction converse pin, the render test). The gate still held, as it must.
- The ID-safety trap fired live during filing: the CLI probe offered an id 12 below the true ceiling, which sat *uncommitted* in another session's worktree; the on-disk sweep caught it.

## Verification

Gate verbatim `PASSED: No regression. 105 metric(s)`; `Tests/RAG_Search` 347/12s; settings+library batteries 546–556 passed across three independent re-reads; collection sweep 568/0; the review independently re-proved the AC#3 and note-b teeth by mutation. Pre-existing reds named and reproduced at the merge-base (3 settings-hub tests; 4 optional-dep imports).

Refs TASK-3502.

---
## Summary by cubic
Makes reranking honest and reviewable. The UI now lets users pick the provider and see cost before enabling; the engine counts failures per call, surfaces skipped/degraded runs, and stops stamping reranked on rows no model scored.

- Engine
  - `tldw_chatbook/RAG_Search/reranker.py` now returns `RerankOutcome(results, failed, total)` instead of a bare list; removes shared-instance counters and threads a per-call tally (fixes cross-search misattribution).
  - Adds per-row `RerankingResult.scored`; failed rows keep original score and score kind; no `rerank_score` stamp when a model did not score the row.
  - Pairwise/Listwise copy-not-mutate contract is pinned by new degraded-path tests; `tldw_chatbook/RAG_Search/simplified/enhanced_rag_service_v2.py` reads `RerankOutcome`, tags `reranking_skipped`/`reranking_degraded`, and logs a single disclosure.

- UI and rollout
  - Settings ▸ RAG adds a Provider Select enumerated from `Chat_Functions.API_CALL_HANDLERS` with “openai (default)”; unknown stored providers are preserved so users can repair bad values.
  - Cost line is visible with reranking off and tracks the staged provider and `top_k` (“up to {top_k} calls per search”); the Evidence region renders one quiet notice when tagged.
  - Docs updated; new tests cover degraded paths, provider enumeration and adapter read/write, notice rendering, and a seam characterization test that binds `chat_api_call`’s argument order (red-on-repair).
  - Known limitation: the current credential/dispatch path reaches 0 of the offered providers due to mis-ordered positional args; the picker ships to expose a visible, spend-free failure. Follow-ups: TASK-17065 (dispatch/credential repair), TASK-17165 (sink-side redaction), TASK-16965 (cross-encoder measurement).

Migration: direct callers of `rerank()` must use `outcome.results` and stop reading `last_rerank_*` counters. Addresses TASK-3502 AC#1–AC#4.

<sup>Written for commit fcbb39bfba3e45ae7506626fb886a7f2586053d4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1751?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
All three findings addressed in `f8e36fc95`.

**Finding 2 was real and is now RED-proven.** An unrecognised stored provider normalises to the default *for display*, so the mount-echo guard folded a user's corrective pick straight back to the broken value and re-saved it — a hand-edited or newer-build provider name could never be repaired from the picker. The fold-back is now restricted to values it actually preserves (blank, or a registered name); an unrecognised loaded value makes the pick a genuine change, which stages. The task-15740 mount-echo guard keeps its real job and both its cases stay pinned. This is the same issue our own final review raised as F5 and deferred — two independent reviewers flagging it moved it from "defer" to "fix", which was the right call.

**Finding 3 is the sharpest thing anyone said about this branch,** because it names the mechanism by which the underlying bug survived: every fake at this seam copies the caller's own parameter order, so a suite full of green reranker tests structurally cannot see that the caller is mis-ordered. It now has a mechanical guard — a characterization test binding the reranker's positional sequence against the **real** `chat_api_call` signature, asserting where each argument lands today:

| passed | lands in |
|---|---|
| the API key | `api_endpoint` |
| the provider | `api_key` |
| the model | `temp` |
| the temperature | `system_message` |
| `max_tokens` | `streaming` |

`model` and `maxp` are never reached at all. The test is **deliberately red-on-repair** and recorded as such on TASK-17065: fixing the caller must break it, and the repair includes rewriting it to assert the correct binding.

**Finding 1:** both new task files carry reproduction snippets; their key-shaped literals are replaced with values that cannot be mistaken for credentials or trip a scanner.

Verification after: `Tests/RAG_Search/` 348 passed / 12 skipped; the three settings files 247 passed; ruff clean on all touched files.

PR #1764, opened 2026-08-17
fix(research): window runs reach the pipeline; make decomposition a measured choice

rmusser01:
## Why

Following the question *"wouldn't that be the sub-question creation/multi-hop that we tested earlier and didn't include?"* — it was. Both mechanisms ship; both were off in every measurement, and one of the two shipped launch paths could not run at all.

## What this changes

**1. Research-window runs never reached the pipeline (TASK-17380, new).** `Research_Window._start_local_engine` built `LocalResearchEngine` with no `search_params`, so `generate_and_search` rejected the run on its own required-key check — an empty dict is falsy, giving `ValueError("Invalid search_params parameter")`. Verified by launching a run exactly as the window does:

```
status      : failed
error/msg   : Invalid search_params parameter
```

The Research screen is reachable from navigation again (task-16322), so this was the shipped path. Console `/research` was unaffected (it assembles params). The same omission left gap-driven replanning inert for window runs: `_default_gap_fn` reads `final_answer_llm` from those params and returns "no gaps" when unset, so multi-hop could never fire regardless of `max_iterations`.

Fix: the window assembles via the shared `deep_search_pipeline_params()` (task-16484) and reports an assembly failure instead of spending a doomed run; the engine pre-flights the pipeline's required params so a failure names the missing keys and `[SearchSettings]` — **only** when the real pipeline is the search function, so injected `search_fn` callers (every engine test, any future non-web lane) keep their contract. The required-key list is exported from the pipeline instead of duplicated.

**2. Decomposition is a measured choice, not a constant (TASK-17370).** The recorder hard-coded `subquery=False, max_queries=1` and launched with no `limits_json` (so `max_iterations` fell to the engine default of 1). It now takes `--max-queries`, `--max-iterations`, `--deadline-s`, derives sub-query generation from the query cap so the two cannot contradict, and stamps every emitted aggregate with the settings it ran under.

**3. The baseline doc now states what its numbers measured.** Every recorded gate number — including 0.29 → 0.42 — was measured with both mechanisms off. The gate is prompted with the run's sub-questions, so those runs handed it an empty list: a repository record had no narrower facet to be relevant to, which makes the recorded "the residual is partly genuine" reading unfalsifiable from those runs.

## Still open

- **The decomposition-on arm is not measured yet.** It needs the same judge model as the recorded arms (local Qwen3.8-27B, `:9191`), which is currently down — only a different model (`:9099`, gemma-4-26B) is up, and swapping judges would confound the one variable under test. No numbers are claimed.
- Fan-out never reaches the paper providers (they loop `round_queries` = `[question]` in round 1) — filed as TASK-17372, and the reason the pending re-run is a clean gate-only experiment.
- Shipped defaults + per-run UI control for decomposition stay filed as TASK-17371 (To Do), deliberately not decided here: the whole point is to set them from numbers.

## Verification

- `Tests/Research Tests/UI/test_research_screen.py Tests/Web_Scraping/test_deep_search_pipeline.py Tests/Tools/test_web_deep_search.py Tests/Helper_Scripts` — **362 passed**
- `ruff check` on all touched files — clean
- New tests: window assembles required keys; engine pre-flight fires with the real pipeline and stays out of the way for injected search functions.

Commit `e83a92f8f` (recorder flags) originated in a parallel session of the user's own and is carried here with its reasoning intact.

---
## Summary by cubic
Restores Research window runs to the deep-search pipeline, makes decomposition a measured per-run choice, and prevents summarizer errors from being stored as evidence. The engine now also fails fast when required LLMs are unset so runs don’t spend before inevitably failing.

**Details**
- Research window: `Research_Window` assembles `deep_search_pipeline_params()`. `LocalResearchEngine` preflights required pipeline keys when using the default pipeline and names missing keys and `[SearchSettings]`. Gap-driven replanning runs when `max_iterations > 1`.
- Early LLM checks: default-pipeline runs without `relevance_analysis_llm` or `final_answer_llm` now fail before phase 1. This makes the “summaries-only” degraded mode unreachable for research runs. Tool and injected `search_fn` callers are unchanged.
- Baseline recorder: adds `--max-queries`, `--max-iterations`, `--deadline-s`, and `--llm-timeout-s`; derives subquery generation from the query cap; records decomposition/timeouts in the aggregate JSON. Defaults remain one query and one iteration.
- Summarization: the `llama.cpp` summarizer prefers `api_settings.llama_cpp.api_url`, normalizes to `/v1/chat/completions`, and parses OpenAI chat-completion responses (falls back to native shapes). `_is_summary_failure` matches provider-prefixed failure phrases and prevents error text from becoming evidence for per-result and chunk summaries.

**Rollout**
- Configure `[SearchSettings]` `relevance_analysis_llm` and `final_answer_llm` to run default-pipeline research; otherwise runs fail fast.
- Summarizer routing prefers `api_settings.llama_cpp.api_url` but falls back to legacy `api_ip`. Recorder defaults are unchanged; injected `search_fn` callers continue to work without pipeline params.

<sup>Written for commit 54a25a2ecfdd39b098ecb915c8d7d18cdc859229. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1764?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/Development/research-report-eval-baseline.md
+++ b/Docs/Development/research-report-eval-baseline.md
@@ -165,6 +165,58 @@ closing the gap would need either a per-kind relevance threshold or
 category-tuned question sets — recorded as the follow-up lever, with the
 fallback (top-3 flagged) still covering the remainder.
 
+## Decomposition measured on the repositories lane (2026-08-17, task-17370)
+
+Same runner, same three questions, same judge (local Qwen3.8-27B on
+`:9191`), same 5-results bound as the 0.29 and 0.42 arms; the only change is
+`--max-queries 3`, which turns on phase-1 sub-question generation:
+
+Command:
+`python3 Helper_Scripts/Benchmarks/record_research_baseline.py --questions 3 --engine duckduckgo --academic --providers repositories --llm-base-url http://127.0.0.1:9191/v1 --max-queries 3 --max-iterations 1 --deadline-s 900`
+
+| Metric | 1 query, strict | 1 query + source-type note | **3 queries + note** |
+|---|---|---|---|
+| `gate_pass_rate` | 0.29 | 0.42 | **0.38** |
+| `citation_accuracy` | 1.00 (73/73) | 1.00 (72/72) | 1.00 (70/70) |
+| `claim_support_rate` | 0.97 | 1.00 | 1.00 |
+| `cited_sentence_ratio` | 0.52 | 0.75 | 0.70 |
+| `quote_grounding` | 0.00 | 0.33 | 0.33 |
+
+Per-question `gate_pass` in the fan-out arm: 0.54, 0.40, 0.20.
+
+**The result is negative: giving the gate narrower facets did not admit more
+repository evidence.** 0.38 against 0.42 is flat to slightly down, and this
+doc already records that gate pass-rate varies between identical runs on this
+model, so the honest reading is "no measurable improvement", not "a
+regression". Citation integrity again held (70/70 markers resolved).
+
+What this arm did and did not test, because it decides how far the result
+generalizes:
+
+- **Tested:** the gate's context. `websearch.result_relevance_eval` takes
+  `sub_questions` as a required placeholder, so the earlier arms rendered it
+  as an empty list and this one rendered real facets. That is a genuine
+  change in what the judge was asked.
+- **Not tested:** retrieval. The web lane returned zero results throughout
+  (the DuckDuckGo bot challenge this doc notes elsewhere), and the academic
+  lane only searches round 1's `[question]` (task-17372), so the repository
+  records under judgment were the SAME set as the 0.42 arm. Fan-out changed
+  how they were judged, not which ones existed.
+
+So the case for decomposition on this lane now rests on retrieval rather than
+on gate context: multi-hop rounds >= 2 do drive retrieval (measured
+separately below), and fan-out reaching the paper providers does not happen
+today at all (task-17372).
+
+One bound on the non-gate rows of every table in this document, including
+this arm's: per-result summarization was silently failing for llama.cpp
+(task-17382), so the synthesis was built from titles and gate reasoning with
+an error string where each source body belonged. `gate_pass_rate` is
+unaffected — the gate judges scraped content before summarization — but
+`cited_sentence_ratio`, `claim_support_rate` and `quote_grounding` were
+graded on reports written without their sources, and should be re-measured
+after that fix.
+
 ## What every number above was measured with (task-17370)
 
 Both of the pipeline's decomposition mechanisms were OFF for every baseline
--- a/Docs/Development/research-report-eval-baseline.md
+++ b/Docs/Development/research-report-eval-baseline.md
@@ -217,6 +217,55 @@ unaffected — the gate judges scraped content before summarization — but
 graded on reports written without their sources, and should be re-measured
 after that fix.
 
+## Multi-hop measured on the repositories lane (2026-08-17, task-17370)
+
+The same arm plus `--max-iterations 2`, so gap analysis gets to spend a second
+round. Read this table per question, NOT by its mean: the three runs did three
+different things.
+
+| | Arm B: 1 round | Arm C: 2 rounds |
+|---|---|---|
+| Q1 RAG | gate 0.54, markers 33/33, cited 0.92 | gate **0.71**, markers **0/0**, cited 0.00 |
+| Q2 MoE | gate 0.40, markers 24/24, cited 0.77 | gate 0.37, markers **39/39**, cited **0.95** |
+| Q3 GNN | gate 0.20, markers 13/13, cited 0.42 | gate 0.07, markers 1/1, cited 0.02 |
+| mean `gate_pass_rate` | 0.379 | 0.381 |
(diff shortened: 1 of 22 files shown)


rmusser01:
Two notes on the checks.

**The `No duplicate backlog task IDs` red is pre-existing on `dev`, not from this branch.** Running the guard's own rule (match on the `" - "` separator, so sub-task ids like `task-10.6` don't false-positive against `task-10`) against a clean `origin/dev` checkout yields exactly one duplicate:

```
task-16812
```

Two files claim it on `dev`: `task-16812 - Console-thinking-levels-and-budget-for-local-providers.md` and `task-16812 - Record-repository-research-graph-and-biomedical-stress-live-baselines.md`. This branch's four task files (17370, 17371, 17372, 17380) are each unique, so every PR against `dev` currently inherits this red.

I did not renumber it here, because it needs a decision rather than a rename: both tasks are `Done`, so "never move a Done task" can't break the tie, and both IDs are cited from live artifacts — ADR-066 links the Console file *by filename*, while the research id is cited from `academic_providers.py`, a test section header, the eval baseline doc, and task-17066's justification. Filed as **TASK-17381** with the constraints written down. Notably the Console file was added by a commit titled *"chore(backlog): resolve duplicate task IDs"* — a renumber that landed on a taken id, which is the second time that has happened here.

**TASK-17380 AC #1 is intentionally left unticked.** The window's assembly and the engine pre-flight are unit-verified, but "a window-launched run reaches the search phase" needs a live run against a configured endpoint, and the local Qwen endpoint every other live verification in this stream used is currently down. Same blocker as the unmeasured decomposition arm. Claiming it from unit evidence is exactly the substitution `lessons-testing-evidence.md` exists to prevent.

rmusser01:
Both findings adjudicated — one fixed, one declined with evidence.

## ✅ Fixed: "Window misses LLM preflight" (`54a25a2ec`)

Correct, and it contradicted this PR's own stated intent of failing before spend: the pre-flight checked the pipeline's required *search* keys but not its LLM identifiers, so a run with no `relevance_analysis_llm` / `final_answer_llm` would spend every phase-1 search and only then fail in relevance scoring or synthesis.

I fixed it in the **engine** rather than at the window, which is the second half of your own suggestion, because the shipped tool path already refuses both cases before phase 1 (`web_tool_impls.py:2369-2373` — `[deep-search-failed] relevance: no relevance_analysis_llm configured` / `synthesis: no final_answer_llm configured`) and the baseline recorder refuses at startup. Matching that contract in one place covers every default-pipeline caller including Console `/research`, instead of each launch site checking for itself. Empty string and `None` both count as unusable, since that's the shape a config with no LLMs actually produces.

**Trade-off recorded rather than glossed:** this makes `analyze_and_aggregate`'s "evidence summaries only" degraded mode unreachable for research *runs*. Deliberate — a run persists an artifact, and an unsynthesized one nobody asked for is worse than a legible refusal. The mode stays reachable for the tool and for injected-`search_fn` callers. Two tests cover both directions (empty/None LLMs refused naming both keys and `[SearchSettings]`; a complete assembly passes, so the check can't refuse every real run).

## ❌ Declined: `_FakeResponse` not PascalCase (rule 497154)

The rule reports a real regex mismatch, but the premise that this reduces consistency is inverted here. Measured on this repo:

```
511    test files containing leading-underscore helper classes
1,344  such classes
```

Leading-underscore test helpers are the established convention across the entire suite, and `_Name` is idiomatic Python for a module-private class. Renaming the 5 new ones would make them the inconsistent ones against 1,344 existing peers, and the cited precedents (PR #96/#97) are not test-helper renames.

If the intent is genuinely to enforce `^[A-Z][A-Za-z0-9]*$` on test helpers, that's a 1,344-site sweep and its own task — not something to start unannounced in a research-pipeline PR. Happy to file it if you want it.

## Note on the review itself

Your first pass returned *"We weren't able to complete the code review on our side"*, and the pass before that was stale (it reported 12 files against the branch's 22). This adjudication is against the `/agentic_review` run. Re-review welcome on `54a25a2ec`.

rmusser01:
Merging. For the record, precisely: Qodo's `/agentic_review` posted two findings at 19:50Z, both adjudicated — the LLM-preflight bug fixed in `54a25a2ec` with tests covering both directions, and the PascalCase rule declined with measured evidence (1,344 leading-underscore test helper classes across 511 files in this repo). I requested a re-review of the fix commit and none arrived in the following 15 minutes; their agent errored once and served a stale 12-file pass earlier today. So **the fix commit itself is not Qodo-reviewed** — it is covered by local verification (550 passed / 3 skipped across the research, window, pipeline, summarizer, recorder and tools suites; ruff clean) and by the review's own suggested remedy, which is what it implements.

The one red check, `No duplicate backlog task IDs`, is dev's pre-existing `task-16812` collision, filed as TASK-17381 — not introduced here.

PR #1765, opened 2026-08-17
feat(agents): note the active workspace in agent prompts when non-default

rmusser01:
## Summary

When a Console session runs in a workspace **other than the built-in default**, the agent (and any sub-agents it spawns) now receives an environment note in its system prompt telling it which workspace it's in and that the launch directory differs. The default workspace adds nothing to the prompt, so the common case is unchanged.

Rendered note (non-default workspace):

```
Note: This session is NOT running in the default workspace.
Active workspace: "Research Notes"
Launched from: tldw_chatbook/
Workspace file roots (relative to the launch directory):
  - data/corpus
  - external-data (outside the launch directory)
  - readonly (read-only)
```

**No absolute host paths are ever emitted** — roots inside the launch tree render as relative subpaths, a root equal to the launch dir renders as `.`, and out-of-tree roots render as their leaf folder name only (no `../..` traversal chain). The launch directory itself is shown by basename only.

## What changed

- **`Tools/workspace_file_roots.py`** — new `workspace_context_note()` shared resolver (used for both the primary agent and, via config propagation, its sub-agents so they stay byte-identical). It applies the *same* drift/symlink filtering as `allowed_file_roots`, so the note reflects what the file tools will actually honor rather than what's merely configured; it degrades to a one-line "not the default workspace" note on any registry failure or unknown id. Also adds set-once `set_launch_cwd()` / `get_launch_cwd()`.
- **`Agents/agent_models.py`** — `AgentConfig.workspace_context_note` (additive field, default `""`).
- **`Agents/agent_service.py`** — appended in `call_model` last each turn (native **and** fence-protocol paths, cache-friendly stable suffix). It touches only the runtime `system_content`, never the stored `config.system_prompt` that `_is_subagent` prefix-matches, so sub-agent identity detection is untouched. Propagated to `child_config`.
- **`Chat/console_agent_bridge.py`** — `run_reply` resolves the running session's workspace **up front** (so the note is present even on the fast path that skips the provider-gated branch) and sets the field.
- **`app.py`** — captures the launch directory once at process entry.

## Testing

- **21 new tests**, all green: 3 launch-cwd accessor, 12 resolver (default/none → empty, in-tree relative, equal → `.`, out-of-tree basename, read-only annotation, no-roots, drifted-symlink exclusion, launch-basename-only, name sanitization, registry-unavailable/unknown-id degrade), 3 primary-prompt integration (native + fence + empty), 1 sub-agent propagation, 2 bridge-wiring.
- **Regression net:** `Tests/Tools` + entire `Tests/Agents` + both Chat bridge suites = **1688 passed, 0 failed**.
- The broad `Tests/Chat` + `Tests/Internal_Prompts` run shows 15 failures in `test_console_provider_continuation.py` — these are **pre-existing baseline failures**, reproduced on a pristine detached `origin/dev` worktree with none of these changes. **Zero regressions from this PR.**

## Notes / follow-ups

- The note wording lives as module-level constants (mirroring `RUN_LOG_PROMPT_SECTION`); moving it into the Internal-Prompts registry for editability is a possible follow-up.
- No user-facing screen changed, so no `Docs/User_Guide` page update is required.

---
## Summary by cubic
Adds a workspace context note to agent system prompts when a session uses a non-default workspace, so agents and sub-agents know the active workspace and launch-relative file roots. Previously no note was added; default-workspace sessions remain unchanged.

- Builds the note in `tldw_chatbook.Tools.workspace_file_roots.workspace_context_note()`: applies the same filtering as `allowed_file_roots` (existing, non-symlink, non-drifted); in-tree roots render relative, out-of-tree roots render as leaf names, read-only roots are tagged; absolute host paths are never emitted. Workspace and path displays collapse whitespace, and the workspace name is JSON-delimited to prevent prompt injection. Unknown/failed registry degrades to a one-line “details unavailable” note.
- Captures the launch directory once via `set_launch_cwd`/`get_launch_cwd`, recorded on both the CLI path (`main_cli_runner`) and the `python -m tldw_chatbook.app` path; fixes the “Launched from: //” case when starting from filesystem root.
- Adds `AgentConfig.workspace_context_note`; `agent_service.call_model` appends it last each turn on native and fence-protocol endpoints. Sub-agents inherit via `spawn`, and the sub-agent identity prefix remains the prompt lead (detection unaffected).
- `console_agent_bridge.run_reply` resolves the session workspace up front and sets the note so it appears on both the fast path and the provider-gated path. Caveat: on the no-gate fast path the enforced file roots can fall back to the active workspace while the note reflects the session workspace.
- No migration for Console sessions. Custom runners can set `AgentConfig.workspace_context_note` using `workspace_file_roots.workspace_context_note(workspace_id)`; leave empty for default-workspace behavior. Backlog: task `TASK-17067` tracks deduplicating root-drift filtering across helpers (no functional change in this PR).

<sup>Written for commit ac53f9fdaee20e47d9e4f1bf14f0d8e8eb49b8cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1765?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Agents/test_agent_service.py
+++ b/Tests/Agents/test_agent_service.py
@@ -630,6 +630,38 @@ def test_spawn_creates_linked_child_with_clean_context(db):
     assert db.count_subagent_runs("c") == 1
 
 
+def test_spawn_propagates_workspace_note_to_child_prompt(db):
+    """A non-default workspace note rides the parent config onto the child's
+    config, so a spawned sub-agent -- which operates on the same workspace
+    roots -- sees it too. Appended last (in call_model), so the sub-agent
+    identity prefix still leads the child's system prompt."""
+    chat = FleetChat(
+        [
+            fence(SPAWN_TOOL_NAME, {"task": "compute 6*7"}),  # parent turn 1
+            "The sub-agent says 42.",  # parent turn 2
+        ],
+        {"compute 6*7": ["sub answer: 42"]},  # CHILD turn 1
+    )
+    service = _service_with(db, chat)
+    cfg = AgentConfig(
+        model="test-model",
+        system_prompt="You are helpful.",
+        allowed_tools=("calculator", "get_current_datetime", SPAWN_TOOL_NAME),
+        workspace_context_note="CHILD_WS_NOTE",
+    )
+    service.run_turn(
+        conversation_id="c",
+        messages=[{"role": "user", "content": "delegate this"}],
+        config=cfg,
+        api_endpoint="llama_cpp",
+    )
+    join_fleet_children(service)
+
+    child_system = chat.child_calls["compute 6*7"][0]["messages_payload"][0]["content"]
+    assert "CHILD_WS_NOTE" in child_system
+    assert child_system.startswith(SUBAGENT_SYSTEM_PROMPT.split(".")[0])
+
+
 def test_run_turn_records_assistant_message_id_on_primary_only(db):
     """The primary run records the assistant_message_id it is handed; a
     spawned sub-agent run records None (it produces no transcript reply)."""
--- a/Tests/Agents/test_workspace_context_note_prompt.py
+++ b/Tests/Agents/test_workspace_context_note_prompt.py
@@ -1,0 +1,74 @@
+# Tests/Agents/test_workspace_context_note_prompt.py
+"""The workspace-context note rides into the agent's system prompt.
+
+A non-default workspace attaches a note (built by
+``workspace_file_roots.workspace_context_note``) to ``AgentConfig``; it must
+reach the model's system prompt on every turn -- on native AND fence-protocol
+endpoints -- while the default-workspace empty string adds nothing.
+"""
+
+from tldw_chatbook.Agents import run_log as run_log_module
+from tldw_chatbook.Agents.agent_models import AgentConfig, RunBudget
+from tldw_chatbook.Agents.agent_service import AgentService
+from tldw_chatbook.Agents.tool_catalog import BuiltinToolProvider, ToolCatalogRegistry
+from tldw_chatbook.DB.AgentRuns_DB import AgentRunsDB
+
+
+def _service(tmp_path, capture):
+    db = AgentRunsDB(tmp_path / "runs.db")
+    registry = ToolCatalogRegistry()
+    registry.register_provider(BuiltinToolProvider())
+    return AgentService(db, registry, chat_call=capture)
+
+
+def _run(service, note, api_endpoint):
+    service.run_turn(
+        conversation_id="c",
+        messages=[{"role": "user", "content": "hi"}],
+        config=AgentConfig(
+            model="m",
+            system_prompt="BASE",
+            budget=RunBudget(),
+            workspace_context_note=note,
+        ),
+        api_endpoint=api_endpoint,
+    )
+
+
+def _capture(prompts):
+    def capture(**kwargs):
+        prompts.append(kwargs["messages_payload"][0]["content"])
+        return {"choices": [{"message": {"content": "ok"}}]}
+
+    return capture
+
+
+def test_workspace_note_reaches_system_prompt_on_native_endpoint(tmp_path, monkeypatch):
+    monkeypatch.setattr(run_log_module, "resolve_log_root", lambda: None)
+    prompts: list[str] = []
+    _run(_service(tmp_path, _capture(prompts)), "WS_NOTE_XYZ", "openai")
+    assert prompts, "expected at least one model turn"
+    assert all("WS_NOTE_XYZ" in p for p in prompts)
+    assert all(p.startswith("BASE") for p in prompts)
+
+
+def test_workspace_note_reaches_system_prompt_on_fence_protocol_endpoint(
+    tmp_path, monkeypatch
+):
+    """The non-native branch reassigns ``system_content`` to splice the fence
+    protocol -- the note must survive that, exactly as RUN_LOG_PROMPT_SECTION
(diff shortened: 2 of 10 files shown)


rmusser01:
## Code-review adjudication (follow-up commit `6fcd056fc`)

Each finding was verified against the code before acting.

| # | Finding | Verdict | Action |
|---|---|---|---|
| 1 | Launch cwd never captured on the `python -m tldw_chatbook.app` path (the documented dev run command) — it doesn't route through `main_cli_runner` | ✅ real | Added `set_launch_cwd()` to the `__main__` block, before anything can `chdir` |
| 2 | Folder-root display strings spliced into the prompt note without the newline sanitization the workspace **name** gets → prompt-injection via a bound folder whose leaf name contains a newline | ✅ real (security) | Collapse whitespace in root display strings (parity with the name guard) + regression test |
| 5 | `Launched from: //` (doubled slash) when launched from the filesystem root | ✅ real (cosmetic) | Use the path anchor for the empty-name case + test |
| 6 | Comment claims `_is_subagent` matches the stored `config.system_prompt`; it actually prefix-matches the **sent** `messages_payload[0]` content | ✅ confirmed | Corrected the comment. The code was already correct: the note is appended **last**, so the sent prompt still leads with the sub-agent identity prefix and detection is unaffected |
| 4 | On the no-gate fast path the note describes the session workspace while file tools fall back to the active workspace, so they can diverge | ⚠️ latent — mitigated in production (`console_chat_controller` always passes `builtin_gate` → the fresh-build branch binds `run_workspace(run_workspace_id)`); reachable only in test/embedding configs | Documented the assumption in a code comment at the note-build site |
| 7 | `set_launch_cwd`'s set-once is a non-atomic check-then-assign; docstring overclaims thread-safety | ⚠️ benign — the value is set once at single-threaded boot | Tightened the docstring to state the actual first-write-wins / single-threaded-boot contract (no lock added for a boot-only value) |
| 3 | The binding drift/symlink filter loop now exists in three places (`workspace_context_note`, `allowed_file_roots`, `folder_binding_roots`) | 🟡 valid maintainability nit | **Deferred.** Consolidating would touch the security-sensitive, ADR-028 `allowed_file_roots` (which interleaves its rw/ro filter) and the differently-logged `folder_binding_roots`; extracting a shared iterator is worthwhile but out of scope for this PR. Tracked as a follow-up |

The two substantive fixes are **#1** (the feature's launch-capture guarantee was unmet on the primary dev command) and **#2** (the exact injection hole the name guard closes was left open for root names).

**Verification:** 285 passed / 0 failed across the touched suites (`Tests/Tools`, `Tests/Agents`, `Tests/Chat` bridge), including 2 new regression tests. The rendered note is byte-identical for the normal case.

rmusser01:
## Qodo review adjudication (commit `505d93168`)

**Rebase:** `origin/dev` is still at `524194c15` — the exact base this branch was cut from — so the branch is already current with `dev` (rebase was a no-op, no history rewrite needed).

| # | Qodo finding | Verdict | Action |
|---|---|---|---|
| 1 | **Workspace name prompt injection** (🐞 Bug / Security, Strong) — name embedded without robust delimiting/escaping | ✅ valid defense-in-depth | **Fixed** in `505d93168`: the name is now rendered via `json.dumps(name, ensure_ascii=False)`, delimiting it as a data literal and escaping embedded quotes/backslashes/control chars so a crafted name can't break out of the quoted field. Layered over the existing whitespace-collapse (which already blocked the newline/fake-section vector). Added a quote-escaping regression test. Normal names render identically. I did **not** add the suggested standalone "untrusted metadata" disclaimer line — it costs tokens every non-default turn and JSON-delimiting already marks the value as data. |
| 2 | **Unvalidated `binding.locator` filesystem access** (📘 Rule 497145 / Security, Moderate) | ⛔ **Declined** (not applicable) | Evidence: (a) the locator is validated + canonicalized at write time in `registry_service.add_folder_binding` (`expanduser().resolve()`, `is_dir()` check, denies filesystem root / home). (b) `workspace_context_note` performs **no file operation** — only `is_dir()`/`is_symlink()`/`resolve()` stat checks to decide *display*, byte-for-byte the same pattern used by `allowed_file_roots` (the actual tool-enforcement point) and `folder_binding_roots`, **neither of which calls `path_validation.py`**. (c) Adding `path_validation` *only* here would make the note diverge from `allowed_file_roots`, re-introducing the "note advertises a root the tools reject" bug this design deliberately avoids. `path_validation.py` guards file *operations* (open/read/write/traverse), not display-only stat filtering of an already-validated path. |
| 3 | **Root launch basename `//`** (🐞 Bug / Correctness, Strong) | ✅ already fixed | Resolved earlier in `6fcd056fc` (uses the path anchor for the empty-name/root case) — Qodo already marks this **✓ Resolved**. |

**Verification:** 286 passed / 0 failed across the touched suites (`Tests/Tools`, `Tests/Agents`, `Tests/Chat` bridge), including the new quote-escape test. Rendered note is unchanged for normal names.

rmusser01:
The one deferred item from the reviews — consolidating the drift/symlink filter loop shared by `allowed_file_roots` / `folder_binding_roots` / `workspace_context_note` — is now filed as **TASK-17067** and committed to this branch (`ac53f9fda`, under `backlog/tasks/`). It's scoped so the note can never diverge from the roots the file tools actually honor.

PR #1766, opened 2026-08-17
feat(research): multi-hop by default for local research runs

rmusser01:
Stacked on **#1764** (`feat/research-shipped-decomposition`), which carries the measurement this decision rests on. This branch's own change is the single commit `11eda0b47`; merging #1764 first makes this diff one file plus its test and config comment.

## The decision

Gap-driven replanning has shipped since task-16324, but `max_iterations` defaulted to `1`, so **no real run ever performed a second round** — the mechanism was dead in practice. TASK-17370 measured what it's worth on the repositories lane, and the two halves of decomposition came out differently:

| Mechanism | What it changes | Measured | Shipped default |
|---|---|---|---|
| Multi-hop (`max_iterations`) | **retrieval** — round 2's queries reach the paper providers | held gate rate; markers 24 → 39; citation density 0.77 → 0.95 | **on (2)** |
| Fan-out (`subquery_generation`) | only the gate's context | 0.42 → 0.38, flat | off |

Fan-out stays off because it measured flat *and* cannot change retrieval on this lane at all while TASK-17372 stands (the academic lane only ever sees round 1's `[question]`).

## How it's wired

`DEFAULT_MAX_ITERATIONS = 2`, resolved through `_configured_max_iterations()`:

- per install: `[SearchSettings] research_max_iterations`
- per run: `limits_json.max_iterations` — always wins
- missing, unreadable or non-positive config falls back to the shipped default, so configuration can never silently disable the mechanism

The per-run override is what keeps recorded baselines honest: the baseline recorder passes `max_iterations` explicitly (default 1), so every recorded number stays reproducible byte-for-byte under the changed default.

## Cost, stated where the user meets it

Multiplicative, and documented on the config key rather than as a general warning: one extra search per gap, each with its own per-result relevance and summarization calls, plus another synthesis and gap analysis per round. **Measured: 3 → 12 search calls across three questions, roughly tripled wall-clock.**

Two interactions this default increases exposure to, both recorded in the eval baseline doc:

- each new source needs a summarization call, which cannot complete inside the shipped 30s `relevance_llm_timeout_s` on a local model (TASK-17382's chain; measured 42–131s per summary)
- larger evidence pools are what trigger map-reduce chunking, where chunk summarization still fails against a local endpoint (TASK-17384)

Both degrade to real source text rather than to nonsense, so the default is safe — but a local-model install will feel this as latency, not as better summaries.

## Verification

- New tests pin all three behaviours: a second round runs with no limits at all, an explicit `max_iterations: 1` still yields exactly one pass, and a configured default is honoured.
- **382 passed** across `Tests/Research`, `Tests/UI/test_research_screen.py`, `Tests/Helper_Scripts`, `Tests/Tools/test_web_deep_search.py`, `Tests/Web_Scraping/test_deep_search_pipeline.py`.
- `ruff check` clean on all touched files.

## Still open

TASK-17371's AC #2 — a real per-run control that shows the setting and persists it — is deliberately **not** in this PR. Today a user overrides per run only by typing limits into the Research window's existing input. That control matters more than the default itself and should be its own change.

---
## Summary by cubic
Local research runs default to multi-hop (two rounds) to improve retrieval, and approved plan-review limits now bound iterations. Previously runs stopped after one pass by default, and a plan-review patch could be ignored for the iteration count, allowing an extra round.

- `_configured_max_iterations()` resolves the default: per-run `limits_json.max_iterations` wins; per-install `[SearchSettings] research_max_iterations` sets the default; non-positive or unreadable falls back to 2. Fan-out stays off.
- `_execute_phases` now receives the run’s effective `limits` (with any approved plan-review patch merged); it no longer re-reads limits from the run record, so the ledger and iteration bound agree.
- Baselines remain reproducible: the recorder still passes `max_iterations: 1` explicitly. Tests pin default second round, single-pass override, configured default of 3, and plan-review bounding.
- Config comments document `research_max_iterations` and `relevance_llm_timeout_s` implications. Meets `TASK-17371` ACs #1, #3, #4; AC #2 (UI control) remains out of scope.
- Expect more searches and longer wall-clock; on local models, summarization often times out at 30s and degrades to source text.

**Migration**
- To keep single-pass behavior, set `limits_json.max_iterations: 1` per run or `[SearchSettings] research_max_iterations = 1`.
- On local models, raise `relevance_llm_timeout_s` if you want summarized evidence instead of raw source text.

<sup>Written for commit 7d79b997558f03877d09be3c0b21d95958e8eb8d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1766?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/Development/research-report-eval-baseline.md
+++ b/Docs/Development/research-report-eval-baseline.md
@@ -392,6 +392,36 @@ chunk's source text and marks it not-generated, so evidence stays real; the cost
 is a wasted call and summarization quality on exactly the large evidence pools
 multi-hop produces.
 
+### The shipped default that came out of this (task-17371)
+
+Local research runs are **multi-hop by default** as of this measurement:
+`DEFAULT_MAX_ITERATIONS = 2` in `local_research_engine.py`, overridable per
+install via `[SearchSettings] research_max_iterations` and per run via
+`limits_json.max_iterations`.
+
+Chosen from the numbers above rather than from caution: the second round is the
+half of decomposition that changed retrieval, and on the one question whose
+synthesis path was intact it held the gate's pass rate while taking resolved
+markers 24 -> 39 and citation density 0.77 -> 0.95. Fan-out stays OFF by
+default, because it measured flat (0.42 -> 0.38) and on this lane it cannot
+change retrieval at all while task-17372 stands.
+
+The cost is real and multiplicative: one extra search per gap, each with its own
+per-result relevance and summarization calls, plus another synthesis and gap
+analysis per round -- the measured arm went 3 -> 12 search calls over three
+questions and roughly tripled wall-clock. Recorded baselines are unaffected
+because the recorder passes `max_iterations` explicitly (default 1), which is
+the property that keeps them reproducible byte-for-byte.
+
+Two known interactions worth stating, since this default increases exposure to
+both: multi-hop enlarges the evidence pool, and (a) each new source needs its
+own summarization call, which cannot complete inside the shipped 30s timeout on
+a local model (task-17382 chain), and (b) larger pools are what trigger
+map-reduce chunking, where chunk summarization still fails against a local
+endpoint (task-17384). Both degrade to real source text rather than to
+nonsense, so the default is safe -- but a local-model install will feel it as
+latency, not as better summaries.
+
 ### Reading gate_pass_rate
 
 It is a ratio, and multi-hop adds to its denominator. Pulling in more marginal
--- a/Tests/Research/test_local_research_engine.py
+++ b/Tests/Research/test_local_research_engine.py
@@ -1105,3 +1105,92 @@ def test_preflight_accepts_fully_configured_params():
             "final_answer_llm": "llama_cpp",
         }
     )
+
+
+# --- multi-hop on by default (task-17371) -------------------------------------
+# Gap-driven replanning shipped but max_iterations defaulted to 1, so every
+# real run was single-pass and the mechanism never ran. task-17370 measured
+# what it is worth: on the one question whose synthesis path was intact,
+# a second round held the gate rate while taking resolved markers from 24 to
+# 39 and citation density from 0.77 to 0.95. Deep research defaults to it now.
+
+
+def _gap_pipeline(question: str, gaps_per_round):
+    """search/analyze/gap fakes recording the queries of every round."""
+    rounds: list[str] = []
+    remaining = list(gaps_per_round)
+
+    def search_fn(q, params):
+        rounds.append(q)
+        return (
+            {"results": [{"title": f"R:{q}", "url": f"https://x.example/{len(rounds)}"}],
+             "warnings": []},
+            {"sub_questions": [], "main_goal": question},
+        )
+
+    async def analyze_fn(wsr, sqd, params, cancel_event=None):
+        return {
+            "final_answer": {"text": "Answer citing [1].", "evidence": [
+                {"id": 1, "url": "https://x.example/1", "title": "R"}],
+                "confidence": 0.5, "chunks": []},
+            "relevant_results": {"0": {"url": "https://x.example/1"}},
+        }
+
+    async def gap_fn(context):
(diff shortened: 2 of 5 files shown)


rmusser01:
Rebased onto `dev` at `4620a3f93` (which now contains #1764), so this PR is down to its own single commit — the engine default, its config key, three tests, and the doc/task records.

Verification after the rebase: **276 passed** across `Tests/Research`, `Tests/UI/test_research_screen.py` and `Tests/Helper_Scripts`; ruff clean on all touched files.

/review

rmusser01:
Both findings adjudicated — the correctness one fixed, the observability one declined.

## ✅ Fixed: "Plan patch ignored" (`7d79b9975`)

Real bug, well spotted, and this PR made it matter. `execute_run` merged an approved plan-review limits patch into the limits used for the `BudgetLedger`, but `_execute_phases` re-read limits from the run record — and approved checkpoints don't mutate stored limits. So the patch bounded the *budget* while the iteration count fell back to the default.

Pre-existing, but this PR flipped its direction from harmless to harmful: with multi-hop defaulting to 2, a user who limits a run to a single pass *during plan review* would still get a second round and spend more than they'd just approved. That's the worst version of the bug, so thank you for catching it before it shipped.

Fixed by passing the effective limits into `_execute_phases` rather than re-deriving them, so the ledger and the iteration bound can't disagree by construction (your second suggested option — I preferred it over mutating the in-memory `run` because one source of truth beats two that must be kept in sync).

**On the test, worth recording:** my first attempt followed your suggestion literally and drove a *checkpointed* run. It read `['q', 'q']` and proved nothing — a resumed checkpointed run restarts the phase machine from the top (documented v1 limitation), so it re-searches round 1 and round counting through that path is ambiguous. The committed test drives an autonomous run with a pre-approved plan patch, which isolates the merge under test. **Mutation-checked**: reintroducing the re-read makes it fail with `['q', 'gap one']`.

## ❌ Declined: "logging.error lacks request context" (rule, Observability)

Two independent reasons.

**It would violate this repo's diagnostic-privacy contract.** That log lives in a file whose diagnostics are tracked in a security-reviewed inventory (`Docs/security/production-diagnostic-inventory.json`, task-492/3750), and `Tests/LLM_Calls/test_summarization_diagnostic_privacy.py` tracks **11** diagnostics in a `private endpoint/path` category with runtime canary tests (`test_runtime_sentinel_hides_private_value[local-path]`) that fail if an endpoint reaches a log line. Adding `api_url` is precisely what those canaries exist to prevent — the suggestion would trade a debugging convenience for a deliberate privacy guarantee.

**It is also out of scope here.** The llama.cpp summarizer change is in #1764, which is already merged; this PR's diff is the engine default, its config key, three tests and the doc/task records. Nothing in it touches that log.

For debuggability without the leak, the response *shape* (key names, not values) would be defensible — but it belongs in a change that owns that file and updates the reviewed inventory deliberately, not here.

Re-review welcome on `7d79b9975`.

rmusser01:
Merging, with the same explicit record as #1764: Qodo's review posted at 19:14Z, both findings adjudicated — the plan-patch correctness bug fixed in `7d79b9975` with a mutation-checked regression test, the observability rule declined on privacy-contract and scope grounds. I requested a re-review of the fix commit and none arrived; **the fix commit itself is not Qodo-reviewed**, and rests on local verification (338 passed across the research, window, recorder and pipeline suites; ruff clean) plus a mutation check proving the new test fails without it.

Note I also caught my own polling script reporting a false 'fresh review' — its timestamp comparison was malformed. The timestamps above are read directly from the API.

The one red check remains dev's pre-existing `task-16812` duplicate (TASK-17381).

PR #1769, opened 2026-08-17
feat(research): show and persist the multi-hop rounds a run will use

rmusser01:
Closes **TASK-17371**'s last open criterion (AC #2). Single commit on top of `dev`, which now contains #1764 and #1766.

## Why

#1766 made multi-hop the shipped default (2 rounds), chosen from measured numbers. But a user could only *change* it by typing `max_iterations=N` into the limits box — a setting that existed and was invisible. AC #2 asked for a control that shows the setting and persists it, and I argued in that PR that the control matters more than the default itself.

## What

A rounds `Select` beside the policy picker in the Research window.

**Its default comes from the engine's own resolver** (`_configured_max_iterations`), not a second default of its own. A number displayed here that could drift from the number a run actually uses would be worse than showing nothing.

**Persistence** rides the existing `save_state`/`restore_state` path, exactly like the academic-lane toggle.

**Precedence:** the control contributes `max_iterations` to the launched run's `limits_json` **only when the limits text doesn't already state one**. A typed value is the more specific statement of intent, and it's what the engine treats as authoritative — the control must not silently overwrite it.

**Spend at the point of decision:** selecting a value states the trade-off in the status line ("each extra round researches the gaps the previous answer left open — more evidence, and proportionally more searches and LLM calls") rather than leaving it to a config comment the user will never open.

## Existing tests changed, with reasons in-line

Three tests pinned the exact persisted state and `limits_json` and now see the new key. Updated with the reason recorded at each site, which is what AC #4 requires:

- the window's persisted state gained `rounds`
- a launched run carries `max_iterations` whenever the limits text doesn't state one

## Verification

- **200 passed** across `Tests/UI/test_research_screen.py` and `Tests/Research`; `ruff` clean.
- Four new unit tests: default sourced from the engine, state round-trip, the value reaching `limits_json`, and typed limits winning.
- One **compose-level** test mounts the window in a Textual app and drives the `Select` — an int-valued `Select` rendering with the right option selected isn't something unit-level state can prove, and that's exactly the kind of thing that breaks silently.

## Note

I have not merged this one. Your merge authorization covered #1764 and #1766 specifically; this is new work, so it's yours to call. Qodo review requested.

---
## Summary by cubic
Shows and persists the multi-hop rounds a research run will use, and fixes two defects in the new control. Previously rounds were hidden and only changeable via limits text; now a visible rounds Select persists and adds max_iterations to limits unless the limits text already sets it.

- Default comes from the engine’s configured max_iterations to avoid drift; selecting a value updates the status line with expected spend. Satisfies TASK-17371 AC #2.
- Persistence: save_state adds rounds; restore_state restores it with a safe fallback and widens the mounted Select’s options to include the current value so high configured/restored rounds do not crash the window.
- Precedence and casing: a typed max_iterations wins over the control regardless of key casing; the value is normalized onto the canonical lowercase key so the engine receives it.
- Tests updated where persisted state and limits_json were pinned; new unit and compose-level tests cover default sourcing, round-trip, payload merge, casing precedence, and mounting with high rounds. 203 tests pass.

<sup>Written for commit 1bb7126326725a1874472804a4b0982acab2d6fe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1769?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_research_screen.py
+++ b/Tests/UI/test_research_screen.py
@@ -1,3 +1,4 @@
+import asyncio
 from types import SimpleNamespace
 
 import pytest
@@ -372,14 +373,16 @@ def test_window_academic_toggle_defaults_off_and_persists_in_state():
     window = ResearchWindow(app_instance=app)
 
     assert window.academic_enabled is False
+    # "rounds" joined the persisted state in task-17371 -- the window now shows
+    # and remembers how many multi-hop rounds it will launch with.
     assert window.save_state() == {"source": "local", "academic": False,
                                    "limits": "", "policy": "balanced",
-                                   "providers": ""}
+                                   "providers": "", "rounds": 2}
 
     window.academic_enabled = True
     assert window.save_state() == {"source": "local", "academic": True,
                                    "limits": "", "policy": "balanced",
-                                   "providers": ""}
+                                   "providers": "", "rounds": 2}
 
 
 def test_window_academic_toggle_restores_from_state():
@@ -463,8 +466,10 @@ async def test_create_run_carries_parsed_limits(monkeypatch):
     await window.create_run({"query": "Budgeted question"})
 
     create_call = [c for c in service.calls if c[0] == "create_run"][0]
+    # max_iterations rides along from the rounds control (task-17371) whenever
+    # the limits text does not state one; the typed budget keys are untouched.
     assert create_call[2]["limits_json"] == {
-        "max_searches": 3, "max_fetched_docs": 10.0,
+        "max_searches": 3, "max_fetched_docs": 10.0, "max_iterations": 2,
     }
 
 
@@ -779,7 +784,7 @@ async def test_create_run_reads_limits_and_providers_from_inputs(monkeypatch):
     create_call = [c for c in service.calls if c[0] == "create_run"][0]
     payload = create_call[2]
     assert payload["provider_overrides"]["academic_providers"] == ["pubmed", "arxiv"]
-    assert payload["limits_json"] == {"max_searches": 3}
+    assert payload["limits_json"] == {"max_searches": 3, "max_iterations": 2}
 
 
 def test_parse_provider_tokens_validates_and_dedupes():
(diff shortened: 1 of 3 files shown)


rmusser01:
All four findings addressed in `1bb712632` — three accepted (two of them real bugs), and the framing of one corrected.

## ✅ Fixed: "Rounds Select value crash"

Serious, and I'd missed it: the Select offered a fixed 1–4 with `allow_blank=False` while its value came from configuration or restored state, so an install setting `research_max_iterations` above 4 — or a restored state holding more — raised `InvalidSelectValueError` and **the entire window failed to mount**. My own mount test passed only because it used the default of 2.

Options are now built to always contain the current value, and `restore_state` widens the mounted control before assigning. I extended the options rather than clamping the value deliberately: clamping would make the control display a different number than the run will actually use, which is worse than the crash it prevents.

## ✅ Fixed: "Case-sensitive max_iterations check"

Also real. `_parse_limits_text` keeps keys verbatim, so a typed `Max_Iterations=1` was invisible to my presence check *and* to the engine, which reads the canonical lowercase key. Net effect: the control injected its own value and took charge of a run the user had explicitly bounded to one pass — precisely the override this feature promises not to do.

Any casing now counts and is normalized onto the canonical key, so the typed value both wins *and* reaches the engine (previously it would have been dropped either way). A later variant wins, as a repeated key would.

Both fixes are covered by compose-level tests that mount the window, since neither defect is reachable from unit state — one is a mount-time exception, the other only appears once parsed limits reach the payload.

## ✅ Accepted: docstrings (findings 2 and 3)

Fair catch rather than noise — the repo's own style guide requires Google-style docstrings on public APIs, and I'd modified four public methods without adding them. Added on `compose`, `save_state`, `restore_state`, `create_run`, and on the two new tests that lacked one.

## Verification

**203 passed** across `Tests/UI/test_research_screen.py` and `Tests/Research`; `ruff` clean.

Re-review welcome on `1bb712632`.

rmusser01:
Merging, with the same explicit record as #1764 and #1766: Qodo's review posted at 20:25Z, all four findings addressed in `1bb712632` — the mount-time `InvalidSelectValueError` and the case-sensitivity override fixed with compose-level tests, both docstring findings accepted. I requested a re-review of the fix commit and none arrived, so **the fix commit itself is not Qodo-reviewed**; it rests on local verification (361 passed across the window, research, recorder and pipeline suites; ruff clean).

That closes TASK-17371 in full — the shipped default from measured numbers (AC #1), a visible and persisted per-run control (AC #2), the spend implication stated at both the config key and the point of decision (AC #3), and the three pre-existing tests updated with reasons recorded in-line (AC #4).

The remaining red check is dev's pre-existing `task-16812` duplicate (TASK-17381).

PR #1772, opened 2026-08-17
feat(research): let generated sub-questions reach the paper providers

rmusser01:
Closes **TASK-17372**, the finding that explained why fan-out measured flat — and the live arm here shows what it unlocks, plus what still blocks it.

## The asymmetry

Sub-question generation lives inside the *web* pipeline, so generated facets only ever drove web searches: the academic lane looped `round_queries`, which is `[question]` in round 1. Fan-out changed how academic evidence was **judged** (facets do reach the relevance prompt via the merged sub-question list) while leaving what was **retrieved** identical. That's why TASK-17370's measurement recorded fan-out as flat and could say nothing about retrieval — the records under judgment were the same set either way.

## The decision, made explicitly

The lane now searches the facets. Stated rather than left incidental: fan-out's only measured effect was on the gate, and it was nil, so unless facets reach retrieval the feature has no demonstrated purpose.

**Bounded twice**, since this multiplies provider calls:

- the total obeys the same `search_default_max_queries` the web lane obeys
- each *extra* query reserves and settles a search against the ledger, so a tight `max_searches` cannot be exceeded by the lane fanning out

The *base* `round_queries` keep today's uncounted accounting deliberately — counting them would silently shrink every existing run's web budget, which is a separate decision, recorded in the helper's docstring rather than smuggled in here.

## Measured: it works, and the synthesis path eats the benefit

Arm G — repositories lane, same questions/judge/bounds as the recorded arms, `--max-queries 3 --max-iterations 1`:

| | Arm B (gate context only) | **Arm G (retrieval too)** |
|---|---|---|
| evidence **admitted** by the gate | 32 | **50** |
| evidence rejected | 21 | 28 |
| `gate_pass_rate` | 0.379 | 0.335 |
| resolved citation markers | 70 | **37** |
| chunk operations / failures | 1 / 0 | 8 / **6** |

**Retrieval-side fan-out works: +56% admitted evidence.** The falling `gate_pass_rate` is a ratio behaving as a ratio — more candidates judged, proportionally more marginal — which is exactly why the absolute admitted count is the number that answers the question.

**But it doesn't reach the report.** Markers *halved* while admitted evidence grew by half again. Not the gate: arm G is the first fan-out arm large enough to trigger map-reduce chunking, and 6 of 8 chunk summarizations failed with `No choices in response data` (**TASK-17384**). The synthesis got fallback text for most chunks and cited a fraction of what was admitted.

That's the third time in this program a retrieval improvement has been masked by the summarization path rather than refuted by the gate, so I raised 17384 to **high** — with multi-hop now shipping on by default, every larger evidence pool hits it.

## Documentation (AC #3)

My own earlier verdict — "fan-out: no measurable benefit" — is now scoped in the baseline doc with a note on every fan-out number: they measured the gate-context half only, so fan-out's retrieval value was **untested, not disproven**. The `search_enable_subquery` config comment says the same. Leaving that unqualified would have let a future reader cite my measurement as evidence against a mechanism it never exercised.

## Verification

- Four new tests: facets reach the providers, the query cap holds, a tight `max_searches` caps fan-out at one extra query, and a facet identical to the question isn't searched twice (matching what the pipeline already does).
- **289 passed** across `Tests/Research`, `Tests/UI/test_research_screen.py`, `Tests/Helper_Scripts`; `ruff` clean.
- Live arm above, on the same judge and question set as every recorded number.

---
## Summary by cubic
Let generated sub-questions drive academic-provider searches, not just web searches, so fan-out changes retrieval as well as judging. Old behavior: academic lane searched only `round_queries` (usually just `[question]`). New behavior: it also searches generated facets, capped and budgeted; budgeting now settles only attempted extra calls.

- Adds `_academic_queries`: builds `[round_queries + facets]`, case-insensitive/trim deduped, capped by `search_default_max_queries` (centralized as `DEFAULT_MAX_QUERIES`). Base `round_queries` remain uncounted to avoid shrinking current web budgets.
- Budgeting fix: extra academic facet queries are reserved at planning and settled after provider calls; unused reservations are released, preventing inflated spend on partial failures. A tight `max_searches` cannot be exceeded.
- Academic lane reads the same resolved per-run params as the web lane. Tests cover provider reach, cap adherence, ledger reserve/settle, dedup, and unusable-cap fallback; 294 passing.
- Docs: baseline now scopes earlier fan-out numbers to gate-only and records +56% admitted evidence with retrieval fan-out; `config.py` clarifies `search_enable_subquery` affects retrieval and judging and that facets cost two provider calls (web + academic). Satisfies `TASK-17372`; `TASK-17384` raised to high priority.

<sup>Written for commit b0f0c7913e60533f9bbebac06622b4b855b10fa7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1772?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/Development/research-report-eval-baseline.md
+++ b/Docs/Development/research-report-eval-baseline.md
@@ -432,6 +432,42 @@ endpoint (task-17384). Both degrade to real source text rather than to
 nonsense, so the default is safe -- but a local-model install will feel it as
 latency, not as better summaries.
 
+### Fan-out WITH retrieval (arm G, 2026-08-17, task-17372)
+
+The repositories lane re-run after the academic lane began searching generated
+facets -- same questions, judge, engine and 5-results bound as every arm above,
+`--max-queries 3 --max-iterations 1`:
+
+| | Arm B: fan-out, gate context only | Arm G: fan-out reaching the providers |
+|---|---|---|
+| evidence ADMITTED by the gate | 32 | **50** |
+| evidence rejected | 21 | 28 |
+| `gate_pass_rate` (mean of runs) | 0.379 | 0.335 |
+| resolved citation markers | 70 | **37** |
+| `citation_accuracy` | 1.00 | 1.00 |
+| `claim_support_rate` | 1.00 | 1.00 |
+| map-reduce chunk operations | 1 | 8 |
+| chunk summarization failures | 0 | **6** |
+
+**Retrieval-side fan-out works: 56% more admitted evidence (32 -> 50).** The
+falling `gate_pass_rate` is the ratio behaving as a ratio -- more candidates
+judged, proportionally more of them marginal -- which is why the absolute
+admitted count is the number that answers the question.
+
+**And the benefit does not reach the report.** Resolved markers HALVED (70 ->
+37) while admitted evidence grew by half again. The cause is not the gate: arm G
+is the first fan-out arm large enough to trigger map-reduce chunking (8 chunk
+operations against arm B's 1), and 6 of those chunk summarizations failed with
+"No choices in response data" (task-17384). The synthesis was handed fallback
+text for most of its chunks, so it cited a fraction of what the gate had
+admitted.
+
+This is the same shape as arm C's Q1, and the third time in this program that a
+retrieval improvement has been masked by the summarization path rather than
+refuted by the gate. task-17384 is therefore on the critical path for BOTH
+decomposition mechanisms now that multi-hop ships on by default: the bigger the
+evidence pool either one produces, the more of it chunking swallows.
+
 ### Reading gate_pass_rate
 
 It is a ratio, and multi-hop adds to its denominator. Pulling in more marginal
--- a/Docs/Development/research-report-eval-baseline.md
+++ b/Docs/Development/research-report-eval-baseline.md
@@ -198,10 +198,12 @@ generalizes:
   as an empty list and this one rendered real facets. That is a genuine
   change in what the judge was asked.
 - **Not tested:** retrieval. The web lane returned zero results throughout
-  (the DuckDuckGo bot challenge this doc notes elsewhere), and the academic
-  lane only searches round 1's `[question]` (task-17372), so the repository
-  records under judgment were the SAME set as the 0.42 arm. Fan-out changed
-  how they were judged, not which ones existed.
+  (the DuckDuckGo bot challenge this doc notes elsewhere), and at the time the
+  academic lane only searched round 1's `[question]`, so the repository records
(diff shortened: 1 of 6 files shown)


rmusser01:
All four findings addressed in `$(git -C /Users/macbook-dev/Documents/GitHub/tldw_chatbook/.worktrees/research-shipped-decomp rev-parse --short HEAD)`.

## ✅ Search spend settled prematurely (correct, and the important one)

`_academic_queries` reserved **and settled** every selected facet at planning time, so a run failing part-way through the lane recorded later, unattempted searches as spent — an inflated budget artifact on exactly the runs whose budget matters most, and a straight violation of the reserve-before/settle-after contract the web lane follows.

It now reserves only and returns the reservation count; the caller settles what it actually attempted and releases the remainder. An attempted provider call counts as spent whether or not it returned, which matches how `_collect_round` settles executed searches.

## ✅ Fan-out cost is understated

Also correct, and worse than a doc nit: with both lanes active each facet costs **two** provider calls, so operators were told roughly half the cost of a setting that spends their money. The comment now states `2 × (cap − 1)` extra calls per round, plus the per-result relevance and summarization calls each returned source needs.

## ✅ Query-cap default duplicated

Now `DEFAULT_MAX_QUERIES`, so the two lanes cannot drift to different caps.

## ✅ `_academic_queries` lacks unit coverage

Fair — it was only exercised through full runs, so its cap, dedup and budget arithmetic were inferred from end-to-end behaviour. Five direct tests now pin: primaries plus facets, reserve-without-settle, the remaining-budget stop, dedup across primaries and facets in mixed case, and the fallback when the configured cap is unusable.

## Verification

**294 passed** across `Tests/Research`, `Tests/UI/test_research_screen.py` and `Tests/Helper_Scripts`; `ruff` clean. Rebased on `dev` at `05802f577`.

rmusser01:
Merging. Record: Qodo's review posted at 22:25Z with 4 findings, all addressed — the premature-settle reliability bug and the understated fan-out cost fixed, the duplicated cap default extracted, and five direct unit tests added for the helper. I requested a re-review of the fix commit and none arrived within the window, so **the fix commit is not Qodo-reviewed**; it rests on local verification (321 passed across the research, window, recorder and pipeline suites; ruff clean) and on implementing the review's own stated remedies.

PR #1774, opened 2026-08-17
fix(llm): give summarization the token budget its caller configured

rmusser01:
Closes **TASK-17384**. The cause is captured, not inferred — and **the cause I wrote into that task when I filed it was wrong**, so this PR corrects the record as well as the code.

## What it isn't

I filed 17384 guessing "an error body returned for an oversized prompt." Disproven:

- the chunk packer caps chunks at **6000 chars** (`_build_chunk_infos`) against a server with a 64k context
- probes at 2,000 and 20,000 chars returned **HTTP 200 with `choices` present**

`"No choices in response data"` was my own parser's fallthrough on **empty content**, never a missing `choices` array.

## What it is

A faithful reproduction — the real 6000-char packed chunk, the verbatim chunk prompt, `max_tokens=4096` exactly as the summarizer resolved it:

```
finish_reason='stop'   content_len=465   reasoning_len=13209
usage: completion_tokens=4028   (of max_tokens=4096)
```

The model spends its completion budget on `reasoning_content` and emits content at the very edge of it. A chunk that reasons slightly longer returns **nothing** — which is why 6 of 8 chunks failed in TASK-17372's arm while per-result calls in the same run succeeded. It's variance at a razor-thin margin, not a size threshold.

Measured headroom, same chunk, two budgets:

| `max_tokens` | reasoning tokens | content | headroom |
|---|---|---|---|
| 4096 | 4028 | 465 chars | **68 tokens** |
| 16384 | 3617 | 564 chars | **12,767 tokens** |

**An earlier probe of mine appeared to refute this hypothesis** — it used a small input and spent 168 completion tokens, 24× less than a real chunk. A token-budget hypothesis cannot be tested on an input that doesn't provoke the reasoning. That's recorded in the task notes, because it's the kind of mistake that repeats.

## The fix

Same root cause as the other three defects in this function: **reading configuration from a place the rest of the app stopped using.** `max_tokens` now prefers the modern `api_settings.llama_cpp` entry before legacy `llama_cpp_api` — a run priming a local endpoint sets 16384 there, while the legacy default pinned summarization to 4096 with the chat path on four times that.

The failure message now names the actual cause: finish reason, tokens spent against the budget, and whether the completion was reasoning-only. The `logging.error` text stays **verbatim** — it's tracked in the reviewed diagnostic inventory (task-492/3750), and this changes the *returned* value a caller surfaces in run warnings. The `"no choices in response"` prefix is preserved so the deep-search failure detector still catches it.

## Verification

- Three new tests: modern budget preferred, legacy fallback intact, and an empty reasoning-only completion reported with its real cause.
- **No governed-diagnostic churn**: the privacy suite fails exactly the 3 manifest-boundary tests clean `dev` fails (3 failed / 254 passed on both sides).
- **390 passed** across `Tests/Web_Scraping`, `Tests/Research` and the summarizer config tests; `ruff` clean.
- A live arm is running now that reproduces TASK-17372's conditions **with** this fix, to test whether the markers that arm lost (70 → 37, while admitted evidence rose 32 → 50) come back. I'll post the numbers here either way.

---
## Summary by cubic
Give chunk summarization the caller’s configured token budget and explain empty content when the model spends the budget on reasoning. Previously it always used a 4096-token limit and returned “Llama: No choices in response data” without context.

- Prefer `api_settings.llama_cpp.max_tokens` over legacy `llama_cpp_api.max_tokens`; validate and fall back to `DEFAULT_SUMMARY_MAX_TOKENS` (4096).
- Returned message now appends `finish_reason`, `completion_tokens/max_tokens`, and a reasoning‑only marker; keep the “Llama: No choices in response data” prefix and the unchanged `logging.error` line for detectors.
- Tests: modern-budget preference, legacy fallback, and diagnostic details; new integration test through `analyze` verifies budget propagation and reasoning‑only failure at the public seam.
- Docs/backlog: update research baseline; add TASK-17386 for synthesis timeouts exposed by the fix.
- Migration: set `api_settings.llama_cpp.max_tokens` (e.g., 16384) to align chunk summarization with chat-path budgets.

<sup>Written for commit 89e0fe2f857a5a7649bcb57b70538a8588de8262. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1774?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/Development/research-report-eval-baseline.md
+++ b/Docs/Development/research-report-eval-baseline.md
@@ -468,6 +468,39 @@ refuted by the gate. task-17384 is therefore on the critical path for BOTH
 decomposition mechanisms now that multi-hop ships on by default: the bigger the
 evidence pool either one produces, the more of it chunking swallows.
 
+### The chunk fix, and where the bottleneck went next (arm H, task-17384)
+
+Arm G's conditions repeated with map-reduce chunk summarization fixed (the
+summarizer was pinned to a 4096 token budget while the chat path ran on 16384,
+and the model spends ~3.6-4k of that on reasoning):
+
+| | Arm G (chunk summaries failing) | Arm H (fixed) |
+|---|---|---|
+| chunk summarization failures | 6 of 8 | **0 of 2** |
+| evidence admitted | 50 | **66** |
+| resolved citation markers | 37 over 3 runs | 55 over 2 runs (**2.2x per run**) |
+| `cited_sentence_ratio` | 0.63 | **1.00** |
+| `citation_accuracy` | 1.00 | 1.00 |
+| questions scored | 3 of 3 | **2 of 3** |
+
+**The retrieval gain now reaches the report.** Citation density hit 1.00 --
+every sentence in both scored reports carried a marker -- and markers per run
+more than doubled. Chunk summarization stopped failing entirely.
+
+**And the bottleneck moved again, to the synthesis call's wall clock.** The
+third question produced NOTHING: its single synthesis call hit the provider read
+timeout (600s, what the recorder primes) and the run left no report and no
+verification payload, so it is absent from the aggregate rather than scored as a
+failure. Filed as task-17386.
+
+That is the fourth time in this program that fixing one stage has revealed the
+next one as the binding constraint -- gate, then chunk summarization, then
+per-result summarization, now synthesis wall clock. The pattern is worth stating
+plainly: **each measured improvement in what the pipeline RETRIEVES raises the
+load on what it must then SUMMARIZE and WRITE**, and every one of those stages
+had a bound calibrated for the single-query runs this document used to record.
+
+
 ### Reading gate_pass_rate
 
 It is a ratio, and multi-hop adds to its denominator. Pulling in more marginal
--- a/Tests/LLM_Calls/test_llama_summarizer_config.py
+++ b/Tests/LLM_Calls/test_llama_summarizer_config.py
@@ -216,6 +216,8 @@ def test_summarize_with_llama_reports_an_unusable_payload(monkeypatch):
 
 
 def test_summarize_with_llama_prefers_the_modern_max_tokens(monkeypatch, captured_post):
+    """A run priming a local endpoint sets the budget in the modern table; the
+    summarizer must spend it rather than the legacy default."""
     settings = _settings()
     settings["api_settings"]["llama_cpp"]["max_tokens"] = 16384
     monkeypatch.setattr(lib, "load_settings", lambda: settings)
@@ -228,6 +230,7 @@ def test_summarize_with_llama_prefers_the_modern_max_tokens(monkeypatch, capture
 def test_summarize_with_llama_falls_back_to_the_legacy_max_tokens(
     monkeypatch, captured_post
 ):
+    """With no modern entry the historical key still governs the budget."""
     monkeypatch.setattr(lib, "load_settings", lambda: _settings())
 
     lib.summarize_with_llama("text", "Summarize.", api_key=None)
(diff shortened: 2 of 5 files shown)


rmusser01:
Arm H is in — the live validation promised above, reproducing TASK-17372's conditions **with** this fix.

| | Arm G (before) | **Arm H (with this fix)** |
|---|---|---|
| chunk summarization failures | 6 of 8 | **0 of 2** |
| evidence admitted by the gate | 50 | **66** |
| resolved citation markers | 37 over 3 runs | 55 over 2 runs (**2.2× per run**) |
| `cited_sentence_ratio` | 0.63 | **1.00** |
| `citation_accuracy` | 1.00 | 1.00 |
| questions scored | 3 of 3 | **2 of 3** |

**The fix does what it claims.** Chunk summarization stopped failing entirely, citation density reached 1.00 — every sentence in both scored reports carried a marker — and markers per run more than doubled. The evidence that fan-out gained now actually reaches the report, which is the whole point.

**And I have to report the other half: only 2 of 3 questions scored.** The third produced nothing at all — its single synthesis call hit the provider read timeout (`ReadTimeoutError ... read timeout=600` on `/v1/chat/completions`) and the run left no report and no verification payload, so it is *absent* from the aggregate rather than scored as a failure. That is not caused by this change, but it is made more likely by it: successful chunk summaries lengthen the reduce step, and the pool is larger than any earlier arm's.

Filed as **TASK-17386** (high), because a run that silently produces nothing is worse than one that fails legibly, and larger pools are now the default rather than the exception.

This is the fourth time in this program that fixing one stage revealed the next as the binding constraint — gate, then chunk summarization, then per-result summarization, now synthesis wall clock. Recorded in the baseline doc as a pattern rather than four unrelated incidents: every measured improvement in what the pipeline *retrieves* raises the load on what it must then *summarize* and *write*, and each of those stages carries a bound calibrated for the single-query runs the document used to record.

rmusser01:
Three of five accepted, two declined with evidence.

## ✅ Token default remains magic
Now `DEFAULT_SUMMARY_MAX_TOKENS`, documented with the measurement that makes it interesting: a real 6000-char chunk spent **4028 of those 4096 tokens on reasoning** before emitting content, so the default is tight rather than arbitrary.

## ✅ New tests lack docstrings
Added.

## ✅ Integration coverage is missing
Fair. A test now drives **`analyze`** — the seam the deep-search pipeline actually calls — instead of `summarize_with_llama` directly, asserting both that the modern budget reaches the request and that a reasoning-only completion surfaces as a failure the caller's detector recognizes.

## ❌ max_tokens ignores environment overrides
The rule cites `env → config.toml → defaults`, but the premise doesn't hold here. Measured on this repo:

- the loader never env-overrides `max_tokens` for **any** provider
- the chat path (`LLM_API_Calls_Local`), which reads the *same* config, has no env read for it either

So this asks me to invent an env convention in one summarizer that exists nowhere else, which would make the summarizer and the chat handler disagree about the same setting — the exact class of split that caused TASK-17382 and TASK-17384 in the first place. If the convention is wanted, it belongs in the loader for all providers at once; happy to file that.

## ❌ Response bypasses Pydantic validation
Also inconsistent with the surrounding code:

- **zero** `BaseModel` usage in either summarization lib
- **13** raw-dict response reads in this module alone

Introducing a Pydantic response model on this one branch would leave eight sibling summarizers parsing raw dicts and one not. The parsing added here is defensive (`isinstance` guards at every level) and the malformed-payload case is covered by a test that asserts the caller's detector catches the result. A response-model refactor across the summarizer family is a reasonable task; it isn't this PR.

## Verification
**249 passed** across the summarizer config, pipeline and research suites; `ruff` clean. Rebased on `dev` at the post-#1772 tip, with two "keep both sections" conflict resolutions in the shared task file and eval doc.

rmusser01:
Merging. Record: Qodo posted 5 rule findings at 22:44Z — three accepted (named constant, docstrings, integration test at the `analyze` seam) and two declined with measured evidence (no env-override convention exists for max_tokens in the loader or the chat path; zero Pydantic models against 13 raw-dict response reads in this module). Re-review requested, none arrived, so **the fix commit is not Qodo-reviewed**; it rests on local verification (249 passed across the summarizer, pipeline and research suites; ruff clean).

PR #1775, opened 2026-08-17
cross_encoder: implement, measure, and say what the measurement actually showed (TASK-16965)

rmusser01:
## The question this arc answers

Does reranking help retrieval in this repo? It has never been measured. TASK-3502 declared it out of scope for a good reason — the three shipped strategies are LLM-driven, and the gated instrument is local and deterministic. A **local cross-encoder** is the one strategy the instrument can see, so this arc implemented it and measured it.

**`sentence-transformers` was already a dependency**, so this cost a new model artifact, not a new library.

## The census came first, and it is what makes the answer interpretable

Before implementing anything:

| mode | queries returning ≥2 rows |
|---|---|
| `plain` | **0 / 60** (39 return nothing, 21 return exactly one) |
| `semantic` | **60 / 60** — full window |
| `hybrid` | **60 / 60** — full window |

Reranking is an order-only change, so `plain` is provably the identity there (a moved plain cell would have been a stop, not a win — and the probe's guard confirms it moved nothing). But semantic and hybrid hand a reranker a full list on every query, so a null could not be dismissed as "nothing to reorder".

## The verdict: HARMED — against a rule fixed before any code was written

| mode | MRR | NDCG |
|---|---|---|
| semantic | 0.804 → 0.772 | 0.804 → 0.780 |
| hybrid | 0.809 → 0.798 | 0.817 → 0.810 |

The strategy demonstrably ran: 3,621 rows scored, 0 failed, **1,950 rows moved**. Regressions beyond tolerance land on `paraphrase` (MRR 1.000 → 0.923) and `vocabulary_mismatch` (1.000 → 0.944).

**But the headline is narrower than it sounds, and the final review caught it.** That averaged row is the instrument's `overall`, which *excludes* `scoped` and `negative` — and `scoped` is exactly where the strategy wins. Averaged over all 53 ground-truthed queries, **hybrid reverses sign: MRR 0.731 → 0.806 (+0.075)**. Every headline site now says which average it means. The pre-registered verdict is unchanged — it is computed from the instrument's own cells and the six regression cells stand — but a reader deserves both numbers.

**The result is strongly bimodal**: hybrid `scoped` MRR **0.163 → 0.929**, `prompt` 0.022 → 0.200, semantic `negation` NDCG 0.000 → 0.105 — paid for by demoting rank-1 answers in the only two categories already saturated at 1.000, where the only available movement was down.

## Owner ruling: keep the code, retire the promise

Asked with the trade-offs shown, the owner ruled (2026-08-17) to keep `CrossEncoderReranker` implemented and selectable while removing every claim that it helps. Implemented faithfully: it is the default, recommendation and profile-strategy of **nothing**, and eight user-visible surfaces now carry the measured result instead of a promise — including `cross_encoder`'s entry under *"Future Enhancements → Planned Features"* in the design doc, and its role as the **example value** in both `config.py` and the shipped example TOML. Per this programme's convention, every site records two adjacent facts: what the rule said, and that the owner ruled otherwise.

Hybrid Full stays on `pointwise` — now documented as a permanent, measured choice rather than a stopgap.

## Two findings about the evidence itself

**The measurement was one monkeypatch away from a fabricated null.** `Tests/conftest.py` sandboxes `HOME`, and huggingface's cache constant is computed from `expanduser("~")` at import — so under pytest the model fails to load *even though it is cached*. Because the reranker degrades rather than raises, every window would have returned unchanged and every metric would have read a flawless 0.000 delta: a perfect-looking null, entirely manufactured. The probe now asserts it did work (`rows_scored > 0`, `rows_failed == 0`), and the review **mutation-verified** it: forcing the load to fail produces the pretty NULL and then refuses it with `3621 cross-encoder scoring attempts FAILED`. The lesson is filed — a measurement whose subject degrades silently must assert, inside itself, that work happened, or its null is unfalsifiable.

**Is the harm an artifact of min-max normalisation?** The sharpest question in the arc, and the review ran the control rather than reasoning about it: with the original-score blend off — ordering then identical to raw logits — the verdict is still HARMED with the **same six regression cells** and deltas matching to three decimals; a 10×-scaled blend agrees. **The harm is the model's ranking, not the normalisation.** (Strictly, ordering is only normalisation-invariant with the blend off; the shipped config blends 30%, and the docstring now says so.)

## Verification

Probe reproduces bit-exactly (225/225 lines vs the committed artifact); the rule's code form is proven pre-registered from git history (byte-identical since before Task 1); Task 3's diff is **zero executable lines**, proven by AST comparison; gate `PASSED: No regression. 105 metric(s)` with no baseline re-stamped; 678 passed / 26 skipped; **zero network, zero provider spend**, confirmed three ways.

## Follow-up filed

**TASK-17600** — `rag_pipelines.toml` declares a `result_reranking` middleware with `enabled = true` whose handler is a bare `pass`, listed by the `high_accuracy` pipeline. Same species as the inert parent-inclusion knobs, worse in one respect: it appears in a pipeline a user picks *because* it promises accuracy. The review's F3 (`reranking_strategy` has zero readers repo-wide) is folded into its scope.

Refs TASK-16965.

---
## Summary by cubic
Implements a local `cross_encoder` reranking strategy with `sentence-transformers`, measures it on the offline gate, and updates docs/configs to state the result. Old: `cross_encoder` unimplemented; new: implemented and selectable but not recommended. Verdict: on the instrument’s averaged row (excludes `scoped`/`negative`) semantic/hybrid regress; over all 53 ground‑truthed queries hybrid reverses sign (MRR 0.731 → 0.806); strategy kept to enable measurement, defaults remain `pointwise`.

- Adds `CrossEncoderReranker` (lazy, cached local model; offline-only; shared apply‑scores/logging in `BaseReranker`; preserves `rerank_score`; failures degrade, not raise). Updates `RAG_Search/reranker.py`, `config_profiles.py`, `library_rag_score_kinds.py`, example/config docs; moves example strategy to `pointwise`.
- Gated probe and tests (`Tests/RAG_Eval`) run fully offline and assert work occurred; result is bimodal: large gains on weak‑retrieval categories (e.g., hybrid `scoped` MRR 0.163 → 0.929) with regressions on already‑saturated categories; HARMED verdict stands.
- Follow‑ups: order‑ and content‑sensitive `_cross_encoder_cache_key` (copies on replay), `local_files_only=True` and optional import routing, robust model path handling, `batch_size` forwarded to `model.predict` (min 1), `timeout_seconds` guards await (timeouts degrade and don’t cache). Re‑run reproduces published deltas exactly.
- Docs: remove “future enhancement” claim; user guide records measured outcomes; mark `result_reranking` middleware as a no‑op; file TASK‑17600.

**Rollout and migration**
- No default changes. If you set `reranking_strategy = "cross_encoder"`, ensure the model artifact is in the HF cache; otherwise reranking degrades to pass‑through.
- `batch_size` and `timeout_seconds` now take effect; timeouts skip caching.
- Do not rely on `result_reranking` middleware; it is currently a no‑op.

<sup>Written for commit 8b361829c67f7195cd5c9c000684c679d0a0d92c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1775?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/Development/RAG/RAG-DESIGN.md
+++ b/Docs/Development/RAG/RAG-DESIGN.md
@@ -503,7 +503,9 @@ Batch processing utilities for efficient indexing.
    - Merge overlapping content
 
 3. **Re-ranking**
-   - Apply cross-encoder models (if enabled)
+   - Apply cross-encoder models (if enabled -- off by default, and measured
+     net-harmful on average; see the note under "Re-ranking" in Future
+     Enhancements)
    - Score based on expanded context
    - Filter by relevance thresholds
 
@@ -2357,7 +2359,18 @@ Future support for domain-specific step types:
 
 1. **Advanced Search**
    - Query expansion with LLMs
-   - Re-ranking with cross-encoders
+   - ~~Re-ranking with cross-encoders~~ — **done and measured, not
+     planned.** TASK-16965 shipped `CrossEncoderReranker` (local
+     sentence-transformers, no provider/credential/network) and ran it over
+     the 60-query golden set on the gated instrument against a rule fixed
+     before the run. It came out **net harmful** on the averaged row
+     (semantic MRR 0.808 → 0.762, hybrid 0.812 → 0.787 at k=10; recall@10
+     +0.022) and **strongly bimodal**: hybrid `scoped` MRR 0.163 → 0.929
+     and `prompt` 0.022 → 0.200, paid for by demoting already-rank-1
+     answers in `paraphrase`/`vocabulary_mismatch` (1.000 → 0.87–0.94).
+     It stays selectable as `reranking_strategy = "cross_encoder"` and is
+     the default of nothing. Numbers:
+     `Docs/superpowers/qa/2026-08-17-cross-encoder/report.md`.
    - Multi-lingual support
 
 2. **Scalability**
--- a/Docs/Development/RAG/RAG-DESIGN.md
+++ b/Docs/Development/RAG/RAG-DESIGN.md
@@ -504,7 +504,7 @@ Batch processing utilities for efficient indexing.
 
 3. **Re-ranking**
    - Apply cross-encoder models (if enabled -- off by default, and measured
-     net-harmful on average; see the note under "Re-ranking" in Future
+     net-harmful on average [CAVEAT: that averaged row EXCLUDES `scoped` and `negative` (`UNAVERAGED_CATEGORIES`), and `scoped` is where this strategy WINS -- over all 53 ground-truthed queries hybrid REVERSES sign (MRR 0.731 -> 0.806, +0.075). TASK-16965 final review F1.]; see the note under "Re-ranking" in Future
      Enhancements)
    - Score based on expanded context
    - Filter by relevance thresholds
--- a/Docs/Development/RAG/RAG-Documentation.md
+++ b/Docs/Development/RAG/RAG-Documentation.md
@@ -818,7 +818,7 @@ max_expansions = 3
 
 Reorder results after retrieval. **This is not a quality win by default.**
 The one strategy that has been measured here (`cross_encoder`, TASK-16965)
-came out net harmful on the averaged row of the gated eval set — big gains
+came out net harmful on the averaged row [CAVEAT: that averaged row EXCLUDES `scoped` and `negative` (`UNAVERAGED_CATEGORIES`), and `scoped` is where this strategy WINS -- over all 53 ground-truthed queries hybrid REVERSES sign (MRR 0.731 -> 0.806, +0.075). TASK-16965 final review F1.] of the gated eval set — big gains
 on weak-retrieval query categories, losses on categories retrieval already
 got right — and the three LLM-driven strategies have never been measured at
 all (they bill a provider per call, which the local gate cannot run). Turn
--- a/Docs/Development/RAG/RAG-Documentation.md
+++ b/Docs/Development/RAG/RAG-Documentation.md
@@ -816,7 +816,13 @@ max_expansions = 3
 
 ### 3. Re-ranking
 
-Improve result quality with re-ranking:
+Reorder results after retrieval. **This is not a quality win by default.**
+The one strategy that has been measured here (`cross_encoder`, TASK-16965)
+came out net harmful on the averaged row of the gated eval set — big gains
+on weak-retrieval query categories, losses on categories retrieval already
+got right — and the three LLM-driven strategies have never been measured at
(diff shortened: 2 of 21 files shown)


rmusser01:
All eight findings addressed in `0389fe2eb` — and finding 1 raised a question that outranked the fixes, so it was answered empirically first.

**Did the cache bug taint the published measurement?** It could not be settled by reasoning: `cache_results` defaults to `True`, so the cache *was* live during the probe run. Re-running the probe against the fixed key reproduces every published number exactly — semantic MRR 0.804→0.772, hybrid 0.809→0.798, arm B semantic 0.808→0.762, the same six category cells (paraphrase −0.077, vocabulary_mismatch −0.056, scoped +0.766, prompt +0.178), and `VERDICT: HARMED`. **The verdict stands, now confirmed post-fix.**

**Finding 1** — `_cross_encoder_cache_key` is order-sensitive (the cached `RerankingResult`s are positional and zipped back by index, so an order-insensitive key replayed scores onto the wrong rows), and it keys on the retrieval scores *and* a digest of the document text: the 0.3 original-score blend consumes the former, and an id whose text changed is a different scoring problem rather than a hit. Replays are copied, not aliased.

**Findings 7 and 8 deserve a word,** because they are the sin this arc spent its life naming: `batch_size` and `timeout_seconds` were exposed knobs reaching nothing, shipped by the same arc whose follow-up (TASK-17600) exists to end exactly that — and the class docstring *enumerates* which fields are no-ops, which actively implied these were honoured. `batch_size` now reaches `model.predict` floored to 1; `timeout_seconds` bounds the await via `asyncio.wait_for` (the executor thread can't be cancelled, but the await is what a search's latency depends on), degrading without claiming a rerank score and without caching a transient failure.

**Findings 2, 3, 6:** `local_files_only=True` so the documented no-network promise fails fast instead of silently fetching; the optional import routes through `Utils/optional_deps.py`; a bare directory or Windows path is no longer silently swapped for the default artifact. **4, 5:** Google-style `Args:`/`Returns:` and a typed `**kwargs`.

19 cross-encoder tests pass; `Tests/RAG_Search/` + `Tests/RAG_Eval/` 691 passed / 26 skipped; gate `PASSED: No regression. 105 metric(s)`; ruff clean.

PR #1782, opened 2026-08-18
fix(research): a synthesis that never returns must not vanish

rmusser01:
Closes **TASK-17386**'s legibility and measurement criteria (#1, #2, #4). **AC #3 is deliberately left open** — see below.

Stacked on **#1774**, so its two commits appear here too; merging that first reduces this to the synthesis change and its records.

## The failure that hid itself

A run whose single synthesis call exceeded the provider timeout returned a generic `"Could not create the report due to an error"`, **completed normally**, and carried no citation verdict. The eval, finding no payload, left the sample out of the aggregate entirely.

That's the part worth dwelling on: metrics computed over surviving runs looked *better* precisely because the failures removed themselves from the sample. Arm H scored 2 of 3 questions and reported `cited_sentence_ratio` 1.00 — a number that would have been quietly averaged over the survivors of exactly this failure.

## Measured, not assumed (AC #2)

| arm | admitted (cumulative) | synthesis |
|---|---|---|
| B | 14 / 26 / 32 | 241s / 328s / 287s |
| G | 22 / 38 / 50 | 185s / 247s / 282s |
| H | 22 / 46 / 66 | 155s / **1200s FAILED** / 970s |

The two large numbers decode the mechanism: against a 600s per-attempt timeout with one retry, **1200s is two timeouts** (the log ends in `MaxRetryError`) and the 970s "success" is a first attempt timing out and the retry landing. Synthesis on a default multi-hop pool routinely exceeds the per-attempt budget — retries hide it sometimes and lose the whole run other times.

## What this changes

`aggregate_results` records the failure **class** — never its message, since a timeout's text carries `host:port` and this value travels into a run's artifacts — along with the pool size it failed on. The engine appends the warning *while the round's warning list is still being built*, so the reason reaches the run's warnings and bundle rather than only a summary written afterwards. `FinalAnswerDict` declares the field rather than growing an undocumented key.

## What this does NOT change (AC #3)

Making a large-pool synthesis actually *complete*. That needs a size-aware budget, and the per-call timeout isn't plumbable through `chat_api_call` — its signature is shared by ~9 providers, so threading one through is a change to the shared dispatcher, not to this pipeline. It belongs in its own task, and the task stays open rather than being marked Done on a partial fix.

The honest interim answer is documented beside the measurement: an install running default multi-hop against a local model needs a provider `api_timeout` well above the 600s these arms used — the 66-source pool took 970s of successful synthesis.

## Verification

- New engine test: a run whose synthesis failed records `synthesis_failed` in its verification summary **and** a warning naming the stage, pool size and failure class.
- An existing contract test pinned the failure answer's exact key set; updated to preserve its intent (typed shape, no `summary` key ever) while asserting the new reason.
- **227 passed** across `Tests/Research` and `Tests/Web_Scraping/test_deep_search_pipeline.py`; `ruff` clean.

---
## Summary by cubic
Makes synthesis timeouts visible and counts runs that produced no report so they no longer vanish from aggregates. Also fixes stage attribution so post-synthesis verification errors are labeled “verification,” not “synthesis.”

- `WebSearch_APIs.aggregate_results` returns `synthesis_failed` with `stage`, `error_type`, `evidence_count`, and `chunk_count`, and appends the failure class to `text`; stage is tracked so verification defects are labeled correctly.
- `FinalAnswerDict` declares `synthesis_failed`.
- `local_research_engine` adds a warning during execution and writes `synthesis_failed` (and warnings) into `verification_summary.json`.
- `record_research_baseline`: runs without `citation_verification` return an `{"__unscored__": ...}` marker, and the final aggregate includes `unscored_runs.count` and `unscored_runs.reasons` so missing reports are counted, not dropped.
- Tests assert the new field and warning, correct stage labeling, and that aggregates state unscored runs; docs record synthesis wall clock vs. pool size. Connected to TASK-17386: closes AC #1, #2, and #4; AC #3 (size‑aware budget) remains open.

**Rollout**
- No code migration required. Callers may read `final_answer.synthesis_failed` and aggregate `unscored_runs` to detect failures explicitly.
- For local models with large evidence pools, set a higher provider `api_timeout` until a size‑aware synthesis budget is implemented.

<sup>Written for commit 865f47819e734376133669934150304b70e6e464. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1782?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/Development/research-report-eval-baseline.md
+++ b/Docs/Development/research-report-eval-baseline.md
@@ -501,6 +501,41 @@ load on what it must then SUMMARIZE and WRITE**, and every one of those stages
 had a bound calibrated for the single-query runs this document used to record.
 
 
+
+### Synthesis wall clock against pool size (task-17386)
+
+Measured from the arms above, pairing each question's admitted-evidence count
+with the wall clock of its single synthesis call:
+
+| arm | admitted (cumulative) | synthesis |
+|---|---|---|
+| B | 14 / 26 / 32 | 241s / 328s / 287s |
+| G | 22 / 38 / 50 | 185s / 247s / 282s |
+| H | 22 / 46 / 66 | 155s / **1200s FAILED** / 970s |
+
+The two large numbers are the shape of the failure. With a 600s per-attempt
+provider timeout and one retry, **1200s is two timeouts** (the run's log ends in
+`MaxRetryError`), and the 970s "success" is a first attempt timing out and the
+retry landing. So synthesis on a default multi-hop pool routinely exceeds the
+per-attempt budget: retries hide it sometimes and lose the entire run other
+times.
+
+A lost run used to be invisible. The pipeline returned a generic "Could not
+create the report due to an error" with no verdict, the run COMPLETED, and the
+eval -- finding no citation payload -- simply left the sample out of the
+aggregate. A metric computed over surviving runs looks better precisely because
+the failures removed themselves from the sample. Runs now carry
+`synthesis_failed` (the failure class and the pool size it failed on) in their
+warnings and verification summary, so the stage that failed is nameable.
+
+**Not fixed here**: making a large-pool synthesis actually complete. The
+per-call timeout is not plumbable through `chat_api_call`, whose signature is
+shared by ~9 providers, so a size-aware synthesis budget is its own change.
+Until then, an install running default multi-hop against a local model needs a
+provider `api_timeout` well above the 600s these measurements used -- the 66
+source pool took 970s of successful synthesis.
+
+
 ### Reading gate_pass_rate
 
 It is a ratio, and multi-hop adds to its denominator. Pulling in more marginal
--- a/Helper_Scripts/Benchmarks/record_research_baseline.py
+++ b/Helper_Scripts/Benchmarks/record_research_baseline.py
@@ -207,14 +207,23 @@ async def _run_question(
     verification = service.get_artifact(run["id"], "verification_summary.json") or {}
     payload = verification.get("content") or {}
     if not payload.get("citation_verification"):
-        print("  [no citation verification on the synthesis branch]")
-        return None
+        # Qodo (PR 1782): returning None here is what made failed runs vanish.
+        # A run with no citation verdict has no metrics to average, but it MUST
+        # still be counted, or the aggregate is computed over survivors and
+        # reads as if every run produced a report (task-17386).
+        failure = payload.get("synthesis_failed") or {}
+        stage = failure.get("stage") or "unknown"
+        error_type = failure.get("error_type") or "no citation verification"
+        print(f"  [no report: {stage} failed -- {error_type}]")
+        return {"__unscored__": {"stage": stage, "error_type": error_type}}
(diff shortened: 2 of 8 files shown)


rmusser01:
Both bugs fixed — and the first one is a fair hit on a fix that claimed to solve exactly what it left in place.

## ✅ Failed runs still vanish

Correct, and the important one. Recording `synthesis_failed` on the **run** was only half the job: the baseline recorder still returned `None` for any payload without `citation_verification`, so a run that produced no report never reached the aggregate and the means stayed computed over survivors. That is precisely the distortion TASK-17386 exists to stop, and my change left it standing while claiming to fix it.

The recorder now returns an unscored marker instead. Such runs contribute no metrics — they genuinely have none — but they are **counted**, with their failing stage named in the emitted aggregate, so `sample_count` can no longer be mistaken for the number of runs attempted.

## ✅ Verification errors mislabeled synthesis

Also correct. `verify_citations` runs after the provider returns, inside the same `try`, so a citation-verifier defect would have been recorded as a *synthesis* failure — corrupting the failure-class measurement the field was added for. The stage is now tracked as the block progresses and reported as whatever actually failed.

Three tests cover both: a failed-synthesis run is counted rather than dropped, the aggregate states how many runs produced no report, and a verification defect is labelled `verification`.

## ❌ Repeated 4096 fallback literal — already fixed

Stale against the current branch. That literal became `DEFAULT_SUMMARY_MAX_TOKENS` in #1774, which merged into `dev` before this rebase: 4 uses of the constant and a single bare `4096` (its definition). Nothing to change.

## ❌ max_tokens bypasses Pydantic validation

Same decline as #1774, same evidence: **zero** `BaseModel` usage in either summarization lib against **13** raw-dict response reads in this module alone. A Pydantic response model on one branch of one summarizer, with eight siblings parsing raw dicts, is inconsistency rather than validation. Worth doing as a family-wide refactor; not here.

## Verification

**321 passed** across the pipeline, research and recorder suites. `ruff` reports 4 findings in `test_deep_search_pipeline.py` — all pre-existing on clean `dev` (unused imports/variables), none from this change.

rmusser01:
Merging. Record: Qodo posted at 01:59Z with 2 bugs and 2 rule findings; both bugs fixed in the pushed commit (unscored runs now counted rather than dropped, and the failing stage attributed correctly), one rule finding declined as stale (the literal became DEFAULT_SUMMARY_MAX_TOKENS in the already-merged #1774) and one declined on the same measured evidence as #1774 (zero Pydantic models against 13 raw-dict reads in that module). Re-review requested, none arrived, so **the fix commit is not Qodo-reviewed**; it rests on local verification (399 passed across the research, pipeline, recorder and summarizer suites) and on implementing the review's own remedies.

PR #1788, opened 2026-08-18
fix(llm): read Kobold and TabbyAPI config from sections that exist

rmusser01:
Closes **TASK-17383**'s reachable criteria (#3, #4, #5). **AC #1 and #2 stay open**, blocked on a larger defect this work uncovered — filed as **TASK-17387**.

## The fix

Both summarizers indexed `api_keys`, `local_api_ip` and `models` — names the loader has **never** built — so like llama.cpp before TASK-17382 they raised before reaching a server and reported failure by *returning* an error string a caller could store as evidence. Both providers have complete modern `api_settings` entries **and** complete legacy sections; nothing was missing but the lookup.

Two shared helpers now resolve by name rather than by index.

**One subtlety worth calling out**, because the contract tests caught me: credential resolution preserves a distinction the old code had by accident — a key declared as an empty string is *configured but blank* (proceed with no `Authorization` header), while absent or null is *absent* (refuse). My first version collapsed both to `None`, which turned a working blank-credential call into a failure. `api_key_env_var` is honoured because tabbyapi's modern table carries only that field and two UI surfaces already read it — following the repo's convention rather than inventing one.

## The audit (AC #4)

By AST over every `summarize_with_*`:

| summarizer | verdict |
|---|---|
| `local_llm` | no direct config reads |
| `llama` | fixed in TASK-17382 |
| `oobabooga`, `vllm`, `ollama`, `custom_openai`, `custom_openai_2` | read sections that exist |
| **`kobold`, `tabbyapi`** | **read never-built sections — fixed here** |

## What I found and did NOT fix

Both functions are **generator functions** — a top-level `yield` in their streaming branches — so calling either returns a generator and runs *none* of the body. No request reaches either server regardless of configuration, and a caller storing the result keeps a generator object where a summary belongs. Worse for the pipeline: the failure detector inspects strings so a generator passes it, and a generator is truthy so the emptiness guard passes too — it would be stored as evidence.

I prototyped the fix (nest each streaming body in a `_stream_generator`, as `summarize_with_llama` already does). It worked, and it **broke seven tests in the security-reviewed diagnostic ledger suite**: the ledger keys entries on `(file, function, message, ordinal)`, so nesting re-attributes ~20 diagnostics to the nested function. I reverted it rather than land a silently-rewritten governed artifact inside a configuration fix, and filed **TASK-17387** with the evidence and the ledger consequence.

That's why AC #1 and #2 ("a request reaches its server") are honestly still open here.

## Verification

- The diagnostic privacy suite fails **exactly** the 3 manifest-boundary tests clean `dev` fails (3 failed / 254 passed on both sides) — no governed-artifact churn.
- **1106 passed** across `Tests/LLM_Calls`, the pipeline suite and `Tests/Research`; `ruff` clean.
- Six new tests drive both providers with each server's real response envelope (Kobold's native `results[].text`, TabbyAPI's full OpenAI envelope — it validates `id`/`created`/`model`/`object`/`usage` before reading content, so a fake carrying only `choices` proves nothing).

---
## Summary by cubic
Reads Kobold and TabbyAPI settings from real config sections and fixes Kobold streaming to use the OpenAI-compatible endpoint. Old behavior: both looked up `api_keys`/`local_api_ip`/`models` and returned an error string before any request; Kobold streaming hit the native generate URL and produced no chunks. New behavior: endpoints and credentials resolve from `api_settings.koboldcpp`/`api_settings.tabbyapi` with fallback to `kobold_api`/`tabby_api`; Kobold streaming prefers `kobold_api.api_streaming_ip` while non-streaming uses the native generate endpoint.

**Review notes**
- Config resolution: new helpers prefer modern tables, fall back to legacy; supports `api_key_env_var` for Tabby; preserves blank-vs-absent credential semantics; clear error when no API URL is configured; legacy retry settings are used.
- Endpoints: Kobold streaming sends to OpenAI-compatible SSE; non-streaming sends to native `.../api/v1/generate`.
- Tests/tasks: adds coverage for modern/legacy resolution, real response envelopes, streaming vs non-streaming, and an `analyze` integration; diagnostic-privacy fixtures rekeyed to `kobold_api`/`tabby_api`. Satisfies TASK-17383 AC #3–#5; AC #1–#2 remain open (blocked by TASK-17387: both functions are still generator functions).
- Migration: none for runtime configs using `api_settings` or `kobold_api`/`tabby_api`. Update any test fixtures keyed to `api_keys`/`local_api_ip`/`models`.

<sup>Written for commit e3ef9b126d79ee5450231420e3d92d496f09d343. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1788?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/LLM_Calls/test_kobold_tabby_config.py
+++ b/Tests/LLM_Calls/test_kobold_tabby_config.py
@@ -1,0 +1,203 @@
+"""Kobold and TabbyAPI summarizer configuration resolution (task-17383).
+
+NOTE ON SHAPE: both functions are GENERATOR functions -- a top-level `yield` in
+their streaming branches makes the whole function one -- so calling them returns
+a generator and the body runs only when it is iterated. That is a separate,
+larger defect (a non-streaming caller receives a generator where a summary
+belongs, and the deep-search pipeline would store it as evidence); it is filed
+on its own because fixing it re-attributes ~20 diagnostics to nested functions
+and rewrites a security-reviewed ledger. These tests therefore drive the
+functions the way they currently behave, and still prove the configuration
+resolution.
+
+Both indexed sections the loader has never built -- `api_keys`,
+`local_api_ip`, `models` -- so like llama.cpp before task-17382 they raised
+before contacting a server and returned an error STRING that callers could
+store as evidence. Both providers DO have complete modern `api_settings`
+entries and complete legacy sections; these tests pin that they read those.
+"""
+
+import pytest
+
+from tldw_chatbook.LLM_Calls import Local_Summarization_Lib as lib
+
+
+class FakeResponse:
+    """Shapes the SERVERS actually return, not the shape the code wants.
+
+    Kobold's native generate endpoint answers ``{"results": [{"text": ...}]}``;
+    TabbyAPI is OpenAI-compatible and answers with ``choices[].message``. A
+    single fake asserting one shape would have passed while the other provider
+    stayed broken -- the failure mode task-17382 was built on.
+    """
+
+    status_code = 200
+
+    def __init__(self, payload=None):
+        self.text = ""
+        self._payload = payload if payload is not None else {
+            # Kobold native generate
+            "results": [{"text": " SUMMARY "}],
+            # OpenAI-compatible envelope -- TabbyAPI validates every one of
+            # these keys before reading the content, so a fake carrying only
+            # "choices" passes nothing.
+            "id": "cmpl-test",
+            "created": 0,
+            "model": "test-model",
+            "object": "chat.completion",
+            "usage": {"completion_tokens": 3},
+            "choices": [{"message": {"role": "assistant", "content": " SUMMARY "}}],
+        }
+
+    def json(self):
+        return self._payload
+
+    def raise_for_status(self):
+        return None
+
+
+@pytest.fixture
+def captured_post(monkeypatch):
+    captured = {}
+
+    class FakeSession:
+        def mount(self, *_args, **_kwargs):
+            pass
+
+        def post(self, url, headers=None, json=None, stream=False, **_kw):
+            captured["url"] = url
+            captured["headers"] = headers or {}
+            captured["json"] = json
+            return FakeResponse()
+
+    monkeypatch.setattr(lib.requests, "Session", lambda: FakeSession())
+    return captured
+
+
+def drain(result):
+    """Run a summarizer to completion and return its final value."""
+    if isinstance(result, str):
+        return result
+    chunks = []
+    while True:
+        try:
+            chunks.append(next(result))
+        except StopIteration as stop:
+            return stop.value if stop.value is not None else "".join(chunks)
+
+
+def settings(**overrides):
+    """A snapshot shaped like the real loader's: the modern per-provider table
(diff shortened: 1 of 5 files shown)


rmusser01:
Two accepted, one declined with the repo's own rule as evidence.

## ✅ Streaming uses native endpoint — real bug, mine

Verified against the shipped config values rather than reasoned about:

```
modern koboldcpp.api_url : .../api/v1/generate      ← native
legacy api_ip            : .../api/v1/generate      ← native
legacy api_streaming_ip  : .../v1/chat/completions  ← OpenAI-compatible
```

The streaming branch parses OpenAI SSE, so preferring the modern url — which is normally populated — would have sent streaming to the native endpoint and produced nothing. Streaming now prefers `api_streaming_ip`; non-streaming keeps the native precedence its `results[].text` parser requires. Two tests pin the pair, shaped like the real config.

## ✅ Configuration integration coverage missing

Added, driving `analyze` — the seam the deep-search pipeline actually uses. Worth noting what it asserts: the request reaches the configured endpoint, **and** the returned value is still unusable, because these functions are generators (TASK-17387). Pinned that way deliberately, so the test starts telling the truth about the return value the moment 17387 lands rather than asserting a summary the code cannot yet produce.

## ❌ Environment credential cannot override config

The ordering is deliberate and matches this repo's documented rule. From `CLAUDE.md`:

> Priority: env vars → config.toml → defaults — **except provider API keys**, where an explicit `api_settings.<provider>.api_key` now outranks the matching environment variable.

The existing consumer agrees: `MediaWindow_v2` reads the config field first and only then falls back to `os.environ[api_key_env_var]` ("If not found, check environment variable"). Putting env first here would make these two summarizers disagree with both the documented rule and the surface that already implements it — the same divergence class that produced TASK-17382 and TASK-17384.

## Verification

Privacy suite fails **exactly** the 3 manifest-boundary tests clean `dev` fails (no governed-artifact churn); **939 passed** across `Tests/LLM_Calls` and the pipeline suite; `ruff` clean.

rmusser01:
Merging. Record: Qodo posted 1 bug and 2 rule findings; the bug (streaming sent to the native endpoint instead of the OpenAI-compatible one) is fixed with tests pinning both endpoint shapes against the shipped config values, the integration coverage was added at the `analyze` seam, and the env-precedence finding was declined citing CLAUDE.md's own rule that an explicit `api_settings.<provider>.api_key` outranks the environment variable. No re-review arrived, so **the fix commit is not Qodo-reviewed**; it rests on local verification — the privacy suite fails exactly the 3 manifest-boundary tests clean dev fails (3 failed / 254 passed both sides), 939 passed across LLM_Calls and the pipeline suite, ruff clean.

AC #1 and #2 of TASK-17383 remain open by design, blocked on TASK-17387 (both summarizers are generator functions, so no request reaches either server regardless of configuration).

PR #1791, opened 2026-08-18
Clarification gate: NULL, killed by a census before it was built (TASK-16072)

rmusser01:
The fifth P2c premise to die before any production code existed — and the first killed by a **census** rather than a probe.

## The bar was registered before the probe

`Docs/superpowers/qa/2026-08-18-clarification-gate/bar.md`, committed first and unchanged since. A gate needs all three of: more than one defensible interpretation, **pointing at different documents in this corpus**, where a user's answer would **change which document is retrieved**. Bar: ≥5 of 60 queries. Kill: fewer than 5 ends the arc with a recorded null — and reaching the bar by *authoring ambiguity fixtures* was pre-declared a kill, not an invitation.

## Step 0, from the corpus

| | queries |
|---|---|
| more than one relevant document | **2** |
| exactly one relevant document | 51 |
| zero (negative controls) | 7 |

**The two candidates fail the third condition, and badly.** `kw-nimbus-rollback` and `kw-calyx-limiter` each have two relevant documents — *both correct*. A clarifying question there wouldn't choose between competing interpretations; it would ask the user to discard one of two right answers. Firing a gate on them makes retrieval worse.

Qualifying count under all three conditions: **0**, against a bar of 5. Kill condition fires.

**Twelve short queries are not twelve gate cases.** `'plant maintenance record'`, `'asset tag QX-8842'`, `'gum disease treatment'` — all ≤3 words, all with exactly one relevant document. Terse, not ambiguous. That's the distinction the bar existed to force: surface vagueness isn't an ambiguity signal unless the corpus supplies a second thing to mean.

## Why this generalises past the fixture

Every failing class fails for a reason a question cannot repair — `negation` because the corpus asserts what the query excludes; `prompt` and the residual zero-row queries on **absent content words**; `paraphrase`/`vocabulary_mismatch` in `plain` on no shared content word. **No answer to any question puts a missing document into an index.**

## The transferable part is the method, not the verdict

Four earlier premises (`expansion`, `acronym`, `compositional`, PRF) died on measurement. This one died on **one query over the fixture** — counting how many golden queries even have the property the feature needs — for a fraction of PRF's full-probe cost.

So **TASK-18155** (granularity router, the next named candidate) is filed with the census required *first*, and with an extra constraint recorded: `include_parent_docs` was retired as inert by TASK-16174 and `expand_document` now covers that capability, so a router would be a *third* surface over one capability and its census must clear a higher bar than "it might help".

No production code was written. The fifth retired premise is recorded beside the others in `Tests/RAG_Eval/README.md`.

Refs TASK-16072.

---
## Summary by cubic
Retires the “clarification gate” via a corrected corpus-first census and records a NULL verdict. This avoids shipping a gate with no measurable cases and establishes census-first for future P2c candidates.

- Method fix: condition 2 now checks corpus text, not relevance labels. Result: 1/60 queries shows an unlabelled alternative reading (a negation), but 0/60 satisfy all three conditions. Below the pre-registered bar (≥5), so the kill fires.
- Artifacts added: `bar.md` (bar/kill), `step0.md` (corrected census and verdict), `per-query-census.md` (all 60 rows), and `census.py` (includes a probe-proof check that document text was read).
- `Tests/RAG_Eval/README.md`: records the fifth retired premise; notes the two >1-label cases have two correct answers, so a gate would worsen retrieval, not disambiguate.
- Backlog: TASK-16072 marked Done with ACs met and NULL outcome; TASK-18155 (granularity router) filed with census-first ACs and notes that `include_parent_docs` was retired by TASK-16174 and `expand_document` covers the capability.
- No production code changed. No migration actions required.

<sup>Written for commit bb46a466296dc01da0b48622d558dad2a1932d76. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1791?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/qa/2026-08-18-clarification-gate/bar.md
+++ b/Docs/superpowers/qa/2026-08-18-clarification-gate/bar.md
@@ -1,0 +1,62 @@
+# Clarification gate: the bar and the kill condition (TASK-16072 AC#2)
+
+**Written before Step 0 runs.** Registered so the verdict cannot be
+renegotiated once numbers exist — the discipline TASK-15965 established and
+the cross-encoder arc reaffirmed.
+
+## What a gate would have to fire on
+
+A clarification gate detects that a query is **ambiguous or underspecified**
+and asks the user a question instead of retrieving on a guess. For it to have
+any measurable effect on this instrument, the corpus must contain golden
+queries where:
+
+1. the query admits **more than one defensible interpretation**, AND
+2. those interpretations point at **different documents** in this corpus, AND
+3. a user's answer to a clarifying question would **change which document is
+   retrieved**.
+
+All three are required. A query that is vague but has exactly one plausible
+target is not a gate case — asking would be noise. A query whose target is
+absent from the corpus is not a gate case either: no answer to any question
+puts a missing document into the index.
+
+## Step 0's question (AC#1)
+
+For each of the 60 golden queries, answered **from the corpus** rather than
+from argument: does it carry a signal a gate could fire on, under the
+three-part test above?
+
+## The admission bar (what licenses production code)
+
+**≥ 5 of 60 golden queries** must satisfy all three conditions — enough that
+a gate could move a scored cell rather than a rounding artifact. Below that,
+any measured "gain" would rest on one or two queries and could not be
+distinguished from fixture noise, which is the failure mode that retired
+`acronym` and `compositional`.
+
+## The kill condition (what ends the arc)
+
+**Fewer than 5 qualifying queries ⇒ the arc ends with a recorded null**, per
+AC#3: a paper analysis is the deliverable and no production code is written.
+
+Two specific outcomes are pre-declared as **kills, not invitations to author
+fixtures**:
+
+- If the qualifying count is zero because every failing class fails for a
+  reason a question cannot repair (absent content words, corpus asserting
+  what the query excludes, no shared content word), that is the finding.
+- If reaching the bar would require **authoring new ambiguity fixtures**,
+  the arc ends instead. `Tests/RAG_Eval/README.md`'s admission protocol
+  admits only what today's pipeline is *measured* to fail; a class invented
+  to give a feature something to show measures the class, not the feature.
+
+## What is NOT admissible as evidence
+
+- A simulated user answering the clarifying question. That is an oracle feed,
+  and an oracle feed measures the fixture (Lessons: a control that holds a
+  second variable fixed measures the pair).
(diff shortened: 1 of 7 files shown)


rmusser01:
All four findings addressed in `bb46a4662`. Two of them changed the analysis; one of them is wrong, and I checked rather than assumed.

**Finding 1 — right about the method, and it changed how the census works.** Counting relevance *labels* asks what the fixture intends, not what the corpus contains, and a gate fires on the corpus. Re-measured against document text: **1** of 60 queries has an unlabelled alternative reading, against a bar of 5. The kill still fires — but now on a corpus-derived measure rather than a proxy. That one query is `negation` ("which outstation does *not* take a standard maintenance record"), whose two extra matches exist precisely *because* the corpus asserts what the query excludes — the one thing a clarifying question cannot repair.

**Finding 1's example is factually wrong, and it takes one query to see.** The claim was that `plant maintenance record` has industrial and botanical readings in the corpus. Exactly **one** corpus document mentions "plant" at all — `note-saltmarsh-hide`, the estuary one — and it **is** the labelled relevant document. There is no industrial-plant document to disambiguate against. Plausible claim ("plant" is the textbook polyseme), which is why it was verified instead of believed.

**Finding 2 — right.** All 60 rows are now committed as `per-query-census.md`, with the generating script beside it.

**Finding 4 — my error, and an embarrassing one.** `kw-nimbus-rollback` is three words with two relevant documents, contradicting my own candidate table two paragraphs earlier. My script printed both facts and I didn't cross-check them. Corrected: of the 12 short queries, 11 have exactly one relevant doc; the twelfth is a *multi-target* query (both docs correct), not an ambiguous one.

**Finding 3 — fixed**, `TASK-18155` canonicalised.

**One method note worth recording.** The corrected census initially reported `match = 0` for nearly every query — which would have read as a clean confirmation of the null. It was matching against **empty strings**: `CorpusDoc`'s text field is `content`, and my probe asked for `body`/`text`. A measure that silently reads nothing looks exactly like a measure that found nothing. `census.py` now prints how many documents it actually read before printing any result — the same guard the cross-encoder probe needed for the same reason.

PR #1798, opened 2026-08-18
Sync cache API keys on the same dimensions as the async one (TASK-15701)

rmusser01:
`SimpleRAGCache.get`/`put` did not accept `keyword_source_types`, `hybrid_fusion` or `fts_match_construction` **at all**, so they rendered a legacy key that omitted three dimensions describing *what search actually ran*.

## Why it matters, and why it got worse

The failure mode is not a missed hit — that's harmless. It's a **wrong hit**: two searches differing only in an omitted dimension collide, and the second is served the first one's rows. A Notes-only keyword search and an all-sources one collide; so do two searches at different fusion weights.

It was *accidentally* safe before TASK-15400, because `and` was both the legacy value and the shipped default — an omitted key part described the search truthfully. The default has since moved twice (`and_stopword_trim`, then `and_then_prefix`), so a sync-path entry would now be stored under a key asserting a construction that did not produce it.

## Kept and corrected, not removed (AC#1)

Both arms were open. The sync methods are the cache's **ergonomic test surface** — 58 call sites across five modules — while production reads and writes go through the async path only. Removing a correct-but-unused API to fix a key bug would have traded a latent trap for 58 rewrites and no behavioural gain. The decision and its reason are in `get`'s docstring, as the AC required.

## Still latent at close (AC#4, re-verified today)

The only `SimpleRAGCache` traffic in `tldw_chatbook/` is `rag_service.py:1301` (`get_async`), `:1395` (`put_async`) and `:4229` (`get_metrics`); the `_global_cache` singleton has no consumer outside its own module. Recorded so a future reader knows this closed **before** anything wired to it — which is the state the task was filed to preserve.

## The tests (AC#2, AC#3)

Three parametrized collision tests, one per omitted dimension, each asserting the wrong-hit case specifically: the identical search must still hit, and the differing one must **not** be served the first search's rows.

AC#3 is *shown* rather than asserted separately — a sync `put` followed by an async `get_async` for the same search. That round trip cannot pass while the two paths render different keys, which is precisely what checking each path's key in isolation would have missed.

All four were RED before the change.

## Verification

354 passed across `Tests/RAG/simplified/` and the three sibling suites that exercise the sync API; ruff clean. The one red (`test_config.py::test_config_loading_from_file`) is pre-existing — baselined in a throwaway worktree at clean dev `e49c5dba8` before being called unrelated.

Refs TASK-15701.

---
## Summary by cubic
Sync `SimpleRAGCache.get`/`put` now key on the same search-defining dimensions as the async API (`keyword_source_types`, `hybrid_fusion`, `fts_match_construction`), preventing wrong-hit collisions. Previously, the sync path used a legacy key that omitted these, so two searches differing only in those dimensions could return the wrong rows. Side effect: legacy sync-written entries won’t be hit and will age out via TTL. No migration required; new parameters are optional.

**Reviewer focus**
- Code: `tldw_chatbook/RAG_Search/simplified/simple_cache.py` expands sync `get`/`put` signatures and forwards new args to `_make_key` to render identical keys as async. Behavior change is limited to the sync path; sync API kept for test ergonomics, while production uses async only (re-verified).
- Tests: `Tests/RAG/simplified/test_simple_cache_sync_key_dimensions.py` adds three collision tests (one per dimension) and a sync-put/async-get parity test.
- Docs: `put_async` and sync method docstrings updated to reflect parity and note the historical hazard now closed.

<sup>Written for commit 46c7d27316608a17e43c1974116490f27dacea35. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1798?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/RAG/simplified/test_simple_cache_sync_key_dimensions.py
+++ b/Tests/RAG/simplified/test_simple_cache_sync_key_dimensions.py
@@ -1,0 +1,83 @@
+"""TASK-15701: the SYNC cache API must key on the same search-defining
+dimensions the async one does.
+
+`_make_key` takes eight inputs, three of which describe *what search was
+actually run* — `keyword_source_types` (which sub-legs the keyword leg
+queried), `hybrid_fusion` (alpha / rrf_k / pool) and `fts_match_construction`
+(how the FTS MATCH expression was built). The async path passes all three;
+the sync `get`/`put` did not accept them at all, so they rendered the legacy
+key.
+
+The failure that matters is not a missed hit (harmless) but a **wrong hit**:
+two searches differing only in an omitted dimension collide, and the second
+is served the first one's rows. Before TASK-15400 the legacy key was
+accidentally truthful — `and` was both the legacy value and the shipped
+default. The default has since moved twice (`and_stopword_trim`, then
+`and_then_prefix`), so an omitted part now asserts a construction that did
+not run.
+"""
+
+import asyncio
+
+import pytest
+
+from tldw_chatbook.RAG_Search.simplified.simple_cache import SimpleRAGCache
+
+
+def _cache() -> SimpleRAGCache:
+    return SimpleRAGCache(max_size=32, ttl_seconds=300, enabled=True)
+
+
+@pytest.mark.parametrize(
+    "dimension,first,second",
+    [
+        ("fts_match_construction", "and", "and_then_prefix"),
+        ("keyword_source_types", ("notes",), ("notes", "media")),
+        ("hybrid_fusion", (0.7, 5, 2), (0.3, 60, 4)),
+    ],
+)
+def test_sync_round_trips_differing_in_one_dimension_do_not_collide(
+    dimension, first, second
+):
+    """AC#2: a search that differs ONLY in one search-defining dimension must
+    not be served the other's rows.
+
+    Args:
+        dimension: The `_make_key` parameter under test.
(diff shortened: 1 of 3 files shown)


rmusser01:
Addressed in the latest push.

**The finding was precise and worth having:** adding the three dimensions to the sync API made `put_async`'s own `Args` block **false** — it still told maintainers the sync twins cannot accept them, key without them, and remain collision-prone. A contract that describes a hazard *after* the hazard is fixed is worse than no note at all: it steers people away from a supported API and invites someone to "fix" it a second time.

Rewritten to the resolved state, with the historical hazard preserved in parentheses for anyone arriving via `git blame` — the wrong-hit mechanism is worth keeping legible even once it's closed.

I also swept the other surfaces the task said qualified this guarantee to the async path (`config.py`'s field comment, `_make_key`'s docstring, `get_async`). `put_async` was the only stale one.

Rebased onto latest dev; 62 passed across the cache suites and the construction tests at the rebased head.

PR #1801, opened 2026-08-18
Untiered merge is the measured choice; TASK-16071's note retires (TASK-17955)

rmusser01:
The task asked me to extend the fixture first so a tiered-vs-untiered comparison could say something. Measuring before authoring showed the extension wouldn't have been enough — and that the question is already answered.

## What tiering actually changes

Tiering makes a primary-form row outrank a fallback-form row instead of interleaving by position. That changes what a consumer sees **only when the merged list is cut** — when `rows_returned >= top_k`. If the window is never full, every row survives whatever the order.

| k | queries whose window is FULL (`rows >= k`) | max rows seen |
|---|---|---|
| 10 | **0 of 60** | 6 |
| 20 | **0 of 60** | 6 |

**The displacement failure mode cannot occur on this corpus.** That's stronger than TASK-17755's review reached: it found the two orderings *coincide* (no query mixes primary and fallback rows); this finds that **no row is ever cut**, so the order cannot change what any consumer receives, whatever the mix.

## Why AC#1 was deliberately not taken

To make the comparison mean anything, an authored query would have to *fill the window* — ≥10 rows on a path whose current maximum is 6. Authoring it means choosing its shape, and its shape decides which ordering wins.

That's precisely the trap **TASK-16072's kill condition** named six days ago in this same programme: *a class invented to give a feature something to show measures the class*. It applies with equal force to a class invented to give a comparison something to distinguish. The eval README's admission protocol admits only what today's pipeline is **measured** to fail — and nothing here fails.

## The decision (AC#3)

Untiered ships as the **measured** choice, not as an unpaid debt — and TASK-16071's "the 15700 tier design would apply if this path ever gained fallback forms" note is retired. It gained them in TASK-17755; the answer is that it doesn't matter here, for a structural reason.

## The merge-site comment (AC#4)

Now states what was measured, when, and — the part that stops this recurring — **the one-command re-check**: if any plain query starts returning `>= top_k` rows, tiering becomes measurable and the decision is due for review. The original note was accurate but undated, which is why it survived two arcs unresolved.

## Method note

The gating census was re-measured on current dev rather than inherited: it was originally taken during TASK-17755's review, *before* `and_then_prefix` shipped to this path, and the distribution has changed since — one query (`ng-mains-supply`) now returns 6 rows where previously nothing exceeded 1. The conclusion held, but it was checked rather than assumed.

19 passed across the two four-seam suites; ruff clean.

Refs TASK-17955.

---
## Summary by cubic
Keeps the four‑seam merge untiered by measurement and retires TASK‑16071’s note. Corrects the earlier cut-based argument: MRR/NDCG are order-sensitive, so tiering only matters when a query has ≥2 rows and at least one relevant row retrieved. On the current corpus, that never occurs, so tiering is unobservable. No runtime behavior changes.

- Add `Docs/superpowers/qa/2026-08-18-merge-tiering/report.md`: shows 59/60 queries return ≤1 row; the single 6‑row query retrieves no relevant document, so any ordering scores identically.
- Add `Docs/superpowers/qa/2026-08-18-merge-tiering/tier_observability_census.py`: runnable re-check that lists multi-row queries and ranks where relevant docs land; this is the future trigger (≥2 rows including a relevant one).
- Update `backlog/tasks/task-17955...md`: mark Done; uncheck AC#1 explicitly; record the corrected criterion and outcome; retire TASK‑16071’s note.
- Update merge-site comment in `tldw_chatbook/Library/library_local_rag_search_service.py`: document the measurement, corrected reasoning, and the script path; code change is comment-only.
- No migrations, config, or rollout. Review focus: skim the QA report; optionally run the census script; confirm code diff is documentation only.

<sup>Written for commit f11cb57b32d906d9c905ea6b70747fe9dc6d51d2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1801?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/qa/2026-08-18-merge-tiering/report.md
+++ b/Docs/superpowers/qa/2026-08-18-merge-tiering/report.md
@@ -3,31 +3,46 @@
 Date: 2026-08-18 · measured on the gated fixture (172 docs, 60 golden
 queries), plain path, current dev. No network, no spend.
 
-## What tiering would change, precisely
+## What tiering would change, precisely (CORRECTED after review)
 
-Tiering makes a primary-form row outrank a fallback-form row instead of
-interleaving with it by position. That changes what a consumer sees **only
-when the merged list is CUT** — i.e. when `rows_returned >= top_k`. If the
-window is never full, every row survives whatever the order, and the two
-merges are observationally identical to any downstream metric.
+**The first version of this report argued from CUTS and that was wrong.**
+It said tiering matters only when `rows_returned >= top_k`, so a corpus that
+never fills the window is safe. Qodo (PR #1801, finding 1) pointed out the
+hole: **MRR and NDCG consume ORDER, not membership.** Moving a relevant row
+from rank 1 to rank 3 changes both, with nothing cut. A cut-based argument
+cannot establish that ordering is unobservable, and a `rows >= top_k` review
+trigger would have left tiering closed through exactly the corpus change that
+made it matter.
 
-## The measurement
+## The correct question, and the measurement
 
-| k | queries whose window is FULL (`rows >= k`) | max rows seen |
-|---|---|---|
-| 10 | **0 of 60** | 6 |
-| 20 | **0 of 60** | 6 |
+Reordering can move a score only for a query that has **≥2 rows** AND **at
+least one retrieved row that is relevant** — otherwise every permutation
+scores identically.
 
-Row distribution on the plain path today: **37** queries return 0 rows,
-**22** return exactly 1, **1** returns 6 (`ng-mains-supply` — a negation
-query, and the single query TASK-17755's fallback rescued into multi-row
-territory).
+| k | queries with >1 row | of those, with ground truth | with a RELEVANT row retrieved |
+|---|---|---|---|
+| 10 | **1** | 1 | **0** |
+| 20 | **1** | 1 | **0** |
 
-**So the displacement failure mode cannot occur on this corpus at all.**
-This is a stronger statement than TASK-17755's review made. That review found
-the two orderings *coincide* (no query mixes primary and fallback rows). The
-measurement above shows something more basic: **no row is ever cut**, so the
-order cannot change what any consumer receives, whatever the mix.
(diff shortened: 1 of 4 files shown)


rmusser01:
All three addressed in the latest push — and **finding 1 is a real error in my reasoning**, not a nitpick.

I argued tiering matters only when rows are **cut**, so a corpus that never fills the window is safe. But MRR and NDCG consume **order**: moving a relevant row from rank 1 to rank 3 changes both with nothing cut. Worse, the review trigger I proposed (`rows >= top_k`) would have left tiering closed through exactly the corpus change that made it matter. That's the kind of wrong-but-plausible argument that outlives the arc that wrote it.

**The conclusion survives, for the narrower reason your finding implies** — and it's measured rather than argued. A reordering can move a score only for a query with ≥2 rows **and a relevant row among them**. This corpus has none:

| k | queries with >1 row | with ground truth | with a RELEVANT row retrieved |
|---|---|---|---|
| 10 | 1 | 1 | **0** |
| 20 | 1 | 1 | **0** |

The single multi-row query (`ng-mains-supply`, rescued by TASK-17755's fallback) retrieves **no relevant document at all**, so every permutation of its six rows scores identically.

**Finding 3** — correct, and it was self-inflicted: the note promised a "one-command re-check" and offered prose. The census is now a committed, runnable script (`tier_observability_census.py`) that prints per depth the >1-row queries and the ranks at which relevant rows landed, and it checks the **corrected** trigger rather than the under-triggering one.

**Finding 2** — also correct. I ticked AC#1 with a regex while the outcome text right below it says the fixture extension was deliberately skipped. That's a false project record; the box is unchecked and the deviation is stated in place.

Rebased onto latest dev; 19 passed across the four-seam suites at the rebased head.

PR #1812, opened 2026-08-18
docs(rag): TASK-18155 granularity router — NULL by census, sixth retired P2c premise

rmusser01:
## Verdict: NULL

**Rescuable: 1, against a pre-registered bar of 5.** No probe was run, no production code exists, the arc ends here. That outcome was pre-authorised by the task.

Full record: `Docs/superpowers/qa/2026-08-18-granularity-census/report.md`. Rerunnable census: `granularity_census.py` beside it (refuses to run without `RAG_EVAL=1`).

## The premise has two mechanisms; the cheap count sees one

1. **Direct** — the query's own relevant document is multi-chunk.
2. **Displacement** — some *other* document eats several top-k slots. `canonicalize.py` states it: *"one document can occupy several of the top-k slots."* The cut happens at **row** level and rows collapse to documents only afterwards, so a 3-chunk doc spends 3 of 10 slots and only 8 distinct documents can appear.

A corpus count is blind to mechanism 2, so the census built a live index for it. `plain` is exempt throughout — it returns whole items and is already document-granular.

## What the corpus is

Measured with the real `ChunkingService` at the harness profile's own settings (`hybrid_basic`, 384 words / 64 overlap), not inferred:

**172 documents → 179 chunks. 168 of 172 are single-chunk.** For 98% of the corpus, a chunk already *is* the document.

## Results

| population | semantic | hybrid |
|---|---|---|
| direct (relevant doc multi-chunk) | 4, of which 3 already HIT | 4, **all HIT** |
| displacement-qualifying | **0** | **0** — no query has ANY duplicate slot |
| **rescuable** | **1** | **0** |
| exposure: already-HIT queries a reorder could only move DOWN | **9** | 0 |

Three things make this solid rather than marginal:

- The single rescuable query, `kw-plant-maintenance-record`, has **zero duplicate slots** — displacement is not its mechanism. It is the fusion arc's known similarity miss (*"semantic ABSENT from top-10, present ~rank 22"*), and **fusion weighting already rescued it**, which is why hybrid reads HIT.
- In the **shipped** mode there is nothing to act on at all.
- The exposure asymmetry is PRF's: no gain available, nine currently-correct queries put at risk.

## The census had a defect, and caught it

The first run reported 2 qualifying queries. Both were artifacts:

- a `negative` query has an empty `relevant_slugs`, so the hit test is False **by construction** — all 7 negatives register MISS regardless of what retrieval did;
- a `prompt` target has **no vector index** (B2 gave prompts an FTS sub-leg deliberately without one), so no freed slot can admit it.

"Not applicable" rendering identically to "failed" — the same species as TASK-18255 one arc earlier. Both exclusions are now explicit in the script rather than implicit in the reader.

## Process deviation, recorded rather than papered over

**I ran the census before registering the bar**, which is the order AC#1 exists to prevent. To keep it non-circular the bar is **inherited verbatim** — ≥5 qualifying queries, the number PRF's clause 1 and the clarification gate both used — rather than chosen after seeing the result. AC#1 is ticked with that deviation stated in the task file.

## Third-surface check (AC#5)

A router would be the third surface over one capability, after the `include_parent_docs` family (retired **inert**, TASK-16174) and `expand_document` (the gated, pull-based replacement). It cleared nothing, and the demand-driven tool already serves the need.

## What would reopen it

This is a property of *this* corpus — 168 of 172 documents fit in one chunk. A long-form corpus would move every number. Re-run the script if the fixture gains substantial long documents; the reachable population should decide, not the idea's appeal.

## Verification

No source or test code changed (`Tests/RAG_Eval/README.md` and the task file are the only edits; the QA directory is new), so the gate cannot move. Census reproduced exactly on a clean re-run; the `RAG_EVAL=1` guard verified to refuse otherwise.

Refs TASK-18155.

---
## Summary by cubic
Retires the “granularity router” premise with a NULL from a guarded, rerunnable census; no probe ran and no production behavior changed. Old: unquantified premise. New: rescuable=1 vs bar=5 → below bar, so no router.

- Adds a guarded census and report: `Docs/superpowers/qa/2026-08-18-granularity-census/{granularity_census.py,report.md}`. Uses the real chunker with conversation parity, measures displacement with a live index, prints mapping health, and refuses a verdict if zero rows were canonicalized or any query errored.
- Pins helpers and slot arithmetic in `Tests/RAG_Eval/test_granularity_census.py`; defers imports into `main` (documents the one module-scope import); updates `Tests/RAG_Eval/README.md` to record the sixth retired P2c premise; marks TASK-18155 Done.
- Findings: 172 docs → 179 chunks; 168 single‑chunk. In `hybrid` (shipped), no query has duplicate doc slots. Rescuable=1 in `semantic` (`kw-plant-maintenance-record`), a similarity miss already fixed by fusion; exposure=9 currently‑HIT queries that reordering could only move down.

<sup>Written for commit 2686023f363c9f3e4b32e3b960f2f4fee5734df4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1812?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/qa/2026-08-18-granularity-census/granularity_census.py
+++ b/Docs/superpowers/qa/2026-08-18-granularity-census/granularity_census.py
@@ -35,6 +35,11 @@ REPO = Path(__file__).resolve().parents[4]
 if str(REPO) not in sys.path:
     sys.path.insert(0, str(REPO))
 
+# E402 is unavoidable and deliberate: this script lives under `Docs/`, not on
+# the import path, so the repo root must be on `sys.path` BEFORE any first-party
+# import can resolve. This is the file's only module-scope local import -- every
+# other one is deferred into `main`, in a single contiguous group, so the pure
+# helpers stay importable (and testable) without building a retrieval stack.
 from Tests.RAG_Eval.harness.canonicalize import canonical_source_type  # noqa: E402
 
 CHUNK_SIZE = 384          # hybrid_basic profile, the harness's own
--- a/Docs/superpowers/qa/2026-08-18-granularity-census/granularity_census.py
+++ b/Docs/superpowers/qa/2026-08-18-granularity-census/granularity_census.py
@@ -1,0 +1,186 @@
+"""TASK-18155 census: could a granularity router change any golden outcome?
+
+Run:  RAG_EVAL=1 python Docs/superpowers/qa/2026-08-18-granularity-census/granularity_census.py
+
+Answers two populations, because "chunk vs document granularity" has TWO
+mechanisms and the cheap count only sees the first:
+
+1. DIRECT -- the query's own relevant document is multi-chunk, so retrieving
+   it whole rather than in pieces changes what is scored for that query.
+2. DISPLACEMENT -- some OTHER document occupies several of the top-k row
+   slots (`canonicalize.py`: "one document can occupy several of the top-k
+   slots"), because the top-k cut happens at ROW level and rows collapse to
+   documents only afterwards. Collapsing at retrieval would free those slots
+   for documents currently pushed out. This one needs a live run.
+
+Only a query whose relevant document is CURRENTLY MISSED can be rescued, so
+the qualifying population is (slots freed > 0) AND (relevant doc absent).
+"""
+from __future__ import annotations
+
+import os
+import sys
+from collections import Counter
+from pathlib import Path
+
+REPO = Path(__file__).resolve().parents[4]
+sys.path.insert(0, str(REPO))
+
+if os.environ.get("RAG_EVAL") != "1":
+    raise SystemExit("refusing to run without RAG_EVAL=1 (this builds a real index)")
+
+import tldw_chatbook
+
+# Provenance: a census that silently measured another checkout would be
+# worthless, and this arc already shipped one instrument that could not see
+# what it was asked about (TASK-18255).
+print(f"PROVENANCE tldw_chatbook <- {tldw_chatbook.__file__}")
+assert str(REPO) in tldw_chatbook.__file__, f"WRONG TREE: expected under {REPO}"
+
+from Tests.RAG_Eval.harness.canonicalize import canonical_source_type, slug_lookup_from
+from Tests.RAG_Eval.harness.goldenset import load_fixtures
+from Tests.RAG_Eval.harness.ingest import build_eval_runtime
+from Tests.RAG_Eval.harness.runner import SOURCE_TYPES, _extract_rows, build_query_scope
+
+CHUNK_SIZE = 384          # hybrid_basic profile
+CHUNK_OVERLAP = 64
+K = 10                    # run_eval's default
+BAR = 5                   # inherited verbatim from the PRF + clarification-gate arcs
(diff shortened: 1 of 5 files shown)


rmusser01:
## Review disposition

**5 of 7 auto-resolved. The two still showing open are both addressed; their rendered bodies quote pre-rewrite code.**

### Both bugs — fixed, and one of them mattered more than it looked

**#6 conversation parity** — correct finding. The census chunked raw `CorpusDoc.content`, but conversations index as `f"{sender}: {content}"`, so it measured text one word short of what exists in the index. Fixed via `parity_text`. Re-measured across all 31 conversation fixtures: **0 chunk counts change**, multi-chunk set identical — so the verdict never depended on it, but a future corpus would not be so forgiving.

**#4 incomplete verdict** — correct, and this one could have produced a false NULL. A skipped query shrank the population while the census still printed NULL, so a failed search could have hidden a qualifying query. It now collects errors, **claims no verdict at all** if any occurred, and exits non-zero. This run prints `errors: 0 -- population COMPLETE (60 queries x 2 modes)`.

### #7 tests — added, and chasing it found a way the census could fake its own headline

`Tests/RAG_Eval/test_granularity_census.py`, **23 pins** over `classify`, `parity_text`, `row_slug` and the slot arithmetic (which now lives in a testable `slot_summary` rather than inline — the snippet in the finding is the old shape).

The finding's stated worry — *"defects in the census could silently change the reported NULL verdict"* — turned out to be concretely true in a way I had not seen. `row_slug` maps each row to a fixture slug, and an **unmapped row becomes a distinct `unknown:*` key**. So if mapping broke, every row would look like a different document, freed-slot counts would collapse to 0, and the census would confidently report "no displacement" having measured nothing. Hybrid's headline *is* a zero — exactly the shape that failure produces.

So the census now reports mapping health beside the verdict:

```
PROBE PROOF: rows canonicalized=1200, unmapped=0 (0.0%)
```

Hybrid's zero is backed by 1200 rows that were actually identified.

**Mutation-verified twice** (a green test that stays green under the bug is a guard, not a proof):
- removing the negative exclusion → reds `test_negative_query_is_excluded_not_qualifying`
- removing the `media_15` prefix-strip fallback → reds `test_strips_the_source_type_prefix_as_a_fallback`

Both green on restore.

### #5 imports — addressed as far as is possible

`main()`'s imports are now one contiguous group placed before any executable statement. One module-scope import remains and **cannot** be removed: this script lives under `Docs/`, not on the import path, so the repo root must reach `sys.path` before any first-party import resolves. That `# noqa: E402` is now documented as deliberate at the site.

### Findings 1–3 (docstrings, type hints) — applied as written.

### Verdict unchanged

`rescuable 1 vs bar 5 → NULL`. No production source changed; `Tests/RAG_Eval` reads **336 passed, 14 skipped**.

PR #1815, opened 2026-08-18
docs(rag): TASK-18514 HyDE — NULL by probe, seventh and last retired P2c premise

rmusser01:
## Verdict: NULL — and this exhausts the five named P2c candidates

HyDE is the **first candidate to clear its census** (11 reachable vs a bar of 5), so unlike the clarification gate (0) and the granularity router (1), it earned its probe. The probe returned **2 rescues, 0 losses**.

| pre-registered clause | measured | result |
|---|---|---|
| rescues ≥ 5 | **2** | **FAIL** (gating) |
| zero currently-hitting queries lose their target | **0** | **PASS** |

**Clause 2 passing is the notable half.** PRF died on exactly that clause, losing 10 of 21 hitters. HyDE is *safe* and merely insufficient — a different failure, recorded as such.

## The bar was registered before any measurement existed

Commit `724f28951`, before the census script was written. That's the direct correction to TASK-18155's recorded process deviation, where I measured first and had to inherit a bar to keep it non-circular. Here the ordering is verifiable in git history rather than asserted in prose.

## Why it failed — mechanistic, and *not* the generator

The generations are good: 60 passages, **0 empty**, mean 71 words, 108s total.

**All 11 reachable targets contain every content word of their query.** I first read them as semantically unrelated from a 190-character truncation and had it backwards. They miss because the words appear in a **different sense, incidentally, inside a document about something else**:

| query | target | how the words appear |
|---|---|---|
| `pump chamber inspection` | Icehouse conservation note | a real pump chamber under the flagstones; an annual *inspection of the fabric* |
| `plant maintenance record` | Estuary bird hide note | a **botanical** plant list; *maintenance* of the hide; a *record* of each visit |
| `meter box key` | Staff allotment strip | the meter box key hangs in the tool shed |

These are the corpus's deliberate lexical traps. **HyDE's mechanism makes the query embedding more topically specific** — it writes a passage about pump-chamber inspection procedure, which moves the vector *further* from an icehouse note. On this corpus the semantic misses are lexical coincidences, not semantic near-misses, so **a better generator makes these strictly worse**.

The two rescues confirm the reading: `ng-mains-supply`'s target describes a solar-and-battery outstation, and the generated passage described exactly that — HyDE working as designed, on `negation`, the one category at 0.000 in every mode.

**So HyDE's true population here is the 3 negation queries — below the bar before a token is generated.** Even granting all three (two were near misses a stronger model might take), 3 < 5. That's why the null doesn't depend on the generator, which the task required me to establish separately.

## Instrument notes

`enable_thinking=False` is load-bearing: the server fronts a **reasoning** model that otherwise spends its entire budget in `reasoning_content` and returns `content=""` — a full token budget producing an empty string, which reads exactly like a refusal. Another instance of this programme's dominant defect.

Both scripts refuse a verdict on any error. Census: 15,820 rows, 0 errors. Probe: 0 empty generations, 0 retrieval errors. All 60 passages are committed **verbatim** in `generations.json` rather than summarized, so the generator can be judged independently of my reading of it.

## A wrong turn, recorded

I hypothesised the 7 scoped misses were a scope-application gap — that semantic mode ignored the allowlist. **Measured: it doesn't.** Scope is applied in both modes (allowlist 100, all 10 returned rows in-scope); hybrid hits those targets because its *keyword* leg finds the incidental term match. The hypothesis was wrong and the measurement corrected it before it reached the report.

## What would reopen this

HyDE is mismatched to *this corpus*, not shown useless in general. `negation` is the only category here with the shape HyDE is built for, and HyDE moved it. **If the negation category is ever expanded, re-run `hyde_probe.py` against it specifically.**

## Verification

No production source changed, so the gate cannot move. `Tests/RAG_Eval`: **336 passed, 14 skipped**.

Refs TASK-18514.

---
## Summary by cubic
Retires HyDE as a P2c premise with a reproducible census and probe; verdict remains NULL (2 rescues, 0 losses) and the named P2c list is exhausted. Review hardens measurement: fixes scoring to exclude empty generations, validates generator endpoint/model at startup, and makes the report consistent (4 topical / 7 lexical); no production changes.

- Adds `Docs/superpowers/qa/2026-08-18-hyde-census/{hyde_census.py, hyde_probe.py, generations.json, report.md}`, updates `Tests/RAG_Eval/README.md`, records TASK-18514 in `backlog/tasks/`, and pins helpers in `Tests/RAG_Eval/test_hyde_probe.py` (`score_arms`, census `classify()`), including prompt-exclusion and ordering guards. Defines and uses `MAX_TOKENS` and `GEN_TIMEOUT_S`.
- Generator is fixed and validated: endpoint `http://localhost:9099/v1/chat/completions` (override via `HYDE_ENDPOINT`), model override `HYDE_MODEL`, temperature 0, `max_tokens=220`, `enable_thinking=False`. Empty generations are excluded from scoring and reported; endpoint/model are validated at startup. Census reachable (semantic) = 11 vs bar 5; probe gains = 2, losses = 0; 60 non-empty generations; 0 retrieval errors.

**How to verify**
- Run: `RAG_EVAL=1 PYTHONPATH=$(pwd) <venv>/bin/python Docs/superpowers/qa/2026-08-18-hyde-census/hyde_census.py` then `.../hyde_probe.py`. Expect census reachable = 11 (semantic); probe gains = 2, losses = 0; 60 non-empty generations; 0 retrieval errors.

<sup>Written for commit d37517c9fbb47d9016bbdf3843bfa87aa7611239. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1815?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## Review disposition — 5 of 6 auto-resolved, all substantively addressed

### Both bugs, and one changed a headline

**#3 topical population miscounted** — correct, and an internal contradiction in my own text: I argued `sc-valve-pit-access` *was* a genuine topical match, then counted only the 3 negation queries as topical. Corrected to **4 topical / 7 lexical** in the report and the README. The conclusion holds — 4 < 5 — but the margin is **one query, not two**, and the report now says so instead of overstating it.

**#1 empty generations become losses** — correct, and latent only because 0 came back empty. Scoring an empty generation as a HyDE miss would convert an **unmeasurable** query into a **LOSS**, corrupting the harm gate — the one clause HyDE actually passed. Empty generations are now excluded from scoring and reported separately.

That one is worth naming precisely: it is this programme's recurring defect in its purest form — *"could not measure"* rendering identically to *"measured, and it got worse"*. Three arcs running, it has appeared in the instrument every time.

### #5 tests — now covering both scripts

22 pins in `Tests/RAG_Eval/test_hyde_probe.py`:

- `score_arms` (probe): gains, losses, negatives excluded, and the empty-generation rule
- `classify` (census): the bucketing that produced the "11 reachable" which **licensed the probe** — an over-count probes a population that does not exist; an under-count kills a real candidate silently
- endpoint/model validation

**Mutation-verified twice**, because a green test that stays green under the bug is a guard, not a proof:
- scoring empty generations as misses → reds both empty-generation tests
- disabling the prompt exclusion → reds `test_prompt_excluded_even_when_found_deeper`

Both green on restore. Both scripts reproduce their numbers exactly after the refactor: census **15,820 rows / 0 errors / 11 reachable**; probe **2 gains, 0 losses, NULL**.

### #2, #4, #6 — applied as written

`Raises:` on `generate`, `MAX_TOKENS`/`GEN_TIMEOUT_S` constants, and startup validation of `HYDE_ENDPOINT`/`HYDE_MODEL`. The last matters beyond style: a malformed endpoint would otherwise surface as *"the generator returned nothing"*, which is exactly the reading this programme keeps getting wrong.

### Verdict unchanged

`2 rescues vs a bar of 5` → **NULL**, harm gate **PASS**. No production source changed; `Tests/RAG_Eval` reads **358 passed, 14 skipped**.

PR #1824, opened 2026-08-19
feat(console): agent run budget in Settings, plus three production defects found investigating it

rmusser01:
Four related commits, each independently green, rebased onto `dev` @ `c1932f4dd`.

Started as "expose the agent run budget in Settings"; the investigation that
followed turned up three production defects, which is most of the diff.

## 1. Agent run budget in Settings (TASK-18600)

Five limits, previously hardcoded, now user-configurable in
**Settings ▸ Console Behavior ▸ Agent run budget** and resolved per run, so a
save applies to the next message with no restart.

| Limit | Was | Now |
|---|---|---|
| Token budget (per run) | 1M | 25,000,000 |
| Wall-clock | 1800 s | 86,400 s (24 h) |
| Per-tool-call | 300 s | 3,600 s |
| Model turns | 30 | 2,000 |
| Steps | 96 | 25,000 |

Floors only, no ceilings (owner call). Engine `RunBudget` defaults unchanged —
the Console is the only production caller.

**The turn cap is deliberately unreachable, and the code says so.** The whole
conversation is re-sent every turn and history eviction is off by default, so
spend is quadratic in turn count: 25M is exhausted around turn 250 for a typical
tool round. Owner decision was to keep spend as the real governor and document
it rather than lower the turn cap to something reachable.

The section is driven by one `AGENT_BUDGET_FIELDS` spec table rather than five
copies of this screen's per-field staging boilerplate.

## 2. Performance (TASK-18602/18603/18604)

**Token estimation was quadratic per run.** `_chars_estimate` counted CJK share
with a per-character Python loop, and every caller re-estimated the whole
append-only history each turn. Adds an ASCII short-circuit plus a single-pass
regex, and memoizes `estimate_tokens`.

Live path — `bound_messages_to_window`, run once per Console send:

| history | before | after |
|---|---|---|
| 60 turns / 178 KB | 62.5 ms | 0.6 ms |
| 120 turns / 354 KB | 153.4 ms | 1.1 ms |
| 240 turns / 707 KB | 223.8 ms | 1.8 ms |

The memo is keyed by hash+length, never by the text, so it holds no strong
reference and cannot pin a large message in memory. Reads are unlocked (a dict
`get` is atomic under the GIL; a race costs one recompute), so the agent worker
and UI thread can estimate concurrently. `Helper_Scripts/Benchmarks/token_estimate_benchmark.py`
reports the two fixes separately — the ASCII path carries 310x on the shipped
default, the memo earns its place on installs with tiktoken.

**The run budget counted cache reads at full price.** Anthropic prompt caching is
on by default for Console sends, so on a long run nearly every input token is a
cache read billed at ~0.1x — yet it consumed the budget at 1.0x, stopping runs
that had spent a fraction of what the number implies. Now weights the input
buckets by their real published rates via the existing `ProviderUsage` +
`pricing_catalog`. Output stays 1:1 deliberately.

**Bounded the live step feed** — it retained one object per step to render five
rows.

## 3. Sixteen pre-existing red tests (TASK-18605)

A full `Tests/Chat` run showed 18 failures; 17 reproduced on clean `origin/dev`.
Four unrelated causes, two of them production defects:

- **Provider continuation resume was broken in production** (11 tests).
  `_continuation_restore_target_for_resolution` normalized `base_url` into its
  full endpoint, but the checkpoint pins what the bridge recorded, which is the
  raw value. `validate_continuation_restore` compares byte-exactly, so **every
  Resume** was rejected with "Pinned provider settings no longer match". Fixed at
  the builder — normalizing both sides would have weakened a deliberate
  exactness guarantee that stops a continuation being replayed against a
  different endpoint.
- **The Console screen eagerly imported the TTS/Audio stack** (1). Moved to
  `TYPE_CHECKING` plus one function-local import: `chat_screen` import
  1301 ms → 1108 ms.
- **Skill-script confirms weren't released by teardown** (3). Tests poked
  `_shutdown_requested` directly, bypassing the headless-Event split; switched to
  `begin_shutdown()`. This is also why that file *hung* rather than failed.
- **A moved symbol silently disarmed a guard** (1) — `monkeypatch.setattr` raised
  `AttributeError` instead of arming the assertion.

## 4. Visual renderer decoupled from Pillow (TASK-18606, amends ADR-054)

The 18th failure was not stale evidence — it was a live bug.
`ImageFont.load_default()` is not a stable input; Pillow changed it to a
proportional face during 10.x. On Pillow 12.1.1 an 82-character line renders
**738 px against 496 px of usable canvas**, so transcript text runs off the edge
and is **lost with no error** (ink at x=511 vs usable edge x=504).

The `pillow==11.2.1` pin did not prevent this — it only narrowed the
configuration where it wouldn't happen. Three couplings removed:

- Font pinned to the frozen legacy fixed-cell font, and its metrics **verified**
  at load — a changed cell size now raises instead of clipping. Ink → x=499.
- Identity hashes raw pixels + mode + size, not encoded PNG bytes.
- `renderer_version` no longer embeds `PILLOW_VERSION` — which was also *drawn
  into every page footer*, making the hash unable to survive a bump by
  construction. Both profiles → a `v3` generation.

`png_sha256` is kept alongside the new `pixel_sha256` on purpose: identity
("would another host render this page") and wire integrity ("are these the exact
bytes I hashed") are different questions, and conflating them is what put the
encoder in the identity.

Pillow relaxed to `>=10.1` in `pyproject.toml` and `requirements.txt`. **Rationale
for relaxing rather than keeping the pin:** Pillow is the image parser this app
points at untrusted input (media ingestion, chat attachments, remote fetch);
holding it a major version back so one evidence file keeps matching is the wrong
trade. Reproducibility is the renderer's job, not the resolver's. ADR-054 carries
the amendment and its reasoning.

The evidence guard is reframed from "the matrix **is** current" to
"**never enables on stale evidence**": matching evidence still gets the full
geometry check; superseded evidence must carry empty `eligible_models`, so it can
never authorize a default. The strict branch reactivates by itself once evidence
is re-captured.

## Verification

- **7,781 passed, 0 failed** (`Tests/Chat`, `Tests/Agents`, settings/config/
  auto-speak suites) against the rebased base.
- **50,761 tests collect** cleanly.
- The 17 pre-existing failures were each confirmed on a clean `origin/dev`
  worktree before being attributed.
- Live-verified in the running app: the Settings section renders, `86400.0` shows
  `= 1d.`, editing the token budget and pressing `s` wrote
  `agent_max_total_tokens` to config.toml and `console_run_budget()` returned it,
  and a low step budget raised the derived-floor warning.

## Not done, deliberately

Re-capturing the gpt-5.6-terra visual evaluation under the v3 renderer. That is a
real model run and a spend decision. The matrix already reads "not ready", so the
safe state holds meanwhile (TASK-18606 AC#7). Also filed: TASK-18601, the agent
run step log is a single JSON blob column that the raised step budget outgrows.

---
## Summary by cubic
Makes the Console agent’s run budget configurable with higher defaults and fixes three production defects. Old: hardcoded limits, flat input pricing, unbounded live feed, resume rejected, renderer coupled to `pillow`; New: Settings-driven per-run limits, cache-aware input pricing, bounded live feed, verbatim resume targets, and a pixel-based renderer identity (requires `pillow>=10.4`).

- Settings: five limits in Settings ▸ Console Behavior ▸ Agent run budget; resolved per run with floors only. Defaults: 25M tokens, 24h wall, 1h per-tool-call, 2k turns, 25k steps. A per-tool-call value of 0 maps to a finite “unlimited” deadline so Stop/cancellation still work. Engine `RunBudget` defaults unchanged.
- Cost accounting: budgets weight input by cache read/write rates via existing usage + pricing catalog; outputs remain 1:1. Gateway normalization currently hides Anthropic cache writes (budget undercounts writes); follow-up filed.
- Performance: CJK count uses ASCII fast-path + single-pass regex; `estimate_tokens` is memoized; live step feed becomes a bounded tail plus counter.
- Continuation resume: carry `api_base_url` verbatim to match checkpoints; Resumes no longer fail on normalized endpoints.
- Visual renderer: pins fixed-cell font and verifies metrics; identity hashes raw pixels; renderer generation v3. Keeps `png_sha256` for wire integrity and adds `pixel_sha256` for identity. Requires `pillow>=10.4`.

- Rollout and migration
  - No restart needed; Settings apply to the next message.
  - Config: new `[console] agent_max_*` keys are optional; invalid or below-floor values fall back to defaults. Tool-call 0 means “no per-call limit” (Stop preserved).
  - Visual evaluation: re-capture renderer evidence under v3 before enabling strict defaults.
  - Environments must allow `pillow>=10.4`; remove any older exact pin.

<sup>Written for commit a65688c3f242fab6af088be9995258377e4f0ae8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1824?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/User_Guide/console/agent-runs-and-tools.md
+++ b/Docs/User_Guide/console/agent-runs-and-tools.md
@@ -1019,6 +1019,9 @@ run budget will not save a crawl, ingest, or build that takes longer than
 the per-tool-call ceiling; that is the limit that kills it. Lowering this
 one below about 186 seconds is the risky direction — a call reported as
 timed out may still be running, and an MCP tool can end up executing twice.
+Setting it to 0 removes the ceiling but not Stop: cancellation is still
+polled every 0.5 s while a tool runs, so pressing Stop interrupts the wait
+(even though the tool's own thread may finish in the background).
 
 **Setting the token budget to 0 means unlimited, and costs you your only
 safety net.** The loop detector only catches a tool called repeatedly with
--- a/Docs/User_Guide/console/agent-runs-and-tools.md
+++ b/Docs/User_Guide/console/agent-runs-and-tools.md
@@ -963,6 +963,63 @@ may send the approved evidence to its client or model.
   warning for that tab, and quitting the app reports the whole fleet. Details in
   [Console agent runs are screen-scoped](../index.md#console-agent-runs-are-screen-scoped).
 
+### Agent run budget — how long and how expensive one reply may get
+
+Every agent run is bounded by five limits, all in **Settings ▸ Console
+Behavior ▸ Agent run budget**. A "run" is one message you send, from your
+prompt to the agent's final reply — however many tool rounds that takes.
+
+| Limit | Default | What it bounds |
+|---|---|---|
+| Token budget (per run) | 25,000,000 | Prompt + completion tokens spent by one run |
+| Wall-clock limit | 86,400 s (24 h) | How long one run may take end to end |
+| Per-tool-call limit | 3,600 s (1 h) | How long a *single* tool call may take |
+| Model turns | 2,000 | Tool-calling rounds per message |
+| Steps | 25,000 | Individual loop steps (a tool round costs 3) |
+
+Changes apply to your next message — no restart.
+
+**The token budget is the one that actually stops a long run.** This is the
+least obvious thing on this page, so it is worth stating plainly: the whole
+conversation is re-sent to the provider on *every* turn, so cost does not
+grow with the number of turns, it grows with the *square* of it. A run
+whose rounds add roughly 800 tokens each will exhaust a 25M budget somewhere
+around turn 250 — nowhere near the 2,000-turn cap. That is expected. The
+turn and step limits are backstops sized so they never become the surprise
+limiter; spend is the real governor.
+
+That same arithmetic is why raising the token budget alone rarely buys many
+more turns: at ~800 tokens a round, the prompt at turn 250 is already about
+200k tokens, which is where a 200k-context model stops accepting it anyway.
+If you want genuinely long runs, the knob that changes the shape of the
+problem is `[agents] run_log_evict_enabled` (below), not a bigger number
+here.
+
+**Sub-agents inherit these limits rather than sharing them.** Each
+sub-agent gets its own full token budget, so one message that spawns two
+helpers can spend about three times the number you set. Their wall-clock is
+the exception — that comes from `[agents] child_max_wall_seconds`, so
+raising the run's wall limit does not extend theirs.
+
+**Raise the per-tool-call limit if you are running long tools.** A 24-hour
+run budget will not save a crawl, ingest, or build that takes longer than
+the per-tool-call ceiling; that is the limit that kills it. Lowering this
+one below about 186 seconds is the risky direction — a call reported as
+timed out may still be running, and an MCP tool can end up executing twice.
+
+**Setting the token budget to 0 means unlimited, and costs you your only
+safety net.** The loop detector only catches a tool called repeatedly with
+*identical* arguments; a loop that varies anything — an incrementing offset,
+a slightly reworded query — walks straight past it. At a 2,000-turn cap the
+token budget is the last thing standing between a stuck agent and an
+unbounded bill.
+
+**If you lower the step budget, lower it deliberately.** A tool round costs
+3 steps (think, call, result) and the closing reply costs 1, so N turns need
+`3*(N-1)+1` steps. Set steps below that and runs stop on "step budget
+exhausted" well before your turn limit — Settings warns you when the two
+disagree.
+
 ## Common tasks
 
 1. **Run two agents in parallel.** Send a prompt, press **Ctrl+T** for a new
@@ -1004,6 +1061,21 @@ Enter). Tab-fleet keys (Ctrl+T, Alt+1…9, Ctrl+K) are covered in
   preview shows). This caps parallel **tabs**, a different knob from
   `[agents] max_live_subagents` below, which caps parallel **sub-agents
   within one reply**.
+- **Settings > Console Behavior > Agent run budget** — the five limits on
+  one run: token budget, wall-clock, per-tool-call, model turns, and steps
+  (saved as `console.agent_max_total_tokens`,
+  `console.agent_max_wall_seconds`, `console.agent_max_tool_call_seconds`,
+  `console.agent_max_model_turns`, `console.agent_max_steps`). No upper
+  bounds. See [Agent run budget](#agent-run-budget--how-long-and-how-expensive-one-reply-may-get)
+  above for why the token budget, not the turn cap, is what stops a long run.
+- **`[agents] run_log_evict_enabled`** in `config.toml` — whether older
+  rounds are trimmed out of what gets re-sent to the provider each turn
+  (default `false`). This is the companion knob to the token budget: with
+  it off, every turn re-sends the whole conversation and spend grows
+  quadratically. It is off by default deliberately — a weaker model whose
+  recent turns are trimmed tends to redo work it already did and end
+  `stuck`, which is worse than overflowing the window. Turn it on for a
+  model you trust to search its own run log. No Settings UI switch.
 - **`[agents] max_live_subagents`** in `config.toml` — how many sub-agents
   of one conversation may run at once, counting any still working from an
   earlier message (default 3; `1` disables the fleet). No
--- a/Docs/User_Guide/console/agent-runs-and-tools.md
+++ b/Docs/User_Guide/console/agent-runs-and-tools.md
@@ -1001,6 +1001,19 @@ helpers can spend about three times the number you set. Their wall-clock is
 the exception — that comes from `[agents] child_max_wall_seconds`, so
 raising the run's wall limit does not extend theirs.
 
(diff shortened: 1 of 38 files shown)


rmusser01:
## Review pass — findings addressed (a9b4b7e26)

Full review done against 3ecf4a2ea; fixes landed in `4746f6d35` + `a9b4b7e26`, rebased onto current `dev` (no new commits since the original base, so a clean fast-forward).

### Findings fixed

1. **`_CHAT_CALL_TIMEOUT_SECONDS` stale justification** — its comment claimed "2× `CONSOLE_RUN_BUDGET.max_wall_seconds` (1800)", which TASK-18600 made false (wall is now user-configurable, default 86400). Rewritten: the tripwire is deliberately fixed and covered by the gateway's 300s read timeout; a budget-derived multiple would follow the user's wall to absurd values.

2. **"Conservative" was backwards for the cache-write gap** — the gateway's normalization folds `cache_creation_input_tokens` into `prompt_tokens`, so writes are budgeted at 1.0× instead of 1.25×. For a *budget*, under-counting is the **permissive** direction (~25% overshoot on the write portion), not conservative. Wording corrected in the test comment and `_budget_weighted_tokens` docstring; **filed TASK-18607** for the normalization fix (a wire-shape change with its own readers, not something to ride this PR).

3. **`agent_max_tool_call_seconds = 0` silently disabled Stop** — the engine's literal 0 bypasses `_call_with_timeout`, which is the *only* Stop poller (0.5s) while a tool runs. The Console resolver now translates 0 to `UNLIMITED_TOOL_CALL_DEADLINE_SECONDS` (finite, unfireable), keeping the wrapper — and Stop — alive. Regression test added (`test_unlimited_tool_call_seconds_still_wrap_the_cancellation_poll`); Settings help, User Guide, and config comments updated to say what 0 means.

4. **Pillow floor was wrong: 10.1 → 10.4** — `ImageFont.load_default_imagefont()` first appeared in **10.4.0** (per Pillow's CHANGES); 10.1.0 is the `load_default()` proportional-face change. The `>=10.1` floor admitted three versions where `renderer_font()` raises at render time. Fixed in pyproject, requirements, module docs, ADR-054, and the floor test. (This was the PR's own "verify the history" gap — the narrative was confidently specific and wrong.)

5. **Backlog hygiene** — TASK-18602/18603/18604/18606 → `Done` (ACs were already checked); TASK-18605 → `In Progress` with AC#6's disposition recorded (superseded by 18606's unpinning; the underlying guard is green on 12.1.1).

### Qodo findings disposition

| # | Finding | Disposition |
|---|---|---|
| 8 (High) | Cache writes undercharged via gateway normalization | **Accepted as real** → documented honestly + filed TASK-18607 (see finding 2) |
| 4, 5 | Missing Returns/Args docstring sections | Fixed (`renderer_font`, budget handler) |
| 7 | No integration test for the UI | Added: mounted edit → save path → `console_run_budget()` round-trip |
| 1, 2, 6 | input_validation.py / pydantic / duplicated defaults | Rejected — the coerce-helper pattern matches every existing `[console]` numeric key; the defaults duplication is deliberate (no top-level config import in the bridge) and pinned by `test_bridge_default_budget_matches_config_defaults` |
| 3 | No env-var overrides for the new keys | Rejected — no `[console]` key has env precedence; `get_cli_setting` has no env layer. Consistent with precedent |

### Verification

- Touched suites: 321 passed (Chat/Agents files) + 153 (UI/Chat) + 28 (settings budget, incl. new round-trip test)
- Full `Tests/Agents/` + qwencloud: **1,565 passed, 0 failed** on the rebased head
- The 4 failures I saw in `test_settings_configuration_hub.py` on my machine **reproduce identically on clean `origin/dev`** — environment-specific, not from this PR
- Pillow history verified against Pillow's own CHANGES.rst

rmusser01:
## CI status — pre-existing red, attributed (rebased onto latest dev @ a65688c3f)

The `Tests` workflow run on the previous head came back red on all three heavyweight jobs. Attribution below; none of it is from this PR.

**UI Tests (cancelled at 45m):** the job runs 13,558 tests on a 2-worker xdist grid and is killed at ~11% progress by the job's wall budget every run. The last **30 of 30** runs of this workflow across all branches — including `dev` itself — are `cancelled`; there is no recent green run anywhere. Pre-existing infra condition.

**Core Tests (55 failures / 20,500 passed, macOS):** every failing test is in modules this PR does not touch — Notes import planner / git push services, TTS audio supervisor and request admission, LLM_Calls summarization privacy, comfyui adapter, Architecture diagnostic inventory, Character_Chat. I reproduced 10 of them identically on a clean `origin/dev` checkout. Same class as the 17 red Tests/Chat tests this PR already fixed: the suite has been red off-branch for a while.

**Filed [TASK-18608]** to make the workflow green again (shard/split the UI job, triage the 55 core failures).

**What was verified for THIS PR instead:** 1,903 tests green locally on the rebased head (full `Tests/Agents/`, all eight touched Chat/UI files), after rebasing onto `dev` @ `12f984605` (research-jobs merge; no textual or semantic conflicts). The PR's own suites are the merge signal here; branch protections permitting, this is ready.

PR #1836, opened 2026-08-20
fix(console): re-key interrupt-bridge retained payloads by round with FIFO head (task-15661)

rmusser01:
## What

Re-keys the three Console interrupt bridges' retained-payload maps from `session_id` to `round_id`, with a per-session FIFO head, and deletes the three order-dependent TOCTOU guards that existed only to compensate for the old key shape.

Bridges: MCP tool approvals, skill-install confirm, skill-script confirm.

## Why

Each bridge retained the payload its card is re-derived from in a `dict[session_id, payload]` — one slot per session, last-armed-wins. Arming a second round for the same session **overwrote the first's payload**: the first round's card vanished from the screen and the round hung undecidable until its own timeout fired and failed it closed.

This was already known and filed as **task-15661**. The codebase documented it in two places — `request_mcp_approvals`' `finally` block ("Accepted scope limitation ... per-round payload storage is a larger change out of scope here") and `remount_pending_approval_for_active_session`'s docstring — and pinned it with a test whose docstring said it existed so there would be "a failing test the day someone fixes it, rather than folklore in a comment."

That test now asserts the fixed contract.

## How

Five generic helpers (`_park_round_payload`, `_head_round_payload`, `_unpark_round_payload`, `_remount_head`, `_head_round_payload_locked`), shared by all three bridges. The mounted card is always the session's FIFO head — its oldest-armed round. A later sibling queues instead of clobbering, and is promoted when the head resolves.

**Why the guards could be deleted rather than adapted.** Each `_clear_pending_*_if_round_is_current` had two parts: "has a newer round claimed the slot" and "is an older sibling still armed". Both existed *only* because a shared slot made *clearing* order-dependent — whether to clear depended on which sibling resolved first, and a worker-thread snapshot of that answer could be stale by the time the UI thread ran it. Re-deriving the head is order-**independent**: a pure function of current state. The failure mode is now unrepresentable rather than defended.

The race-proofing principle is preserved exactly — the decision still runs inside the callable passed to `call_from_thread`, never from a worker-thread snapshot. Only the decision itself got simpler.

## Scope beyond the obvious

- Four approvals re-derive sites, including the public `remount_pending_approval_for_active_session`
- Both revocation paths (`_revoke_tool_approval_rounds`, `_revoke_skill_script_rounds`)
- A consumer outside the controller: `ChatScreen._current_park_round_ids`
- The script guard had a second caller in `revoke_approval_rounds_for_run` — deleting it without that would have raised `AttributeError` on every cancellation carrying a live confirm

## Behaviour change

No change to the single-round path — card content, timeouts, badges, and decisions are all identical.

Concurrent same-session rounds now queue FIFO instead of clobbering. A non-head round on the **active** session is silent (no card, no toast) until promoted; a non-head round on a **background** session still toasts. That silent-queue gap is recorded in the design doc as a known item for the follow-on unified-interrupt-surface work.

## Testing

New `Tests/UI/test_console_parked_payload_rekey.py` — six tests, written RED before the fix, covering per-bridge FIFO behaviour and cross-bridge independence.

Pre-existing concurrency suites (task-581 / TASK-910) updated: their pinned assertions asserted the *old* single-slot semantics. Where a call-count proxy was used, it was replaced with a content assertion on which round is displayed — under FIFO the re-derive seam legitimately fires on every teardown, so counting calls would fail on correct behaviour.

`170 passed, 4 failed` across the eight suites this branch touches. All four failures are pre-existing on `dev` and were reproduced at the merge base in a separate worktree before any change here:

- `test_skill_install_concurrent_confirms.py::test_bare_shutdown_flag_alone_denies_a_real_session_round_within_one_poll_interval`
- `test_skill_install_concurrent_confirms.py::test_shutdown_flag_alone_denies_both_unregistered_sessions_rounds_and_cleans_accounting`
- `test_console_parallel_runs.py::test_navigating_away_with_busy_fleet_confirms_and_records_teardown`
- `test_console_parallel_runs.py::test_navigation_guard_survives_stay_then_renavigate_then_leave_by_coordinates`

The full project suite was **not** run end to end.

## Net effect

Production is **−144 lines**. Three bespoke guard methods carrying three fix-rounds of accumulated docstrings are replaced by five short helpers no bridge can special-case.

## Design docs

Included in the branch: `Docs/superpowers/specs/2026-08-19-console-user-interaction-design.md` (this is PR0 of that program) and `Docs/superpowers/plans/2026-08-19-parked-payload-rekey.md`.

---
## Summary by cubic
Re-keys Console interrupt bridges’ retained-payload maps from per-session to per-round with a per-session FIFO head, so concurrent same-session rounds queue instead of clobbering (task-15661). Previously a second round overwrote the first and the first hung; now the oldest-armed mounts first and siblings promote on resolve. Mounted cards now show remaining timeout at mount/promotion, and legacy session_id=None cards clear correctly after session switches.

- Review notes
  - Applies to all three bridges: MCP tool approvals, skill-install confirm, skill-script confirm. Deletes the `_clear_pending_*_if_round_is_current` guards; replaces with shared helpers: `_park_round_payload`, `_head_round_payload`, `_unpark_round_payload`, `_remount_head` (decisions still run on the UI thread).
  - Timeout display: payloads carry `deadline_monotonic`; `_head_round_payload` returns a fresh remaining-time snapshot at each re-derive. No-deadline rounds are unchanged.
  - Legacy behavior: `_remount_head(None)` now resolves for the active session when the callback runs, clearing stale cards post-switch and restoring that session’s head; revoke paths benefit from the same cleanup.
  - Active vs background: non-head rounds on the active session are silent; non-head on background sessions still toast. Bridges keep independent heads (cross-bridge independence tested).
  - `ChatScreen._current_park_round_ids` now scans payloads by `session_id` and is key-shape agnostic. Tests updated; new `Tests/UI/test_console_parked_payload_rekey.py`. Known baseline: 4 pre-existing failures on `dev`.

- Required migration
  - Replace any `_parked_*_payloads[session_id]` lookups with `_head_round_payload(store, session_id)` or scan payloads by their `session_id`.
  - Do not count re-derive calls; assert mounted content and FIFO order instead.

<sup>Written for commit 10ed5f84f4e13bbabebd16661e635cef715e2ba0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1836?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/User_Guide/console/agent-runs-and-tools.md
+++ b/Docs/User_Guide/console/agent-runs-and-tools.md
@@ -117,7 +117,10 @@ scoped to the one run that raised it, so deciding one card never resolves or
 touches another's. Cancelling that sub-agent (see
 [Stopping & leaving](#stopping--leaving)) withdraws only its own still-pending
 cards; a sibling sub-agent's card, or the parent's, is left exactly as it
-was.
+was. Within one session, Console shows only one card at a time, oldest-armed
+first — a second round arming for the same session while another is still
+pending queues silently and mounts its own card only once the earlier round
+is decided.
 
 ### Interrupted provider tool runs — Resume, Take over, or Discard
 
@@ -1285,6 +1288,14 @@ advances to `· 5s` on the next poll
 live-walked in a terminal — the states and the elapsed are covered by that
 suite plus mutation testing; the rest of this page's content is unchanged
 from the prior stamp.*
+
+*The one-card-at-a-time-per-session sentence added @ HEAD — 2026-08-19
+(task-15661, PR0 parked-payload re-key: documentation-only for this page.
+A second same-session approval round used to evict the first's card from
+the pre-PR0 shared slot; the round-keyed FIFO map now leaves it mounted
+and queues the new round's card invisibly behind it until the head
+resolves. The rest of this page's content is unchanged from the prior
+stamp.)*
 
 *Headless-wake sections (auto-wake off-screen, wake at launch, headless
 approval, the kill switch) re-verified against dev @ 524194c15 —
--- a/Docs/superpowers/plans/2026-08-19-parked-payload-rekey.md
+++ b/Docs/superpowers/plans/2026-08-19-parked-payload-rekey.md
@@ -584,7 +584,13 @@ Finally, rewrite `_remount_parked_skill_install`'s body to use the head:
 - [ ] **Step 4: Run the tests**
 
 Run: `.venv/bin/python -m pytest Tests/UI/test_console_parked_payload_rekey.py Tests/UI/test_skill_install_concurrent_confirms.py Tests/UI/test_console_skill_install_confirm.py -v`
-Expected: all PASS, including every pre-existing TASK-910 test.
+
+Expected: your new tests PASS and every pre-existing TASK-910 test PASSES, **except** these two, which already fail on `origin/dev` before any PR0 change and are NOT yours to fix:
+
+- `test_skill_install_concurrent_confirms.py::test_bare_shutdown_flag_alone_denies_a_real_session_round_within_one_poll_interval`
+- `test_skill_install_concurrent_confirms.py::test_shutdown_flag_alone_denies_both_unregistered_sessions_rounds_and_cleans_accounting`
+
+The baseline is exactly `2 failed, 26 passed` for `test_console_headless_approval.py` plus `test_skill_install_concurrent_confirms.py` together. A THIRD failure in that pair is a real regression. Do not "fix" the two known ones; if your change makes them pass, say so in your report rather than assuming it is unrelated luck.
 
 - [ ] **Step 5: Commit**
 
--- a/Docs/superpowers/plans/2026-08-19-parked-payload-rekey.md
+++ b/Docs/superpowers/plans/2026-08-19-parked-payload-rekey.md
@@ -167,6 +167,13 @@ def test_the_queued_round_mounts_when_the_head_resolves(controller):
     assert _wait_until(lambda: len(_round_ids(controller)) == 2)
     round_2 = [r for r in _round_ids(controller) if r != round_1][0]
 
+    # Pre-condition: the head still owns the card. Without this, the test
+    # cannot tell FIFO promotion from the arm-time clobber it exists to
+    # catch -- round_2 would already be mounted and the post-assert below
+    # would pass for the wrong reason, on both sides of the fix.
+    time.sleep(0.1)
+    assert _mounted_round(controller) == round_1
+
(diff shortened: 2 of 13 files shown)


rmusser01:
Both Qodo findings addressed in 10ed5f84f (branch also rebased onto current dev tip, clean, no conflicts).

**Finding 1 — queued cards show expired deadline: fixed.** Each bridge's payload now carries `deadline_monotonic` (the same ADR-067 conditional the worker enforces — `None` when no deadline is armed), and `_head_round_payload` returns a shallow remaining-time snapshot at every re-derive. The retained payload is never mutated, so each later mount computes its own fresh remainder; `_park_round_payload`'s `head is payload` identity check goes through the locked variant and is unaffected. This is the single choke point every late mount flows through — promotion at head-resolve, switch-back to a parked session, and headless attach — so it also fixes the pre-existing instance of the same staleness on the TASK-910 background-park path, which predates this PR.

One scope note: only the approval card renders the countdown (`format_approval_deadline`, TASK-1844); the skill confirm cards carry `timeout_seconds` but never display it, so for them the snapshot is inert but uniform.

**Finding 2 — legacy cards survive session switches: fixed, with one narrowing note.** `_remount_head(session_id=None)` now means "resolve the session being viewed when the callback runs"; legacy teardowns pass `None` instead of the arm-time `owning_session_id` snapshot. This clears the stranded card and restores the now-active session's own head if it has one — strictly better than the pre-PR0 unconditional clear, which could wipe a real sibling's card. The revocation paths' existing `None`-session calls pick up the same cleanup for free (previously a silent no-op with the same stranding potential).

The narrowing note: both production wirings bind `session_id` via `functools.partial` (`_compose_mcp_provider`, the review-hook construction), so the no-session path is reachable only from legacy/test callers today. The fix is still warranted — the pre-PR0 code deliberately preserved unconditional-clear semantics for exactly these callers — but the live exposure was lower than the finding's severity suggests.

Both fixes were written test-first: `test_promoted_round_mounts_with_remaining_time_not_original_timeout` and `test_legacy_round_teardown_clears_its_card_after_a_session_switch`, each failing on the exact defect before the fix. Suite result across the eight files this PR touches: 172 passed, 4 failed — the four failures are the pre-existing dev baseline documented in the PR description, unchanged (CI's own run independently shows the same four in this area).

PR #1883, opened 2026-08-21
feat(console): Conversation Inspector — exchange capture + unified cost/context review (task-18300)

rmusser01:
## What & why

Clicking the Console token/cost chip previously opened a numbers-only breakdown; the full next-send payload viewer was hidden behind Ctrl+Shift+P; and nothing recorded what actually went over the wire on past turns. Users could not answer *"what exactly is being sent back and forth in my conversation, and what did each piece cost?"*

This ships a unified **Conversation Inspector** and the capture pipeline behind it.

Spec: `Docs/superpowers/specs/2026-08-18-console-conversation-inspector-design.md` (six recorded owner decisions)
Plan: `Docs/superpowers/plans/2026-08-18-console-conversation-inspector.md` (11 TDD tasks)
Task: **task-18300**

## What shipped

**UI** — one modal, opened from the cost chip, Ctrl+Shift+P, and the command palette:
- **Costs** — per-message rows + totals (unchanged ground truth), each expanding into that turn's captured provider calls with per-call cost.
- **Exchange** — per turn → per call → System prompt / Messages / Tools / Response / Tool calls / Sampling & routing, plus an "omitted by capture policy" line, status badges (`[stopped]` / `[error]` / `[abandoned regeneration]`), per-call Copy/Save. Per-piece token figures are labelled **estimates**; provider-reported buckets stay authoritative.
- **Next Send** — the former Ctrl+Shift+P viewer, ported intact.
- `ConsoleCostModal` and `ConsoleContextModal` are retired, their behaviour pins migrated.

**Capture** — every Console provider call (each agent tool-loop iteration, direct sends, and the llama.cpp branch) is captured request+response at the gateway seam and persisted locally per turn.
- Requests are **allowlist-sanitized by construction** (`CAPTURE_REQUEST_ALLOWLIST`); `api_key`/`api_key_resolved` are deliberately off it and surface as `omitted_keys`, so a future secret-bearing kwarg is safe the day it is added. Pinned by a property test.
- Binary/base64 content is stubbed `[mime, size, sha256:…]` — identity verifiable without the bytes.
- Stopped turns keep partial content marked `stopped`; abandoned regenerations are **kept** and marked; ephemeral sessions keep captures in memory only.

**Storage** — ChaChaNotes **v43** `message_exchanges`: local-only (no `sync_log`, no FTS), upsert on `(message_id, run_tag, seq)`, zlib blobs, cascade delete.

**Config** — `[console] exchange_capture = true` kill-switch, read through the resolved settings layer; when off, no capture objects are built at all.

## Honesty caveats (stated in the UI and the User Guide)

- Capture happens at the **provider-adapter boundary**, so provider-internal HTTP framing and adapter-injected `cache_control` markers are not shown. The llama.cpp branch is the exception — there the capture *is* the literal wire payload.
- Turns recorded before this feature, with capture disabled, or where capture failed render an explicit "no capture recorded" row.

## Included upstream repair (disclosed)

Migration `.sql` files are enumerated by name in **four** packaging registries. Our v40→v41 file *and* the pre-existing **v39→v40** file (task-17169) were missing from all four — an installed wheel/sdist could not open the database, and the release checker validates against the same list so it could not catch it. Both are now registered; verified against a real `uv build` sdist + wheel.

## Testing

Targeted gate, green on the merged tree: **774 passed** across capture/gateway/inspector (303), DB/store/controller (312), UI wiring + modal inventory (159). Full programme gate additionally ran `Tests/DB/` (968 passed; one pre-existing `numpy`-missing failure, environment-only).

**Live verification against a real provider (~$0.006), reported honestly:**

| Scenario | Result |
|---|---|
| Multi-call agent turn with a tool | **PASS** — 4 calls; tool schemas in call 0; tool-role message in a later call; correct answer |
| System prompt byte-for-byte | **PASS** |
| Kill-switch (both directions, after restart) | **PASS** |
| Mid-stream Stop | **PARTIAL** — `[stopped]` badge confirmed twice, but only with empty partial content |
| Regenerate-then-Stop | **NOT VERIFIED** — four honest attempts raced past natural completion |

`task-18300` is deliberately left **In Progress** at 4/8 acceptance criteria; each unticked criterion states exactly what was and was not demonstrated. Unit coverage exists beneath all three gaps (`test_consumer_abort_mid_stream_closes_capture_as_stopped`, `test_variant_restored_message_keeps_captures_marked_abandoned`, and siblings).

## Rebased onto dev (2026-08-21)

Rebased from `1bdbcac61` onto current `dev` (657 commits of drift). Four things worth a reviewer's attention:

- **The schema version race happened.** `dev` took v41 (persona visual) *and* v42 (Console project context) during review, so this branch's `v40→v41` collided head-on with a migration of the identical name. Renumbered to **v42→v43**: migration method, `.sql` file, all four packaging registries, and the DB tests. `dev`'s `_migrate_from_v40_to_v41` is untouched.
- **`dev` built a feature into the modal this PR deletes.** `ConsoleContextModal` gained a project-instructions panel (`00814de48`, `a6709f169`, `b236d056c`, `12c6e3b62`). Keeping the deletion alone would have silently dropped it, so it is ported into the Next Send tab — panel, `AUTO_FOCUS`/focus-order handling, the recovery flow, and **the privacy guarantee**: Copy JSON and Save to File drop any message tagged `EPHEMERAL_ORIGIN_KEY == "project_instructions"`, so the disposable preview is the only surface that ever shows an instruction body. All 15 of dev's tests re-targeted, none weakened.
- **Two resurrection bugs caught.** `dev` deleted `_run_dictionary_summary_off_thread` and the `collections_tag_window` module; taking "ours" wholesale would have revived both. Dropped.
- **`dev`'s improvements preserved over ours where they overlapped** — session-scoped draft capture and the `session_id`-aware snapshot call now live inside our factory helper, and `dev`'s robust `assert current_schema_version > 35` replaced this branch's brittle `== 41`.

Post-rebase gate: capture+gateway **276**, store+controller **305**, UI inspector suites **88**, `Tests/DB/` **1053 passed**. Collection sweep 51,748 tests, and the 2 collection errors this branch introduced are fixed (the remaining 30 are missing optional deps — `numpy`, Confluence, MLX — and reproduce on clean `dev`).

Pre-existing failures confirmed against a clean `origin/dev` checkout, **not** from this branch: 6 media privacy tests, `test_chat_image_db_compatibility` (numpy), `test_sql_validation::test_no_missing_tables` (dev's four `persona_visual_*` tables are absent from `VALID_TABLES` — worth a follow-up on dev), and 2 `test_console_modal_dismissal` cases (dev's `ProjectInstructionNoticeModal`/`SetupModal` have no dismissal contract).

Sibling version pins that hard-coded the current schema number were made dynamic against `_CURRENT_SCHEMA_VERSION` — they had already been re-hardcoded at 35, 39, 40 and 41, and still catch an accidental downgrade.

## Reviewer notes

- `Docs/security/production-diagnostic-inventory.json` is stale repo-wide (red before this branch). Note its entry for `console_context_modal.py`, **a file this PR deletes** — that is not merge drift.

## Post-rebase review: a Critical privacy defect, found and fixed

The rebase follow-on work got its own independent review (the rest of the branch had already had one). It found that the project-instructions port faithfully reproduced `dev`'s export filter for the **Next Send** tab — but this branch adds a **second** export surface `dev` never had: the Exchange tab's per-call **Copy JSON / Save to File**, which serialized the raw capture. Since `CAPTURE_REQUEST_ALLOWLIST` admits `messages_payload` verbatim, the auto-loaded `AGENTS.md` body was reaching the clipboard and `~/Downloads`, was **displayed** on the Exchange tab, and was **persisted at rest** in `message_exchanges` — with capture on by default. That falsified this branch's own shipped guarantee ("Copy JSON and Save to File omit automatic instruction bodies… saved conversation state carry only content-free metadata") in three ways at once.

Fixed at the **capture layer** rather than in the two export methods — redaction now happens in `build_request_capture`, the same seam that keeps `api_key` out, so export, display and at-rest are all covered by one change. Filtering only the exports would have closed one hole and left the other two claims false.

Verified independently of the implementer's tests:

```
tagged row content : '[project instruction body omitted by capture policy -- 39 chars]'
tagged role kept   : user
untagged preserved : True          # identical text, untagged -> NOT redacted
api_key omitted    : True | in blob: False
SENTINEL count in serialized capture: 1   (the untagged row only)
```

Red-proofed: reverting the redaction makes the new tests fail with the body visible in the serialized capture.

Same review also fixed: focus-stealing off the Next Send tab (a slow snapshot load yanked focus to **Close** mid-interaction, and the cost chip landed focus on Close so Enter dismissed the modal); a determinism hole where the stub length-gate measured the raw value while the hash used canonical bytes; an uncleared blob cache in `restore_state`; two docstrings that overstated what their guard catches; and a missing save-block re-check.

## Known red, not from this branch

The **"No duplicate backlog task IDs"** check fails on seven pre-existing collisions between other contributors' tasks (16320, 16322, 16323, 16324, 18912, 18913, 18915) — each pair is present on `origin/dev` already (verified: 2 files per id). This PR's own ids (18300, 19324, 19325) are unique. Renumbering other people's in-flight tasks is not this PR's call; flagging for an owner decision.

## Follow-ups filed

- **task-19324** — llama.cpp `stream`→`complete` fallback makes a second, uncaptured HTTP request.
- ~~**task-19325** — exchange flush re-compresses every prior capture~~ — **fixed in this PR** (Qodo finding 4) and closed.

Four low-severity items from the final review are recorded in the branch history rather than filed: a `deepcopy` sitting just outside its guard, a widened base64 false-positive class, truncated captures no longer badged, and a negative test pin that would go vacuous if one call site is refactored.

## Found by the feature itself

During live verification the inspector revealed that an intermediate **tool-role message's `content` carries the tool's static description string instead of the JSON result** (the final answer was still correct, which is why it had gone unnoticed). This PR cannot have caused it — the branch changes zero lines in `Agents/` and `console_agent_bridge.py`, and the capture layer is read-side only. Worth its own investigation.


rmusser01:
/review

Re-review requested: the branch has changed substantially since the last Qodo pass, which reviewed the pre-rebase code.

Since then it was rebased onto current `dev` (657 commits of drift; schema renumbered v40→v41 ⇒ **v42→v43** after dev took both), all four previous Qodo findings were fixed, and an independent review of the post-rebase work found and fixed a **Critical** privacy defect: the Exchange tab's per-call Copy/Save exported auto-loaded `AGENTS.md` bodies, which were also displayed and persisted at rest. Redaction now happens in `build_request_capture` (the same seam that keeps `api_key` out), so export, display and at-rest are covered by one change.

Areas most worth your attention: `console_exchange_capture.py` (the redaction + base64 canonicalisation), the Exchange-tab export paths in `console_conversation_inspector.py`, and the rebase resolutions in `chat_screen.py` / `ChaChaNotes_DB.py`.

rmusser01:
Thanks — both focus areas checked against the code rather than answered from intent.

**1. Silent Data Loss — deliberate, and the trade-off is the spec's.**

This is by design, not an oversight. The binding constraint is *"capture must never break a send"*: a capture failure degrades to "no capture" and the conversation continues. Losing an observability record is strictly better than failing a turn the user paid for, and the same posture is applied at every layer (builder failure → warning; serialization failure → the row-building sits *inside* the try, deliberately; write failure → `False`, logged; blob-decode failure → that row skipped, rest of the turn intact).

The loss is also not silent to the user: a turn with no captures renders an explicit **"No capture recorded for this turn (recorded before capture existed, capture disabled, or capture failed)"** row rather than a blank, and a loader failure renders a distinct *retryable* message that un-marks the row so re-expanding retries. Not adding a retry for the attach path: it would have to run on the send path, which is exactly what the constraint forbids.

**2. Repeated Persistence — real concern, already fixed in this PR.**

You flagged this in the previous pass (finding 4) and it was fixed in `4261f6d32`. Re-attaching the full collection no longer re-serializes: blob compression is memoized per message, keyed `(run_tag, seq, status)`, so an unchanged capture is compressed exactly once no matter how many flushes occur. `status` is part of the key on purpose — a `stopped` snapshot superseded by the closed capture *must* re-compress — and the cache is evicted at five sites (message delete, node removal ×2, `restore_state`, plus per-flush pruning).

Verified rather than asserted — `_attach_stream_usage` does have four call sites (`console_chat_controller.py:10798, 11056, 11087, 11118`), so repeat attaches are reachable, and the behaviour is pinned by tests:

```
test_persist_exchanges_only_compresses_each_capture_once_across_flushes
test_persist_exchanges_only_recompresses_a_superseded_stopped_capture
test_exchange_blob_cache_does_not_leak_between_messages
test_restore_state_clears_exchange_blob_cache_and_abandoned_run_tags
→ 5 passed
```

The DB write is an upsert on `(message_id, run_tag, seq)`, so a repeat attach is idempotent at rest as well.

PR #1903, opened 2026-08-21
refactor(console): unified interrupt-round host (sub-project C1)

rmusser01:
## What

Extracts the Console's interrupt-round lifecycle into a new `tldw_chatbook/Chat/console_interrupt_rounds.py` (`InterruptRoundHost`) and migrates all three bridges onto it: MCP approvals, skill-install confirm, skill-script confirm.

**No user-visible change.** This is the foundation for the Console agent↔user interaction program; the follow-on sub-projects add the surfaces.

## Why

Each bridge carried its own inlined copy of the same lifecycle — arm → park/mount → 1s poll → resolve/timeout/revoke → teardown re-derive — plus its own registry, its own lock, and its own re-derive helper. Three copies of a concurrency-sensitive loop is three places for them to drift, and the recent parked-payload defect (task-15661, PR #1836) had to be fixed three times for exactly that reason.

## How

`InterruptRoundHost` owns one lock, per-kind storage, the five payload helpers moved from PR #1836, and one generic `run_round(kind, …)`. Per-bridge asymmetries ride explicit hooks rather than branches:

| | approvals | skill-install | skill-script |
|---|---|---|---|
| `check_revoked` | default (True) | **False** — never swept | default (True) |
| `announce_detached` | detached-view leg | — | — |
| `on_cancelled` / `on_timeout` | stamps decisions + audit-logs | — | — |

The three bridge methods keep their exact public names, signatures, and return types. Nine legacy attribute names (three locks, three registries, three payload maps) are **aliased onto the host's own objects**, because 11 test files and production `ChatScreen._current_park_round_ids` read them directly — so no consumer changed.

Four re-derive call sites (`new_session` / `switch_session` / `close_session` / `remount_pending_approval_for_active_session`) collapse from three per-kind calls each to one.

## Invariants, verified rather than asserted

- **UI-thread decision rule** — the mounted-card decision is computed inside the `call_from_thread` callable, never snapshotted on a worker thread. This is what three prior fix rounds converged on. It is now slightly *stronger*: the host re-reads the setter inside the callable where the old code bound it as a worker-thread argument, so a setter torn down between enqueue and execution is caught.
- **One non-reentrant lock behind three names** — the branch's sharpest hazard. All 16 acquisition sites (10 production, 6 test) are flat; every call that re-takes the lock is dedented outside the `with`. Notably both revoke sweeps *already* ran their unpark outside the critical section, with a comment explaining why, before this branch merged the locks.
- **FIFO by construction** — the only production write into a payload map is `park_round_payload`; the only write into a registry is `run_round`.
- **Alias integrity** — exactly ten assignments repo-wide, all in `__init__`, all bare references (never copies), zero reassignments.

## Deliberate deferrals

- **The revoke sweep stays in the controller.** It operates entirely through the aliased registries and host helpers, so moving it changes no behavior until a fourth round kind needs sweeping. The follow-on sub-project that adds one will move it.
- **Six now-dead helpers are left in place** (`_marshal_pending_*`, `_head_round_payload_locked`, and two shims). Deleting them would edit the existing test battery — one is named in a test docstring — and that battery is this refactor's byte-parity oracle. The next sub-project touches this layer and owns the cleanup.

## Testing

Strict parity: the existing interrupt battery passes **unchanged** — no existing test file was modified. Baseline at the merge base was 0 failed / 220 passed, so any failure would be a regression rather than something to excuse.

`236 passed, 0 failed` (220 pre-existing + 16 new host unit tests) on the rebased tree.

The new `Tests/Chat/test_console_interrupt_rounds.py` exercises the lifecycle against a 20-line `FakeSeams` with no `ConsoleChatController` involved — the testability that motivated extracting to a module rather than generalizing in place.

Five tests fail on this branch **and on clean `origin/dev`** (verified in a detached worktree at `e33e59762`): two in `test_console_skill_script_confirm.py`, three in `test_console_fleet_wake_safety.py`. Pre-existing, not introduced here.

## Honest accounting

The lifecycle collapse is real: three inlined copies became one 93-line `run_round`. The payload layer, though, is *displacement* rather than de-duplication — PR #1836 had already de-duped it — and raw production line count is roughly a wash (−334 controller / +286 host). What the move actually buys is unit-testability of a concurrency path that previously could only be tested through a 12,000-line controller.

## Design docs

Included: `Docs/superpowers/specs/2026-08-20-console-interrupt-host-design.md` and `Docs/superpowers/plans/2026-08-20-console-interrupt-host-c1.md`.

---
## Summary by cubic
Unifies the Console interrupt-round lifecycle for approvals, skill-install, and skill-script into a single host with one lock and shared storage, with no user-visible changes. This removes three divergent concurrency loops to reduce defects and enable direct unit testing.

- New `InterruptRoundHost` in `tldw_chatbook/Chat/console_interrupt_rounds.py` owns a non-reentrant lock, per-kind registries/payload maps, and a generic `run_round(kind, ...)`.
- Bridges now delegate to the host; public method names, signatures, and return types are unchanged.
- Nine legacy attributes (three locks, three registries, three payload maps) alias the host's objects to preserve existing consumers and tests.
- Re-derive sites in the controller collapse to single calls; the revoke sweep stays in the controller for now.
- Tests: existing interrupt suites pass unchanged; new unit tests cover the host against a minimal seams double. Design and plan docs added under `Docs/superpowers`.

<sup>Written for commit 09dbc187e8908ab309d6be088160d95fcf10683a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1903?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Qodo's finding is correct and is fixed in 09dbc187e8 (docstring only, no code change).

The module claimed `FakeSeams` was "the full surface the host may touch." That was true when written — Task 2 added only the payload helpers — and went stale the moment Task 3 landed `run_round`, which also calls `_is_session_cancelled`, `add_pending_round`/`discard_pending_round`, and reads `park_pending_approval`. The docstring now splits the surface by entry point: payload helpers and `remount_head` need `FakeSeams`; `run_round` needs the `FakeSeamsFull` superset. I enumerated all eight `_seams` accesses in the module to confirm the new text is complete rather than merely less wrong.

Worth noting this is the third stale-docstring finding in this program's lineage, all the same shape: a claim that was true when written and falsified by a later commit in the same branch.

---

**Status note for whoever picks this up: the branch is now substantially stale and I do not recommend merging it as-is.**

Since this PR was opened, `dev` has advanced **3,053 commits**, **157** of which touch the two files C1 changes. `console_chat_controller.py` has grown from 12,482 to 24,045 lines — it roughly doubled.

I attempted the rebase and aborted it. Four conflicts, one spanning 800+ lines — but the textual size is not the problem. The problem is semantic: `dev` has rewritten the approvals bridge's verdict keying from `llm_name`-keyed to call-id-keyed (native `call_id` when present, falling back to `llm_name`), a ~245-line change inside exactly the code C1's migration restructures. C1's `_on_cancelled`/`_on_timeout` hooks stamp that decisions map, so rebasing is not conflict resolution — it is re-deriving the migration against semantics that changed underneath it.

Also relevant to any re-verification: the interrupt battery's baseline on clean `dev` is now **15 failures / 231 passed**, up from 0 at this branch's merge base three weeks ago. Two consecutive runs of the same `dev` tree gave 15 and 14, so at least one of those is flaky. Any future parity claim for this work needs a fresh detached-worktree baseline, not this PR's numbers.

The design (`Docs/superpowers/specs/2026-08-20-console-interrupt-host-design.md`) is unaffected by all of this — the three-bridge duplication it removes is still there, and dev's own churn in this area arguably strengthens the case. What is stale is the implementation's fit against the current file.

rmusser01:
Closing without merging. This branch went stale in a way that isn't a rebase problem: `dev` rewrote the approvals bridge's verdict keying (call-id-keyed, ~245 lines) inside exactly the code this refactor restructures, and the controller file doubled in size underneath it. Carrying it forward would mean re-deriving the migration, not resolving conflicts.

The design it implemented is preserved at `Docs/superpowers/specs/2026-08-20-console-interrupt-host-design.md`, and the durable pieces — the self-contained `InterruptRoundHost` module and its unit tests — are recoverable from this branch's history if the extraction is picked up again.

More to the point, the PRD that now defines the two features this was foundation for (#2368, merged) puts a unified interrupt-surface refactor explicitly out of scope: both features ride the existing per-kind round machinery as a fourth kind. If a later feature needs the unification, it should start from the spec and current `dev`, not from this branch.

PR #1914, opened 2026-08-22
Change review: git commit/push/PR modes (TASK-16801)

rmusser01:
## What this adds

A `Working tree (current)` mode on the Change Review screen that reads the **real** git repository (not the shadow tracker), plus confirmed **commit**, **push** and **open-PR** actions. This is arc B of the V2 turn-file-card design; arc A (the review rail) merged as #1855.

Spec: `Docs/superpowers/specs/2026-08-20-console-review-git-modes-design.md` · Plan: `Docs/superpowers/plans/2026-08-20-console-review-git-modes.md` · Task: TASK-16801

- **Commit** is file-picked from a *fresh* working-tree read at modal-open, needs a message, offers an optional new branch, warns (never blocks) on detached HEAD and main/master, and is refused while an agent run is active or a merge/rebase/cherry-pick is in progress.
- **Push** is confirmed, sets upstream on first push, and — unlike commit — is *not* refused during an active run, since it ships already-committed state.
- **Open PR** opens a compare URL for github.com, gitlab.com, bitbucket.org and codeberg.org via `app.open_url`.
- Kill switch `[change_review] git_actions` (default on) removes the feature entirely.
- Snapshot-only features (revert, comments) gate off in current mode, so the pseudo-row's `id=-1` can never reach the notes DB.

## Security: four repository-supplied argv vectors, found in review

None of these put a dangerous flag in our own argv — the **repository** supplied it, which is why the existing no-force assertions all passed. Each was reproduced destroying real data before being fixed:

| Vector | Proven damage |
|---|---|
| Remote named `--force` / `--mirror` | `+ 33b7b99...4a3c4a7 main -> main (forced update)`, another clone's commit gone |
| Branch named `--mirror` via `.git/HEAD` | `refs/heads/release` deleted, a private branch leaked |
| `remote.push` / `remote.mirror` / `push.default=matching` | forced update, deleted refs, and an unrelated branch's private commit published while the modal named another branch |
| Pathspec magic in a filename (`:!nothing`) | a 1-file selection committed 4 files |

Fixes prefer *only known-good values reach argv* over asserting the bad thing is absent: refuse option-shaped remote and branch names, pass an explicit fully-qualified refspec on both push forms, and set `GIT_LITERAL_PATHSPECS=1`. `diff.external`/textconv could also make the review pane render a fabricated diff (or a blank one for a genuinely changed file), so both `git diff` sites now pass the machine-safe flags.

Note for reviewers: the engine-side refusals are load-bearing on their own. The UI check cannot catch the remote vector, because the hostile name genuinely *is* a detected remote — please don't drop the engine guard as redundant.

## Verification

- **629 tests** across the arc's suites, two consecutive foreground runs; `--collect-only Tests/` = **52,273 collected, 0 errors**; ruff clean.
- Every engine and end-to-end test drives a **real temp repository with a real local bare remote** — no mocked git.
- Guards were **mutation-tested, not trusted**: reviewers independently killed the index-hijack pin, both argv refusals, the active-run refusals, the stale-land guards and the mode gating, confirming a test dies for each. Security fixes were proven red on the *destruction* (the other clone's commit, branch and tag surviving), not on a raised exception.

## Deliberately not done

TASK-19700 (agents can write `.git/` inside a workspace root — the upstream cause of the threat model above), TASK-19701 (`pushurl`/`pushInsteadOf` redirects), TASK-19702 (Default workspace never sees Change Review at all — predates this arc), TASK-19703 (three honesty/polish items).

Lesson recorded in `backlog/docs/lessons-testing-evidence.md`: *"no dangerous flag in our argv" is not a guarantee — audit by tracing each argv slot to its source, never by asking whether a validator exists somewhere.*

---
## Summary by cubic
Adds a Working tree (current) mode that reads the real repository and enables confirmed Commit, Push, and Open PR actions. Previously review was snapshot-only and read-only; now users can ship changes from the console with hardened git calls and honest outcomes.

- Behavior: review now has dual modes; snapshot-only affordances disable in current mode; Commit requires a fresh status read and a message, can optionally create a branch, and is refused during an active run or while merge/rebase/cherry-pick is in progress; Push sets upstream on first push and is allowed during an active run; Open PR supports github.com, gitlab.com, bitbucket.org, and codeberg.org; commit path handles staged renames/deletions safely.
- Security: refuse option-shaped remote and branch names; push always uses an explicit fully-qualified refspec from `%(upstream:remoteref)` and refuses when upstream cannot be resolved; commit runs with `GIT_LITERAL_PATHSPECS=1` and scrubs conflicting pathspec env; diffs pass machine-safe flags to ignore `diff.external`, textconv, and color; never force-push.
- Qodo review fixes: unborn HEAD status/diff routing now handles staged adds correctly; untracked preview no longer follows symlinks outside the workspace; `[change_review] git_actions` now uses boolean coercion so string values cannot defeat the kill switch.
- Wiring: new `tldw_chatbook/Workspaces/git_workspace.py` (detection/status/diff/commit/push/PR URL); `ChatScreen` passes live `workspace_roots` so current mode appears without prior turns; worker landings catch teardown `RuntimeError` only; kill switch `"[change_review] git_actions"` (default on) hides the feature entirely when disabled.
- Rollout: no migrations or API changes; feature gated by `"[change_review] git_actions"`; current mode appears only when the conversation’s workspace roots resolve to a git repo; satisfies TASK-16801 acceptance criteria.

<sup>Written for commit a782670c51a6f055962e356e33484ed3cc811ce4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1914?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## Rebased onto dev `3d79dcb83`

Two conflicts, both resolved and verified rather than assumed:

**Generated CSS bundle** — regenerated via `python -m tldw_chatbook.css.build_css`, never hand-merged (repo rule).

**`test_console_modal_dismissal.py` reachable-modal count** — a semantic conflict git could not see: dev unwired two Console modals (42) while this branch adds two (the git commit and push/PR modals). Resolved to **44**, and since that assertion sits *after* a line that fails for an unrelated pre-existing reason, I verified the number by driving the real walk directly rather than trusting arithmetic:

```
ACTUAL reachable modal types: 44
git modals present: ['ChangeGitCommitModal', 'ChangeGitPushModal']
```

## Gate after rebase: 639 passed, 2 failed

Both failures are **pre-existing on dev**, not from this branch — baselined against a throwaway worktree pinned to the exact dev SHA `3d79dcb83` (not a moving ref), which reproduces them identically at *2 failed, 115 passed*:

- `test_console_modal_launch_declarations_match_runtime_construction`
- `test_console_modal_inventory_matches_runtime_ast_and_transitive_launches`

Dev's own in-file comment documents this gap: `ProjectInstructionNoticeModal` and `ProjectInstructionSetupModal` are real Console modals that no contract declares, left un-widened deliberately so the assertion stays honest. This branch neither causes nor papers over it.

rmusser01:
## Qodo review addressed — all three, commit `3b328d0de`

Each finding was reproduced empirically before fixing, and each fix is pinned by a test proven red first. Gate after: **645 passed** (was 639, +6), ruff clean.

### 3. Unborn head diff breaks — High ✅ correct, fixed
Confirmed exactly as described. On a fresh repo, `git add staged.txt` yields porcelain `A  staged.txt`, which is not `??` and so absent from `untracked`; it then reached `git diff HEAD`, which is fatal before the first commit. Red-first output:

```
AssertionError: a staged add on an unborn HEAD must not render as unavailable:
"diff unavailable: git diff failed (128): fatal: bad revision 'HEAD'"
```

Root cause was a wrong assumption in my own docstring — *"unborn ⇒ every file is untracked"* — which held in the existing test only because that test created untracked files exclusively. Now routes on `GitWorkspaceInfo.unborn`, per your screen-side suggestion, with the docstring corrected to say why.

### 1. `untracked_preview()` bypasses `path_validation` — Medium ✅ correct, and worse than rated
Reproduced: an untracked symlink pointing outside the workspace had its target's content rendered into the review pane.

```
PREVIEW OUTPUT: 'new file: escape_link\n\n+SECRET-OUTSIDE-CONTENT\n'
LEAKED OUTSIDE CONTENT: True
```

Worth flagging why this is more than a display bug: agent write tools can plant such a link in the very root this feature operates on, and the review pane's text flows **back to the model** through the V1.5 annotate/delivery loop — so a rendered `~/.ssh/id_rsa` is an exfiltration path. Fixed via `path_validation.is_safe_path` (which resolves symlinks before comparing, so it covers traversal segments and link escapes alike), matching CLAUDE.md's boundary rule. A control test pins that legitimate nested paths still render.

### 2. Kill switch mis-coerced — Medium ✅ correct, fixed with one deviation
`bool("false")` is indeed True. Now uses `coerce_bool_setting` as you suggested.

Two notes from verifying rather than assuming its behaviour: it is **asymmetric** — `"off"`/`"no"` read as False but `"on"` does *not* read as True — so my test pins the shared helper's real vocabulary rather than a wider one this feature would have invented. And `coerce_bool_setting(None, True)` returns `None`, not a bool, which would read as falsy and silently disable a feature that ships ON; added a `None` guard so the function keeps its `-> bool` contract.

---

The two `test_console_modal_dismissal.py` failures in the gate are **pre-existing on dev** (`ProjectInstruction*` modals undeclared), baselined against a worktree pinned to dev tip `599544bad` at an identical *2 failed, 115 passed*.

PR #1916, opened 2026-08-22
docs(backlog): TASK-19602 tranche 3 notes — reader-seam bug fixed, remaining-16 mapped

rmusser01:
Docs-only: records PR #1915's outcome (4 cleared, 1 real bug — the prompt-field reader dropping unsaved edits) and the diagnostic map for the remaining 16 failures.

---
## Summary by cubic
Documents tranche 3 outcomes for TASK-19602 and the fix to the library prompt editor reader: previously copy/save/export ignored unsaved edits (persisted detail won); now live, non-empty textarea values take precedence with a fallback for structured/foreign lanes to avoid blanking.

- Clears 4 more failures (41 of ~50 total now cleared); maps the remaining 16 for the next tranche.
- Aligns three label expectations with recent relabels.
- Test updates:
  - Use the editor’s edit seam (`_change_field` with per-lane publish) instead of textarea assignment to avoid reverts on lane change.
  - Filter or wait out the lifecycle “Library tools are now available.” toast to prevent double-notify assertions.
  - Update selectors/flows for the prompts rail (collapsed/renamed at some sizes) and a gated/moved export button.
  - Triage `app_config` WorkerFailed and `file_notes` issues (keyboard source choices, database-files return, False-is-True cases).

<sup>Written for commit a647df9efdf97b86a28115041ca2f896574031d8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1916?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #1917, opened 2026-08-22
fix(tests): TASK-19602 tranche 4 — lifecycle-profile wrappers clear the rail-row stranding (16→5)

rmusser01:
Tranche 4: the big unlock was diagnosing why rail-row failures passed in bare asyncio probes but failed under pytest — the sandbox factory builds a NEW Library profile (lifecycle UNKNOWN → 2-row landing rail) while bare runs read the real config's graduated state. Legacy-profile wrappers installed in both affected files (mirroring test_library_shell.py's), plus export-refusal seeding, toast-tolerant notify asserts, and the discard clean-state alignment. Differential: 16 → 5 unique failing tests, no regressions. Remaining 5 mapped with root-cause evidence in the task notes.

---
## Summary by cubic
Force legacy Library profile in pytest to match the real graduated lifecycle and unblock UI tests. Previously the pytest sandbox built a new profile (lifecycle UNKNOWN → 2-row landing rail), which stranded rail-row interactions and broke the Database-Notes switch; now tests default to the legacy profile with an opt-out.

- Add local `_build_test_app` wrappers in `test_library_file_notes_workspace.py` and `test_library_prompts_canvas.py` that set `library_new_profile_admission=False` by default; pass `preserve_profile_admission=True` to test new-profile behavior (TASK-19602).
- Relax notification assertions to tolerate an earlier lifecycle graduation toast; check message content on the latest call instead of enforcing a single call.
- Seed a placeholder Prompt in the export-refusal test so the action row renders in server mode.
- Align Discard button expectations to the always-enabled clean state; only in-flight writes disable it.
- Result: unique failing tests drop from 16 to 5; remaining cases are tracked under TASK-19602.

<sup>Written for commit b2127e613220f41cc187fbe2750adb9767f21d13. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1917?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_library_file_notes_workspace.py
+++ b/Tests/UI/test_library_file_notes_workspace.py
@@ -62,7 +62,22 @@ from Tests.UI.test_library_shell import (  # noqa: E402
     _two_conversations,
     _wait_for_library_shell,
 )
-from Tests.UI.app_factory import _build_test_app  # noqa: E402
+from Tests.UI.app_factory import _build_test_app as _build_tldw_test_app  # noqa: E402
+
+
+def _build_test_app(*args, **kwargs):
+    """Force the legacy (graduated) Library profile this suite assumes.
+
+    TASK-19602: under the pytest sandbox the factory builds a NEW profile
+    (lifecycle UNKNOWN -> landing surfaces), which strands the
+    Database-Notes switch contract; mirrors test_library_shell.py's
+    wrapper. Pass ``preserve_profile_admission=True`` for new-profile
+    tests.
+    """
+    app = _build_tldw_test_app(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


rmusser01:
Guard queued under the standing congestion; exact-expression check on this branch: 0/0 duplicate IDs. Differential evidence in the task notes. Merging on local evidence like #1900–#1916.

PR #1922, opened 2026-08-22
fix(notes): preserve the losing side of a sync conflict and implement disk_wins (task-19554)

rmusser01:
## Summary
Closes task-19554 (P0, data loss). Notes sync overwrote the losing side of a `both_changed` conflict **wholesale, with no recoverable copy** — the `SyncConflict` dataclass carried both texts but only hashes were persisted, and the `sync_conflicts` table had no content columns. No `.bak`, no note history. Aggravating: the Library UI deliberately offers only destructive strategies, and auto-sync runs unattended every 300 s.

Second defect: `"disk_wins"` was offered in the UI but `DISK_WINS` appeared in the engine exactly once — its enum definition. Selecting it recorded a conflict and applied nothing, which the user reads as "synced".

## Preservation design: sidecar first, DB row second, fail-closed
Before any resolution overwrites a side, that side's text is written verbatim to `<note>.conflict-<UTC>-<db|disk>.bak` beside the note file, through the existing pinned-root boundary. Only then is the winner applied.

**The sidecar is primary** because it is the only surface a user can already reach with no new UI: it lands in the folder they are already syncing, sorts beside the file it came from, names which side and when, and recovery is a `mv`. It is deliberately **byte-exact** — no header, no diff markers — because anything prepended turns recovery from a rename into an edit.

**The DB row as well** (`losing_side` / `losing_content` / `preserved_file_path`, schema v43→v44) because the criterion asks for reconstruction, not a hash: it survives the folder being tidied up and is the only programmatically reachable copy. Written **only when a side is actually discarded**, never on mere detection — `sync_log` is already an unpruned full-content shadow of `notes`, and a second unconditional one would be a real regression.

**Fail-closed is load-bearing**: an unwritable root does not stop `disk_wins` from destroying the note (that path only needs the DB), so without the rule, preservation would fail in exactly the case where destruction still succeeded.

**What surfaces it**, since a preserved copy nothing can reach is barely better than none: the Library activity log names the sidecar; `get_conflicts_for_session()` returns the new columns; and a new `get_preserved_conflict_copies(limit)` lists every preserved copy across all sessions — the entry point that needs no session id, because after an unattended sync the user knows only that a note changed under them.

Rejected: DB-only (nothing renders `sync_conflicts`, so recovery would be "write a SQL query"); sidecar-only (fails the reconstruction criterion and ties the only copy to a directory the user actively manages); a notes-history table (the right long-term answer, but a schema-plus-UI programme that would delay a live P0).

`DISK_WINS` is **implemented, not removed** — removing it would break persisted preferences, a CHECK constraint that already lists it, and the public enum. One shared resolver now serves all three directions, and a policy whose winner cannot be written leaves the conflict **open** rather than claiming to have settled it.

## Three more unfiled defects found and fixed
`DB_WINS` was equally a no-op in bidirectional; `NEWER_WINS` was a no-op in `db_to_disk` even when the note was newer; and `disk_to_db` overwrote note bodies **with no conflict row at all**. The panel also reported every *recorded* conflict as "resolved" — it now counts what was actually applied.

## Evidence
13/13 born-red at base, including `assert 0 == 1` on the conflict count for `disk_to_db` — proving that direction had no conflict detection whatsoever. `Tests/Notes/` **2431 passed / 0 failed** post-rebase; every other suite's failure set is identical to `origin/dev`.

Three traps closed along the way: the existing `test_conflict_detection` was **hollow** (a stale expected version meant the baseline hash was never stored, so it "detected" a conflict for any never-synced note); the repo's one exact schema-version pin lived in an unrelated older file, so every future bump would edit a file that does not own it; and a sidecar written into the sync root **would have been re-ingested as a note** on the next pass, compounding one duplicate per conflict.

Owner questions recorded, not acted on: whether to re-offer `ASK` (meaningless while nothing in the Library prompts — re-offering today would recreate the selectable-but-inert shape this task removes), and whether bidirectional `disk_wins` should unlink a note whose file was deleted.

---
## Summary by cubic
Preserves the losing side on Notes sync conflicts and makes `DISK_WINS` actually apply. Previously we overwrote the loser with no recoverable copy and “Disk wins” did nothing; now we save the loser as a sidecar and apply the chosen policy safely.

- Conflict preservation and safety: write `<note>.conflict-<UTC>-<db|disk>.bak` before any overwrite; claim names atomically via `PinnedSyncRoot.create_new_text` (O_CREAT|O_EXCL|O_NOFOLLOW); recognize sidecars only when both the `.conflict-` marker and `.bak` suffix are present to avoid dropping legitimate notes; exclude sidecars from scans; fail closed if preservation cannot be written.
- Policies and detection: implement `DISK_WINS`; fix `DB_WINS` and `NEWER_WINS`; add conflict detection for `disk_to_db`; leave the conflict open if the winner cannot be applied.
- Data/UI/Docs: add `losing_side`, `losing_content`, and `preserved_file_path` to `sync_conflicts` (written only when a side is discarded); `get_conflicts_for_session` exposes them; the activity line counts applied vs unresolved and names the policy; docs describe policies and sidecar location.
- Migration: bump schema to v44; run the DB migration (task-19554).

<sup>Written for commit 4d525bbb84b2cbb09347f44e0d000ba63b85fba4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1922?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/Features/notes_bidirectional_sync.md
+++ b/Docs/Features/notes_bidirectional_sync.md
@@ -91,6 +91,45 @@ both default to `ConflictResolution.ASK` when called with no explicit
 value. The panel never relies on that default: it always resolves and
 passes one of its three exposed policies explicitly.
 
+### What each policy does to the losing copy
+
+A resolution that overwrites one side always keeps that side first
+(task-19554). What each offered policy does on a `both_changed` conflict:
+
+| Policy | Winner | What happens to the other copy |
+|---|---|---|
+| Newer wins | The side with the later timestamp (note `last_modified` vs. file mtime) | Saved as a sidecar next to the note file, then overwritten |
+| Disk wins | The file on disk | The note body is saved as a sidecar, then replaced |
+| Library wins | The note in the Library | The file's text is saved as a sidecar, then overwritten |
+| Ask (engine-only, never offered by the panel) | Neither | Nothing is written; the conflict row stays open |
+
+The sidecar is named
+`<file name>.conflict-<UTC timestamp>-<db|disk>.bak` and holds the
+discarded text **verbatim** — no header, no diff markers — so recovering
+it is a rename. Sidecars are excluded from every subsequent scan (by the
+`.conflict-` marker, not just the `.bak` suffix), so a preserved copy
+never becomes a note of its own.
+
+The same text is also written to the conflict's own row —
+`sync_conflicts.losing_side` / `losing_content` /
+`preserved_file_path` — so the discarded version survives the sidecar
+being moved or deleted. Those columns are written **only when a side is
+actually discarded**, never on mere detection, so the table does not
+become a second unbounded content shadow of `notes`.
+
+Preservation is fail-closed: if the sidecar cannot be written (an
+unwritable root, a rejected path), the overwrite does **not** happen. The
+conflict is left open, an error is recorded on the run, and both copies
+survive.
+
+Direction narrows which side a policy may write, and a policy whose
+winner cannot be written in the chosen direction applies nothing rather
+than pretending to have settled anything — e.g. "Disk wins" during a
+`db_to_disk` push has nothing to do, because the disk copy already
+stands. `SyncConflict.applied` is the engine-level fact behind this, and
+the panel's activity log distinguishes "N conflicts resolved" from "N
+conflicts left unresolved — both copies kept as they are" on it.
+
 ## Path safety
 
 Every read and write goes through `PinnedSyncRoot`
@@ -125,6 +164,20 @@ Plus two standalone tables the engine writes on every run:
 policy, status, file/conflict/error counts) and `sync_conflicts` (one row
 per detected conflict, linked to its session).
 
+`sync_conflicts` also carries the preservation columns added in schema
+v44:
+
+```sql
+losing_side TEXT,          -- 'db' | 'disk' -- which side was discarded
+losing_content TEXT,       -- that side's text, verbatim
+preserved_file_path TEXT   -- the sidecar written next to the note file
+```
+
+All three stay `NULL` for a conflict nothing was discarded for.
+`resolution` (`use_db` / `use_disk`, or `NULL` while open) and
+`resolved_at` are stamped by the engine only once a side was actually
+applied.
+
 None of `is_externally_synced`, `file_path_on_disk`, or the other sync
 columns currently drive any per-note UI indicator — the Library notes list
 shows no synced/conflict badge on individual rows (confirmed: nothing in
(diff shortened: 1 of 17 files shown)


rmusser01:
Qodo round addressed in `dcf8e8018`.

**1. `is_conflict_sidecar` over-matched.** Now requires the marker **and** the `.bak` suffix. The silence was what made this bad — `_scan_directory` dropped the match with no error and no skip row. Born-red on behaviour rather than a missing symbol: `meeting.conflict-notes.md` created zero notes with `skipped=[]`. Both halves are pinned by a single test, so a suffix-only predicate reds on `archive.bak` and a marker-only one reds on the legitimate note — neither half can be dropped later.

**2 + 3 — one change, as you'd expect.** A new `PinnedSyncRoot.create_new_text` claims the name with `O_CREAT|O_EXCL|O_NOFOLLOW`, so check-and-claim is a single syscall. `FileExistsError` means "advance to the next ordinal", never "overwrite"; an exhausted name space **raises** rather than falling back to a replacing write, which keeps the fail-closed rule intact (a raise means "could not preserve", so the destructive path does not run). Cleanup only unlinks a file *this call* created, so a name another run holds is never removed.

That deletes finding 3 as a side effect: `_write_conflict_sidecar` now makes **no filesystem call of its own** — the `.exists()` probe is gone and everything routes through `PinnedSyncRoot`. That is the right compliance target here rather than `path_validation.py`: `sync_paths.py` exists precisely because the descriptor-anchored boundary is stronger than lexical validation, which is why the sync engine has never imported the latter.

**Concurrency evidence, deliberately not threads.** Real threads would be non-deterministic and could pass on a machine that happened not to interleave. The window is opened through a `_before_create` seam following the module's existing `_before_replace` idiom, and since the seam only exists post-fix, the honest evidence is mutation of the single flag that carries the fix — `O_EXCL` → `O_TRUNC`:

    assert "another run's preserved copy" in {'Modified on disk'}
    assert 'Modified in database' == 'Modified on disk'

The first is the clobber itself; the second shows the same mutation *also* breaks fail-closed — the squatter silently overwritten **and** the destructive path allowed through. Restored via Edit, both green again.

`create_new_text` is a new public boundary primitive, so it got boundary tests beside `write_text`'s: refuses an existing name untouched, refuses to follow a symlinked name, refuses a missing parent, writes `0o600`, and leaves nothing behind if the write fails after the create.

`Tests/Notes/` **2440 passed / 0 failed** (was 2431); 53,887 collected repo-wide.

PR #1924, opened 2026-08-22
fix(logs): close the in-app log share path (task-19555)

rmusser01:
## Summary
Closes task-19555 (P1, privacy). ADR-029's "persistent logs are metadata-only" guarantee is enforced by `PersistentDiagnosticFilter`, attached in exactly two places — **both the rotating file handler**. The file sink is airtight. But `PersistentLogHandler` was installed unconditionally at level NOTSET with **no filter**, feeding an **unbounded** buffer; the Logs window did no sanitization; and "Copy all" joined the entire session to the system clipboard — under empty-state text telling the user to reproduce the problem and share their logs.

The privacy suite proved this **by omission**: it attaches both a filtered file handler and an unfiltered collector, and asserts only against the file.

## Design — deviates from the filed suggestion, and the AC was rewritten with the original struck through
`PersistentDiagnosticFilter` is an all-or-nothing **admission** filter, not a redactor: 12 `persist_event` call sites against ~6,081 loguru call sites, so attaching it at `emit` would drop >99.8% of records. The chips, regex filter, counters and next-error jump would all operate on the output of 12 call sites. That is not a privacy fix with a usability cost — it is deleting the screen.

Instead, the two stores already fed by the one `emit` each get **one job**:
- **live view / "Copy visible logs"** — descriptive, but never credentials and never the OS account name (`/Users/janedoe/x.pdf` → `~/x.pdf`);
- **"Copy all (redacted)"** — metadata-only, decided by the *same filter class* the file handler uses, so clipboard and disk cannot drift.

Buffer bounded to 10,000 records; lines capped at 2,000 characters **before** redaction — strictly less data, not a bypass, and 45× faster on a 100 KB line.

## Two real bugs found alongside
- `log_sanitizer._is_sensitive_log_key` computed a hyphen normalization and then passed the **raw** key to an underscore-suffix checker, so `x-auth-token`, `x-session-key` and `x-client-secret` — exactly what provider request logging emits — were classified harmless.
- The home-path pattern was case-sensitive throughout, so `c:\users\<name>` and `\\SERVER\Users\<name>` kept the account name. Now split into POSIX (case-sensitive, so REST URLs are untouched) and Windows (fully case-insensitive, plus a UNC form).

## Review: FIX-FIRST, then addressed
The reviewer adjudicated the design deviation **in the implementation's favour** and strengthened it with the call-site census above. Its substantive catch: the two **stores** were pinned but the live **feed** was not — its mutation survived 111 passed / 0 failed. No production code was wrong there; the tests were missing. Both feeds are now pinned and that exact mutation reds.

It also caught an absolute claim — "API keys … are always removed" — shipping in five places while several common token formats survived into the live view. In a task whose thesis is *disclose what you cannot remove*, that was the one sentence that could not ship. It now reads "recognised API-key formats", and the denylist was widened by six shapes rather than only softening the words. On bare `key=`, a census (all five such labels in the package are non-secrets, and redacting them would blind a diagnostic reporting *which* config key failed to decrypt) settled it as a wording fix.

## Note on the branch shape
Collapsed to a single commit: GitHub push protection rejected an earlier commit over a **synthetic** test fixture in the real Slack token format. The scanner was right to fire, so the fixture is now assembled at runtime — the redactor under test receives the identical string — rather than clicking through the unblock URL. A repo that trains people to bypass push protection is worse off than one with an awkward fixture.

882 passed / 0 failed on the targeted sweep; 53,857 collected repo-wide.

---
## Summary by cubic
Closes TASK-19555 by sealing the in‑app log share path and bounding the session buffer to uphold ADR‑029. Previously the in‑app collector buffered unfiltered logs and “Copy all” exported raw text; now the live feed is redacted and “Copy all (redacted)” exports metadata only.

- Apply sink-side redaction in `PersistentLogHandler.emit`: strip recognised credentials (Bearer, URL userinfo, labelled values, and standalone shapes for OpenAI/Anthropic/OpenRouter, Google, GitHub, Hugging Face, AWS, Slack, JWT) and home-directory usernames; fix hyphen-normalised key checks.
- Build the share artifact with `PersistentDiagnosticFilter` (metadata-only); keep the live view descriptive but never with credentials or OS account names; relabel UI (“Copy all (redacted)”) and notifications.
- Bound `_log_buffer` and `_log_records` to 10,000; truncate oversized lines at a token boundary, redact the remainder, withhold a single unbroken token larger than the cap, and append a length notice.
- Anchor home‑path redaction: segment‑bounded literal home replacements; POSIX case‑sensitive roots; Windows case‑insensitive drive and UNC forms.
- Pin tests to the real collector and both live feeds; add astride‑boundary and path‑anchoring regressions; assemble synthetic credentials at runtime; update docs to reflect DEBUG scope and “Sharing logs safely”.

<sup>Written for commit 012ed6ab81ff74339d1b60ed557a12d2d69f354e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1924?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/Development/RAG/RAG-OCR.md
+++ b/Docs/Development/RAG/RAG-OCR.md
@@ -821,6 +821,19 @@ import logging
 logging.getLogger("tldw_chatbook.Local_Ingestion").setLevel(logging.DEBUG)
 ```
 
+Read the output in the **Logs** screen (`F8`); the rotating log file is
+metadata-only by design (ADR-029) and does not carry this detail.
+
+> **What `DEBUG` exposes.** It widens the in-app log well past the default:
+> ingestion source paths, extracted document text, OCR output and SQL
+> parameter previews all become visible. Recognised API-key formats and your
+> account name are stripped before anything is buffered — a denylist, so not
+> a guarantee for every secret — and **Copy all (redacted)** exports
+> only timings, loggers and error types — but **Copy visible logs** copies the
+> log text as shown, so read it before attaching it to an issue. Turn `DEBUG`
+> back off when you are done. See
+> [Sharing logs safely](../../User_Guide/logs.md#sharing-logs-safely).
+
 ### Validation Tools
 
 ```python
--- a/Docs/Features/ChatDictionaries-Documented.md
+++ b/Docs/Features/ChatDictionaries-Documented.md
@@ -460,7 +460,18 @@ Enable debug logging in `config.toml`:
 log_level = "DEBUG"
 ```
 
-Check logs at: `~/.share/tldw_cli/logs/`
+Read the output in the **Logs** screen (`F8`). The rotating file under
+`~/.share/tldw_cli/logs/` is metadata-only by design (ADR-029) and does not
+carry dictionary matches or replacement text.
+
+> **What `DEBUG` exposes.** It widens the in-app log well past the default:
+> matched keywords, replacement content, file paths and SQL parameter
+> previews all become visible. Recognised API-key formats and your account
+> name are stripped before anything is buffered — a denylist, so not a
+> guarantee for every secret — and **Copy all (redacted)** exports only
+> timings, loggers and error types — but **Copy visible logs** copies the log
+> text as shown, so read it before sharing. Turn `DEBUG` back off when you are
+> done. See [Sharing logs safely](../User_Guide/logs.md#sharing-logs-safely).
 
 ## Comparison with World Books
 
--- a/Docs/Features/World-Lore-Books-Documented.md
+++ b/Docs/Features/World-Lore-Books-Documented.md
@@ -507,7 +507,18 @@ Enable debug logging to see world book processing:
 log_level = "DEBUG"
 ```
 
-Check logs at: `~/.share/tldw_cli/logs/`
+Read the output in the **Logs** screen (`F8`). The rotating file under
+`~/.share/tldw_cli/logs/` is metadata-only by design (ADR-029) and does not
+carry the lines below.
+
+> **What `DEBUG` exposes.** It widens the in-app log well past the default:
+> matched keywords, world-book entry content, file paths and SQL parameter
+> previews all become visible. Recognised API-key formats and your account
+> name are stripped before anything is buffered — a denylist, so not a
+> guarantee for every secret — and **Copy all (redacted)** exports only
+> timings, loggers and error types — but **Copy visible logs** copies the log
+> text as shown, so read it before sharing. Turn `DEBUG` back off when you are
+> done. See [Sharing logs safely](../User_Guide/logs.md#sharing-logs-safely).
 
 Look for:
 - "World info keyword found"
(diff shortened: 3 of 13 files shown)


rmusser01:
Qodo round addressed in `247fa55b1`. Finding 1 was mine to own — the truncate-before-redact ordering came from my dispatch, and Qodo is right that it can slice a token out of regex range.

**1. Redact-then-truncate, with a correction to the framing.** I implemented it as *cut on a token boundary, then redact everything that survives the cut*, and re-measurement is why. **The old cost note was wrong in a second way**: it blamed key/value density, but a 100 KB line costs 16–21 ms *whatever its shape* —

| shape | 2 KB | 16 KB | 100 KB |
|---|---|---|---|
| kv-dense | 0.445 / 0.446 ms | 0.445 / 3.434 ms | **0.448 / 20.898 ms** |
| JSON | 0.384 / 0.384 ms | 0.378 / 3.004 ms | **0.389 / 18.518 ms** |
| prose | 0.332 / 0.334 ms | 0.342 / 2.644 ms | **0.335 / 16.252 ms** |

(token-cut / uncapped; typical 140-char lines are 0.032 ms either way.)

So pure redact-first costs 16–21 ms **on the emitting thread, including the UI thread**, for any large line. The token-aligned cut is equally sound and holds cost flat at ~0.4 ms: a credential never contains whitespace, so cutting on one leaves it either wholly inside (redacted) or wholly outside (discarded) — no fragment is reachable. A single unbroken token longer than the cap has no safe cut, so its body is **withheld rather than sliced**, since a 2,000-char prefix of an opaque token is most of a secret.

**2. Home-path — worse than described.** With `$HOME=/Users/jan`, the path `/Users/janedoe/Notes/x.pdf` became **`~edoe/Notes/x.pdf`**: half of a *different* account's name, with the `/Users/` prefix destroyed so the regex fallback couldn't recover it. Also `/srv/appdata-backup/db` → `~-backup/db`. Fixed by anchoring the literal candidates on path-segment boundaries, longest-first. **The Windows literal had the identical defect** and is now pinned separately.

**3. Fixture sweep — all of them.** Every synthetic credential is assembled at import. Swept all 13 changed files with 16 detectors (the redactor's own patterns plus Stripe, npm, PyPI, SendGrid, PEM): **0 contiguous token-shaped literals**, including five pre-existing ones Qodo didn't flag.

**Two vacuous drafts of the astride test were caught by mutation, not by reading** — one hand-computed padding that ignored the formatter's prefix so the key never landed astride; the other used `logger.info("%s %s", ...)` when **loguru formats with `str.format`**, so it logged the template and dropped the payload. Both passed against the broken implementation. The shipped test measures the prefix width off a probe record, sweeps every straddle position, and carries an anti-vacuity control.

889 passed / 0 failed on the targeted sweep; 53,901 collected repo-wide.

PR #1925, opened 2026-08-22
fix(security): stop custom auth headers surviving cross-origin redirects (task-19557)

rmusser01:
## Summary
Closes task-19557 (P1, security). Custom auth headers survived cross-origin redirects. `httpx` pops only `Authorization` and `Cookie` on a cross-host redirect; `requests` deletes only `Authorization` in `rebuild_auth`. Neither touches `X-API-KEY`/`x-api-key`.

- `tldw_api/client.py` set `X-API-KEY` on a client with `follow_redirects=True` — undocumented since the client's first commit. API-key is the **default** auth mode, and this client backs Notes, Chatbooks, Character/Persona, Sync and MCP. One `302` from the configured server, or a MITM on an `http://` base URL, exfiltrates the key.
- Anthropic's `x-api-key` posted with `requests` and no `allow_redirects=False`, and its endpoint is user-configurable, so an LLM proxy in the path can redirect.

No test pinned the behaviour, so this was a gap rather than an accepted decision.

## Fix: refuse outright, mirroring the pattern the repo already chose
The repo had **already fixed this exact hazard** for another provider, with a comment naming it. Rather than invent a second mechanism, this reuses that shape:
- the shared `AsyncClient` now uses `follow_redirects=False`, with a `_raise_if_redirected` helper called in **all five** request-issuing methods (`_request`, `_binary_request`, `_headers_request`, `_stream_request`, `_sse_request` — the client has five distinct request shapes, not one);
- both Anthropic `session.post` calls, including the cache-control retry, pass `allow_redirects=False`, and a 3xx raises;
- the summarization sibling does the same, returning an error string to match that function's own convention.

**Alternatives rejected with reasons.** A hop-loop would reinvent httpx's and requests' redirect engines with real edge-case risk. Routing the httpx client through the shared guarded-fetch helper would be *strictly better* — it also closes SSRF on the same seam — but that helper is GET-only today and this client has five request shapes and dozens of callers; either choice would also have violated the criterion mandating reuse of the existing pattern.

## Evidence
Born-red in-process — an httpx `MockTransport` and a patched `HTTPAdapter.send`, **no real sockets** (the suite has a network guard). Verified by patch revert/restore with a byte-identical diff afterwards: all three tests fail at base with the sentinel key visibly present in the cross-origin host's headers.

`Tests/tldw_api/` 489 passed; `Tests/LLM_Calls/` 1049 passed with 3 pre-existing failures reproduced identically with this change fully reverted; 53,824 collected repo-wide.

## Sibling found and deliberately not fixed here
**`Utils/egress.py`'s `_STRIP_HEADERS` does not include `x-api-key`** — a live gap in the *shared* guarded-fetch primitive, reachable through user-configurable feed auth headers in the subscriptions path. Fixing two callers while the shared primitive stays leaky would have been the smaller half of the job, so it is recorded for its own task rather than folded in. Also unfixed and recorded: the Kobold local-call path, and Bing/Brave/Serper/Exa in the web-search module.

---
## Summary by cubic
Stops custom API-key headers from leaking on cross-origin redirects by refusing redirects in the TLDW API client and Anthropic call sites. Old behavior: `httpx`/`requests` followed 3xx and forwarded custom headers; new behavior: never follow and raise/return before processing, explicitly closing redirect responses, excluding 304, and omitting the redirect Location.

- TLDW API client: constructs `httpx.AsyncClient` with `follow_redirects=False`. Async `_raise_if_redirected` checks only 301/302/303/307/308, awaits `response.aclose()`, never echoes `Location`, and is called in `_request`, `_binary_request`, `_headers_request`, `_stream_request`, and `_sse_request`. 304 is not treated as a redirect.
- Anthropic calls: both `requests.post` paths now pass `allow_redirects=False`. On 3xx, `chat_with_anthropic` closes the response and raises `ChatProviderError`; `summarize_with_anthropic` closes the response and returns an error string. No new logging added.
- Tests/fixtures: redirect-leak regressions assert the credential is absent on the second hop; streaming/SSE and Anthropic tests assert the redirect response is explicitly closed (count=2); a 304 path is pinned as non-redirect. Updated doubles accept `allow_redirects` and expose `status_code`.
- Rollout: callers relying on redirects will now see failures; ensure endpoints are correct and same-origin. Meets task-19557 AC.

<sup>Written for commit e85598b512080725f1238a9060f5d8731337f3c7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1925?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Chat/test_chat_functions.py
+++ b/Tests/Chat/test_chat_functions.py
@@ -289,7 +289,16 @@ class _CapturedSession:
         self.closed = True
         return None
 
-    def post(self, url, *, headers=None, json=None, stream=False, timeout=None):
+    def post(
+        self,
+        url,
+        *,
+        headers=None,
+        json=None,
+        stream=False,
+        timeout=None,
+        allow_redirects=None,
+    ):
         self._captured.update(
             {
                 "url": url,
--- a/Tests/Chat/test_console_provider_gateway.py
+++ b/Tests/Chat/test_console_provider_gateway.py
@@ -4462,7 +4462,16 @@ class _CapturedURLSession:
     def mount(self, *_args, **_kwargs) -> None:
         return None
 
-    def post(self, url, *, headers=None, json=None, stream=False, timeout=None):
+    def post(
+        self,
+        url,
+        *,
+        headers=None,
+        json=None,
+        stream=False,
+        timeout=None,
+        allow_redirects=None,
+    ):
         self._captured["url"] = url
         return _FakeAnthropicPostResponse()
 
--- a/Tests/LLM_Calls/test_anthropic_redirect_credential_leak.py
+++ b/Tests/LLM_Calls/test_anthropic_redirect_credential_leak.py
@@ -1,0 +1,167 @@
+"""task-19557: ``x-api-key`` must never survive a cross-origin redirect.
+
+Both Anthropic call sites (``chat_with_anthropic`` in ``LLM_API_Calls.py``
+and ``summarize_with_anthropic`` in ``Summarization_General_Lib.py``) send
+the API key in the custom ``x-api-key`` header over ``requests``.
+``requests`` strips only ``Authorization``/``Cookie`` on a cross-host
+redirect -- it has no notion of ``x-api-key`` -- so a redirecting or
+compromised endpoint could otherwise capture the real key verbatim.
+
+Fixed by passing ``allow_redirects=False`` on every ``session.post``/
+``requests.post`` call that carries the header and explicitly refusing any
+3xx response before the body is processed, mirroring the already-shipped
+``x-goog-api-key`` fix for Google (``LLM_API_Calls.py``
+``chat_with_google``, task-686/lines ~3528-3549).
+
+Born-red: reverting either ``allow_redirects=False`` guard (and the
+explicit 3xx check that follows it) makes the corresponding test below fail
+by showing the sentinel key delivered to the cross-origin host.
+
+Transport is faked at ``requests.adapters.HTTPAdapter.send`` -- the layer
+just above the real socket -- so ``requests``' own redirect/session
+machinery (``Session.resolve_redirects``, ``rebuild_auth``,
+``rebuild_headers``) still runs for real; only the actual network I/O is
+replaced. No test in this module makes a real network call.
+"""
+
+from __future__ import annotations
+
+from urllib.parse import urlsplit
+
+import pytest
+import requests
+import requests.adapters
+
+import tldw_chatbook.LLM_Calls.LLM_API_Calls as llm_api_calls
+import tldw_chatbook.LLM_Calls.Summarization_General_Lib as summarization_lib
+from tldw_chatbook.Chat.Chat_Deps import ChatProviderError
+
+# Synthetic sentinel -- never a real credential.
+_SENTINEL_KEY = "sentinel-test-x-api-key-must-never-leak"
+
+
+def _fake_response(
+    status_code: int, headers: dict[str, str], body: bytes = b""
+) -> requests.Response:
+    resp = requests.Response()
+    resp.status_code = status_code
+    resp.headers = requests.structures.CaseInsensitiveDict(headers or {})
+    resp._content = body
+    resp.encoding = "utf-8"
+    return resp
+
+
+def _install_redirecting_adapter(
+    monkeypatch: pytest.MonkeyPatch,
+    seen_headers_by_host: dict[str, dict[str, str]],
+    *,
+    good_host: str,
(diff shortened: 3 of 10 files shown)


rmusser01:
Qodo round addressed in `f9baf7081`. All three findings were scoped to `_raise_if_redirected`; the Anthropic/Google sites were confirmed already correct on all three counts.

**1. Unclosed response.** `_raise_if_redirected` is now async and awaits `response.aclose()` before raising, with all five call sites updated. Proving this on the streaming paths needed care: httpx's own `client.stream()` context manager already has a `finally: await response.aclose()`, so `.is_closed` is True either way and cannot discriminate. Verified instead with a call-counting spy wrapping the response's `aclose` inside the transport handler — 2 calls with the fix, 1 without.

**2 & 3.** Narrowed to `{301, 302, 303, 307, 308}` and dropped the raw `Location` from the exception. Worth recording that finding 2 was **live, not theoretical**: `get_user_profile_catalog(if_none_match=...)` sends a conditional GET, so a legitimate 304 was being refused as a redirect. Pinned by a new test.

Mutation-verified all three in isolation (reverting whole-band + Location together showed `evil.example` leaking into the exception message and a spurious error on 304; reverting just the `aclose()` showed `observed 1 aclose() call(s)`), each restored byte-identical against a saved patch.

`Tests/tldw_api/` 492 passed (489 + 3 new).

rmusser01:
Second Qodo round addressed in `92ec80656`. **Qodo was right and my earlier assessment was wrong** — I re-checked each Anthropic refusal site individually instead of re-asserting the prior summary.

| Site | What closes the response |
|---|---|
| `LLM_API_Calls.py` `chat_with_anthropic` | `response.close()` before the raise — correct, unchanged |
| `LLM_API_Calls.py` `chat_with_google` (the reference) | `response.close()` before the raise — correct, unchanged |
| `Summarization_General_Lib.py` `summarize_with_anthropic` | **nothing** — returned the error string with no close. Fixed. |

Docstrings added to both Anthropic redirect tests, stating the defect pinned rather than what the code does.

**The mutation evidence is the interesting part.** My first attempt asserted `len(close_calls) >= 1` and it **passed with the fix removed**. Rather than accept that, I found why: `requests.Session.send()` always calls `resolve_redirects(..., yield_requests=True)` to populate `Response._next` *even under* `allow_redirects=False`, and that generator's bookkeeping unconditionally closes any response carrying a `Location` — so a genuine 302 always gets exactly one incidental close from the library itself. Tightened both tests to `== 2` (library's incidental close plus the site's own), then verified in both directions: with the fix removed the summarization test reds (`observed 1 close() call(s)`) while its sibling stays green; restored, both green.

That investigation surfaced a **second, hollow-test bug**: the bare `requests.Response()` double defaulted `raw=None`/`_content_consumed=False`, so an unpatched `.close()` raised `AttributeError` — which the production code's outer `except Exception` absorbed and rewrapped into *the same error type the test was asserting on*. The original test therefore passed whether or not `close()` worked at all. Fixed by setting `_content_consumed=True` on the double.

`Tests/tldw_api/` 492 passed; `Tests/LLM_Calls/` 1049 passed with 3 failures reconfirmed pre-existing by reverting this round's diff and re-running.

rmusser01:
Both remaining items are addressed in `e85598b51` (pushed after the last Qodo pass, so the review index above predates it).

- **Finding 1** — `summarize_with_anthropic` now calls `response.close()` on the 3xx refusal path before returning its error string. Per-site audit: `chat_with_anthropic` and `chat_with_google` already closed correctly; the summarization site was the only gap, and my earlier "all three counts correct" claim for the Anthropic sites was wrong.
- **Finding 2** — docstrings added to both Anthropic redirect tests, stating the defect pinned.

Worth recording why the mutation evidence took two attempts: `requests.Session.send()` calls `resolve_redirects(..., yield_requests=True)` even under `allow_redirects=False`, and that generator closes any response carrying a `Location` — so a 302 always gets one incidental close and `>= 1` was vacuous. Tightened to `== 2`. That also surfaced a hollow test double: a bare `requests.Response()` has `raw=None`, so an unpatched `.close()` raised `AttributeError`, which production's outer `except Exception` rewrapped into the same error type the test asserted on — the test passed whether or not `close()` worked.

`Tests/tldw_api/` 494 passed.

PR #1929, opened 2026-08-22
perf: efficiency burn-down batch 1 (task-692, task-19324, task-19573 AC, task-13158)

rmusser01:
## What & why

First burn-down batch from the efficiency backlog. Four small, independently reviewable changes; each one verified against real code before being touched, and each behaviour change red-proofed.

## task-692 — note tools re-opened the DB on every call

The three note tools built a fresh `NotesInteropService` per `execute()`.

The original write-up said this paid "a full DB-open and schema-check cost per call", and the code looked like it did *not* — all three already pass `global_db_to_use=chachanotes_db`. But that only supplies the **path template**. `NotesInteropService._db_instances` is an *instance* attribute, so a per-call service missed its own cache every time and `_get_db` constructed a brand-new `CharactersRAGDB(db_path=…, client_id=user_id)` — a real DB open plus schema-version check — then threw it away when the call returned.

Measured: **~1.8 ms per construction** on an already-created DB, plus a `verify_trusted_directory` filesystem check per call. `TASK-545 P2` put these tools on the agent worker thread, where one run can invoke them many times.

Now built lazily once and reused, keyed on the resolved DB path so re-pointing the data dir still rebuilds. **Attribution semantics preserved deliberately**: caching the *service* keeps `_get_db`'s per-user `client_id` behaviour, whereas reusing `chachanotes_db` directly would have silently changed the `client_id` written into rows.

## task-19324 — the llama.cpp retry was invisible in the Inspector

When a llama.cpp stream opens fine but yields nothing, `stream_llamacpp_chat` retries non-streaming. That retry is a **second HTTP request**, issued below the seam that captures `stream_chat`'s call — so the Conversation Inspector showed one row for a turn that really made two, understating what was sent on exactly the degraded turn a user opens the Inspector to look at.

The retry now gets its own `ExchangeCapture` carrying its real wire payload (`stream=false`) and a `retry_of` marker. Chose AC #1's primary clause (a separate capture) over "amend the existing one": the promise is that the tab shows every request a turn made, and one row carrying two requests doesn't show that. The hook is best-effort — capture must never break a send, and this path is already degraded.

## task-19573's last AC — duplicate task ids now fail locally

`Tests/CI/test_backlog_task_id_uniqueness.py`. The Backlog Guard workflow already rejects duplicate ids, but only *after* a push — a collision costs a CI cycle and a PR round-trip to find, and the workflow log is the only place it's visible. This has recurred at ids 152+, 196-203, 246-256, `task-3401`, and most recently a batch of seven.

Checks both namespaces the workflow does (filename prefix and first frontmatter `id:`, which can disagree after a hand-edited rename), plus one it doesn't: **every task file must declare an id**, so deleting an `id:` line can't silence a collision. A module fixture asserts the tasks directory is non-empty, so a wrong `PROJECT_ROOT` can't make the gate vacuously green.

## task-13158 — closed, resolved upstream

Its `task-3401` collision was cleared in PR #1465 (Console-rail renumbered to `TASK-14650`); the remaining six went under TASK-19573. Verified by running the workflow's own check **verbatim** against this branch: `fail=0 over 2325 task files`.

## Red-proof evidence

Every behaviour claim was verified by breaking it first:

| Change | Reverted → |
|---|---|
| notes service cache | reuse test sees **3** service builds instead of 1 |
| llama.cpp retry capture | capture count drops **2 → 1** |
| duplicate-id gate | each of the three defects (filename collision, frontmatter-only collision, missing `id:`) fails its own assertion and only that one |

## Testing

`Tests/Tools/` + the new gate: **625 passed**. Full `test_console_provider_gateway.py`: **264 passed** (263 on dev + the one added here).

Pre-existing on clean `origin/dev`, confirmed in a throwaway worktree — **not** from this branch: 2 failures in `test_glob_grep_files.py`, and a tiktoken-download teardown error in `test_console_provider_gateway.py` (25 errors on dev, 25 here).

## Left for an owner call (not touched)

`task-13262` and `task-14650` are the **same task** — identical title, identical `created_date`, identical body except only 14650 carries the renumber provenance — and **both are still `In Progress`**. Different ids, so the duplicate-id guard cannot see them, but someone picking up the work sees two live copies. 14650 is canonical (referenced by task-3793, task-4000, and its own design plan); 13262 is referenced only by TASK-19573. Not archived unilaterally.

---
## Summary by cubic
Speeds up note tools, improves Inspector accuracy for llama.cpp fallbacks, and adds a local test gate to block duplicate backlog IDs before push.

- Notes tools: Previously reopened the DB on each call; now reuse a single `NotesInteropService` cached by the resolved chachanotes DB path. Per-user client_id attribution is unchanged. Adds `_reset_notes_service_cache()` and an autouse test reset.
- llama.cpp fallback capture: The non-streaming retry now gets its own ExchangeCapture with `stream=false` and a `retry_of` marker. Implemented via an internal `on_fallback_retry` hook; closes the call-scoped signals and logs only the exception type. Best-effort and non-blocking.
- Backlog guard: Adds `Tests/CI/test_backlog_task_id_uniqueness.py` to fail locally on duplicate IDs in filenames or frontmatter, and on missing `id:` lines. Run pytest locally to catch collisions before pushing.

<sup>Written for commit d3e4253b8feb730d714d620c7a38a88ea863f17c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1929?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/CI/test_backlog_task_id_uniqueness.py
+++ b/Tests/CI/test_backlog_task_id_uniqueness.py
@@ -38,10 +38,24 @@ _FILENAME_ID = re.compile(r"^(task-[0-9]+(?:\.[0-9]+)*) - .*\.md$")
 
 
 def _task_files() -> list[Path]:
+    """Return every backlog task file, sorted for stable failure output.
+
+    Returns:
+        Sorted ``backlog/tasks/task-*.md`` paths.
+    """
     return sorted(TASKS_DIR.glob("task-*.md"))
 
 
 def _format(dupes: dict[str, list[str]]) -> str:
+    """Render duplicate ids and their files as an actionable failure message.
+
+    Args:
+        dupes: Task id -> the file names that claim it.
+
+    Returns:
+        A multi-line message naming every colliding file plus the owner rule
+        for resolving it.
+    """
     lines = []
     for task_id, names in sorted(dupes.items()):
         lines.append(f"--- {task_id} ---")
@@ -57,6 +71,11 @@ def _format(dupes: dict[str, list[str]]) -> str:
 
 @pytest.fixture(scope="module")
 def task_files() -> list[Path]:
+    """Every task file, with a guard against a mis-resolved project root.
+
+    Returns:
+        Sorted ``backlog/tasks/task-*.md`` paths.
+    """
     files = _task_files()
     # Guard the guard: a wrong PROJECT_ROOT would make every assertion below
     # vacuously true, which is how a green gate hides a real collision.
@@ -65,6 +84,11 @@ def task_files() -> list[Path]:
 
 
 def test_no_duplicate_task_ids_in_filenames(task_files: list[Path]) -> None:
+    """Fail when two task files share an id in their FILENAME prefix.
+
+    Args:
+        task_files: Every ``backlog/tasks/task-*.md`` file, from the fixture.
+    """
     by_id: dict[str, list[str]] = defaultdict(list)
     for path in task_files:
         match = _FILENAME_ID.match(path.name)
@@ -75,6 +99,14 @@ def test_no_duplicate_task_ids_in_filenames(task_files: list[Path]) -> None:
 
 
 def test_no_duplicate_task_ids_in_frontmatter(task_files: list[Path]) -> None:
+    """Fail when two task files share an id in their frontmatter.
+
+    Checked separately from the filename because the two can disagree after a
+    hand-edited rename, and the backlog CLI resolves by frontmatter.
+
+    Args:
+        task_files: Every ``backlog/tasks/task-*.md`` file, from the fixture.
+    """
     by_id: dict[str, list[str]] = defaultdict(list)
     for path in task_files:
         for line in path.read_text(encoding="utf-8", errors="replace").splitlines():
--- a/Tests/CI/test_backlog_task_id_uniqueness.py
+++ b/Tests/CI/test_backlog_task_id_uniqueness.py
@@ -1,0 +1,108 @@
+"""Local gate for duplicate backlog task ids (TASK-19573's last AC).
+
+Two branches minting the same id and merging separately produce two task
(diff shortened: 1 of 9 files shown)


rmusser01:
Rebased onto latest `dev` and addressed all six Qodo findings.

**Bugs (3), all real:**
- **#4 unclosed retry usage call** — the llama.cpp fallback opened a call-scoped signals object with `new_usage_call()` and never closed it, leaving its entry in `_active_usage_payloads`. Harmless today because the retry records no usage, but the moment one is added that stuck entry gets billed by `usage_payloads()`'s in-flight tail. Now closed alongside the exchange close.
- **#5 inconsistent DB path snapshot** — resolved once and derived, so a profile switch between the two calls can't key a cached service against a database it wasn't bound to.
- **#6 fallback capture hides errors** — now logs the exception *type*, which is actionable and, unlike a traceback, cannot carry payload from the frame's locals (this branch has a hard no-payload-in-logs rule).

**Rule violations (3):** Google docstrings on the new guard tests/helpers, type hints on the fallback test.

One extra cleanup while there: #5's fix made two monkeypatches in the notes-service test inert — they patched `_notes_db_base_dir`, which is no longer on that path. Removed, since an inert patch implies a coupling that no longer exists.

Gate: notes + backlog-guard tests **13 passed**; `test_console_provider_gateway.py` **264 passed** (4 errors are the pre-existing tiktoken-network teardown, unrelated).

PR #1932, opened 2026-08-22
perf: efficiency burn-down batch 2 — stop blocking the event loop (task-3503, task-585, task-19562 measurement)

rmusser01:
## What & why

Second efficiency burn-down batch: the **blocking-the-loop** family. Two fixes plus one measurement that settles an open design question. Both fixes are red-proofed, and both were verified against the real code before being touched — one task's stated premise turned out to be wrong, which is reported rather than papered over.

## task-3503 — the settings cache rebuilt once per *thread*, not once per miss

`load_settings()` cleared `_SETTINGS_CACHE` **inside** the cache lock and then ran the rebuild **outside** it. Every thread that arrived during the miss window ran the entire rebuild — re-reading and re-parsing the TOML, deep-merging defaults, re-ensuring directories — and the last writer won.

**Measured**, counting `_load_cli_config_bootstrap` on one invalidation:

| | bootstrap loads |
|---|---|
| 8 threads, before | **32** |
| 8 threads, after | **4** |
| 1 thread (the floor) | 4 |

The duplication was exactly linear in thread count, so the saving scales with concurrency. After the fix, cross-thread duplication is **zero** — the remaining 4 is one rebuild's own nested config reads, a separate single-threaded inefficiency this task doesn't address (worth its own follow-up).

Implemented as a thin locking wrapper over the original body (renamed `_load_settings_uncached`) rather than re-indenting ~1400 lines into a `try/finally`. The rebuild lock is an **RLock on purpose**: the rebuild reaches helpers that read config again, so a plain `Lock` would deadlock the rebuilding thread against itself. The cache-hit path is untouched — one short lock, rebuild lock never taken.

**The task's AC #1 does not reproduce, and I'm flagging that rather than quietly satisfying it.** It asks that `load_settings()` never hand a concurrent caller `None`. It never did: the function doesn't return the cache cell during the window, and no module outside `config.py` reads `_SETTINGS_CACHE` directly (grep-verified). A thread arriving mid-rebuild always took the miss branch and got a real mapping — it just paid for a redundant rebuild to get it. The duplicated work was the real defect. `None`-freedom is pinned anyway, so a future refactor that starts returning the raw cell fails.

## task-585 — Confluence HTTP blocked the event loop

`ConfluenceAuth.make_request` is a synchronous `requests` call, and `_extract_page_id_from_url` falls back to a blocking fetch. Both ran inline from `async def`, so a slow or hanging Confluence server stalled the **whole loop** for up to the full request timeout (30s since task-328 bounded it; unbounded before).

All seven `make_request` call sites across the scraper and crawler, plus the page-id helper, now run via `asyncio.to_thread`. Chose the executor route over an httpx port deliberately: it satisfies the AC at a fraction of the risk, leaving session/auth/retry handling untouched.

The tests assert the **property** — the blocking callable runs on a different thread than the loop — rather than asserting a particular wrapper, so they survive a later move to httpx async. A fourth test pins the symptom a user would feel: the loop keeps ticking while a slow request is in flight.

## task-19562 — the `busy_timeout` question, measured not assumed

That task has an AC requiring this be *measured*. Done, on this branch's base:

```
SubscriptionsDB connection busy_timeout (ms): 5000
journal_mode: wal
second writer blocked for 1.08s -> acquired   (lock held 1.0s)
```

The premise is **confirmed** — no `busy_timeout` is set anywhere, so it inherits Python sqlite3's 5 s default, and a writer collision really does block for the lock's duration up to that ceiling.

With one **narrowing the design needs**: `journal_mode = WAL`, so readers and writers don't block each other. The 5 s stall exposure is **writer-vs-writer only**, not all 22 async methods as the task's wording could be read to imply. The read-only methods still block the loop for their own query duration — a separate fix — but cannot be *stalled* by a concurrent writer. Lowering `busy_timeout` only converts the stall into an earlier exception, so it is not a substitute for moving the work off the loop.

task-19562 itself is **left open**: it is an arc, not a quick win (7,117 lines across two files, 40 `async def`, bundling a user-visible duplicate-alert bug, the 22-method offload, a non-reentrant `transaction()`, and a leaked connection). The task file now carries a recommended 3-way split.

## Red-proof evidence

| Change | Reverted → |
|---|---|
| settings rebuild lock | `8 concurrent callers rebuilt the settings cache 32 times; one uncontended rebuild costs 4` |
| Confluence offload | all 4 loop-thread tests fail |

One test-quality note recorded rather than hidden: the Confluence responsiveness test initially passed against the *unfixed* code, because it sampled the heartbeat counter after awaiting the heartbeat — which always completes. It now samples before draining and fails with the fix reverted, like the other three.

## Testing

Config suites **123 passed**; `Tests/Web_Scraping/` **226 passed, 3 skipped**.

The 8 `Tests/DB/` failures on this branch were A/B'd properly — same worktree, same venv, only `config.py` reverted to dev's: **8 before, 8 after**, so none are mine.

Note for local runs: the Confluence package imports `playwright` and `trafilatura` transitively, so its tests (including the two pre-existing ones) don't collect without those optional deps.

---
## Summary by cubic
Stops blocking the event loop on Confluence I/O, rebuilds the settings cache once per miss rather than once per thread, and records SQLite `busy_timeout`; also serializes the shared Confluence `requests.Session` to keep newly offloaded calls thread-safe. Old Confluence behavior: sync `requests` ran inline and froze the loop; new behavior: call sites run via `asyncio.to_thread` and `session` access is guarded by a lock. Old config behavior: concurrent misses duplicated full rebuilds; new behavior: a reentrant lock serializes miss → rebuild → store. Measured DB behavior: default 5s `busy_timeout` with WAL; stalls are writer-vs-writer only. (task-585, task-3503, task-19562)

- Confluence: executor offload only; no API changes. `ConfluenceAuth` now guards its `requests.Session` with a lock in both `make_request` and `test_authentication` to avoid concurrent access introduced by `asyncio.to_thread`. Tests assert off-loop execution, loop responsiveness under slow requests, and single-entry session use. Local runs of `Tests/Web_Scraping/Confluence/` require optional deps `playwright` and `trafilatura`.
- Config: `load_settings()` wraps `_load_settings_uncached` with a module-level `RLock` to serialize rebuilds; a double-check on miss avoids duplicate work. Cache-hit path is unchanged. Tests cover concurrency, never returning `None`, and warm-cache behavior.
- DB: code unchanged; task-19562 doc adds measured `busy_timeout` = 5000 ms and WAL mode, narrowing offload scope to writer-vs-writer stalls.

<sup>Written for commit a8b424cec11be60ee1e020c861cb42283f4c8433. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1932?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Web_Scraping/Confluence/test_confluence_no_blocking_io_on_loop.py
+++ b/Tests/Web_Scraping/Confluence/test_confluence_no_blocking_io_on_loop.py
@@ -1,0 +1,141 @@
+"""Confluence HTTP must not run on the event loop thread (TASK-585).
+
+`ConfluenceAuth.make_request` is a synchronous `requests` call and
+`ConfluenceScraper._extract_page_id_from_url` falls back to a blocking fetch.
+Both were invoked inline from `async def` methods, so a slow or hanging
+Confluence server stalled the whole loop for up to the full request timeout
+(30s since task-328 bounded it; unbounded before that), starving every other
+concurrent operation in the app.
+
+These tests assert the property directly -- the blocking callable must
+execute on a DIFFERENT thread than the running loop -- rather than asserting
+that some particular wrapper was used, so they keep holding if the code later
+moves to httpx async instead of a worker thread.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import threading
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import pytest
+
+from tldw_chatbook.Web_Scraping.Confluence.confluence_crawler import ConfluenceCrawler
+from tldw_chatbook.Web_Scraping.Confluence.confluence_scraper import ConfluenceScraper
+
+
+class _ThreadRecordingAuth:
+    """Stands in for ConfluenceAuth, recording where make_request ran."""
+
+    def __init__(self) -> None:
+        self.threads: list[int] = []
+        self.base_url = "https://example.invalid/wiki"
+
+    def make_request(self, method: str, endpoint: str, **kwargs):
+        self.threads.append(threading.get_ident())
+        response = MagicMock()
+        response.status_code = 200
+        response.json.return_value = {
+            "id": "123",
+            "title": "T",
+            "body": {"storage": {"value": "<p>x</p>"}},
+            "version": {"number": 1},
+            "space": {"key": "S", "name": "S"},
+            "results": [],
+            "size": 0,
+        }
+        response.content = b""
+        return response
+
+
+@pytest.mark.asyncio
+async def test_scrape_page_by_id_does_not_block_the_loop_thread():
+    auth = _ThreadRecordingAuth()
+    scraper = ConfluenceScraper(auth)
+
+    loop_thread = threading.get_ident()
+    await scraper.scrape_page_by_id("123")
+
+    assert auth.threads, "make_request was never called; the test proved nothing"
+    assert loop_thread not in auth.threads, (
+        "ConfluenceAuth.make_request ran on the event loop thread: a slow "
+        "Confluence server would stall every other concurrent operation."
+    )
+
+
+@pytest.mark.asyncio
+async def test_crawler_child_page_lookup_does_not_block_the_loop_thread():
+    auth = _ThreadRecordingAuth()
+    crawler = ConfluenceCrawler(auth)
+
+    loop_thread = threading.get_ident()
+    await crawler._get_child_pages("123")
+
+    assert auth.threads, "make_request was never called; the test proved nothing"
+    assert loop_thread not in auth.threads
+
+
+@pytest.mark.asyncio
+async def test_page_id_extraction_does_not_block_the_loop_thread(monkeypatch):
+    """The URL helper falls back to fetching the page, so it blocks too."""
+    auth = _ThreadRecordingAuth()
+    scraper = ConfluenceScraper(auth)
+
+    seen: list[int] = []
+
+    def recording_extract(url: str):
+        seen.append(threading.get_ident())
+        return "123"
+
+    monkeypatch.setattr(scraper, "_extract_page_id_from_url", recording_extract)
+
+    loop_thread = threading.get_ident()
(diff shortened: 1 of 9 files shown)


rmusser01:
Rebased onto latest `dev` and addressed both findings.

**#1 shared `requests.Session` across threads — a real second-order consequence of this PR's own change, and a good catch.** `requests.Session` isn't thread-safe (mutable cookies, connection pool, headers). That was harmless while every call ran inline on the loop: `scrape_many`'s `asyncio.gather` *looked* concurrent, but each blocking `make_request` serialized the others anyway. Moving them to `asyncio.to_thread` made the concurrency real. Measured with the lock removed: **6 concurrent entries into one session**.

Fixed with a lock rather than thread-local sessions, deliberately: `self.session` also carries the auth configuration (`configure_api_token` sets `.auth`/`.headers`/`.cookies` on it), so per-thread copies would silently stop inheriting any configuration applied after a thread's first call. Serialized-but-off-loop still delivers what the offload was for — the UI no longer freezes for the request duration. Both session request sites are guarded.

**#2** test double renamed to PascalCase.

Two notes on the regression test, since both were mistakes worth recording:
- My first version wrapped a **fake** auth's `make_request`, which bypassed the production lock entirely — it passed while measuring only the stub. It now drives the real `ConfluenceAuth.make_request` with only `session.request` stubbed.
- It asserts a request actually *reached* the session, which immediately caught the egress policy rejecting the synthetic host before the session was ever involved.

Gate: `Tests/Web_Scraping/` + config concurrency **230 passed, 3 skipped**.

PR #1933, opened 2026-08-22
feat(console): make Library retrieval and recovery truthful

rmusser01:
Design only. One file, no code.

## The problem, as found in the code

Console can consult a user's Library in three ways and distinguishes none of them on screen.

A user may press **Search Library**, which stages evidence for the next send. A switch buried in a modal, `rag_auto_retrieve_on_send`, does the same automatically on every send, using the draft as the query. And the model itself may call a Library tool, which `[console].direct_library_tools` registers **by default** whenever tools run at all. These are not variants of one feature: they are three mechanisms with three different owners — the user, the application, the model.

A fourth path, `get_rag_context_for_chat`, is dead. Every reference to it outside its own definition is a docstring.

Four faults follow:

- **The chip misreports itself.** `Library search: on|off` looks like a switch. It is a status — "on" means evidence is staged. Nothing can be toggled there.
- **The query is not the message.** A separate input, backed by `_console_draft_looks_like_rag_query`, a heuristic that guesses whether a draft counts as a query.
- **Automatic retrieval is global.** One switch governs every conversation, indefinitely. Enable it for research and unrelated chats quote your files weeks later, with no indication why.
- **Review is unusable in practice.** The staged tray emits 3–4 rows per source into a box capped at ten. A five-source retrieval renders ~22 rows; two are visible without scrolling. The count lie this caused was fixed before; the density was not.

## Why the remedy is mostly consistency

This codebase already treats **what** to retrieve as a property of a conversation — `rag_scope` in `conversations.metadata`, alongside `active_dictionaries`, with context policy following the same per-conversation pattern. Only **whether** to retrieve stayed global, and that is the more dangerous half: scope narrows the blast radius, mode decides whether there is a blast.

## What the design proposes

**An explicit per-conversation mode — Off / Manual / Auto** — stored beside `rag_scope`, sparse, `None` inherits the global default. The global setting becomes the default for new conversations rather than a live switch reaching into old ones.

**Off withholds the Library tool from the model** for that conversation. Disclosure without control was rejected: the complaint this answers is chats including library material the user did not ask for, and the model's own tool does exactly that, on by default. The seam already exists — `tool_configuration` is captured per turn and `_library_provider_for_context` already receives it.

**One row per source** in the tray, snippet on expand, with agent-initiated retrievals shown in the same place and marked as the assistant's doing.

**The draft-sniffing heuristic is removed.** In Manual and Auto the draft *is* the query.

## The hazard the spec records prominently

`write_conversation_scope` fails soft on corrupt metadata — the write is skipped, a warning logged. Prudent for scope, since the alternative risks erasing `active_dictionaries`. For mode it is close to disastrous: a user selects **Off**, is told nothing, and on resume the conversation inherits **Auto** and the Library is read anyway.

This design exists to stop controls asserting outcomes they did not produce. Routed through that seam unchanged, the storage layer would reintroduce that exact fault beneath the switch meant to cure it. The spec requires the mode write to fail loudly.

## Out of scope, with reasons

- **Unifying evidence with attachments** — cleanest mental model, but needs per-source send granularity the code deliberately declines to advertise.
- **Deleting the dead path** — housekeeping, no user surface.
- **TASK-406 / TASK-2375** — separate defects about content reaching the model at all.

## Next step

Review this file. On approval it goes to the writing-plans skill for an implementation plan. No code is touched before then.

---
## Summary by cubic
Makes Console Library retrieval and assistant access explicit per conversation and adds durable, recoverable sends. Previously a global auto-retrieve switch and default Library-tool registration controlled behavior; now a per‑conversation policy (Off/Manual/Auto) and a single immutable pre‑dispatch retrieval gate govern sends, with durable turn acceptance, device‑local checkpoints, and ADR‑063 continuation handoff.

- Persists policy in schema v45 beside `rag_scope` in `conversations.metadata`. Legacy databases must be opened with `ConsoleLibraryMigrationSeed` to sanitize `auto_retrieve_on_send`; changing the global default does not alter existing conversations.
- Governs assistant access via policy: `tool_configuration["direct_library_tools"]` reflects the per‑conversation setting; Off/Blocked withholds the Library provider; Library tool names are reserved; provider authority is bound per instance; boundaries fail closed.
- Retrieval and review: in Manual/Auto the draft is the query; one immutable automatic Library preparation gates eligible sends; the staged tray renders one row per source and marks assistant‑initiated retrievals.
- Durable acceptance and recovery: accept turns in one transaction before provider entry; write device‑local dispatch checkpoints; support resume, explicit Retry/Discard, and post‑commit interruption recovery; after the first tool batch, ownership transfers to ADR‑063 continuations; assistant generation state is preserved and exported.
- Destination disclosure: classify and record resolved egress as on‑device/private/public/unknown from the normalized endpoint only.

Addresses TASK-19900 (Console Library controls: explicit per‑conversation retrieval and assistant access).

<sup>Written for commit 1662c9383f452afb776f33e9fdfc6d702de7726d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1933?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
CI repair c25ea897c regenerates the consolidated recovery CSS and diagnostic inventory. Statement-level diagnostic review: removed entries are the replaced temporary-save and automatic-retrieval paths; added entries contain only bounded identifiers, event literals, or draft length—no prompt content, credentials, paths, or URLs. Local evidence: 70 targeted generator/workflow/recovery tests passed; all five derived-artifact guards, Ruff, formatter, and diff checks passed.

PR #1934, opened 2026-08-22
Security: refuse agent writes into a repository's .git metadata (TASK-19700)

rmusser01:
## What this fixes

An agent's write tools could create or rewrite `.git/config` and `.git/HEAD` inside a workspace root. That is the **upstream cause** of the four repository-supplied argv vectors fixed defensively in TASK-16801 (PR #1914) — with a hostile `.git/config`, an ordinary `git push` became a forced update or a ref deletion, and an option-shaped remote or branch name reached git's argv without any dangerous flag appearing in our own code.

Adds `is_git_metadata_write()` to `Utils/sensitive_paths.py` — the module both write families already import — matching `.git` as an exact path **component**, and covering the `.git` *file* a linked worktree carries as well as the directory.

## Policy (AC #1)

| | |
|---|---|
| Writes under any `.git` component | **Refused** |
| `.gitignore`, `.gitattributes`, `.github/` | **Allowed** — ordinary tracked files sharing only a name prefix |
| Reads | **Unchanged** — ADR-032 adopted `allow_hidden` so a coding agent can read dotfile config |

Both exceptions named when the task was filed dissolve on inspection: "fix the gitignore" edits `.gitignore`, not `.git/`; and the app tooling that writes `.git/info/exclude` is the shadow tracker, which owns a private `GIT_DIR` and never routes through these tools. `.gitattributes` has a security history (textconv drivers) but that vector was closed in TASK-16801 with machine-safe diff flags, so blocking the file would cost usability for no remaining gain.

## The task's premise was half wrong — verified, not assumed

The task said neither write family excluded `.git/`. **Only one did.** Mutation testing while implementing showed `Tools/file_operation_tools.py` refuses *every* hidden component already ("Access to hidden files/directories is not allowed"), so `.git/` was unreachable there — disabling the new check leaves its test green.

The real gap was `Tools/local_tool_impls.py`, whose `resolve_workspace_path` sets `allow_hidden=True` per ADR-032 precisely so a coding agent can read dotfiles. That is where the guard is load-bearing and where the five refusal tests are red-first. The check is still wired into the other family as defense-in-depth against a future `allow_hidden` adoption, and its tests **name the mechanism actually doing the work today** rather than crediting the new one for a refusal it didn't produce.

## Test-quality note

The first `patch_files` test used another tool's `*** Begin Patch` envelope, which this parser rejects as `invalid_diff` — so it passed without ever reaching the guard. It now uses a real unified diff, paired with a control proving the same payload *applies* to an ordinary path.

## Verification

- **682 passed** across `Tests/Tools/` + `Tests/Utils/test_sensitive_paths.py`; ruff clean.
- Refusal tests proven red before the fix; the not-load-bearing check proven not-load-bearing by mutation.
- Blast-radius audit (AC #5) in the task file: which features shell out to git against agent-writable roots, and one residual recorded honestly — the shadow tracker diffs the working tree without machine-safe flags, reachable only via a driver in the user's *global* gitconfig, which an agent cannot write.

Follows TASK-16801 (PR #1914). Read-side question deliberately **not** addressed: `.git/config` can embed a credential in a remote URL, which is a disclosure problem with a different shape.

---
## Summary by cubic
Refuses agent writes into a repository’s .git metadata within workspace roots. Previously `local_tool_impls` could create or edit `.git/config` and `.git/HEAD`; now writes to any path with a `.git` component are rejected (component-exact, case-insensitive), preventing repo reconfiguration that could turn safe pushes into destructive operations.

- Adds `is_git_metadata_write` in `tldw_chatbook/Utils/sensitive_paths.py`, matching `.git` as an exact path component case-insensitively and covering the `.git` file used by linked worktrees.
- Enforces the guard in `tldw_chatbook/Tools/local_tool_impls.py` for write intent; also applied in `tldw_chatbook/Tools/file_operation_tools.py` as defense-in-depth (that family already blocks hidden components).
- Keeps `.gitignore`, `.gitattributes`, and `.github/` writable; reads unchanged per ADR-032.
- Tests refuse writes to `.git/config`, `.git/HEAD`, hooks under `.git/`, the worktree `.git` file, and case-variant paths; a control proves the same patch applies outside `.git`. Errors name the refused path.
- No migration required; satisfies TASK-19700 acceptance criteria with policy and blast-radius audit recorded in the task.

<sup>Written for commit 2dbf5b2a8878c021be82653eea9ee5d360da0ff4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1934?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Tools/test_file_tool_sandbox.py
+++ b/Tests/Tools/test_file_tool_sandbox.py
@@ -409,3 +409,58 @@ def test_default_sandbox_configuration_still_works_end_to_end():
     grep_result = asyncio.run(fot.GrepFiles().execute(pattern="hi there"))
     assert "error" not in grep_result, grep_result
     assert any(m["path"].endswith("hello.txt") for m in grep_result["matches"])
+
+
+# ---------------------------------------------------------------------------
+# TASK-19700: `.git/` metadata and THIS family.
+#
+# Verified by mutation while writing these: this family does NOT have the
+# gap. `validate_path_multi` refuses EVERY hidden component ("Access to
+# hidden files/directories is not allowed"), so `.git/config` was already
+# unreachable here -- disabling the new `is_git_metadata_write` check leaves
+# the refusal test passing. The real gap was in the OTHER family
+# (`local_tool_impls`), which sets `allow_hidden=True` per ADR-032 so a
+# coding agent can read dotfiles; that is where the guard is load-bearing
+# and mutation-proven (see `Tests/Tools/test_local_tool_sensitive_paths.py`).
+#
+# The check is still wired here as defense-in-depth -- if this family ever
+# adopts `allow_hidden` the way ADR-032 drove the other one to, the guard is
+# already in place -- but these tests are honest about which mechanism does
+# the work today, rather than crediting the new one for a refusal it did not
+# produce.
+# ---------------------------------------------------------------------------
+
+
+def test_write_file_tool_refuses_git_config_today_via_the_hidden_rule(
+    monkeypatch, tmp_path
+):
+    """Regression pin, NOT red-first: the hidden-component rule is what
+    refuses this today. Pinned so a future `allow_hidden` adoption here
+    cannot silently open `.git/` writes without a test noticing."""
+    sandbox = tmp_path / "sandbox"
+    (sandbox / ".git").mkdir(parents=True)
+    (sandbox / ".git" / "config").write_text("[core]\n")
+    monkeypatch.setattr(fot, "_tool_sandbox_root", lambda: sandbox.resolve())
+
+    result = asyncio.run(
+        fot.WriteFileTool().execute(
+            file_path=".git/config", content='[remote "--force"]\n'
+        )
+    )
+    assert "error" in result, f"the write was allowed: {result}"
+    assert (sandbox / ".git" / "config").read_text() == "[core]\n"
+
+
+def test_is_git_metadata_write_predicate_is_exact_on_components():
+    """The shared predicate itself -- the part that IS load-bearing in the
+    other family. Prefix lookalikes must stay writable."""
+    from tldw_chatbook.Utils.sensitive_paths import is_git_metadata_write
+
+    assert is_git_metadata_write(Path("/repo/.git"))
+    assert is_git_metadata_write(Path("/repo/.git/config"))
+    assert is_git_metadata_write(Path("/repo/.git/hooks/pre-commit"))
+    assert is_git_metadata_write(Path("/repo/nested/.git/config"))
+    assert not is_git_metadata_write(Path("/repo/.gitignore"))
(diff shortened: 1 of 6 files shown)


rmusser01:
## Qodo finding addressed — commit `2dbf5b2a8`

### 1. Case-variant `.git` bypass — High ✅ correct, and confirmed exploitable

Reproduced before fixing, on this macOS machine (case-insensitive by default):

```
FS case-insensitive? True
WRITE ALLOWED (bypass!)
actual .git/config now: '[remote "--force"]\n'
```

The hostile remote landed in the **real** `.git/config` — the guard was defeated entirely. Now casefolded, still component-exact so `.GITIGNORE` stays writable. Mutation-proven: reverting the casefold fails the new test. **684 passed** across `Tests/Tools/` + `Tests/Utils/test_sensitive_paths.py`, ruff clean.

---

## ⚠️ While verifying this, I found the same bypass in the **pre-existing** denylist — filing separately

This is **not** introduced by this PR; it is in the TASK-19551 credential denylist that `is_sensitive_path` implements, and it is materially more serious than the finding above.

```
~/.ssh/id_ed25519    -> sensitive=True
~/.SSH/id_ed25519    -> sensitive=False
~/.aws/credentials   -> sensitive=True
~/.AWS/credentials   -> sensitive=False
```

End-to-end through `fs_read`, against the app's own denylisted config file:

```
tldw_cli/config.toml   -> refused (Refused: '...' is a protected path...)
TLDW_CLI/config.toml   -> ALLOWED, returned 32782 chars   <-- BYPASS
```

That is the user's real config — the file holding their provider API keys — read straight through the denylist by changing the case of one path component. The same shape applies to `~/.ssh`, `~/.aws`, and `mcp_permissions.json`, and bypassing that last one turns every `ask` into `allow`, which is the permission-gate bypass TASK-19551 existed to close. Reachable by prompt injection, on the two platforms where case-insensitive filesystems are the default.

I am **not** folding it into this PR: it is a different defect with its own test surface (~11 config accessors, dynamic dirs, symlink resolution), and burying a credential-disclosure fix inside a `.git` guard would be the wrong place for it to get reviewed. Filing it as a high-priority task and taking it as the very next PR, ahead of the remaining polish follow-ups.

Worth noting the general question it raises, which that task will settle rather than guess: case is one filesystem normalization, and Windows also strips trailing dots and spaces (`.git.` → `.git`). The denylist should normalize deliberately, not per-call-site.

PR #1935, opened 2026-08-22
perf: efficiency burn-down batch 3 (task-19560 fixed; 2902 + 18601 measured)

rmusser01:
## What & why

Third efficiency batch. One task fixed end to end, and **three tasks where the honest deliverable was a measurement** — including one that would otherwise have been re-implemented for a third time.

## task-19560 — Kokoro downloads: unbounded, on the loop, and non-atomic

Three defects shared one code path:

* `requests.get(url, stream=True)` carried **no timeout**, so a half-open connection hung forever — no error, no way out;
* the call ran inline inside `async def`, so a several-hundred-megabyte transfer blocked the event loop (the whole TUI) for its duration;
* the model/voice downloads streamed **straight to the final path**, so an interrupted transfer left a truncated file that the next run's `os.path.exists()` check accepted as a complete model.

All four download sites now route through one `_kokoro_stream_download` helper: connect+read timeouts (read is per-chunk, so a slow-but-progressing transfer isn't killed), a `.part` file promoted with `os.replace` only once fully read, partial cleanup on any `BaseException` including cancellation, and progress logging with a percentage when `content-length` is present. The two ONNX sites keep their existing checksum-verify and move logic — only the transfer moved.

**Not done, flagged rather than counted:** progress goes to the log, not a UI progress bar. Wiring a visible bar needs a caller that consumes a callback, and none exists today.

Also bounded the two summarization POSTs the task names, using the config-driven idiom the OpenAI path already used.

### A runtime bug caught mid-implementation

`Local_Summarization_Lib` reads its timeout via `get_cli_setting` — in a module that never imported it. That would have raised `NameError` on the first real call, and **no AST-level test would have seen it**. The import is added, and a test now pins that the accessor resolves *and* returns a usable value.

### Scope finding: 27 more unbounded calls, reported not fixed

The task names 2 timeout-less requests. A static audit of those two modules finds **29**. Bounding an arbitrary two of twenty-nine leaves the same hang-forever hazard everywhere else, and the right answer is a session-level default (requests has no native per-session default, so it needs a small `Session` subclass or adapter) — a design change deserving its own task. Full site list is in the task notes.

## task-2902 — re-measured, and **do not re-attempt the deferral**

This was on my own shortlist as "owner ruled proceed now". Reading its notes first was the difference between a day of work and a paragraph: it has been **implemented and reverted twice**. Round 2 measured the widget deferral at **4–8%** and concluded Console is not widget-mount-bound, filing the real levers as task-3010 (sync storm) and task-3011 (1,352 SQLite connections per push). **Both are now Done.**

So I re-measured, as the task asks. Under the shared harness, 235×52, 5 interleaved runs each:

```
chat    median 0.221s
notes   median 0.218s
```

Console is no longer distinguishable from a structurally simpler screen, against round 2's 2.27–2.54s.

**And that measurement does not close the AC — stated plainly in the task.** `_build_test_app` fakes every real I/O seam including the DB, and task-3011 was *specifically* about `Workspace_DB` opening 1,352 real connections per push. A faked-DB harness cannot observe the cost 3011 removed. The AC asks for a live measurement; this is not one. The task now names the live probe as the next step, with an explicit warning not to re-attempt the deferral.

## task-18601 — the premise, quantified

`append_steps` re-reads, re-parses and rewrites the **entire** step blob on every append — O(n) per step, O(n²) per run. Measured against a real DB:

| append # | cost |
|---|---|
| 1 | 0.05 ms |
| 500 | 0.49 ms |
| 2000 | **2.18 ms** (44× the first) |

2,000 appends take 2.09s. Extrapolated to the 25,000-step budget AC #1 names: **~5.4 minutes** of write churn for one run, before the viewer reads any of it back.

Not started — it's an arc (child table + schema bump + compat read path + viewer paging). The task now carries a 3-way split so it can be planned against a number rather than an adjective.

## Red-proof evidence

| Change | Reverted → |
|---|---|
| Kokoro helper | 3 of 5 fail (timeout, atomicity, off-loop). The other two are explained in the commit: the partial revert left the cleanup path intact, and the hasher test covers an orthogonal property. |
| summarization timeouts | 2 of 3 fail, including the missing-import test |

## Testing

New tests **8 passed**; all three changed modules import cleanly.

Pre-existing and unrelated, verified by A/B with dev's `kokoro.py` (**13 failed both ways**): `Tests/TTS/test_kokoro_backend_profile_paths.py`. Several other TTS modules don't collect without `numpy`.

---
## Summary by cubic
Hardens Kokoro asset downloads and bounds two summarization POSTs to eliminate hang-forever paths and fix a Python 3.11 import break. Previously `requests` ran on the event loop with no timeouts and wrote to final paths; now downloads run off-loop with connect/read timeouts, complete atomically, and clean up on error.

- Kokoro (`tldw_chatbook/TTS/backends/kokoro.py`): all model/voice fetches use `_kokoro_stream_download` with `(connect, read)` timeouts, `asyncio.to_thread` offload, `.part` + `os.replace` atomicity, partial cleanup, destination `validate_path`, and progress logged at most every 2s. ONNX flows keep checksum-verify/move but now use `tempfile.mkstemp` for unique temp paths; removes a 3.12-only f-string that broke imports on 3.11 and avoids concurrent clobbering.
- Summarization (`tldw_chatbook/LLM_Calls/Summarization_General_Lib.py`, `tldw_chatbook/LLM_Calls/Local_Summarization_Lib.py`): added config-driven timeouts to the Anthropic and local-LLM POSTs; fixed a missing `get_cli_setting` import. Tests pin both timeouts and the import.
- Guardrail: new `Tests/Architecture/test_python_floor_syntax.py` compiles every module under the declared Python floor to catch future 3.12-only syntax.
- Scope note: a static audit found 27 additional unbounded `requests` calls; recorded for a follow-up that adds a session-level default timeout.
- Tasks: closes TASK-19560; records a re-measurement for TASK-2902 (live probe still required) and quantifies TASK-18601’s O(n²) write cost (planning only).

<sup>Written for commit ea4fa2130c5e575c025ef81566a4990507da7c0f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1935?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Architecture/test_python_floor_syntax.py
+++ b/Tests/Architecture/test_python_floor_syntax.py
@@ -1,0 +1,98 @@
+"""Every shipped module must parse on the project's MINIMUM Python (TASK-19560 review).
+
+`pyproject.toml` declares `requires-python = ">=3.11"`, but development here
+happens on a much newer interpreter. That gap hid a real defect: a nested
+same-quote f-string (`f"...{"literal"}..."`) is legal only from Python 3.12
+(PEP 701), so `TTS/backends/kokoro.py` failed to import **entirely** on 3.11
+while every local test passed on 3.14.
+
+Two things made it invisible:
+
+* `ast.parse(..., feature_version=(3, 11))` does NOT reproduce it -- that
+  argument does not downgrade the tokenizer, so it happily accepted the
+  3.12-only form. A green `feature_version` check is not evidence.
+* The test suite ran on the developer's interpreter, where the syntax is
+  valid, so the module imported fine and its tests passed.
+
+This guard compiles every module under a real minimum-version interpreter
+when one is available, and SKIPS with a clear reason when it is not -- rather
+than silently passing and re-creating exactly the blind spot it exists to
+close.
+"""
+
+from __future__ import annotations
+
+import re
+import shutil
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+PROJECT_ROOT = Path(__file__).resolve().parents[2]
+PACKAGE = PROJECT_ROOT / "tldw_chatbook"
+
+
+def _declared_floor() -> tuple[int, int]:
+    """Read `requires-python` from pyproject rather than hardcoding it.
+
+    Returns:
+        The (major, minor) minimum version the project claims to support.
+    """
+    text = (PROJECT_ROOT / "pyproject.toml").read_text(encoding="utf-8")
+    match = re.search(r'requires-python\s*=\s*"[^"]*?(\d+)\.(\d+)', text)
+    assert match, "could not read requires-python from pyproject.toml"
+    return int(match.group(1)), int(match.group(2))
+
+
+def _floor_interpreter(floor: tuple[int, int]) -> str | None:
+    """Locate an interpreter matching the declared floor, if one exists."""
+    exact = shutil.which(f"python{floor[0]}.{floor[1]}")
+    if exact:
+        return exact
+    uv = shutil.which("uv")
+    if not uv:
+        return None
+    found = subprocess.run(
+        [uv, "python", "find", f"{floor[0]}.{floor[1]}"],
+        capture_output=True, text=True,
+    )
+    candidate = found.stdout.strip()
+    return candidate if found.returncode == 0 and candidate else None
+
+
+def test_every_module_compiles_on_the_declared_python_floor() -> None:
+    floor = _declared_floor()
+    if sys.version_info[:2] == floor:
+        pytest.skip("already running on the declared floor")
+
+    interpreter = _floor_interpreter(floor)
+    if interpreter is None:
+        pytest.skip(
+            f"no Python {floor[0]}.{floor[1]} available to check the declared "
+            "floor against; install one (`uv python install "
+            f"{floor[0]}.{floor[1]}`) to enable this guard"
+        )
+
+    probe = (
+        "import sys, pathlib\n"
+        "bad = []\n"
+        f"for f in pathlib.Path({str(PACKAGE)!r}).rglob('*.py'):\n"
+        "    try:\n"
+        "        compile(f.read_text(encoding='utf-8'), str(f), 'exec')\n"
+        "    except SyntaxError as e:\n"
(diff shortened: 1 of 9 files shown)


rmusser01:
Rebased onto latest `dev` and addressed all four findings. **#1 was a genuine shipped breakage that my environment hid — thank you.**

The ONNX temp path used a nested same-quote f-string, `f"kokoro-download-{os.getpid()}-{"model.onnx"}"`. That's legal only from Python 3.12 (PEP 701), and `pyproject` declares `requires-python = ">=3.11"` — so `TTS/backends/kokoro.py` **failed to import entirely** on the project's minimum supported Python. Confirmed against a real 3.11:

```
SyntaxError line 383: f-string: expecting '}'
```

Every local test passed because this venv is 3.14.

**#3** (same lines): the path was pid-deterministic, so two concurrent initialisations in one process clobbered each other's in-flight download and the checksum/move then operated on the wrong file. Both fixed together with `tempfile.mkstemp` — unique per call, created 0600.

**#4**: `_kokoro_stream_download` now validates its destination through `path_validation`. **#2**: docstrings added.

I also added `Tests/Architecture/test_python_floor_syntax.py`, which compiles every shipped module under a *real* minimum-version interpreter. Two things had made this class of defect invisible, and both are documented in the guard:

- the suite runs on the developer's newer interpreter, so the module imports fine;
- `ast.parse(..., feature_version=(3, 11))` does **not** downgrade the tokenizer — it accepted the 3.12-only form. A green `feature_version` check is not evidence.

The guard skips with an explicit reason when no floor interpreter is installed, rather than passing silently and re-creating the exact blind spot it exists to close. Red-proofed by reintroducing the f-string: it fails naming `kokoro.py:397`.

Gate: **9 passed** across the Kokoro, summarization-timeout and new floor-guard suites.

PR #1936, opened 2026-08-22
Security: sensitive-path denylist bypassable by path case (TASK-19800)

rmusser01:
## A live credential-disclosure bypass in the existing denylist

The credential/gate-state denylist added by TASK-19551 compared resolved paths with **exact, case-sensitive equality**. macOS and Windows filesystems are case-insensitive by default, and `Path.resolve()` does **not** canonicalise case on them — it resolves symlinks and `..` but preserves whatever spelling the caller typed. So a case-variant reaches the same file while matching no denylist entry.

Verified end-to-end before the fix, through `fs_read`, against this app's own denylisted config:

```
tldw_cli/config.toml   -> refused (Refused: '...' is a protected path...)
TLDW_CLI/config.toml   -> ALLOWED, returned 32782 chars    <-- BYPASS
```

That is the real config file — the one holding provider API keys — read straight through the denylist by changing the case of one path component. The predicate agrees:

```
~/.ssh/id_ed25519   -> sensitive=True      ~/.SSH/id_ed25519   -> sensitive=False
~/.aws/credentials  -> sensitive=True      ~/.AWS/credentials  -> sensitive=False
```

The same shape reaches `~/.ssh`, `~/.aws` and `mcp_permissions.json`. Bypassing that last one turns every `ask` into `allow` — the permission-gate bypass TASK-19551 exists to prevent — and it is reachable by prompt injection on the two platforms where case-insensitive filesystems are the default.

**How it was found:** Qodo flagged the identical weakness in TASK-19700's brand-new `.git` write guard (PR #1934). Checking whether the older, shipped denylist shared it showed that it did.

## The fix

Every denylist comparison now goes through `_compare_key` — casefolded, component-wise — via `_same_path`/`_is_within`, covering denied files, database paths and their WAL/SHM/journal sidecars, denied directories, and the direct-child container rule.

**Casefolding is unconditional**, not platform-gated or probed. Platform is only a proxy for the real question (macOS can be configured case-sensitive; Linux can mount a case-insensitive volume), a per-path probe would add I/O to a check that runs on every candidate, and the error directions are not symmetric: over-refusing a genuinely distinct `~/.SSH` costs one explained refusal of a very unusual path, while under-refusing leaks a credential.

## Confinement is deliberately NOT folded

The two checks fail in **opposite directions**, so applying this "consistently" would be a security regression:

| Check | Folding produces | Direction |
|---|---|---|
| Denylist — "is this forbidden?" | extra refusals | fails **safe** |
| Confinement — "is this inside the root?" | extra **admissions** (`/Root/evil` counts as inside `/root`) | fails **open** |

`is_within` in `Tools/file_operation_tools.py` is left case-sensitive for that reason, and the asymmetry is documented at `_compare_key` so a later consistency pass doesn't undo it.

## Verification

- **688 passed** across `Tests/Tools/` + `Tests/Utils/test_sensitive_paths.py`; ruff clean.
- Three bypass tests proven **red first**; removing the casefold turns them red again (mutation-proven).
- Ancestry and lookalike behaviour pinned unchanged: `~/.sshfoo` is still not `~/.ssh`.

## Deliberately not addressed

Windows also normalises trailing dots and spaces (`.ssh.` → `.ssh`). That is a real second normalization, but it is Windows-only — on POSIX `.ssh.` is a genuinely different directory, so stripping there would over-refuse for no gain — and it wants a platform-gated rule of its own rather than being folded into this change.

Follows TASK-19700 (#1934) and TASK-19551.

---
## Summary by cubic
Case-variant paths could bypass the sensitive-path denylist on case-insensitive filesystems. Previously we compared resolved paths case-sensitively; now we compare component-wise with unconditional casefolding across denylist checks and the folder-binding overlap gate to prevent credential and permission-gate leaks.

- Routes all denylist comparisons through `_compare_key`/`_same_path`/`_is_within` for denied files, DB paths and sidecars, denied directories, and the direct-child container rule in `tldw_chatbook/Utils/sensitive_paths.py`.
- Applies the folded comparison to `find_root_binding_conflict` so binding a case-variant root that overlaps a protected path is rejected; equality is excluded in ancestry cases to preserve most-specific reporting.
- Keeps confinement checks case-sensitive by design; do not refactor them to use these helpers.
- Adds tests that fail red without the fix, including end-to-end `fs_read` refusal of a case-variant config; lookalikes like `~/.sshfoo` remain allowed. No migrations; expect added refusals for case-only-distinct paths on case-sensitive volumes. Linear: TASK-19800.

<sup>Written for commit c89038e06e602f387aba328dff30b04683ab89f0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1936?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Tools/test_local_tool_sensitive_paths.py
+++ b/Tests/Tools/test_local_tool_sensitive_paths.py
@@ -795,3 +795,81 @@ def test_case_variants_do_not_over_refuse_gitignore(tmp_path):
     (root / ".git").mkdir(parents=True)
     write_file(".GITIGNORE", "build/\n", workspace_root=root)
     assert (root / ".GITIGNORE").read_text() == "build/\n"
+
+
+# ---------------------------------------------------------------------------
+# Shape 6: case-variant spellings of a denylisted path (TASK-19800).
+#
+# Found while fixing TASK-19700's `.git` guard for the same weakness. macOS
+# and Windows filesystems are case-insensitive by DEFAULT, and
+# `Path.resolve()` does NOT canonicalise case there (it resolves symlinks
+# and `..`, but preserves what the caller typed) -- so `~/.SSH/id_rsa` opens
+# the same file as `~/.ssh/id_rsa` while comparing unequal to every
+# denylist entry.
+#
+# Reproduced end-to-end before the fix against the app's OWN config file:
+#   tldw_cli/config.toml -> refused
+#   TLDW_CLI/config.toml -> ALLOWED, returned 32782 chars
+# i.e. the user's provider API keys, read straight through the denylist by
+# changing the case of one path component. The same shape reaches `~/.ssh`,
+# `~/.aws` and `mcp_permissions.json` -- and bypassing that last one turns
+# every `ask` into `allow`, which is the permission-gate bypass TASK-19551
+# exists to prevent.
+# ---------------------------------------------------------------------------
+
+
+def test_is_sensitive_path_matches_case_variants_of_a_denied_directory():
+    from tldw_chatbook.Utils.sensitive_paths import is_sensitive_path
+
+    _plant_ssh_key()
+    home = _home()
+    for spelling in (".ssh", ".SSH", ".Ssh", ".sSh"):
+        assert is_sensitive_path(home / spelling / "id_rsa"), (
+            f"~/{spelling}/id_rsa must be denied: on a case-insensitive "
+            f"filesystem it IS ~/.ssh/id_rsa"
+        )
+
+
+def test_is_sensitive_path_matches_a_case_variant_of_a_denied_file():
+    from tldw_chatbook import config as app_config
+    from tldw_chatbook.Utils.sensitive_paths import is_sensitive_path
+
+    real = app_config._get_effective_config_path()
+    real.parent.mkdir(parents=True, exist_ok=True)
+    real.write_text("[API]\nopenai_api_key = \"SYNTHETIC-19800\"\n")
+    variant = real.parent.parent / real.parent.name.upper() / real.name
+    assert is_sensitive_path(real)
+    assert is_sensitive_path(variant), (
+        f"{variant} must be denied: it is the same file on a "
+        f"case-insensitive filesystem"
+    )
+
+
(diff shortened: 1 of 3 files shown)


rmusser01:
## Qodo finding addressed — commit `c89038e06`

### 1. Binding gate still case-sensitive — High ✅ correct, and I missed it

You're right, and the reasoning is exactly right: `find_root_binding_conflict()` enforces the **same** protected-path policy as `is_sensitive_path`, just at folder-bind time rather than per-access — so leaving it case-sensitive left the identical bypass one layer up, where the consequence is a workspace root that overlaps a protected directory and widens tool reachability into it.

Proven red first:

```
FAILED test_binding_gate_detects_a_case_variant_conflict
```

All three of its cases now compare through the same folded key — root **is** protected, root **nested inside** a protected dir, root **contains** a protected dir/file — with equality excluded from the two ancestry cases so the "most specific match" reporting behaviour is unchanged (deepest enclosing / shallowest contained).

Also noting you explicitly flagged that confinement checks should stay unchanged. Agreed, and that asymmetry is documented at `_compare_key`: folding a denylist adds refusals and fails safe, folding confinement adds *admissions* and fails open.

## Verification

- **1070 passed** across `Tests/Tools/` + `Tests/Utils/test_sensitive_paths.py` + `Tests/Workspaces/`; ruff clean.
- The new binding-gate tests are red-first; a control pins that an unrelated root still binds cleanly.
- One unrelated failure appears in a broad `-k "registry or binding"` sweep — `Tests/UI/test_library_ingest_retry_last.py::test_registry_ticks_only_reflow_footer_when_retry_availability_changes`. It is **pre-existing on dev**, confirmed against a throwaway worktree pinned to the current dev tip (1 failed, 12 passed there). It matches only on the keyword "registry" and concerns Library ingest footer reflow.

PR #1937, opened 2026-08-22
fix/perf: watchlists duplicate-alert guard, off-loop sqlite, nesting-safe transaction (task-19562 A, B, C)

rmusser01:
## What & why

Fourth efficiency batch: **parts A and C of task-19562**. One user-visible correctness bug fixed, one silent-data-loss hazard closed — and one of the task's own claims refuted by measurement.

## Part A — overlapping feed checks fired the alert twice

`_check_url_guarded` (task-16838) serializes concurrent checks of the same source, but only for the url-family arms (`url`, `url_list`, `sitemap`). The **feed** arm and the **API** arm registered no claim at all — and feeds are the commonest source type. A scheduler tick overlapping a manual "Check Now" therefore ran the check **twice**: the alert notification fired twice and statistics double-counted.

Both arms now claim through the same registry, scoped to the whole source with a NUL-prefixed sentinel in the URL slot — no real URL can contain NUL, so it cannot collide with a url-family claim for the same subscription.

### The load-bearing part is the disposition, not the skip

A skipped feed check returns a `DISPOSITION_SKIPPED_IN_FLIGHT` disposition rather than the arm's usual `None`. That is not cosmetic. `_entirely_skipped_dispositions` → `execute_run` reads it to **skip source-health accounting**. Without it, a turned-away check takes `record_check_result`'s SUCCESS branch and resets the auto-pause breaker, clears `last_error`, and stamps `last_successful_check` — recording a successful check for a run that never contacted the source. Returning `[]` items with `None` dispositions would have been indistinguishable from a clean "nothing new" result.

Two comments in the file asserted the old invariant (*"feed/API runs can never skip"*, *"`None` for the feed and API arms"*). Both are corrected rather than left to rot into folklore.

One trap avoided deliberately: the claim key stores `id(db)` and holds **no reference** to it, so the file's own comment warns that extracting the claim into a helper taking only a precomputed key silently drops the liveness guarantee. The new helper therefore takes `db` and holds it in its own frame across the await.

## Part C — a nested transaction committed the outer one early

`SubscriptionsDB.transaction()` committed unconditionally on exit, so a nested `with self.transaction()` had its **inner** exit durably commit the **outer** transaction's work. A later failure in the outer scope could no longer roll back what the inner block had already written — silent partial persistence, no error anywhere.

Nesting is now tracked per thread (the connection is thread-local, so the depth must be too), mirroring `ChaChaNotes_DB`'s `TransactionContextManager` rather than inventing a savepoint scheme. Depth is cleared in a `finally`, so a raise cannot strand it and turn every later transaction into a no-op joiner.

### The task's named call site is refuted

The task states `record_check_result` nests `transaction()`. It **does not**. Instrumenting `transaction()` across a real call observed **depth 1, one entry**. The lane's CONFIRMED-LATENT rating was right about the hazard and wrong about that call site.

The fix stands anyway — it converts a silent-partial-persistence trap into a structural impossibility instead of relying on nobody ever nesting — but the claim is corrected in the task rather than quietly satisfied.

## Part B — 22 async methods did synchronous sqlite on the event loop

Every one blocked the loop (the whole TUI) for its query duration. They now route through `db_offload.run_db_off_loop`, the helper this codebase **already had** and already used in several methods — not a new mechanism.

The four `transaction()` sites were handled differently on purpose: that helper's contract forbids holding a transaction open across the thread boundary, so each block was extracted whole into an offloaded function rather than offloading statements inside an open transaction.

**A counting note worth recording.** My AST sweep for `db.<method>()` / `self._db().<method>()` found 19 methods. Three more — `get_alert_rule`, `list_runs`, `list_alert_rules` — reach the database through a bare `db.conn.cursor()` and were invisible to that pattern. Adding them gives **22**, which matches this task's original figure; the narrower scan would have quietly under-reported and left three methods blocking the loop. Verified after the change: **zero** inline db calls remain in any `async def` in the file.

Red-proofed: reverting `cancel_run`'s offload fails its test with `cancel_run's transaction opened on the event-loop thread`.

Note for anyone extending these tests: `run_db_off_loop` deliberately runs **inline** when `is_memory_db is True`, so a test DB must be file-backed or the offload correctly won't happen and the test proves nothing.

## Red-proof evidence

Both fixes were verified against the real unfixed code, on behaviour rather than vocabulary:

| Change | Reverted → |
|---|---|
| feed/API guard | `the feed was fetched 2 times for one overlap; the second entrant was not turned away` |
| nesting-safe transaction | `the nested block's write survived a failure in the outer scope: its exit committed the outer transaction early` |
| sqlite offload | `cancel_run's transaction opened on the event-loop thread: [8692277440]` |

The interleave is forced for real — the check blocks on an `asyncio.Event` and the second entrant starts while the first is still inside its await. The wait is deterministic rather than a race: the first entrant is gate-blocked and *cannot* progress, so the sleep only gives the second one loop time to reach the guard. (My first attempt used `sleep(0)`, which yielded once and let the second arrive too late — that produced a false pass, and is why the timing is spelled out in the test.)

## Testing

`Tests/Subscriptions/` **781 passed, 3 skipped** (from a 759 baseline, +22 new tests). Subscription-related DB tests **133 passed**. The pre-existing `Tests/UI/test_watchlists_check_now_skipped.py` still passes.

## Still open on task-19562

* **C residue** — the leaked thread-local connection that is never closed and never checkpoints the `-wal`. Untouched here.

### DIFF
--- a/Tests/Subscriptions/test_subscriptions_transaction_nesting.py
+++ b/Tests/Subscriptions/test_subscriptions_transaction_nesting.py
@@ -1,0 +1,109 @@
+"""`SubscriptionsDB.transaction()` must be safe to nest (TASK-19562 part C).
+
+It used to commit unconditionally on exit, so an INNER `with
+self.transaction()` durably committed the OUTER transaction's work too: a
+later failure in the outer scope could no longer roll back what the inner
+block had already written. Silent partial persistence, no error anywhere.
+
+Recorded for accuracy: the call site the task named (`record_check_result`)
+was instrumented and found NOT to nest -- observed depth 1. The hazard was
+latent, not live. These tests make it structurally impossible instead of
+relying on nobody ever nesting.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from tldw_chatbook.DB.Subscriptions_DB import SubscriptionsDB
+
+pytestmark = pytest.mark.unit
+
+
+@pytest.fixture
+def db(tmp_path):
+    return SubscriptionsDB(tmp_path / "subs.db", "test")
+
+
+def _names(database: SubscriptionsDB) -> set[str]:
+    with database.transaction() as conn:
+        return {
+            row["name"]
+            for row in conn.execute("SELECT name FROM subscriptions").fetchall()
+        }
+
+
+def test_failure_after_a_nested_block_rolls_the_whole_unit_back(db):
+    """The defect, stated as behaviour: the inner write must not survive."""
+    with pytest.raises(RuntimeError):
+        with db.transaction() as conn:
+            conn.execute(
+                "INSERT INTO subscriptions (name, type, source) VALUES (?,?,?)",
+                ("outer", "rss", "https://e.invalid/outer.xml"),
+            )
+            with db.transaction() as inner:
+                inner.execute(
+                    "INSERT INTO subscriptions (name, type, source) VALUES (?,?,?)",
+                    ("inner", "rss", "https://e.invalid/inner.xml"),
+                )
+            # Fails AFTER the inner block exited. Pre-fix, the inner exit had
+            # already committed both rows, so this rollback did nothing.
+            raise RuntimeError("outer scope fails")
+
+    surviving = _names(db)
+    assert "inner" not in surviving, (
+        "the nested block's write survived a failure in the outer scope: its "
+        "exit committed the outer transaction early"
+    )
+    assert "outer" not in surviving
+
+
+def test_nested_success_still_commits_once_the_outer_block_exits(db):
+    with db.transaction() as conn:
+        conn.execute(
+            "INSERT INTO subscriptions (name, type, source) VALUES (?,?,?)",
+            ("outer", "rss", "https://e.invalid/outer.xml"),
+        )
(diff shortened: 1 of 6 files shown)


rmusser01:
Rebased onto dev `e0f0b8b74` and addressed Qodo's review.

**Qodo: 0 bugs, 1 rule violation.** The single finding was that `transaction()`'s docstring omitted `Returns`/`Raises`. Fixed in `9f39869bb` — and worth more than a style tick here, because the two things a caller most needs to know about a *nesting-safe* transaction are exactly what was missing: the same connection object is yielded to an inner block (so an inner `with` joins the outer transaction rather than starting its own), and only the **outermost** block rolls back, with an inner one re-raising unchanged for the outermost to handle. Documentation only; no behaviour change.

**Rebase note.** One conflict, in `backlog/tasks/task-19562 - ...`, against the busy_timeout measurement #1932 added to the same file. Resolved by keeping **both** sections — they answer different questions (that measurement narrowed the exposure to writer-vs-writer under WAL; this branch's notes record parts A and B as shipped).

Gate: `Tests/Subscriptions/` **781 passed, 3 skipped**.

PR #1941, opened 2026-08-22
Change review: explain an empty review instead of implying nothing changed (TASK-19702)

rmusser01:
## The problem

A Default-workspace conversation shows an empty Change Review screen reading **"No file changes recorded for this conversation."** — copy that asserts the agent changed nothing. The app cannot know that, because it never watched anything.

Premise re-verified against the real registry rather than taken from the filing:

```
add_folder_binding(DEFAULT_WORKSPACE_ID, ...) -> WorkspaceRegistryServiceError:
    "Default workspace does not allow runtime bindings."
add_folder_binding("ws-x", ...)               -> OK, bindings = 1
```

Folder bindings delegate to `save_runtime_binding`, which refuses Default. No bound folder ⇒ no tracked roots ⇒ the bridge records no snapshots ⇒ empty screen. Same honesty rule the working-tree pane already follows by saying *unavailable* rather than *clean* when every root fails (TASK-16801).

## What changed

**The empty state now distinguishes the two things it can mean.** With no tracked roots it names the cause, states explicitly that this is *not* a report that nothing changed, and points at Settings ▸ Workspaces plus the Default limitation. With roots present the original copy is untouched — there it is true.

**The write/visibility mismatch is disclosed** (AC #3). Confirmed in `console_chat_controller.py`: the agent's file-tool confinement root comes from `tool_configuration["workspace_root"]` or `[console] workspace_root`, **falling back to `os.getcwd()`** — while change tracking follows bound folders, with nothing keeping them in agreement. An agent can therefore edit files this screen will never show, precisely where the user comes to check what it did. When a configured root isn't covered by a tracked one, the banner now names it.

## What deliberately did NOT change

No fallback to `[console] workspace_root` or the CWD to manufacture a reviewable root (AC #2). That was rejected in TASK-16801's whole-branch review and is worth restating: it would offer confirmed **commit and push** against a repository the workspace never bound — worse than showing nothing.

The Default workspace's inability to bind folders is also untouched. That's a deliberate registry rule; the remit here was that the user must not be left with a silent gap, not that the rule should be relaxed.

## Verification

- **533 passed** across the change-review suites + `Tests/Workspaces/`; ruff clean.
- Both new behaviours **mutation-proven**: collapsing the empty-state discrimination fails the new test (and the pre-existing one), and emptying the disclosure list fails the banner test.
- One pre-existing assertion updated: `test_pseudo_entry_absent_without_candidate_roots` asserted the old copy in exactly this scenario. Its subject — no pseudo-entry without candidate roots — is unchanged.

Follows TASK-16801 (#1914).

---
## Summary by cubic
Explains why Change Review is empty instead of implying nothing changed, and discloses agent‑writable but untracked directories (including the CWD fallback) to prevent silent gaps in Default-workspace conversations.

- Empty state: when no tracked roots, the copy names the cause and states it is not a “nothing changed” report; when roots exist, the original “No file changes recorded” copy remains.
- Disclosure banner: identifies an agent-writable root from `[console] workspace_root` or the process CWD when it is not covered by a tracked root; normalizes paths via `tldw_chatbook.Utils.path_validation.validate_path_simple`; logs and skips on config read errors; does not fabricate roots or relax the Default workspace rule (satisfies TASK-19702 AC #1–#3).
- Tests: new cases for untracked CWD disclosure and a control ensuring a tracked CWD yields no warning; existing assertion updated for the new empty-state copy (AC #4).

<sup>Written for commit 82fe5c42abeea367df9a57a447bb6caa4ec49ca0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1941?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_change_review_current_mode.py
+++ b/Tests/UI/test_change_review_current_mode.py
@@ -429,7 +429,10 @@ async def test_pseudo_entry_absent_without_candidate_roots(
         await _wait_for_detection(pilot, screen)
 
         assert _select_values(screen) == []
-        assert "No file changes recorded" in screen.diff_pane_text()
+        # TASK-19702: with no tracked roots the empty state now names the
+        # CAUSE rather than asserting nothing changed. This test's subject
+        # (no pseudo-entry without candidate roots) is unchanged.
+        assert "No folder is bound" in screen.diff_pane_text()
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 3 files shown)


rmusser01:
## Qodo review addressed — commit `82fe5c42a`

Three fixed, one declined with reasoning. **155 passed** across the change-review suites; ruff clean.

### 3. Cwd mismatch not disclosed — Bug ✅ correct, and the sharpest of the four
Right, and slightly embarrassing: the PR description itself says the agent's tools fall back to `os.getcwd()`, and the code then returned early exactly when `[console] workspace_root` is unset — omitting the disclosure in the one case that needed it most. It now mirrors the controller's own resolution, CWD included. Red-proven:

```
FAILED test_untracked_cwd_is_disclosed_when_workspace_root_is_unset
```

Paired with a control asserting a **tracked** cwd produces no warning — a spurious banner would just train users to ignore it.

### 2. Unvalidated path `resolve()` — Rule violation ✅ correct
I was ready to argue this one, since the value is user config used only for a display comparison with no filesystem access. Checking first was the right move: `path_validation.validate_path_simple` exists precisely for base-less cases ("without requiring a base directory"), which is exactly this — there is no confinement root here, only a path to normalise before comparing. Now routed through it.

### 4. Silent broad exception swallow — Observability ✅ correct
Fixed. A swallowed failure there silently drops a disclosure, which is the same masking class this screen exists to prevent — that pattern has been caught repeatedly across this workstream, so it is a fair hit.

### 1. Test docstrings — Rule violation ⚠️ half taken
One of the three new tests genuinely lacked a docstring; added.

Declining the accompanying `Args:`-section request. **Zero** tests in this file document their pytest fixtures that way, and only 58 of 624 UI test files contain `Args:` anywhere (helpers, not test functions). Adding `Args: monkeypatch, tmp_path` here would break the file's convention and add noise without information — the Google-style requirement in CLAUDE.md targets public APIs, not pytest fixture parameters. Happy to revisit if the convention is meant to cover tests too, but that should change the file as a whole rather than these three functions.

PR #1943, opened 2026-08-22
perf(agents): move run step log out of a rewritten JSON blob (task-18601 part A)

rmusser01:
## What & why

**task-18601 part A**: the agent run step log lived in a single JSON blob column that `append_steps` re-read, re-parsed, extended and rewrote **on every append** — O(n) per step, O(n²) per run. Steps now live in a child table and an append is an INSERT.

## The defect, measured before and after

| ms per append | #1 | #500 | #2000 |
|---|---|---|---|
| before | 0.05 | 0.49 | **2.18** (44× growth) |
| after | 0.088 | 0.024 | **0.024** (flat) |

At the 25,000-step budget AC #1 names, that is roughly **5.4 minutes of write churn reduced to under a second**.

## Compatibility is the part that could destroy data, so it was probed not assumed

AC #4 requires existing blob-format runs to stay readable. Verified against a real database, independently of the implementation's own tests:

* a run whose steps live **only** in the old blob reads back exactly — 5/5 steps, identical content;
* a **mixed** run (legacy blob + new appends) returns all of them with blob steps first and new steps last, so a legacy run appended to after this change is neither reordered nor truncated.

## Design choices worth reviewing

* The four-line-deep read path matters: `_row_to_dict` is a `@staticmethod` with no DB access, so the child-table read could not go there — it had to move to the sites that fetch rows.
* `run_db_off_loop`-style concurrency was not needed here, but `seq` allocation had to stay correct under the agent worker thread; see the commit for the approach.
* `list_runs` / `undelivered_wake_runs` deliberately still take the full-hydration path. Disproportionate for part A, and documented in the code rather than left silent.

## Red-proof

Reverting `append_steps` to the read-modify-write shape fails the scaling test (`0.1080s` not `< 0.0286s`); restoring it (diff-verified identical) passes 9/9.

## Testing

`test_agent_runs_db.py` **53 passed**. `Tests/DB/` **1067 passed** with the same **8 pre-existing failures** this branch base already has (6 media-privacy, 1 numpy-missing, 1 `sql_validation` persona_visual). `Tests/Chat/ -k agent` was A/B-verified byte-identical against pristine `origin/dev`.

## Incidental find, fixed

`Tests/Chat/test_console_agent_swap.py::_all_runs` bypassed the DB API and read `agent_runs` with raw SQL, so it under-reported steps once they moved tables. A pre-existing test-only bug — worth naming as a pattern: **a test that reaches around its own API stops testing the API and starts pinning the storage layout**, and then breaks on any legitimate change to that layout.

## Still open

**AC #3** — the run-log viewer paging, so a long run isn't held in memory at once. Untouched here; it's part B of the split recorded in the task file, along with an optional backfill of historical blobs as part C.

---
## Summary by cubic
Moves agent run steps from the `agent_runs.steps` JSON blob to a child table `agent_run_steps`; `append_steps` is now an INSERT. This removes O(n²) write churn, keeps existing runs readable, and avoids parsing the step log when callers only need metadata.

- Satisfies task-18601 AC #1, #2, and #4; AC #3 (viewer paging) remains open for part B.
- Behavior change: old `append_steps` re-read/rewrote the blob per call; new `append_steps` inserts `(run_id, seq)` rows inside a single `BEGIN IMMEDIATE` transaction. Per-append cost is now flat (e.g., ~44× faster by append #2000; 25k-step runs drop from ~5.4 min to <1s).
- Read path: dual source. Hydration concatenates legacy blob steps first, then child rows in `seq` order. Mixed runs preserve order; fresh runs keep an empty blob and hydrate only from the child table.
- Metadata-only reads for AC #2: `get_run_metadata`, `get_run_metadata_fresh`, and `latest_primary_run_metadata` return no `steps` key and never parse the step log. Call sites updated: `ConsoleFleetWakeCoordinator._rows_for`, `AgentService.send_to_agent`, and latest-primary-run lookups in `Chat/console_agent_bridge.py`.
- Multi-row reads: batch-hydrate steps and chunk `IN (...)` clauses to 900 ids to stay under SQLite’s host-parameter ceiling across builds; avoids N+1 and “too many SQL variables”.
- Schema v13: new `agent_run_steps` with `ON DELETE CASCADE`. No backfill/migration required; legacy blobs remain as-is. Dropped a redundant `(run_id)` index (left-prefix of the PK).
- Intentional: `list_runs` and `undelivered_wake_runs` still fully hydrate steps; caller behavior unchanged.
- Tests: scaling, compatibility, and chunking coverage; fixed a test helper that bypassed the DB API and under-reported steps after the move.

<sup>Written for commit 428ca9f86ce05eabee25e9b0b164bb57fbda9db7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1943?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Chat/test_console_agent_swap.py
+++ b/Tests/Chat/test_console_agent_swap.py
@@ -382,9 +382,29 @@ async def test_citation_repair_agent_real_genuine_fallback_copy_does_not_bypass(
 
 
 def _all_runs(db):
-    """Read every persisted run record directly (AgentRunsDB has no list-all)."""
+    """Read every persisted run record directly (AgentRunsDB has no list-all).
+
+    task-18601 part A: ``append_steps`` now inserts into the
+    ``agent_run_steps`` child table instead of rewriting the
+    ``agent_runs.steps`` blob, so a raw ``SELECT * FROM agent_runs``
+    alone under-reports a run's steps for any run appended to after
+    that change. Reproduces ``AgentRunsDB``'s own dual-read (blob steps
+    first, then child rows in ``seq`` order) so ``row["steps"]`` stays a
+    JSON string of the FULL step list, exactly as callers here already
+    expect from ``json.loads(row["steps"])``.
+    """
     with db.connection() as conn:
-        return [dict(r) for r in conn.execute("SELECT * FROM agent_runs").fetchall()]
+        rows = [dict(r) for r in conn.execute("SELECT * FROM agent_runs").fetchall()]
+        for row in rows:
+            blob_steps = json.loads(row["steps"] or "[]")
+            child_rows = conn.execute(
+                "SELECT payload FROM agent_run_steps WHERE run_id = ? ORDER BY seq",
+                (row["id"],),
+            ).fetchall()
+            row["steps"] = json.dumps(
+                blob_steps + [json.loads(c["payload"]) for c in child_rows]
+            )
+        return rows
 
 
 def _join_fleet_threads(timeout=5.0):
--- a/Tests/DB/test_agent_run_steps_child_table.py
+++ b/Tests/DB/test_agent_run_steps_child_table.py
@@ -14,10 +14,14 @@ the 25,000-step budget AC#1 names for a single run.
 from __future__ import annotations
 
 import json
+import sqlite3
+import tempfile
 import time
+from pathlib import Path
 
 import pytest
 
+from tldw_chatbook.DB import AgentRuns_DB as agent_runs_db
 from tldw_chatbook.DB.AgentRuns_DB import AgentRunsDB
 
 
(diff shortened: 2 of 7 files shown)


rmusser01:
Rebased onto dev `869634f5a` and worked through all five Qodo findings. **Three fixed in `428ca9f86`, two declined with evidence.**

### Fixed

**#1 SQLite param-limit overflow — real, fixed.** `_batch_hydrate_steps` bound one host parameter per run id and no caller bounds the list (`ConsoleAgentController.subagent_runs` asks for every run in a conversation). Now chunked at 900.

One correction to the finding's premise, which changed the fix: the limit is **not** "commonly 999" here — measured on this machine, SQLite 3.50.4 reports `SQLITE_LIMIT_VARIABLE_NUMBER = 32766`, and 32767 placeholders is the first value that raises. But it is **build-dependent** — 999 on SQLite < 3.32, and this project's floor is Python 3.11, which can ship either. So the chunk is sized under the *old* ceiling: correct on every build, not just the newest.

That same fact is why the regression test lowers the connection's own limit via `conn.setlimit(...)` rather than trusting the local build. Unchunked, 1,200 ids pass fine on SQLite 3.50 — a naive ">999 ids" test would have been green against the broken code and proven nothing about the machines that actually break. Red-proofed:

```
E   sqlite3.OperationalError: too many SQL variables
FAILED ...::test_batch_hydrate_survives_more_run_ids_than_the_old_param_ceiling
```

**#5 Redundant index — real, fixed.** Confirmed by `EXPLAIN QUERY PLAN` with the index absent: `SEARCH s USING INDEX sqlite_autoindex_s_1 (run_id=?)`. The PK `(run_id, seq)` already serves the lookup, so `idx_agent_run_steps_run` bought nothing and cost an extra b-tree write per appended step — directly undercutting the point of this PR. Dropped. No shipped database has it; the table is new in this unmerged branch.

**#4 Docstring** — `Args`/`Returns`/`Raises` added to `get_run_metadata_fresh`.

### Declined, with evidence

**#2 "Schema change outside migrations."** This DB deliberately has no `migrations/` directory — `CREATE TABLE IF NOT EXISTS` inside `_initialize_schema` **is** its migration mechanism, documented in the class docstring: *"this DB has no migration framework — `_initialize_schema` runs CREATE TABLE IF NOT EXISTS plus guarded idempotent ALTERs, and appends one `INSERT OR IGNORE INTO schema_version` row per version."* The PR follows that contract exactly and bumps `_CURRENT_SCHEMA_VERSION` 12 → 13, which is enforced by `test_schema_version_constant_agrees_with_the_version_table`. Routing this one table through a different mechanism than its twelve predecessors would make the schema *harder* to audit, not easier.

**#3 "Reads bypass `transaction()`."** The repo rule is about write discipline; `transaction()` here is explicitly a **write** helper. Its docstring: *"Yield a write connection inside an immediate transaction… uses `BEGIN IMMEDIATE` (not the deferred default)."* Wrapping a pure `SELECT` in it would take a write lock for a read — serializing readers against writers on a hot path, in a PR whose entire purpose is removing agent-run DB contention. The module docstring states the split (`connection()` for reads, `transaction()` (BEGIN IMMEDIATE) for writes), 20 existing read sites already follow it, and an in-file audit comment records it: *"every `connection()` site is read-only except `_initialize_schema`."* The new methods match the file's convention.

### Gate

`Tests/DB` + `Tests/Agents`: **1713 passed**, 2 tests added.

Three pre-existing reds are unrelated to this branch and reproduce identically with dev's `AgentRuns_DB.py` swapped in (same 1 failed + 2 errors): they are the tiktoken encoding download hitting the test network guard, plus `test_local_tool_provider.py` needing `jsonschema`, which isn't installed in this worktree's venv.

PR #1949, opened 2026-08-22
fix(llm): config-driven default-timeout session, closing the hang-forever hazard (task-19830)

rmusser01:
## What & why

**task-19830**, filed during this burn-down after task-19560 named *two* unbounded `requests` calls and an audit found the problem is systemic.

`requests` has **no per-session default timeout**, so every call site has to remember one individually. Measured on dev:

```
requests.Session() construction sites: 56   (34 with no timeout on their calls)
timeout-less request calls:            40
files involved:                        13
```

A call with no timeout waits forever on a half-open connection: the calling coroutine hangs with no way to cancel, and several of these run on the event loop, so the whole TUI freezes with no error.

## The fix

`create_default_session()` in `Utils/egress.py` — the module that already owns this app's HTTP policy. A `requests.Session` subclass that fills in a `(connect, read)` timeout **only when the caller omitted one**; an explicit `timeout=` always wins, so no deliberate per-provider timeout changes behaviour.

Verified by live probe rather than by reading:

```
omitted timeout   -> injected: (10.0, 30.0)
explicit timeout=7 -> preserved: 7
```

Config-driven via `[web_security] request_connect_timeout_seconds` / `request_read_timeout_seconds`, falling back to 10s/30s.

**A tuple, not one number, on purpose:** the read half is re-armed per chunk on a streamed response, so a slow-but-progressing stream is only killed by a stall *between* chunks — never by total elapsed duration. A single total timeout would have broken long streaming completions.

## Scope: deliberately bounded, with the remainder made visible

All 6 `LLM_Calls/` files converted — 49 `Session()` sites plus 2 bare `requests.post` calls. Independently verified: **zero** bare `requests.Session()` remain in that package.

One deliberate exception: `summarize_with_anthropic` streams, and was given an explicit `timeout=` instead of being wrapped in a session — closing a session mid-stream risks tearing down the connection pool under an in-flight read.

The other 6 files are **not** converted. Converting 13 files across Embeddings, TTS, Character_Chat and Web_Scraping in one mechanical diff is a regression risk not worth taking here. They are named in the guard's `EXEMPT_FILES` set so the remaining work is visible rather than implied.

## The guard

`Tests/Architecture/test_default_timeout_session_guard.py` — an AST guard failing on any new bare `requests.Session()` or timeout-less request call, with `LLM_Calls/` required fully clean.

Red-proofed: injecting a bare `requests.Session()` into `qwencloud.py` fails both assertions and correctly names `LLM_Calls/qwencloud.py:1300`; removing it returns 11/11 green.

## The incidental find, which is the important part of this PR

**91 test-double mocks across 12 files silently stopped intercepting.**

They patched `requests.Session` / `requests.post` **by name**. The moment production called `create_default_session()` instead, those patches bound to a symbol nobody used any more — the tests kept passing while testing nothing. Only 8 of the 91 were in `Tests/LLM_Calls/`; the rest were scattered where nobody would look for them.

Found by alias-grepping the entire `Tests/` tree rather than the obvious directory, and all fixed. Confirmed afterwards that none had started reaching the real network: **0 network-guard violations** across the 1,009 passing tests.

A generalised lesson is recorded in `backlog/docs/lessons-testing-evidence.md` — this is the third instance of the same class in this burn-down alone (an AST scan that pinned call *syntax* and missed 3 methods; a test that pinned raw SQL and broke when a table moved; and now mocks that pinned a *symbol name* and went inert). **A check that pins the implementation's shape rather than its behaviour stops being a check the moment the shape changes — and does so silently.**

## Testing

`Tests/Architecture/` **173 passed** (5 pre-existing failures). `Tests/LLM_Calls/` **1009 passed** (7 pre-existing). A broad `-k "summariz or llm or api_call"` sweep: **2013 passed**, and every one of the 13 failures / 31 errors was individually triaged and A/B-confirmed pre-existing (diagnostic-privacy ledger, an anthropic-redirect fake-response gap, mlx_lm kwarg drift, missing optional deps, tiktoken-network teardown).

---
## Summary by cubic
Prevents LLM HTTP calls from hanging by defaulting timeouts. Previously, many `requests` calls had no timeout and could block forever; now calls through `create_default_session()` get a config-driven `(connect, read)` timeout unless the caller supplies `timeout=`. Streaming behavior remains unchanged; explicit timeouts still win. (task-19830)

- What to review
  - `tldw_chatbook/LLM_Calls/*` now uses `create_default_session()`; two bare posts were wrapped. `summarize_with_anthropic` keeps a direct call but adds `timeout=default_session_timeout()` to avoid session-close risks during streaming.
  - `tldw_chatbook/Utils/egress.py` adds `DefaultTimeoutSession`, `create_default_session()`, and `default_session_timeout()`; defaults come from `[web_security] request_connect_timeout_seconds`/`request_read_timeout_seconds` (fallback 10s/30s).
  - `Tests/Architecture/test_default_timeout_session_guard.py` blocks new bare `requests.Session()` and timeout-less calls; it also flags `**kwargs` expansions unless a timeout is forwarded explicitly, caches its scan, requires `LLM_Calls/` to be clean, and treats stale exemptions as failures.

- Migration
  - Use `create_default_session()` for new HTTP calls; do not construct `requests.Session()` directly.
  - Update tests to patch `<module>.create_default_session` instead of `<module>.requests.Session` or `requests.post`.
  - Optional: set `[web_security] request_connect_timeout_seconds` and `request_read_timeout_seconds` to tune defaults.

<sup>Written for commit f43bcc2ccb4584be4d1040908c48ed6f9cb0232f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1949?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Architecture/test_default_timeout_session_guard.py
+++ b/Tests/Architecture/test_default_timeout_session_guard.py
@@ -39,6 +39,7 @@ from __future__ import annotations
 
 import ast
 from dataclasses import dataclass
+from functools import lru_cache
 from pathlib import Path
 
 REPO_ROOT = Path(__file__).resolve().parents[2]
@@ -76,10 +77,11 @@ EXEMPT_FILES: frozenset[str] = frozenset(
         # as the LLM_Calls fix is a regression risk this task isn't taking.
         "Embeddings/Embeddings_Lib.py",
         "Local_Inference/ollama_model_mgmt.py",
-        # task-19560 (open, separate task) already covers the Kokoro
-        # download timeouts specifically -- exempted here rather than fixed
-        # incidentally by this guard's presence.
-        "TTS/backends/kokoro.py",
+        # (task-19560 shipped the Kokoro download timeouts while this branch
+        # was in review -- `TTS/backends/kokoro.py` used to be exempted here
+        # and is now clean, so the entry was removed. This guard's
+        # stale-exemption check is what caught that, which is the point of
+        # naming files individually instead of skipping a whole directory.)
         "Web_Scraping/Confluence/confluence_auth.py",
         "Web_Scraping/WebSearch_APIs.py",
         # Utils/egress.py's OWN internals (`guarded_fetch_requests`'s
@@ -187,8 +189,15 @@ def _scan_source(source: str, *, relative_path: str) -> list[Finding]:
                     session_vars.add(arg.arg)
 
     def has_timeout(call: ast.Call, leaf: str) -> bool:
+        # A `**kwargs` expansion deliberately does NOT count. It MIGHT carry
+        # a timeout, and the AST cannot tell -- but a guard that assumes the
+        # safe case whenever it can't see is a guard that passes on exactly
+        # the call sites hardest to eyeball. `requests.post(url, **opts)` is
+        # the shape most likely to hide a missing timeout, not least likely.
+        # A site that really does forward one through `**kwargs` states it
+        # explicitly (`timeout=opts.get("timeout", ...)`) and stays green.
         for kw in call.keywords:
-            if kw.arg == "timeout" or kw.arg is None:  # `timeout=` or `**kwargs`
+            if kw.arg == "timeout":
                 return True
         idx = _REQUEST_POSITIONAL_TIMEOUT_INDEX.get(leaf)
         return idx is not None and len(call.args) > idx
@@ -231,8 +240,20 @@ def _scan_source(source: str, *, relative_path: str) -> list[Finding]:
     return findings
 
 
+@lru_cache(maxsize=1)
 def _scan_package() -> dict[str, list[Finding]]:
-    """Every finding under ``tldw_chatbook/``, keyed by path relative to it."""
+    """Every finding under ``tldw_chatbook/``, keyed by path relative to it.
+
+    Cached: three tests in this module each need the whole picture, and
+    re-parsing every file in the package three times cost ~9s of a suite
+    run for three identical answers (14.2s -> 5.0s for this module). Safe
+    because nothing here mutates the tree between tests -- the package is
+    read-only input to a pure function of its own source.
+
+    Returns:
+        ``{relative_path: [Finding, ...]}`` for files with at least one
+        finding. Callers must treat the value as read-only; it is shared.
+    """
     by_file: dict[str, list[Finding]] = {}
     for path in sorted(PACKAGE_ROOT.rglob("*.py")):
         relative = path.relative_to(PACKAGE_ROOT).as_posix()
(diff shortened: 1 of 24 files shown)


rmusser01:
Rebased onto dev `fc5c9b57e` and addressed all four Qodo findings in `f43bcc2cc`.

### Fixed

**#1 Guard misses `**kwargs` timeouts — real, fixed.** `has_timeout()` counted a `**kwargs` expansion as "timeout supplied", so `requests.post(url, **opts)` passed unexamined. The AST genuinely cannot see whether `opts` carries one — but resolving that ambiguity as *safe* blinds the guard to exactly the call shape where a missing timeout is hardest to spot by eye. It now flags the expansion; a site that really does forward one says so (`timeout=opts.get("timeout", ...)`) and stays green.

Worth recording: tightening this flagged **nothing new** in the package, so it costs zero conversions — it closes a future hole, not a present one. Red-proofed both ways — reverting the rule fails the new self-test, and a second self-test pins the escape hatch so the rule can't silently become a wall.

**#4 Guard scans the package 3× — real, fixed.** Three tests each re-parsed every file under `tldw_chatbook/` for three identical answers. `_scan_package` is now `lru_cache`d: **14.2s → 5.0s** for this module.

**#2 / #3 Type hints and docstrings** — added to `DefaultTimeoutSession.__init__` / `.request` and `create_default_session`.

### Caught by this branch's own guard during the rebase

`TTS/backends/kokoro.py` was listed in `EXEMPT_FILES` as remaining work, but **task-19560 shipped its download timeouts while this branch was in review**, so the file is now clean and the exemption was stale. `test_exempt_files_are_current` failed and named it. Entry removed, task census updated. That the guard fails on a *stale* exemption — not only on a new violation — is the whole argument for naming files individually instead of skipping a directory; a directory-level skip would have quietly kept fixed work on the books.

### Rebase conflicts (3), resolved by keeping both sides

task-19560 (merged meanwhile) had added per-provider `api_timeout` config keys at two sites this branch converts to the shared session. Both are kept: the session is the safety net for calls that forget a timeout, and `DefaultTimeoutSession.request` never overrides an explicit `timeout=` — so the user-facing `local_llm.api_timeout` / `anthropic_api.api_timeout` knobs still govern those calls. That is the designed contract, not a workaround.

### Gate

Guard + egress: **119 passed**. `Tests/LLM_Calls/` + guard + egress: **1131 passed, 7 failed**.

All 7 failures are **pre-existing on dev** and reproduce identically with every one of this branch's 24 files reverted to dev's version — two independent causes, both toolchain drift rather than product bugs:

* `test_summarization_diagnostic_privacy.py` (5) — the frozen ledger stores `ast.dump()` output, whose rendering of empty `keywords=[]` changed in a newer Python (3.14.2 here).
* `test_anthropic_redirect_credential_leak.py` (2) — requests 2.34.2's `rebuild_auth` reads `response.request.url` unconditionally, and the test double leaves `.request` as `None`, so the refusal path raises `AttributeError` before it can be observed.

Neither is touched by this branch. Happy to file both as separate tasks.

PR #1954, opened 2026-08-22
fix(backlog): repair 32 task files that no backlog read could parse

rmusser01:
## What

32 task files on `dev` cannot be parsed, so every backlog read skips them behind
a single loguru warning. They are absent from `task list`, the board, search, and
the MCP read tools:

```
visible tasks before  2358
visible tasks after   2390
```

## Why they were invisible rather than loud

`_load_tasks_from_dir` logs `Skipping unreadable task file …` per file and
continues — correct for a read command, useless as a signal. The warnings
interleave with output, so the problem reads as two or three files until the
whole set is swept.

## The three shapes

| | count | example |
|---|---|---|
| unquoted title containing `": "` | 28 | `title: Console transcript: skip move_child for rows already in position` |
| single-quoted title with an undoubled apostrophe | 3 | `title: 'Impersonate drafts the user's next reply'` |
| owned section with no `:END` marker | 3 | one `DESCRIPTION`, two `NOTES` |

None are producible by the backlog writers, which emit frontmatter through
`yaml.safe_dump`; they come from hand edits, agents, and older Backlog.md
versions. Two of the unterminated sections only became visible once the title
parsed, since the YAML error aborted the parse first.

## How

`backlog-py doctor --fix` (shipped in backlog-md-py #171), which re-parses a
file before writing it and leaves anything it cannot repair untouched.

## Verification

- **Every one of the 32 titles carries byte-identical text**; only the quoting
  changed. Checked programmatically against `HEAD` by stripping `title:` and any
  surrounding quotes from both sides — 32 files, 0 differing strings.
- Each added `:END` marker sits at end of file, closing a section that ran to EOF.
- `backlog-py doctor` → `0 unreadable file(s)`.
- The duplicate-id guard's own logic still passes on this branch (0 filename
  dupes, 0 frontmatter dupes).
- Diff is 34 insertions / 31 deletions across 32 files: 31 title lines and 3
  markers. Nothing else.

## Related

`doctor` also reports one duplicate id on `dev` that the CI guard cannot see —
`TASK-2157` is held by a live task in `backlog/tasks/` and an archived one in
`backlog/archive/tasks/`, while the guard only scans `backlog/tasks/`. That is a
separate PR.

---
## Summary by cubic
Repairs 32 unreadable backlog task files so reads include them again. Previously reads skipped these files with a warning; now YAML titles are correctly quoted and the lone unterminated DESCRIPTION is closed, increasing visible tasks from 2358 to 2390.

- Fixed 31 frontmatter titles by adding/correcting quotes (and doubling apostrophes where needed); title text is unchanged.
- Closed one DESCRIPTION section by adding `<!-- SECTION:DESCRIPTION:END -->` at EOF in task-15390; no other content changes.
- Produced with `backlog-py doctor --fix`; `backlog-py doctor` reports 0 unreadable files and the duplicate-id guard passes. No migration required.

<sup>Written for commit 2f5c1987d0d2b6845e50cdf8b4dcfc4cba47ca48. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1954?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backlog/tasks/task-15390 - Search-RAG-gate16-evidence-heading-test-fails-on-clean-dev.md
+++ b/backlog/tasks/task-15390 - Search-RAG-gate16-evidence-heading-test-fails-on-clean-dev.md
@@ -27,6 +27,7 @@ Nobody in this arc root-caused it (out of scope both times it surfaced); it is N
 long-standing known-ambient list (the old ~45 shell-geometry failures were fixed on dev
 separately), so it is presumably a recent regression or a test drifted from a deliberate
 Search/RAG copy change.
+<!-- SECTION:DESCRIPTION:END -->
 
 ## Acceptance Criteria
 <!-- AC:BEGIN -->
--- a/backlog/tasks/task-15453 - Console-transcript---skip-move_child-for-rows-already-in-position.md
+++ b/backlog/tasks/task-15453 - Console-transcript---skip-move_child-for-rows-already-in-position.md
@@ -1,6 +1,6 @@
 ---
 id: TASK-15453
-title: Console transcript: skip move_child for rows already in position
+title: 'Console transcript: skip move_child for rows already in position'
 status: Done
 assignee:
   - '@claude'
--- a/backlog/tasks/task-15454 - Console-rail-search---move-DB-work-inside-its-debounce-and-re-guard-the-workspace-tray.md
+++ b/backlog/tasks/task-15454 - Console-rail-search---move-DB-work-inside-its-debounce-and-re-guard-the-workspace-tray.md
@@ -1,6 +1,6 @@
 ---
 id: TASK-15454
-title: Console rail search: move DB work inside its debounce and re-guard the workspace tray
+title: 'Console rail search: move DB work inside its debounce and re-guard the workspace tray'
 status: Done
 assignee:
   - '@claude'
--- a/backlog/tasks/task-15455 - Console-transcript---windowed-mount-for-long-conversation-load.md
+++ b/backlog/tasks/task-15455 - Console-transcript---windowed-mount-for-long-conversation-load.md
@@ -1,6 +1,6 @@
 ---
 id: TASK-15455
-title: Console transcript: windowed mount for long-conversation load
+title: 'Console transcript: windowed mount for long-conversation load'
 status: Done
 assignee:
   - codex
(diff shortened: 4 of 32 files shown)


rmusser01:
Rebased on latest `dev` (`22d4f1228`) and **fixed the Qodo finding at its source**.

The bot was right: `task-15460` and `task-15462` each came out with one `SECTION:NOTES:BEGIN` and **two** `:END` markers. The cause was a bug in `backlog-py doctor --fix`, not in these files — it inserted an END at the first heading after a BEGIN, on the assumption that a heading ends the section. Both files carry a NOTES block with its own `###` sub-headings and were already closed correctly at the bottom, so the insert was pure corruption.

Fixed upstream in [backlog-md-py#173](https://github.com/rmusser01/backlog-md-py/pull/173) (merged): the repair now looks ahead for a matching END and leaves any section that closes anywhere below its BEGIN untouched. Regression test uses exactly this shape.

This branch was regenerated from `dev` with the fixed tool. The diff is now **32 files, 32 insertions, 31 deletions**:

- 31 title requotes
- **1** marker insertion — `task-15390`'s `DESCRIPTION:END`, which is genuinely unterminated on dev (no END anywhere in the file)

Re-verified: all 32 titles byte-identical in text, `backlog-py doctor` → 0 unreadable, `scripts/check_backlog_task_ids.py` → no duplicates across 2405 files.

rmusser01:
Merging on local verification — GitHub runners have not picked up this PR's jobs (queued ~10 min; dev's own Derived Artifacts runs are being cancelled on the same treadmill).

Ran the required check's four steps locally on this branch:

| checker | result |
|---|---|
| `check_bundle_sync.py` | exit 0 |
| `check_profile_owned_path_inventory.py` | exit 0 |
| `check_persistent_diagnostic_inventory.py` | exit 1 — **pre-existing on dev** |
| `check_backlog_task_ids.py` | exit 0 |

The inventory failure reproduces identically on the untouched dev base and names Python sources (`Notes/sync_engine.py`, `Chat/console_provider_gateway.py`, `Notes/Notes_Library.py`, …) that this branch does not touch — its diff is 32 backlog markdown files, titles only plus one section marker.

PR #1958, opened 2026-08-22
Change review: report submit bugs, guard the detection landing, align the in-progress refusal docs (TASK-19703)

rmusser01:
Three items from TASK-16801's whole-branch review, batched. Each is a place a failure was reported less honestly than the rest of the feature manages.

## 1. The detection landing wasn't guarded (AC #1)

`_dispatch_git_detection`'s `call_from_thread` was the one landing not routed through `_land_on_ui` — the shared helper that tolerates **only** Textual's teardown `RuntimeError`. A genuine bug in `_land_git_detection` therefore died quietly on the worker thread instead of surfacing as a `WorkerFailed` traceback. Now routed.

## 2. Both modals' confirm button could go silently dead (AC #2)

Each `_submit` wraps its body in a broad `except` so it can never raise into a Textual handler — correct — but then returned **silently**. Pressing Commit or Push did *nothing at all*: no action, no error, no dismissal. Indistinguishable from a dead button, and the same silent-masking class this workstream has caught repeatedly.

The broad catch stays; the failure is now visible. The commit modal reuses its inline `#change-git-commit-error` Static; the push modal has no such widget and reports through `notify` rather than inventing one (plus CSS) for a path only a bug reaches. Both proven red first, and mutation-checked — removing the two reports fails both new tests.

## 3. Code, spec and Guide disagreed on the in-progress refusal (AC #3)

The spec listed the merge/rebase/cherry-pick check as a **modal-open gate**, and the User Guide told users it fires *"before the dialog even opens"*. It actually ships as the engine's `in-progress-check` step — at confirm time. The Guide was the user-facing inaccuracy.

**Resolved by amending the documents, not the code.** The engine is the right home: the repository can enter a merge between modal-open and submit, so a pre-modal check could only ever be advisory — and an advisory check that *passes* and is then refused at submit is more confusing than one honest refusal, while duplicating it in both places invites the copies to drift.

The Guide sentence had bundled this with the **active-run** refusal, which genuinely *is* pre-modal, so the two are now stated separately with the reason for the difference. Stamped honestly: not driven live, corrected by reading the shipped code against the page's own claim.

## Verification

- **272 passed** across the change-review suites; ruff clean.
- Both new submit tests red-first and mutation-proven.
- The 2 failures in `test_console_modal_dismissal.py` are **pre-existing on dev** — confirmed against a throwaway worktree pinned to the current dev tip (2 failed, 115 passed there).

Closes TASK-19703. Follows TASK-16801 (#1914).

---
## Summary by cubic
Surfaces real landing failures as WorkerFailed, reports commit/push submit errors to users, and aligns docs with engine behavior. Also distinguishes teardown from landing bugs using `App.is_running` to prevent silent failures.

- Detection landing now uses `_land_on_ui`; the guard re-raises when the app is running and swallows genuine teardown, so real bugs surface.
- Consolidated landing policy: the current-mode worker’s local `_land` delegates to `_land_on_ui`, keeping per-root isolation and a single rule.
- Commit modal `_submit` shows its inline error; push modal `_submit` reports via `notify`. Both still catch broadly and keep the modal open.
- Spec and User Guide updated: mid-merge/rebase/cherry-pick refusal occurs at confirm (`in-progress-check`). Tests cover the new reporting and landing behavior. No migration required.

<sup>Written for commit d703080fc57c103318424f6d80929f89cbae9cc0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1958?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/User_Guide/console/agent-runs-and-tools.md
+++ b/Docs/User_Guide/console/agent-runs-and-tools.md
@@ -416,9 +416,13 @@ optional "create branch first" field checks out a new branch before
 staging. The dialog also shows non-blocking **warnings** — they never stop
 the commit — for a detached HEAD ("this commit will not be on any
 branch") and for committing straight to `main` or `master`. Committing is
-**refused outright** (before the dialog even opens) while an agent run is
-active on that workspace, and separately while the repository is mid
-merge, rebase, or cherry-pick — finish or abort that operation first.
+**refused outright before the dialog even opens** while an agent run is
+active on that workspace. A repository that is mid merge, rebase, or
+cherry-pick is also refused, but at the moment you confirm rather than
+before the dialog opens — finish or abort that operation first. (The check
+runs then because the repository can enter a merge while the dialog is
+open, so checking earlier could only promise something that later stops
+being true.)
 
 **Push (`p` / the Push… button).** Always confirms in a dialog first,
 naming the repository, the branch, and where it's going. A branch with no
@@ -1659,3 +1663,5 @@ conversation with no turns yet may show no `current` entry; this is a
 pre-existing characteristic of the change-tracking roots this arc reused
 verbatim, not something Task 9 changed, and it is not re-verified live
 here.)*
+
+*Git-actions placement corrected @ TASK-19703 — 2026-08-22: the mid-merge/rebase/cherry-pick refusal was documented here as happening "before the dialog even opens", which is true of the active-run refusal but not of this one — it fires when you confirm. Not driven live; corrected by reading the shipped code (`commit_selected`'s `in-progress-check` step) against this page's claim, and the design spec was amended to match rather than the code changed (a pre-modal check could only be advisory, since the repository can enter a merge while the dialog is open).*
--- a/Docs/superpowers/specs/2026-08-20-console-review-git-modes-design.md
+++ b/Docs/superpowers/specs/2026-08-20-console-review-git-modes-design.md
@@ -280,15 +280,25 @@ Button `Commit…` + binding, visible/enabled only in current mode with
    - warnings (never blocks): detached HEAD ("commit will not be on
      any branch"), branch in `{main, master}` ("committing directly
      to <branch>");
-   - an operation in progress (`rev-parse --verify -q` on
-     `MERGE_HEAD`, `REBASE_HEAD`, `CHERRY_PICK_HEAD` — check=False)
-     → commit REFUSED with reason ("finish or abort the
-     merge/rebase/cherry-pick first") — git refuses a partial
-     (pathspec) commit during a merge OR a cherry-pick, and the raw
-     git error is worse copy.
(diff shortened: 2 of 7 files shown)


rmusser01:
## Qodo review addressed — commit `d703080fc`

One fixed (and it improved the change), two declined with reasoning. **158 passed** across the change-review suites; ruff clean.

### 2. RuntimeError bug can hide — Bug ✅ correct, and the sharpest catch here

You're right, and this one genuinely made the PR better. `_land_on_ui` tolerates `RuntimeError` because that's Textual's teardown signal — but `call_from_thread` also re-raises whatever the *callback* raised, so a real bug that happens to be a `RuntimeError` was read as shutdown and vanished. Routing detection through that helper (AC #1) therefore **narrowed** visibility for exactly that case, which is the opposite of this PR's stated goal.

`App.is_running` is the real discriminator — it's precisely what `call_from_thread` consults before raising "App is not running" — so the guard now re-raises when the app is still alive. Red-proven, and mutation-proven after.

Chasing it surfaced two things worth flagging:

- **A second copy of the same rule.** The current-mode worker had its own local `_land` helper carrying a duplicate of the teardown policy, which the sharpening would have left behind — so a landing bug still vanished on the very path the fix targeted. It now delegates to the single implementation. Two copies of one policy is how this drifted in the first place.
- **The existing teardown test was a bad proxy.** It simulated shutdown by raising `RuntimeError("App is not running")` *while the app was still running* — indistinguishable from a genuine bug, which is the hole you found. It now drives `_land_on_ui` directly against a stub app that is genuinely not running, so both halves test the real condition.

### 3. Raw exception shown to user — declined, with reasoning
This is a single-user local TUI, not a multi-tenant surface: the operator *is* the user, and this very screen already prints absolute repository paths in its banner and git's raw stderr in push-failure notifications — both deliberately, for honesty. Redacting only this one message would be inconsistent, and a generic "could not submit" whose detail lives in a log file the user is unlikely to find is *less* honest, not more — the failure mode this PR exists to fix. `str(exc)` is already the message rather than a repr/traceback. Happy to revisit if you see a path where this string crosses a trust boundary I've missed.

### 1. Test docstring style — declined, as on #1941
The docstrings are present and explain the failure mode each test pins. Declining the `Args:`-section part: **zero** tests in these files document their pytest fixtures that way, so complying would break local convention and add noise — CLAUDE.md's Google-style requirement targets public APIs, not fixture parameters. If the standard is meant to cover tests, that should be a repo-wide change rather than three functions.

PR #1959, opened 2026-08-22
Change review: name the effective push destination in the confirm dialog (TASK-19701)

rmusser01:
## The gap

`remote.<name>.pushurl` and `url.<other>.pushInsteadOf` both send a push to a **different host** than the repository fetches from. The confirm dialog named the remote's *alias* and never its destination — so a terminal running `git remote -v` told the user more than the dialog whose entire job is to state what a button will do before they press it.

## Decision (AC #1): surface, don't refuse

Both settings are legitimate, widely used git configuration — fetching over https while pushing over ssh is a standard corporate setup. Refusing would break ordinary workflows to guard against nothing this app is entitled to override. What was wrong was the silence, not the redirect.

## No new git call was needed

Verified against real git that both settings are already resolved into `git remote -v`'s `(push)` line — and detection parses exactly that line, so the effective URL was in hand all along:

```
baseline          origin  https://fetch.example/repo.git          (push)
+ pushurl         origin  ssh://git@PUSH-TARGET.example/repo.git  (push)
+ pushInsteadOf   origin  ssh://git@INSTEADOF.example/repo.git    (push)
```

The dialog now reads `pushes to origin (ssh://…)`.

**Shown for every push, redirected or not.** A warning that fires only when something is odd trains people to fear the odd case; a line that always states the destination just makes the dialog honest — and it means the disclosure can't be missed by someone who has never seen the plain version to compare against.

## Verification

- **137 passed** across the change-review suites; ruff clean.
- Three new tests drive **real repositories** configured with `pushurl` and with `pushInsteadOf`, plus a control asserting the plain destination is named when no redirect exists.
- Mutation-proven: dropping the URL from the label fails all three.
- User Guide updated and stamped honestly (not driven live; verified via real-repo tests).

Closes TASK-19701 — the last follow-up from TASK-16801 (#1914).

---
## Summary by cubic
Names the effective push destination in the push confirmation dialog so users see where a push will land. Previously the dialog showed only the remote alias; now it displays "pushes to <remote> (<url[, url...]>)" for every push, including redirects via `remote.<name>.pushurl` or `url.<other>.pushInsteadOf`, and lists all destinations for multi-`pushurl` remotes (TASK-19701).

- Behavior change: always surface the actual destination(s); do not block; show URL(s) for all pushes, not only redirected ones.
- Parsing fixes: stop truncating URLs containing spaces by stripping the trailing " (push)"; collect every `(push)` URL per remote.
- Data shape: keep `remotes` de-duplicated (one entry per remote); add a defaulted `remote_push_urls` field carrying all push destinations.
- Implementation: no new Git calls; reuse `git remote -v` (push) output already parsed; `_destination_label` formats "<name> (<url[, url...]>)" and falls back to the name when no URL is available.
- Tests/Docs: UI tests cover `pushurl`, `pushInsteadOf`, no-redirect, and multi-`pushurl`; workspace tests cover space-in-URL and multi-destination parsing; User Guide updated to document the disclosure.

<sup>Written for commit b1ef8289ec98bd4b0e69d8ecba3a771db666c738. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1959?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/User_Guide/console/agent-runs-and-tools.md
+++ b/Docs/User_Guide/console/agent-runs-and-tools.md
@@ -425,7 +425,14 @@ open, so checking earlier could only promise something that later stops
 being true.)
 
 **Push (`p` / the Push… button).** Always confirms in a dialog first,
-naming the repository, the branch, and where it's going. A branch with no
+naming the repository, the branch, and where it's going — including the
+remote's actual **URL**, not just its name. That matters when the
+repository redirects pushes to a different host than it fetches from
+(`remote.<name>.pushurl`, or `url.<other>.pushInsteadOf`), which is a
+perfectly normal setup — fetching over https and pushing over ssh, say.
+Those redirects are honoured rather than refused; the dialog simply tells
+you where the push will actually land, so the confirmation says as much as
+`git remote -v` would. A branch with no
 upstream yet gets one set on this push; with an upstream already
 configured, push targets exactly that upstream's remote and ref — never
 "whatever the repository's push configuration would have done" (see the
@@ -1665,3 +1672,5 @@ verbatim, not something Task 9 changed, and it is not re-verified live
 here.)*
 
 *Git-actions placement corrected @ TASK-19703 — 2026-08-22: the mid-merge/rebase/cherry-pick refusal was documented here as happening "before the dialog even opens", which is true of the active-run refusal but not of this one — it fires when you confirm. Not driven live; corrected by reading the shipped code (`commit_selected`'s `in-progress-check` step) against this page's claim, and the design spec was amended to match rather than the code changed (a pre-modal check could only be advisory, since the repository can enter a merge while the dialog is open).*
+
+*Push-destination disclosure added @ TASK-19701 — 2026-08-22: the confirm dialog now names the remote's effective push URL. Not driven live; verified against real repositories in tests configured with `remote.<name>.pushurl` and with `url.<other>.pushInsteadOf`, plus a control with no redirect.*
--- a/Tests/UI/test_change_review_push_ui.py
+++ b/Tests/UI/test_change_review_push_ui.py
@@ -1588,3 +1588,102 @@ async def test_a_bug_inside_push_submit_reports_itself(monkeypatch, tmp_path):
     assert any("Could not submit" in note for note in notes), (
         f"a bug in push submit must tell the user; got {notes!r}"
     )
+
+
+# ---------------------------------------------------------------------------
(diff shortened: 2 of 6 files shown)


rmusser01:
## Qodo review addressed — commit `a61c15817`

**542 passed** across `Tests/Workspaces/` + the change-review suites; ruff clean.

### 2. Push URL disclosure incomplete — Bug ✅ correct on both counts, and it mattered

Both sub-claims reproduced against live git before fixing, and both would have made this dialog state something **false** — which is worse than the silence it replaced, because the user would believe it:

```
# multiple pushurl -> git emits one (push) line per destination
origin  ssh://git@FIRST.example/r.git  (push)
origin  ssh://git@SECOND.example/r.git (push)
    parsed: 'origin' -> 'ssh://git@FIRST.example/r.git'      # second one dropped

# local path containing a space
spaced  /…/scratchpad/with space.git   (push)
    parsed: 'spaced' -> '/…/scratchpad/with'                  # a path that does not exist
```

Fixed by recovering the URL from the trailing `" (push)"` **suffix** rather than splitting on the first space, and by collecting every `(push)` line per remote.

One design note on the shape: the de-duplicated `remotes` tuple deliberately keeps **one entry per remote**, because `_resolve_push_remote` and the dialog's "must I ask which remote?" check both key off its *length* — letting a multi-`pushurl` remote appear twice would make a single remote look like a choice between two. The full destination set lives in a new defaulted `remote_push_urls` field, and the dialog joins them.

New tests cover both at the engine level and the multi-destination case end-to-end through the dialog.

### 1. Magic `run_test` size tuple — declined, with counts

The finding's rationale is consistency, and extracting a constant here would reduce it. `size=(160, 48)` appears **28 times in this one file**, 26 in each of the two sibling change-review files, and across **66 test files** repo-wide — against exactly **one** named-constant precedent in the entire suite (`_FOCUS_TERMINAL_SIZE` in `test_focus_mode.py`).

Converting only the four new call sites would leave 28 literals sitting beside 1 constant in the same file, which is strictly worse to read than 32 identical literals. Happy to do the file-wide (or suite-wide) sweep as its own mechanical change if that's wanted — it just doesn't belong inside a PR about push destinations.

PR #1960, opened 2026-08-22
fix(tests): repair six pre-existing dev test reds (four of them Python-version drift)

rmusser01:
Eleven tests were failing on dev before this branch, from three unrelated causes. **None was a product defect** — but two were tests that had quietly stopped being able to observe the thing they guard, which is worse than a plain red.

## 1. `ast.dump()` is not a pure function of the source (5 tests)

Python 3.13 added `show_empty` to `ast.dump()` and defaulted it to `False`, so empty fields stopped rendering: `Call(func=..., args=[...], keywords=[])` on ≤3.12 became `Call(func=..., args=[...])` on ≥3.13. The summarization diagnostic ledger **freezes** dumped shapes, so every reviewed row containing a no-keyword call stopped matching — on a 3.14 interpreter, while still passing on the repo's own 3.12 venv that generated the ledger.

Measured on four real interpreters, digesting all 229 diagnostic shapes in `Local_Summarization_Lib.py`:

| Python | digest |
|---|---|
| 3.11 | `92a85a3a989ffd13` ← what the committed ledger holds |
| 3.12 | `92a85a3a989ffd13` |
| 3.13 | `cd27a1b6c1fdf001` ← diverges exactly where `show_empty` landed |
| 3.14 | `cd27a1b6c1fdf001` |

New `Tests/ast_shape.stable_dump()` pins the pre-3.13 rendering; all four then produce `92a85a3a989ffd13`.

Pinning the **old** rendering rather than regenerating against the new one is deliberate: those ledger rows are a privacy review, not a checksum. Regenerating would invalidate all of them at once *and* break the 3.11 floor (`requires-python = ">=3.11"`) instead of fixing anything.

## 2. A response double `requests` can no longer drive (2 tests)

`test_anthropic_redirect_credential_leak.py` installs a fake `HTTPAdapter.send` but never set `response.request` — a real adapter sets it in `build_response`. requests 2.34.2's `Session.send` peeks ahead through `resolve_redirects` → `rebuild_auth`, which dereferences `response.request.url`. So the double raised `AttributeError` **inside requests**, before the production refusal path ran, and the production `except Exception` re-reported it as a generic failure. The tests looked like they exercised the redirect refusal and never reached it.

Fixed by giving the double its `.request` — then **red-proofed**, because a credential-leak test that merely stops failing is worth nothing:

* `allow_redirects=False` → `True`: fails, printing the sentinel `x-api-key` arriving at `evil.example`.
* refusal-path `response.close()` removed: fails on the close count.

## 3. Reviewed-inventory drift (4 tests)

The persistent-diagnostic inventory had drifted after a Notes sync refactor and several merges (`owner_files` 521→519, sinks 7→8, 11 changed files). Regenerated **only after reading every row** via the artifact's own `--statements --since` review, not a blind `--write`. What that review found:

* The new sink `Notes/notes_sync_coordinator.py` **writes nothing** — zero write calls in the file; the handle is used only for `portalocker`, and the lock filename is a SHA-256 of `(st_dev, st_ino)`, not a path. Clean.
* `library_screen.py` is a net **privacy improvement**: three warnings that interpolated import paths were replaced with `type(exc).__name__`.
* **One real regression, mine, from #1935**: kokoro's partial-file cleanup logged `{partial}`, and two of that function's four call sites pass the user's *configured* model/voice directory — the same class of leak as the two `Downloaded model to {self.model_path}` lines removed in that very change. Now logs the label and error type. Fixed here.
* `change_review_screen.py` interpolates the raw `[console] workspace_root` into a debug line. Not mine and not in scope — **filed as TASK-19936** rather than absorbed.

Also dropped a curated `REVIEWED_METADATA_ONLY_DIAGNOSTICS` entry naming a diagnostic dev's Notes-import refactor deleted, and re-pinned the manifest-boundary digest (both halves agree — the committed and freshly-generated inventories project to the same SHA).

## Verification

`257` privacy + `65` inventory + `2` redirect + `4` new `ast_shape` + `5` kokoro — all pass. The `ast_shape` guard is version-travelling by construction: it asserts the rendering that fails on ≥3.13 without the wrapper and passes on 3.11/3.12 either way.

Out of scope: the `chat_screen.py` size ratchet, which is a separate pre-existing red the owner asked to leave alone.

---

## Update: three more reds found while verifying

Sweeping wider turned up three additional pre-existing failures, all now fixed in this branch.

### 4. A `None` line number took 98 TTS tests out at collection

`Tests/TTS/test_profile_store_lock.py` builds a module-level constant from `dis.findlinestarts()`, comparing each line against a source boundary. That generator's line is **Optional** — an instruction may carry no source position, and CPython emits more of those as it evolves. Measured on the very function the file inspects (`ProfileStoreLease.acquire`): **3.11 yields 0 positionless entries, 3.14 yields 12**.

So on a newer interpreter it raised `TypeError: '<=' not supported between instances of 'int' and 'NoneType'` while building a module-level constant — not a test failure but a **collection error**, so the entire file (98 tests) silently stopped running. Skipping positionless instructions is also semantically right: an instruction with no source line cannot be inside a source-line boundary.

### 5. A concurrency guard that had stopped measuring its subject

`test_invoke_by_name_takes_exactly_one_catalog_snapshot` was failing `assert 2 == 1`. **The production code is correct.** Tracing both calls:

```
call 1  tool_catalog.py:1118  list_catalog      -> _ensure_catalog_cache()
call 2  tool_catalog.py:1209  invoke_by_name    -> _owner_record_for_name -> _ensure_catalog_cache()
```

Call 1 is the *test's own* `name = registry.list_catalog()[0].name`, taken after the counter was installed. `list_catalog` was since refactored onto the shared `_ensure_catalog_cache()` helper — good for consistent locking — and from that moment the assertion read 2 regardless of how `invoke_by_name` behaved. Fixed by resolving the name before installing the counter; red-proofed by reintroducing a second snapshot, which fails with the original `assert 2 == 1`.

### 6. `jsonschema` was never declared

`Tests/Agents/test_local_tool_provider.py` imports `Draft202012Validator` at module level, but `jsonschema` reached environments only **transitively** — via `chromadb`, `docling-core`, `mcp-unified`, all *optional* extras. A plain `pip install -e ".[dev]"` could not collect that module at all. Now declared in the `dev` extra.

## Cross-version verification

Every repair was checked on **both** interpreters, which matters because four of the six causes were version drift:

* Python 3.14: **434 passed, 0 failed** across all touched suites.
* Python 3.12 (the repo's own venv): **322 passed** for the ledger suites, plus 104 for the rest.

## Not fixed, and why

* **`chat_screen.py` size ratchet** — excluded by the owner.
* **Five TTS collection errors** (`numpy`/`torch`/`soundfile`) — these are a thin worktree venv, not a repo red: all pass under the full main venv. Unlike `jsonschema`, these deps *are* declared under real extras.
* **A full-suite run was not completed.** The suite is large enough that a whole-repo `pytest Tests/` was killed at 9% after 26 minutes, so "no other reds exist anywhere" is not a claim this PR makes — the sweep covered `Tests/LLM_Calls`, `Tests/Architecture`, `Tests/Agents`, `Tests/TTS`, `Tests/DB`, `Tests/Subscriptions`, and `Tests/Utils`.

### DIFF
--- a/Docs/security/production-diagnostic-inventory.json
+++ b/Docs/security/production-diagnostic-inventory.json
@@ -2560,6 +2560,13 @@
       "reason": "remaining Chatbook production diagnostic owner"
     },
     {
+      "call_count": 1,
+      "diagnostic_digest": "86a8bc66a943d2aef621",
+      "owner": "TASK-494",
+      "path": "tldw_chatbook/UI/Screens/model_remote_view.py",
+      "reason": "remaining Chatbook production diagnostic owner"
+    },
+    {
       "call_count": 186,
       "diagnostic_digest": "abf6fdc6c9eef6dd4f9c",
       "owner": "TASK-494",
@@ -3225,6 +3232,13 @@
       "reason": "remaining Chatbook production diagnostic owner"
     },
     {
+      "call_count": 1,
+      "diagnostic_digest": "01bf025ed01c65569ea6",
+      "owner": "TASK-494",
+      "path": "tldw_chatbook/Widgets/Console/console_run_inspector.py",
+      "reason": "remaining Chatbook production diagnostic owner"
+    },
+    {
       "call_count": 2,
       "diagnostic_digest": "fd3c40d9f3d798e91fc4",
       "owner": "TASK-494",
@@ -3840,9 +3854,9 @@
   "schema_version": 2,
   "scope": "tldw_chatbook/**/*.py",
   "summary": {
-    "owner_files": 519,
+    "owner_files": 521,
     "persistent_sink_files": 8,
     "task_492_calls": 1222,
-    "task_494_calls": 7187
+    "task_494_calls": 7189
   }
 }
--- a/Docs/security/production-diagnostic-inventory.json
+++ b/Docs/security/production-diagnostic-inventory.json
@@ -1980,7 +1980,7 @@
     },
     {
       "call_count": 70,
-      "diagnostic_digest": "04c4f3f8a3dc9eeadbdd",
+      "diagnostic_digest": "c8c97abdd441f878754e",
       "owner": "TASK-494",
       "path": "tldw_chatbook/TTS/backends/kokoro.py",
       "reason": "remaining Chatbook production diagnostic owner"
@@ -2557,13 +2557,6 @@
       "diagnostic_digest": "eb735f3bcd9aea1fcade",
       "owner": "TASK-494",
       "path": "tldw_chatbook/UI/Screens/model_installed_view.py",
-      "reason": "remaining Chatbook production diagnostic owner"
-    },
-    {
-      "call_count": 1,
-      "diagnostic_digest": "86a8bc66a943d2aef621",
-      "owner": "TASK-494",
-      "path": "tldw_chatbook/UI/Screens/model_remote_view.py",
       "reason": "remaining Chatbook production diagnostic owner"
     },
     {
@@ -3847,9 +3840,9 @@
   "schema_version": 2,
   "scope": "tldw_chatbook/**/*.py",
   "summary": {
-    "owner_files": 520,
+    "owner_files": 519,
     "persistent_sink_files": 8,
     "task_492_calls": 1222,
-    "task_494_calls": 7188
+    "task_494_calls": 7187
   }
 }
--- a/Tests/Agents/test_tool_catalog_concurrency.py
+++ b/Tests/Agents/test_tool_catalog_concurrency.py
@@ -34,6 +34,14 @@ def test_invoke_by_name_takes_exactly_one_catalog_snapshot():
     registry.register_provider(BuiltinToolProvider())
     registry._ensure_catalog_cache()  # warm the cache first
 
+    # Resolve the name BEFORE the counter is installed. `list_catalog()`
+    # routes through `_ensure_catalog_cache()` itself, so taking the name
+    # inside the counting window charges the test's own setup to the
+    # subject and the assertion reads 2 no matter how `invoke_by_name`
+    # behaves -- it stopped measuring anything the moment `list_catalog`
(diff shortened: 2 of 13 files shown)


rmusser01:
Rebased onto dev `93c18eac2` and reviewed all feedback. **No findings to address — nothing was raised.**

* **Qodo code review: 0 bugs, 0 rule violations, 0 requirement gaps.** Its summary review weighed three alternatives to this approach and recommended keeping the PR's: regenerating the ledgers against the ≥3.13 rendering would invalidate large sets of reviewed rows at once and break the 3.11 floor; inlining `show_empty=True` at each call site doesn't compile on 3.11/3.12; raising `requires-python` to ≥3.13 is a product decision a test-only failure doesn't justify. Agreed on all three, which is why the wrapper exists.
* **CodeRabbit: review skipped** — auto-reviews are disabled for base branches other than the default, so this is a configuration notice, not a verdict.

**On the review's currency.** Those Qodo passes ran at 17:17/17:19 UTC, before the final rebase at 17:55, so they technically reviewed the previous SHA. I verified the verdict still applies rather than assuming it: all 13 changed files are **byte-identical** between the reviewed commit `ba8b0b267` and the current head `655090b0b` — the rebase moved the base only. The two dev commits in between (`52b1b3462` exact-width rail, `7dca9a889` modal inventory) touch no diagnostics, and `check_persistent_diagnostic_inventory.py` re-verifies clean against the new tip: **521 owners, 1222 TASK-492 calls, 7189 TASK-494 calls, 8 sink files.**

Final gate on the rebased branch: **434 passed, 0 failed** (3.14), and the ledger suites separately confirmed at **322 passed** on the repo's 3.12 venv — which matters here, because four of the six causes were Python-version drift and a fix verified on one interpreter would prove nothing about the others.

PR #1968, opened 2026-08-22
test(architecture): give the Python floor guard teeth when no floor interpreter exists

rmusser01:
A floor guard already existed (`Tests/Architecture/test_python_floor_syntax.py`, added in #1935). It has a hole, and the hole is the same shape as the bug it was written for.

## The hole

It compiles every module under a real floor interpreter — authoritative and complete — but **skips when none is installed**, and a skip reports exactly as much as a pass. Demonstrated rather than argued, by injecting a genuine PEP 701 construct into `Utils/egress.py`:

```
with uv/python3.11 reachable          ->  1 failed
with PATH stripped to /usr/bin:/bin   ->  SKIPPED ... 1 passed   <- green, with a broken module present
```

So on any machine or CI runner without 3.11, the guard said nothing while a module that cannot import on the declared floor sailed through — which is precisely how `TTS/backends/kokoro.py` shipped in the first place.

## What this adds

`Tests/floor_syntax.py`: an interpreter-independent detector for the class that actually shipped a broken module, plus a test that runs it over every shipped module **with no skip path**. It's deliberately partial — the compile check stays the real gate when an interpreter is available; this is the floor under the floor.

Both PEP 701 relaxations reduce to inspecting the `STRING` tokens nested inside an f-string, so it needs no parser of its own:

* **Same delimiter** — a nested string or f-string reusing the enclosing one. Compared as the *full* delimiter, because `f"""{ d["k"] }"""` is legal on 3.11 (delimiter is `"""`, not `"`).
* **Backslash in the expression** — 3.11 rejects `f"{ 'a\nb'.strip() }"` even though the quotes differ, while a backslash in a *format spec* (`f"{v:\>10}"`) is fine and is not flagged.

**Verdicts come from real interpreters, not from the PEP text.** The detector agrees with a real 3.11 on all 15 cases, and `test_detector_agrees_with_the_real_floor_interpreter` re-derives every pinned verdict from the floor itself — so the table can't quietly rot into folklore. Writing it this way earned its keep immediately: my first implementation was 14/15, missing `f"{ f"{inner}" }"` because a nested *f-string* emits `FSTRING_START`, not `STRING`. The ground-truth comparison caught that; my reasoning had not.

## It also hardens the existing compile check

* Builds the file list with `iter_source_files`, pruning `.venv` / `site-packages` / caches. `.gitignore` does not affect `Path.rglob` — this is the **TASK-19906** class, already in the lessons file. Load-bearing, not theoretical: a planted third-party module in a nested venv **is** reported by a naive sweep (1801 files vs 1800, one false finding).
* Stops `except Exception: pass` swallowing read errors. A file the sweep could not read is a file it did not check, and reporting that as success is the exact failure mode this module exists to prevent.
* Verifies the located interpreter **really reports the floor version**. Both `python3.11` on `PATH` and `uv python find 3.11` are *claims*; an unverified one silently turns the guard into a second run of the developer's own interpreter, which always passes.
* Asserts the probe's own exit status rather than inferring success from empty stdout.

## Verification

Red-proofed three ways:

1. A floor break is caught **with no interpreter present** — the scenario that was previously all-green.
2. Corrupting one pinned verdict fails the reality check, naming the disagreement: `real 3.11 says SyntaxError, table says OK`.
3. A planted floor-breaking module inside a nested venv is correctly **ignored**.

18 passed on 3.12 and on 3.14. On 3.11 itself the detector correctly returns nothing — `FSTRING_START` tokens don't exist there, and an interpreter at the floor rejects these constructs itself at import, which the ordinary suite reports far more clearly.

Companion to #1965, which records the interpreter-drift class in the lessons file. That PR is docs; this one is the guard.

---
## Summary by cubic
Give the Python floor guard teeth when no floor interpreter exists and harden the compile-on-floor path. Previously the architecture test skipped without 3.11 and looked green; now an interpreter-independent scan fails on PEP 701 f-string constructs the floor cannot parse.

- Adds `Tests/floor_syntax.py`, which detects two PEP 701 cases inside f-strings (reusing the enclosing delimiter; backslashes in expressions). Pinned verdicts match real floor behavior, and a test re-derives them from the floor interpreter; the compile check remains the authority when a floor interpreter is available.
- Strengthens `Tests/Architecture/test_python_floor_syntax.py`: verifies the located interpreter actually reports the floor via `_reports_version` (marker-based parsing, timeout), asserts the probe exit status, and feeds a pruned, explicit source list over stdin instead of re-globbing.
- Fixes sweep pruning: `iter_source_files` now excludes only directories below the repo root, and the test asserts the sweep is non-empty. It prunes `.venv`/`site-packages`/caches and reports unreadable files instead of ignoring them.
- Addresses TASK-19560 (prevent shipping floor-incompatible modules) and TASK-19906 (avoid env dirs in sweeps).

<sup>Written for commit bdebb653e1de47c8da3721a7f42c8b314d63f969. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1968?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Architecture/test_python_floor_syntax.py
+++ b/Tests/Architecture/test_python_floor_syntax.py
@@ -75,16 +75,36 @@ def _reports_version(interpreter: str) -> tuple[int, int] | None:
     claim it never checked -- a wrong interpreter here turns the check into a
     second run of the developer's own version, which always passes.
     """
-    probe = subprocess.run(
-        [interpreter, "-c", "import sys;print(sys.version_info[0],sys.version_info[1])"],
-        capture_output=True, text=True,
-    )
+    try:
+        probe = subprocess.run(
+            [
+                interpreter,
+                "-c",
+                "import sys;print('FLOORPROBE',*sys.version_info[:2])",
+            ],
+            capture_output=True,
+            text=True,
+            timeout=60,
+        )
+    except (subprocess.TimeoutExpired, OSError):
+        # A hung or unrunnable interpreter must not stall the suite. Treating
+        # it as "not available" is right: an interpreter we cannot question is
+        # one we cannot verify.
+        return None
     if probe.returncode != 0:
         return None
-    parts = probe.stdout.split()
-    if len(parts) != 2 or not all(p.isdigit() for p in parts):
-        return None
-    return int(parts[0]), int(parts[1])
+    # Scan for the marker line rather than parsing all of stdout. A site
+    # customisation, a warning routed to stdout, or a venv activation notice
+    # would otherwise make a perfectly good floor interpreter look unusable --
+    # and that failure is silent, because it degrades into the skip this whole
+    # module exists to avoid.
+    for line in probe.stdout.splitlines():
+        fields = line.split()
+        if len(fields) == 3 and fields[0] == "FLOORPROBE" and all(
+            field.isdigit() for field in fields[1:]
+        ):
+            return int(fields[1]), int(fields[2])
+    return None
 
 
 def test_every_module_compiles_on_the_declared_python_floor() -> None:
@@ -187,7 +207,12 @@ PEP701_CASES = [
 def test_detector_matches_the_pinned_floor_verdicts(
     source: str, expected_break: bool
 ) -> None:
-    """The detector's verdict on each pinned case, with no interpreter needed."""
+    """The detector's verdict on each pinned case, with no interpreter needed.
+
+    Args:
+        source: A one-line module whose floor-compatibility is pinned.
+        expected_break: Whether a real floor interpreter rejects ``source``.
+    """
     found = find_floor_breaks(source, path=Path("synthetic.py"))
     assert bool(found) == expected_break, (
         f"detector said {bool(found)} for {source!r}; expected {expected_break}"
@@ -256,8 +281,15 @@ def test_no_shipped_module_uses_syntax_the_floor_cannot_parse() -> None:
     which is precisely how `TTS/backends/kokoro.py` shipped unimportable on the
     declared floor.
     """
+    swept = list(iter_source_files(PACKAGE))
+    assert swept, (
+        "swept zero source files -- the sweep is misconfigured, and a guard "
+        "that checks nothing reports the same green as one that checks "
+        "everything"
+    )
+
     findings = []
-    for path in iter_source_files(PACKAGE):
+    for path in swept:
         try:
             source = path.read_text(encoding="utf-8")
         except OSError as exc:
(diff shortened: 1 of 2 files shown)


rmusser01:
Rebased onto dev `08264004a` and addressed all four Qodo findings in `bdebb653e`. **Two were real bugs, and both reintroduced the exact green-while-unchecked mode this PR exists to remove** — which is about the sharpest finding you can land against a guard like this.

### #1 Absolute-path exclusion bypass — real, reproduced

`iter_source_files` matched `EXCLUDED_DIRECTORY_NAMES` against `path.parts`, which includes the components *above* the repo. Reproduced directly:

```
repo under a dir named 'build':
  files swept: 0   <- should be 1
  the floor-breaking module is INVISIBLE
```

So a checkout living anywhere under `build`, `dist` or `venv` swept **zero files** and the always-runs guard passed green with a floor-breaking module sitting in it. Now matched against the parts *below* `root` only, and the test asserts the sweep is non-empty — a guard that checks nothing reports the same green as one that checks everything, so a zero-file sweep must be a failure, not a pass. Both halves re-verified: the real module is seen, a nested `.venv` is still pruned.

### #4 Interpreter probe brittle — real, measured

`_reports_version` ran an external interpreter with **no timeout** and required stdout to be exactly two digit tokens. Any noise — a venv activation notice, a warning routed to stdout — returned `None`, and `None` degrades into precisely the skip this module exists to eliminate. Measured with a shim that echoes one line before delegating to a real 3.11:

```
OLD parser -> None    (guard SKIPS despite a valid 3.11 present)
NEW parser -> (3, 11)
```

Now scans for a marker line and bounds the call at 60s, treating a hung interpreter as unavailable rather than stalling the suite.

### #2 / #3 — style, applied

Return type on `iter_source_files`; `Args:` on the parametrized test's docstring.

### Verification

Red-proofed the new behaviour, not just the absence of failures: forcing a zero-file sweep fails with the new assertion, and the noisy-interpreter shim now parses where the old code gave up. The headline proof still holds — with a PEP 701 construct injected and `PATH` stripped of `uv`/`python3.11`, `test_no_shipped_module_uses_syntax_the_floor_cannot_parse` **fails** where the pre-PR guard was all-green.

**18 passed on 3.12 and on 3.14.**

Companion docs entry #1965 is merged.

PR #1987, opened 2026-08-23
fix(startup): stop Actor Pack recovery from killing app construction when the ChaChaNotes DB cannot be opened

rmusser01:
Closes TASK-20970. **This is a live breakage on `dev`**, not a latent finding.

## The defect

`TldwCli.__init__` → `_wire_character_persona_services` calls `persona_actor_pack_coordinator.recover()` inside `except PersonaActorPackCoordinatorError`. With no ChaChaNotes database the repository raises `AttributeError: 'NoneType' object has no attribute 'execute_query'` — and **`PersonaActorPackCoordinatorError` subclasses `ValueError`, so the `AttributeError` walks straight through the `except` written to contain it.** App construction dies.

Measured at `origin/dev`: **294 of 294** failures across `Tests/App`, `Tests/Scheduling`, `Tests/Subscriptions` and `Tests/Watchlists` end in that identical frame chain, with no second error type present, plus **4 uncollectable `Tests/UI/test_settings_*` suites**. Introduced by TASK-19057.

**Not test-only.** The constructor has always tolerated a missing ChaChaNotes DB — it logs and sets `None`, and the sibling wiring guards that same value. And **TASK-19860 is a shipped cause**: migration `.sql` files are missing from the built wheel, so a PyPI install cannot initialize that database at all. The two findings compound — the user gets a database that will not build and an app that will not start.

## Two repairs, deliberately separate

Collapsing them would leave one layer still lying.

**1. The wiring decides.** `_wire_character_persona_services` reads the database once and branches, matching `_wire_chat_conversation_services` exactly: on `None` it nulls the Actor Pack trio, records `actor_pack_recovery_error`, and logs. It is an `if/else`, **not** an early `return` — the first draft used `return` and would have silently skipped the chat-dictionary and scope-service wiring below it, trading one degraded-mode bug for a quieter one. Review diffed both branches and confirmed they are symmetric, with everything below the `if/else` running in both.

**2. The repository raises its own typed error at construction.** `ActorPackRepository(None)` now raises `ActorPackRepositoryError`. The alternatives were weighed and rejected: `except Exception` at the call site swallows genuine programming errors in a startup path, and adding `AttributeError` to ~10 methods' except-tuples means every method must remember while also swallowing real attribute bugs *inside* the repository. One construction-time check is a true statement about the class's contract rather than a patch on one caller. The call site's `except` is widened to `(PersonaActorPackCoordinatorError, ActorPackRepositoryError)`.

Review confirmed no bypass exists: one production construction site, no classmethods, no alternate constructors, no `__new__`/`__reduce__`/`__copy__`, `self.db` written in exactly one place, and no reassignment anywhere. It cannot fire on a healthy start — proven live, with zero ERROR records during a normal `__init__`.

## The degraded app runs, not merely constructs

Verified live rather than reasoned: isolated HOME/XDG, a real `tldw_chatbook_ChaChaNotes.db` of non-SQLite bytes, **nothing patched** — constructed, mounted, navigated to the Personas screen, drove the real Characters-mode flow, and captured the notification `{"msg": "Actor Pack creation is unavailable.", "severity": "error"}`. Nothing blew up a frame later. Review swept for other consumers of the three attributes and found exactly one outside `app.py`, already guarded.

Born-red for both the no-database and the real-unopenable-file cases; the real-file test asserts the unpatched loader genuinely returns `None` before construction is attempted.

**Three-way mutation, each half independently load-bearing**: repository check removed only → the repository test reds while both app tests stay green; wiring guard disabled only → both app tests red with the *typed* error; both removed → the original `AttributeError` verbatim.

## Review added the missing half of the pin

The fix made this wiring conditional for the first time and pinned only the **degraded** side. Measured gap: nulling all three services at the end of the healthy `else:` branch left 389 tests passing. The only existing guard there is an `inspect.getsource()` ordering assertion — and because the shared app factory patches the database loader to `None`, **no test in the repo constructed `TldwCli` against a working ChaChaNotes database at all.** A new test drives the wiring against a real `CharactersRAGDB` and asserts the trio is live and correctly cross-wired; with it, that mutation reds.

## Also fixed: a fifth shadow, pre-existing

`Tests/Character_Chat/test_character_persona_scope_service.py::test_app_wires_character_persona_services` was already red at `origin/dev` — it passed a bare `object()` as the database, harmless while the wiring only *forwarded* that value, fatal once TASK-19057 made it *query*. Replaced with a two-line double. Review's independent call: every assertion is unchanged and the identity assertion would have gone vacuous under a `None` stand-in, so this is the stronger of the two available repairs.

## Verification

| Gate | `origin/dev` | this branch |
|---|---|---|
| `Tests/App` + `Scheduling` + `Subscriptions` + `Watchlists` | **294 failed**, 1795 passed | **2091 passed**, 1 skipped |
| 4 × `Tests/UI/test_settings_*` | 4 collection errors, 0 collected | **22 passed** |
| repo-wide `--collect-only` | 56,524 collected, **5 errors** | 56,550 collected, **1 error** |

Post-rebase, including `Tests/Actor_Packs`: **2187 passed, 1 skipped**. The surviving collection error is TASK-20972's parametrize mismatch. The diagnostic inventory was regenerated after reading the changed rows with `--statements`: one moved, one re-wrapped, one genuinely new — an all-constant string with no interpolation.

## Honest caveats, both recorded rather than papered over

The operator diagnostic appears verbatim at ERROR, but `PersistentDiagnosticFilter` allowlists the persistent log to structured `diagnostics.*` events, and the in-app Logs handler attaches at mount — so nothing replays `__init__`-era records. In a real TUI run the operator sees this only via `textual console`; their real signal is the point-of-use toast. This branch regresses nothing (it matches the sibling exactly), but "operator-legible" is weaker in practice than it reads.

And "makes *every* method fail in the repository's own typed category" is true for a **missing** database; a non-`None` but wrong-shaped one still raises a raw `AttributeError`.

**Filed separately**: `Tests/UI/test_console_runtime_ownership.py::test_app_fences_console_then_drains_buddy_before_profile_teardown` fails identically before and after (`'TldwCli' object has no attribute 'notes_sync_runtime_owner'`) — a hand-built `object.__new__(TldwCli)` double that never gained an attribute a later drain step started requiring. The other four reds in that module were this defect and are now green.

---
## Summary by cubic
Prevents startup failure when the ChaChaNotes DB cannot open by degrading Actor Pack wiring instead of dereferencing a missing DB. Previously startup built `ActorPackRepository(None)` and `recover()` raised `AttributeError`; now the app starts and Personas features degrade safely.

- `_wire_character_persona_services` branches on `chachanotes_db`. When `None`, it sets the repository/coordinator/creation service to `None`, records `actor_pack_recovery_unavailable`, and logs one ERROR. The rest of app wiring still runs.
- With a working DB, wires repository/coordinator and runs recovery. Catches `(PersonaActorPackCoordinatorError, ActorPackRepositoryError)` and records `actor_pack_recovery_failed` if recovery fails; warns on quarantined intents as before.
- `ActorPackRepository.__init__` now raises `ActorPackRepositoryError("actor_pack_repository_unavailable")` if constructed with `None`, preventing raw `AttributeError` escapes.
- Added tests for degraded startup (factory and real unopenable file) and a repository test; fixed a scope-service test by using a minimal DB double. Regenerated diagnostics inventory and updated docs.
- No migrations or config changes. User-visible effect: when the DB is unavailable or corrupt, Actor Pack creation is unavailable; the UI already guards and logs one operator-legible message.

Addresses TASK-20970.

<sup>Written for commit 99729c736445ea4eb021229ab39ae8d6b455aac6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/1987?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/security/production-diagnostic-inventory.json
+++ b/Docs/security/production-diagnostic-inventory.json
@@ -3666,8 +3666,8 @@
       "reason": "remaining Chatbook production diagnostic owner"
     },
     {
-      "call_count": 328,
-      "diagnostic_digest": "5a92956a5766f8b313e0",
+      "call_count": 329,
+      "diagnostic_digest": "09c4ee210584f33664ef",
       "owner": "TASK-494",
       "path": "tldw_chatbook/app.py",
       "reason": "remaining Chatbook production diagnostic owner"
@@ -3906,6 +3906,6 @@
     "owner_files": 528,
     "persistent_sink_files": 8,
     "task_492_calls": 1222,
-    "task_494_calls": 7241
+    "task_494_calls": 7242
   }
 }
--- a/Tests/Actor_Packs/test_actor_pack_creation.py
+++ b/Tests/Actor_Packs/test_actor_pack_creation.py
@@ -314,3 +314,71 @@ def test_app_owns_creation_only_after_recovery_and_before_surfaces() -> None:
     creation = wiring.index("ActorPackCreationService(")
     scope = wiring.index("CharacterPersonaScopeService(")
     assert recovery < creation < scope
+
+
+def test_app_wires_live_actor_pack_services_when_the_database_opens(
+    monkeypatch, tmp_path: Path
+) -> None:
+    """TASK-20970 review: the *open-database* half of the new wiring branch.
+
+    The sibling above asserts source-text ordering, which a behaviour-only
+    regression walks straight past: nulling all three services at the end of
+    the ``else`` branch leaves that assertion (and every other test in
+    `Tests/Actor_Packs`, `Tests/App`, `Tests/UI/test_actor_pack_creation_
+    workflow.py`) green, because `TASK-20970`'s own coverage only pins the
+    *degraded* side and the shared app factory patches
+    `get_chachanotes_db_lazy` to `None` for every app-building test in the
+    repo. Nothing behavioural asserted that a working database still yields
+    working Actor Pack services. This does.
+    """
+
+    from unittest.mock import Mock
+
+    from tldw_chatbook import app as app_module
+
+    database = CharactersRAGDB(tmp_path / "wiring.db", client_id="wiring")
+    try:
+        monkeypatch.setattr(
+            app_module.ServerCharacterPersonaService,
+            "from_server_context_provider",
+            Mock(return_value=Mock()),
+        )
+        monkeypatch.setattr(
+            app_module.ServerChatDictionaryService,
+            "from_server_context_provider",
+            Mock(return_value=Mock()),
+        )
+
+        fake_app = Mock()
+        fake_app.chachanotes_db = database
+        fake_app.service_policy_enforcer = object()
+        fake_app.server_context_provider = object()
+
+        app_module.TldwCli._wire_character_persona_services(fake_app)
+
+        assert isinstance(fake_app.actor_pack_repository, ActorPackRepository)
+        assert fake_app.actor_pack_repository.db is database
+        assert isinstance(
+            fake_app.persona_actor_pack_coordinator, PersonaActorPackCoordinator
+        )
+        assert (
+            fake_app.persona_actor_pack_coordinator.repository
+            is fake_app.actor_pack_repository
+        )
+        assert isinstance(
+            fake_app.actor_pack_creation_service, ActorPackCreationService
+        )
+        assert fake_app.actor_pack_creation_service.database is database
+        assert (
+            fake_app.actor_pack_creation_service.repository
+            is fake_app.actor_pack_repository
+        )
+        assert (
+            fake_app.actor_pack_creation_service.persona_coordinator
+            is fake_app.persona_actor_pack_coordinator
+        )
+        # Recovery ran and found nothing to reconcile: no fixed category is
+        # recorded, so the degraded/blocked/failed states stay distinguishable.
+        assert fake_app.actor_pack_recovery_error is None
+    finally:
+        database.close_connection()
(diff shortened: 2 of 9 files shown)


rmusser01:
Superseded by task-21106, which landed on `dev` while this was in review and fixes the same crash by a different and better route: it moves Actor Pack crash recovery out of `TldwCli.__init__` entirely, onto a deferred background thread gated ahead of the Personas screen's first read. Verified on current `dev` — `Tests/App` + `Tests/Scheduling` = 504 passed, `Tests/Subscriptions` + `Tests/Watchlists` + the previously-uncollectable settings suites = 1596 passed / 1 skipped, zero failures. The 294-failure cluster and all 4 collection errors are gone.

The two approaches are semantically incompatible, so this cannot simply be rebased: this branch made `ActorPackRepository(None)` raise at construction, while task-21106 deliberately constructs it with `None` and defers the query. Theirs is the better shape — nothing queries during `__init__` at all, which also removes synchronous SQLite from every boot.

Two things from this branch's review are worth keeping and are being carried to a follow-up rather than lost:

1. **No test constructs `TldwCli` against a working ChaChaNotes database.** The shared app factory patches `get_chachanotes_db_lazy` to `None`, so the healthy wiring path is unexercised — measured here by nulling all three Actor Pack services in the healthy branch and finding **389 tests still passing**. That gap is independent of which fix won, and task-21106's deferral arguably widens it.
2. `Tests/UI/test_console_runtime_ownership.py::test_app_fences_console_then_drains_buddy_before_profile_teardown` remains red on `dev` for an unrelated reason — a hand-built `object.__new__(TldwCli)` double that never gained `notes_sync_runtime_owner`, which a later drain step started requiring.

PR #2027, opened 2026-08-23
fix(ci): stop the required Derived Artifacts gate from cancelling its own PR runs (TASK-21250)

rmusser01:
## The problem

`Derived Artifacts` is the single context branch protection requires. It is not reporting.

Measured over 60 consecutive runs today: **45 cancelled / 14 success / 1 failure — a 23% success rate.** Cancellations hit every branch, including other contributors' work and direct pushes to `dev`, so **merges to `dev` are currently blocked at random for everyone**.

## Root cause

The workflow's own concurrency rule:

```yaml
group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```

Its stated justification — *"a merge is gated on the HEAD commit's checks, and the newest run is the one that reports for HEAD"* — is correct as far as it goes, but it assumes the newest run gets to **finish**.

For a `pull_request` event `github.ref` is `refs/pull/N/merge`, and GitHub recreates that merge ref **whenever the base branch moves**. With `dev` absorbing 23–50 merges/day, every open PR's gate is cancelled repeatedly *without anyone touching the PR*. Queue time on the shared pool measured **4–59 minutes** against ~33 s of actual work, so the newest run rarely survives long enough to report.

Observed directly while landing the 2026-08-22 perf burn-down: three PRs sat unmergeable through repeated reruns and rebases, each fresh run cancelled within ~2 minutes, while the same workflow succeeded on a quiet branch.

## The change

```yaml
cancel-in-progress: ${{ github.event_name == 'push' && github.ref != 'refs/heads/main' }}
```

- `push` to `dev` still supersedes — the 50-obsolete-run pileup the rule was written to prevent does not come back.
- `push` to `main` still never cancels.
- `pull_request` now always runs to completion.

The trade-off is explicit and cheap: a rapidly-updated PR may stack a few redundant ~90-second ubuntu runs. That is far better than a required check that blocks merges 77% of the time — and it is the shape TASK-19572 was reaching for: *"small enough to run on every PR, which is what makes it usable as a REQUIRED status check."*

The measurement and the merge-ref mechanism are recorded in the workflow comment so the justification travels with the code.

This PR's own gate run is the first test of the fix.

---
## Summary by cubic
Stops the `Derived Artifacts` required check from cancelling its own PR runs by limiting cancellation to push events. Before: any non-`main` ref cancelled in-progress runs (including `pull_request`) when the base moved; now PR runs complete, pushes to `dev` still supersede, and pushes to `main` never cancel. Addresses TASK-21250.

- Sets `cancel-in-progress` to `${{ github.event_name == 'push' && github.ref != 'refs/heads/main' }}`; concurrency group unchanged; adds inline rationale in the workflow.
- Side effect: fast-updating PRs may run a few redundant ~90 s jobs; no rollout or config changes needed.

<sup>Written for commit 1659097e58f97cbff16040bfc961495b368c862a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2027?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/derived-artifacts.yml
+++ b/.github/workflows/derived-artifacts.yml
@@ -31,17 +31,30 @@ on:
 # buys nothing here, because the job is ~90 s: just run it. If you are tempted
 # to add `paths:` below, you are about to brick merges -- don't.
 
-# One run per workflow per ref: a new push supersedes the previous run instead of
-# queueing another alongside it. Without this, every push to `dev` stacked a full
-# run - 50+ obsolete runs accumulated, all competing for scarce macOS runners.
-# The event name is part of the group so different trigger types (push,
-# pull_request, schedule) never cancel each other, and `main` is never cancelled
-# so its history stays complete. Cancellation is safe for a required check: a
-# merge is gated on the HEAD commit's checks, and the newest run is the one that
-# reports for HEAD.
+# One run per workflow per ref for PUSH events: a new push to `dev` supersedes
+# the previous run instead of queueing another alongside it. Without this, every
+# push to `dev` stacked a full run - 50+ obsolete runs accumulated, all competing
+# for scarce macOS runners. The event name is part of the group so different
+# trigger types never cancel each other, and `main` is never cancelled so its
+# history stays complete.
+#
+# TASK-21250: pull_request runs are deliberately NOT cancelled. The original rule
+# reasoned that cancelling is safe because "the newest run is the one that reports
+# for HEAD" - true, but it assumed the newest run gets to FINISH. For a
+# pull_request event `github.ref` is `refs/pull/N/merge`, and GitHub recreates
+# that merge ref every time the BASE branch moves. On a repo absorbing 23-50
+# merges/day, every open PR's gate was therefore cancelled repeatedly without
+# anyone touching the PR, and because queue time for the shared pool runs
+# 4-59 minutes (vs ~33 s of actual work), the newest run rarely survived long
+# enough to report. Measured over 60 consecutive runs before this change:
+# 45 cancelled / 14 success / 1 failure - a 23% success rate for a REQUIRED
+# check, i.e. merges to `dev` were blocked at random for everyone.
+#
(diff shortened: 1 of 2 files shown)


rmusser01:
## Correction: this PR's own run was also cancelled

I claimed above that this PR's gate run would be the first test of the fix. It ran and was **cancelled within ~2 minutes** — so PR-scoped `cancel-in-progress` was evidently not the only cause, and I do not want that claim standing uncorrected.

What still holds:
- The measurement is real: **45 cancelled / 14 success / 1 failure over 60 consecutive runs**, across every branch including direct pushes to `dev`.
- The merge-ref mechanism is real and is *a* contributing cause: for `pull_request` events `github.ref` is `refs/pull/N/merge`, which GitHub recreates whenever the base moves, so an untouched PR's gate gets superseded repeatedly on a repo taking 23–50 merges/day.
- This change is still strictly correct on its own terms and costs only a few redundant ~90 s runs.

What I got wrong: treating it as sufficient. There is evidently a second source of cancellation — `css-bundle-guard.yml`'s own comment refers to the suite being "left in the intentionally-cancelled state", which suggests cancellation is at least partly deliberate here. If that is an external/manual practice, then requiring `Derived artifacts reproduce from their sources` under branch protection while runs are routinely cancelled makes `dev` unmergeable through the normal path at random — which is what I hit trying to land four burn-down PRs today (each fresh run cancelled within ~2 minutes; four earlier PRs did get a verdict and merged normally).

Flagging for the owner rather than working around it: I have not used `--admin` on anything, and won't.

PR #2033, opened 2026-08-24
Shard the core test job and give the summary check a verdict (TASK-21411)

rmusser01:
The Tests workflow has produced no verdict since 2026-06-26. Two independent faults keep it
that way, and neither is visible from the outcome alone. This does not make the suite green
— it makes it legible, which everything else needs.

## 1. `core-tests` has never finished

Every leg ends at *exactly* 2h00m. That is `timeout-minutes: 120` expiring — and **GitHub
reports a timeout as `cancelled`**, so from the outside it reads like the run being
superseded by a newer push. It isn't.

**Too slow, or hung?** Those look identical from the outcome and have different fixes, so I
read the killed job's own log (run 32647831275, ubuntu leg):

- job starts `15:39:14`
- pytest begins emitting `15:44:55` (~5.7 min checkout + install)
- **still printing progress steadily, no stall,** when the cap kills it at `17:39` — at **60%**

Steady progress at the kill means slow, not hung. Extrapolated: **~190 minutes** on a 4-vCPU
runner. (Public repos get 4 vCPU, so `-n auto` is 4 workers; the same work takes 30.8 min
locally on 8 faster cores — the arithmetic accounts for the gap without a hang.)

No timeout below GitHub's 6-hour ceiling makes one job the right container for that. **Split
six ways with pytest-shard** — the same deterministic contract the UI job has used since
TASK-18608 — at ~32 min per slice, with xdist still parallelizing within each.

**Verified rather than assumed.** Collecting each shard and unioning them:

| | |
|---|---|
| unsharded core collection | **42,811** node ids |
| union of the 6 shards | **42,811**, zero overlap |
| union vs unsharded | **identical** |

Nothing dropped, nothing run twice.

**ubuntu only, deliberately.** macOS is the scarce pool here — on run 32673270908 the macOS
core job waited **42 minutes** to start, and in an earlier run a UI shard waited ~90.
Sharding the scarcest pool six ways would spend more in queueing than it buys. macOS breadth
already lives in `nightly-deep`, and a new test pins both halves of that trade.

## 2. `test-summary` could not fail

It declares `needs:` and `if: always()`, which schedules it after those jobs *whatever they
did* — it never inspected what they did. On run 32647831275 it reported **`success` while
all twelve UI shards were red**. It is also the check most likely to be marked required, so
a summary that cannot go red is worse than no summary.

The verdict is now its own step, placed **last**, so a red suite still leaves the PR comment
behind rather than exiting before it is posted.

## Shape test updated in lockstep — and one finding inside it

Three new tests, each mutation-proven. Removing the verdict step, breaking the shard
partition to ids `1..6`, moving the verdict ahead of the PR comment, and putting macOS back
each fail exactly one test; the restored tree is 18/18 green.

Fixing an existing test was itself a finding. `test_ui_job_is_sharded_to_fit_its_time_budget`
located the shard id list by searching the **whole workflow** for the first `shard: [` —
which was the UI job only because it was then the only sharded job. With core sharded, it
began checking core's ids against the UI job's divisor. A partition assertion that can
silently start describing a different job is worse than no assertion. It now reads from the
UI job block.

## What this cannot fix, and why

`nightly-deep` has still never run, and **no branch to `dev` can change that**. GitHub
registers `schedule:` only from the default branch, which is `main`; main's workflow has no
schedule block, and **main is 11,528 commits behind dev**. Because cron would execute
*main's* workflow, main would have to carry the nightly-deep job itself. That is a decision
about `main`, not a workflow edit, so I've left it to you. Until it's made, `--run-slow`
(41 marks), the thorough Hypothesis profile, the CSS-cache-off soak, Windows and Python
3.11/3.13 run nowhere.

## Note on required checks

Sharding renames the core job, so the check names change to
`Core Tests (all but UI) - shard N/6`. Nothing currently requires them — worth knowing before
any are marked required, which is best done per-job as each goes green rather than all at
once.

---
## Summary by cubic
Shards the core test job into six deterministic pytest slices on Ubuntu and gives the test-summary job a real verdict. Previously, core-tests always timed out at 120m and test-summary always reported success; now each core shard completes in ~32m and the summary fails when any gated job fails while still posting its comment.

- Satisfies TASK-21411 by making the suite finishable and the summary truthful.

**Review notes**
- Core: 6 shards (ids 0–5) on `ubuntu-latest`; `pytest-xdist` stays enabled; per-shard artifact names prevent clobbering; 120m remains headroom; job name is Core Tests (all but UI) - shard N/6; macOS runs stay in nightly-deep to avoid queue contention.
- Summary: adds a final “Require every gated job to have succeeded” step; it fails on any non-success among core-tests, ui-tests, textual-minimum, and artifact-lease-gate; the verdict step runs last so the PR comment is always posted.
- Tests: workflow-shape tests pin core sharding, Ubuntu-only choice, per-shard artifacts, and the summary’s fail/positioning; the UI test now reads shard ids from the UI block.

**Rollout / required actions**
- Update branch protections to the new per-shard core job names and/or the summary check.
- Nightly remains disabled from `dev`; enable `schedule:` on the default branch to run nightly-deep.

<sup>Written for commit 3c6e83f0418ea7ad7072e7f78fe02e1e7241f8b4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2033?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,30 +42,40 @@ jobs:
     # task-1465: full coverage on every PR replaces `-m unit` (which selected
     # 27 of ~950 files while ~590 files ran in no PR-triggered job). The UI
     # shard is its own job below; OS/Python breadth lives in nightly-deep.
-    name: Core Tests (all but UI) - Python ${{ matrix.python-version }} - ${{ matrix.os }}
-    runs-on: ${{ matrix.os }}
-    # 120, not 60: the suite grew past the old budget -- the ubuntu leg was
-    # killed by the 60-minute cap mid-run (TASK-18608) while the macOS leg
-    # finished at ~58m. 2x headroom over the observed ~60m worst leg.
+    #
+    # TASK-21411: sharded, because the 120-minute budget stopped being the
+    # question. Read from a killed run's own log: the ubuntu leg started
+    # pytest at 15:44:55 and was still emitting progress steadily -- no
+    # stall -- when the cap killed it at 17:39 having reached **60%**. So the
+    # suite needs ~190 minutes on a 4-vCPU runner, and no timeout this side
+    # of GitHub's 6-hour ceiling makes a single job a sane place to put it.
+    # Six deterministic pytest-shard slices are ~32 min each, roughly a
+    # quarter of the budget, and xdist still parallelizes within each slice.
+    #
+    # ubuntu only, on purpose. macOS runners are the scarce ones -- queue
+    # waits of 42-90 minutes are routine here -- and multiplying the scarcest
+    # pool by six would cost more in queueing than it buys in coverage. macOS
+    # breadth already lives in nightly-deep, which is where it belongs.
+    name: Core Tests (all but UI) - shard ${{ matrix.shard }}/${{ strategy.job-total }}
+    runs-on: ubuntu-latest
+    # Kept at 120 as headroom: the per-shard estimate is ~32 min, so this is
+    # ~4x margin rather than a target. A shard that approaches it is a signal
+    # to resize the matrix, not to raise the cap again.
     timeout-minutes: 120
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - os: ubuntu-latest
-            python-version: "3.12"
-          - os: macos-latest
-            python-version: "3.12"
+        shard: [0, 1, 2, 3, 4, 5]
 
     steps:
     - uses: actions/checkout@v4
       with:
         fetch-depth: 0
 
-    - name: Set up Python ${{ matrix.python-version }}
+    - name: Set up Python 3.12
       uses: actions/setup-python@v5
       with:
-        python-version: ${{ matrix.python-version }}
+        python-version: "3.12"
         cache: 'pip'
 
     - name: Install all dependencies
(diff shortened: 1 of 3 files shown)


rmusser01:
### The split is confirmed by this PR's own run

Run `32676835700`, on this branch. Two shards got runners and **both finished**:

| shard | window | duration | conclusion |
|---|---|---|---|
| 4/6 | 02:19:33 → 02:53:58 | **34m25s** | `failure` |
| 0/6 | 02:56:13 → 03:30:43 | **34m30s** | `failure` |

Two things worth separating there.

**They completed.** The estimate in the PR body was ~32 min per slice, derived from the
killed job reaching 60% in ~114 minutes. Measured: 34.5 min — close enough that the sizing
holds, and roughly a quarter of the 120-minute budget, so the cap is headroom rather than a
target.

**`failure`, not `cancelled`.** That is the point of the change. These are the first core
verdicts this repository has produced. The reds they report are the pre-existing ~144 core
failures inventoried in #2030's baseline; this PR does not claim to fix them, only to make
them visible.

The remaining four shards show `cancelled` with `startedAt` equal to the run's creation
time — they were still **queued** when a later push superseded the run, never having started.
That is the runner-starvation problem described in the PR body, and it is the reason this
change deliberately does not shard macOS as well.

PR #2038, opened 2026-08-24
fix: keep Windows startup alive when Actor Pack cleanup is unsupported

rmusser01:
## Summary

- treat usable-but-unverified Actor Pack staging as a non-destructive startup cleanup no-op
- preserve cleanup-denied behavior for unusable staging and authenticated cleanup on verified platforms
- add a regression proving service construction succeeds without enumerating or deleting unverified candidates
- file TASK-21252 for secure native Windows Actor Pack import, ACL, reparse-point, cleanup, and CI support

## Security boundary

The guard returns before `os.scandir()` or candidate authority helpers. Import-time `verified_private` checks remain unchanged, so this restores Windows app startup without enabling Actor Pack import on Windows. ADR-074 remains governing; no new ADR is required for this bug fix.

## Verification

- `PYTHONPATH=. /Users/macbook-dev/Documents/GitHub/tldw_chatbook/.venv/bin/python -m pytest Tests/Actor_Packs -q` — 202 passed, 1 dependency warning
- Ruff format check — 2 files already formatted
- Ruff lint — all checks passed
- compileall for the modified Python files — passed
- `git diff --check origin/dev...HEAD` — passed

Repository-wide `pytest -q` produced no terminal result after roughly 100 minutes and was interrupted. This is neither recorded as a pass nor a failure; TASK-21251 remains In Progress for that completion gate. Non-failing post-summary pytest cleanup warnings for leftover temporary directories were also observed.

## Backlog

- TASK-21251: In Progress pending repository-wide test completion
- TASK-21252: To Do — secure native Windows Actor Pack import support

---
## Summary by cubic
Keeps Windows startup alive by making startup cleanup a no‑op when Actor Pack staging is usable but unverified, and returns a stable disk‑unavailable error during import preflight when staging cannot be verified. Previously startup raised cleanup‑denied and aborted; preflight raised a generic error.

- In `sweep_staging()`, if staging is usable but not `verified_private`, return 0 before any `os.scandir()` or candidate helper calls; POSIX‑verified cleanup is unchanged; unusable staging still errors.
- `_preflight_space()` now raises `ActorPackImportError("actor_pack_import_disk_unavailable")` when staging privacy is unverified; `inspect_archive()` creates no staging candidates on this path and reports disk‑unavailable.
- Adds regressions: startup no‑op path skips enumeration and leaves candidates intact; inspect‑on‑unverified reports disk‑unavailable with an empty staging root.
- Security boundary is unchanged and Windows import remains unsupported under ADR‑074. Addresses TASK‑21251; native Windows support is tracked in TASK‑21252.

<sup>Written for commit 7c24fd3d684559e262e7cfff5f4b023a7391270f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2038?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/superpowers/plans/2026-08-23-task-21251-windows-actor-pack-startup-cleanup.md
+++ b/Docs/superpowers/plans/2026-08-23-task-21251-windows-actor-pack-startup-cleanup.md
@@ -207,12 +207,18 @@ archive validation, activation, schema, dependency, or configuration changes.
 - [ ] **Step 5: Update TASK-21251 through the Backlog CLI**
 
 Check all four acceptance criteria, add concise implementation notes including
-the commands and results above, state that no new lesson was needed, retain the
-ADR-074 link and ADR check, and set the task to `Done` only if every required
-check is green:
+the results above, state that no new lesson was needed, retain the ADR-074 link
+and ADR check, and set the task to `Done` only if every required check is green.
+Use `backlog task edit 21251 --notes ... --plain` to write the notes. They must
+name the exact commands, actual pass counts, and warnings observed in Steps 1-3;
+do not substitute generic text and do not mark the task `Done` if that evidence
+is unavailable. Then check the acceptance criteria, inspect the rendered task to
+confirm the evidence is present, and only then change the status:
 
 ```bash
-backlog task edit 21251 --check-ac 1 --check-ac 2 --check-ac 3 --check-ac 4 --notes "Implemented the usable-but-unverified startup cleanup no-op in ActorPackImportService, with regression coverage proving service construction succeeds, sweep returns zero, and no candidate enumeration or deletion occurs. Existing authenticated cleanup behavior remains covered. ADR required: no; ADR-074 continues to govern the fail-closed boundary. Verification: record exact passing commands and counts here. Lessons: no new general lesson; the incident is specific to the documented platform capability split." -s Done --plain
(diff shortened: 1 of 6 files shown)


rmusser01:
Fixed in a5a7f8c78. `_preflight_space()` now raises `actor_pack_import_disk_unavailable` when private staging is unverified, and `inspect_archive()` propagates that category unchanged. A red-green regression proves a valid pack receives the storage-unavailable category before any staging candidate is created. Verification: `Tests/Actor_Packs` passed 203/203; Ruff format/lint and `git diff --check` are clean.

rmusser01:
Final merge evidence on rebased head 7c24fd3d6 (base 7c1808002): Qodo reports 0 bugs / 0 rule violations and the review thread is resolved; `Tests/Actor_Packs` passed 203/203; the focused startup selection passed 2/2; Ruff lint/format, compileall, and diff checks passed. The required Derived Artifacts gate was also reproduced locally with its exact five Python checks, all passing. Two GitHub Actions waves were cancelled before runner assignment with no job logs, so the merge uses the owner/admin bypass for that runner-side cancellation rather than treating it as a test failure.

PR #2046, opened 2026-08-24
Forbid model downloads in the test sandbox — 188 of one CI shard's 189 errors (TASK-21562)

rmusser01:
> **Opened for CI evidence, not ready to merge on local results.** The point of this change is
> a CI-only failure mode; only a CI run can show whether the affected tests then pass offline
> or fail fast. Local results below establish only that the change is *neutral* here.

## What CI was doing

The sandbox points the HuggingFace cache at a directory inside itself
(`<sandbox>/home/.cache/huggingface/hub`) which **never exists**, and `HF_HUB_OFFLINE` was
**False** — so any code path reaching the hub attempted a real download, the network guard
refused it, and `huggingface_hub` retried five times before the guard's record failed the
test at teardown.

From core shard 0/6 of run `32676835700`:

| | |
|---|---|
| errors in that shard | **189** |
| of which egress-blocked | **188** |
| recorded attempts, CDN address | 709 |
| recorded attempts, `huggingface.co:443` | 65 |

The attempt count exceeds the error count because each call retries — `Retrying in 2s
[Retry 2/5]` is in the log. One shard of six.

## Why nobody has seen it

The same tests record **zero** egress attempts locally: `test_console_agent_bridge.py` is 254
passed / 0 blocks here against 19 errors in CI. This was invisible for as long as the core job
could not finish — which, until #2033, was always.

## The fix, and why it has two halves

`huggingface_hub` freezes `constants.HF_HUB_OFFLINE` at *its own* import time, so an env var
written from a fixture is read too late to matter; `is_offline_mode()` consults the constant
on every request. So:

- the **pre-import bootstrap** sets `HF_HUB_OFFLINE=1` before anything imports the client
- an **autouse fixture** patches the constant when the module is already in `sys.modules`

`Tests/RAG_Eval/conftest.py` already documents exactly this split, with the measurement, for
its own scope; this generalises it. The fixture uses a `sys.modules` lookup and never imports
the hub stack — importing it in every session to assert a property of sessions that never
touch it would cost more than the guard is worth.

## What I could not establish, stated plainly

**Why the hub path runs in CI and not here.** The cache is cold in both, so the cold cache is
not the differentiator, and I did not find what is. The likely candidate is the dependency
set — CI installs `.[embeddings_rag,websearch,chunker]` while this venv has every optional
group, and optional-dependency branches decide whether the hub path is entered — but that is
**unconfirmed**.

An earlier hypothesis blaming tiktoken's temp-dir cache was **wrong**, recorded so it is not
retried: a cold `TIKTOKEN_CACHE_DIR` reproduced nothing (269 passed), and the CI log shows
`huggingface_hub` retrying, not tiktoken. The tiktoken `ERROR` lines there are incidental.

## Evidence

**Neutral locally** — two A/Bs, offline forced on versus opted out:

| suites | offline on | opted out |
|---|---|---|
| the three CI-erroring files | 633 passed | 633 passed |
| `RAG_Search` + `Chunking` + `Transcription` | 4 failed / 1214 passed / 74 skipped / 1 xfailed | identical |

**Mutation-proven.** Replacing the bootstrap condition with `if False` reddens the
environment assertion. Notably the other two guard tests still passed under that mutation,
because the autouse fixture patched the constant — the halves are genuinely belt-and-braces,
not one guard written twice.

**Opt-out.** `TLDW_TEST_ALLOW_HF_DOWNLOADS=1`. The guard module carries a `pytestmark` skipif
on it rather than asserting the flag is unset, so asking for live fetches does not redden the
suite — and it *skips* rather than returning early, so the opt-out appears in the run summary
instead of looking like three tests that passed while asserting nothing. Verified: 3 passed
normally, 3 skipped when opted out.

## A local measurement I discarded rather than report

A broader local sweep crashed xdist workers repeatedly. Concurrent sessions had 12–21 pytest
processes running at the time, and **CI shard logs show zero `node down` events**, so those
crashes are this machine and not the code. Reporting that sweep's numbers would have been
reporting contention.

---
## Summary by cubic
Forbids model downloads in the test sandbox to resolve CI-only network retries and make failures deterministic (TASK-21562). Previously, CI tests attempted real `huggingface_hub` downloads and retried before the network guard failed at teardown; now tests run offline and fail fast if they reach the hub.

- Sets `HF_HUB_OFFLINE=1` in `Tests/conftest.py` before any import; adds an autouse fixture that patches `huggingface_hub.constants.HF_HUB_OFFLINE` if the module was imported early (the client freezes the constant at import; `is_offline_mode()` reads it per request).
- Adds `Tests/test_huggingface_offline.py` to assert both halves and to skip clearly when opted out; the fixture uses a `sys.modules` lookup and avoids importing the hub stack.
- CI context: one shard saw 188/189 errors from blocked egress to `huggingface.co:443` and a CDN; local runs were neutral on targeted suites.
- Opt-out for tests that need live fetches: set `TLDW_TEST_ALLOW_HF_DOWNLOADS=1` (skips appear in the summary).

<sup>Written for commit f9128f9bc493904f37e8bac8f78d28f9c3f0b139. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2046?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/conftest.py
+++ b/Tests/conftest.py
@@ -17,6 +17,7 @@ _SANDBOXED_ENV_NAMES = (
     "XDG_CONFIG_HOME",
     "TLDW_CONFIG_PATH",
     "PYTHON_KEYRING_BACKEND",
+    "HF_HUB_OFFLINE",
     _TEST_CONFIG_ROOT_ENV,
     _TEST_CONFIG_OWNER_ENV,
 )
@@ -79,6 +80,27 @@ os.environ["TLDW_CONFIG_PATH"] = str(_BOOTSTRAP_CONFIG_PATH)
 # unpatched. `TLDW_TEST_REAL_KEYRING=1` is the deliberate opt-out.
 if os.environ.get("TLDW_TEST_REAL_KEYRING") != "1":
     os.environ["PYTHON_KEYRING_BACKEND"] = "keyring.backends.null.Keyring"
+
+# TASK-21562: no test downloads a model. `huggingface_hub` freezes
+# `constants.HF_HUB_OFFLINE` at ITS import time, so this has to be written here
+# -- in the pre-import bootstrap -- to be read at all; an env var set from a
+# fixture arrives far too late (the measurement is in
+# `Tests/RAG_Eval/conftest.py`'s offline-latch comment).
+#
+# Why globally, when two sub-conftests already do it for their own scope: the
+# HF cache resolves INTO this sandbox (`.../home/.cache/huggingface/hub`), which
+# never exists, so any code path reaching the hub attempts a real download. That
+# is invisible on a developer machine -- whatever triggers it locally is not
+# triggered -- and very visible in CI, where one core shard alone recorded 188
+# egress-blocked errors against `huggingface.co:443` and a CDN address, each one
+# `huggingface_hub` retrying five times before the guard's record failed the
+# test at teardown. Offline turns that into an immediate, deterministic, local
+# failure instead of a slow CI-only one.
+#
+# `TLDW_TEST_ALLOW_HF_DOWNLOADS=1` opts out, for a test that genuinely needs a
+# live fetch.
+if os.environ.get("TLDW_TEST_ALLOW_HF_DOWNLOADS") != "1":
+    os.environ["HF_HUB_OFFLINE"] = "1"
 
 import pytest  # noqa: E402
 import pytest_asyncio  # noqa: E402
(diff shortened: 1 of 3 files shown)


rmusser01:
### CI evidence: the HuggingFace half is fixed, and it was the smaller half

Core shard 5/6 of this branch's run (`32696030401`) completed — 60 failed, 7,200 passed,
87 skipped, 106 errors in 29:07. Fetched per-job via the API, since run-level logs are
withheld until every UI shard finishes.

**What the change did, measured on the same workflow and the same guard:**

| | baseline (shard 0/6, run 32676835700) | this branch (shard 5/6) |
|---|---|---|
| `huggingface.co:443` attempts | 65 | **0** |
| `huggingface_hub` retry lines | present | **0** |

So the offline latch works exactly as intended. No test reaches the hub any more.

### But the dominant egress source was never HuggingFace

The same log shows **1,156 blocked attempts to a single address**, `57.150.97.129:443`, and
the log names it outright: **2,322 references to `openaipublic.blob.core.windows.net`** and
1,161 to `tiktoken_encoding`.

That address was in the baseline too — **709 of its 774 attempts**, against HuggingFace's 65.
So there were always two sources, tiktoken being roughly 90% of them, and this PR fixes the
10%.

**This corrects something I wrote earlier in this PR's description.** I had recorded the
tiktoken hypothesis as *disproven*. It was not: I tested it by pointing `TIKTOKEN_CACHE_DIR`
at an empty directory for a single file that, as it turns out, does not exercise the
token-counting path — and my machine carries a warm `data-gym-cache` in its temp directory
regardless. The conclusion outran the experiment. `Utils/token_counter.get_tiktoken_encoding`
downloads a BPE file on a cold cache, swallows the failure, and returns `None`; the guard
records the attempt the swallow hides.

### What that means for this PR

It stands on its own: it removes a real source of egress, is provably neutral locally, and
converts a class of CI-only failure into a deterministic one. It just is not the whole
story, and the description overstated its share.

The tiktoken source needs its own fix and is filed separately — it is the larger prize, and
the shape of the fix is not the same, because production genuinely wants an encoding where
HuggingFace access was incidental.

rmusser01:
### Merge note: this conflicts with the sibling egress PR

#2046 (HuggingFace offline) and #2061 (vendored tiktoken tables) both add to the same
pre-import bootstrap block in `Tests/conftest.py`, in two places:

- the `_SANDBOXED_ENV_NAMES` tuple (`HF_HUB_OFFLINE` vs `TIKTOKEN_CACHE_DIR`)
- the block of env writes just after the keyring pin

I hit this cherry-picking both onto a measurement branch. **Both conflicts are purely
additive — the correct resolution is to keep both sides**, in either order. Verified: with
both applied, `Tests/test_huggingface_offline.py` + `Tests/test_tiktoken_vendored_cache.py`
give 8 passed.

Whichever merges first, the second will need that trivial resolution rather than a choice.

PR #2054, opened 2026-08-24
Retire or repair the tests that cannot fail (TASK-21563)

rmusser01:
Six files whose tests passed unconditionally — not weak checks that might miss a subtle
regression, but checks whose assertions were never written, so no behaviour of the code under
test could make them red.

## The security one

`test_cookie_db_path_validation` patched `os.path.expanduser` to return `/etc/passwd`, called
`get_chrome_cookies` inside `try: ... except Exception: pass`, and asserted **nothing**. Its
premise was also not a threat model — an attacker who can patch `expanduser` does not need a
cookie bug — and the module has no path validation to pin.

The real property is narrower and checkable: both paths come from platform literals, and the
caller's `domain_name` reaches only a parameterized `LIKE` query. So **no caller input should
appear in any path opened**. Observed at the Local State `open`, which every platform branch
reaches unconditionally, patched in the module's own namespace rather than `builtins`.

**Demonstrated, not asserted.** Against a product mutated to concatenate `domain_name` into
the path:

| | result |
|---|---|
| original test | **passes** |
| replacement | **fails, all 4 parametrizations** |

## Five tests that logged their verdicts instead of asserting them

`Tests/RAG/test_rag_ui_integration.py` recorded results as `logger.success("✅ …")` /
`logger.error("❌ …")` and ended in `return True`, which pytest ignores — so all five passed on
the very paths their own `❌` strings call failures. They were **running**, not skipped.

Converting the claims into assertions exposed three things:

1. One test looked for the notification `"RAG search error"`. **The product has never emitted
   that** — it says `"RAG search failed"`. Invisible because the check was an `if` around a log
   line. Fixed.
2. An empty search returns the sentence `"No relevant context found."` — not `None`, not empty
   framing. Pinned, since that is what a user sees.
3. Two now fail for a real reason: the hand-built app double predates the Console Library
   policy work, so `_authorize_local_results_for_prompt` discards the candidate with
   `reason=not_currently_authorized`. Marked `xfail(strict=False)` against **TASK-21564** rather
   than deleted, so the coverage stays on the record instead of vanishing.

Separately, `test_ui_settings_reach_the_search_unchanged` kept its only assertions *inside* the
search double — a double that was never called checked nothing, and the test still passed. It
now asserts the invocation itself.

## Deletions, each justified

- **`Tests/Chat/test__zz_probe.py`** — a 30-iteration timing probe ending in `assert True`,
  committed under a `_zz_` name so it sorts last.
- **`Tests/UI/verify_command_palette.py`** — 268 lines of `test_*` functions that `print()` and
  return booleans. Verified **0 nodes collected** when the directory is collected; naming the
  file directly is what makes it look collected. The palette has **76 real collected tests**
  across four sibling files.
- **`test_performance_summary`** — printed "All performance tests completed successfully" and
  asserted nothing. A test whose behaviour was reporting success.

## One renamed and given a body

`test_core_imports_without_optional_deps` had every import deleted down to bare comments and
ended in `assert True`. It cannot make that claim where every optional group is installed, so
it is now `test_core_modules_import` and actually imports the four. Mutation-proven: pointing
one entry at a non-existent module reddens it.

## A number I corrected before acting on it

A naive AST scan put the value-returning-test population at **4,413**. That was wrong — it
descended into helper functions defined *inside* tests. Counting only each test's own scope
gives **28**, and most are fixtures misnamed `test_*` (`test_db`, `test_app`, `test_config`),
which pytest collects as tests *and* uses as fixtures. That is a separate defect and is not
touched here.

## Evidence

`Tests/Web_Scraping` + `Tests/RAG` + `Tests/RAG_Search` + `Tests/unit`: **1,438 passed, 1
failed, 22 skipped, 2 xfailed**. The single failure
(`test_fts5_match_forms_shared.py::test_plain_search_and_rag_answer_answer_the_same_inflection_miss`)
fails identically on a clean `dev` checkout — it is TASK-19642.19.1's known FTS5 parity gap.

Test-only; no product code changes.

---
## Summary by cubic
Repairs or removes tests that could never fail so they now assert real behavior, adds a cookie path safety property test, and tightens RAG UI and eval orchestrator tests to reflect real outcomes.

- Test-only; no product code.
- Security: replaces `test_cookie_db_path_validation` with `test_caller_input_never_reaches_the_cookie_database_path`. Patches `open` in `tldw_chatbook.Web_Scraping.cookie_scraping.cookie_cloner`, parametrizes hostile inputs, and asserts the `domain_name` never appears in any opened path.
- Evals: fixes `test_run_evaluation_tracking` to assert behavior. Expects `DatasetLoadingError`, verifies `create_run` and `update_run_status` are called, and asserts `_active_tasks` is empty after failure.
- RAG UI integration: converts five logged “✅/❌” checks into assertions in `Tests/RAG/test_rag_ui_integration.py`.
  - Fixes notification string to match product (“RAG search failed”).
  - Pins empty-result message (“No relevant context found.”).
  - Asserts the search double is invoked when validating settings.
  - Marks two tests `xfail(strict=True)` due to the Console Library prompt-authority gate not modeled by `MockApp` (tracked in TASK-21564). Removes an orphaned `@pytest.mark.asyncio`.
- Deletions: `Tests/Chat/test__zz_probe.py` (timing probe with `assert True`), `Tests/UI/verify_command_palette.py` (not collected; 76 real palette tests exist elsewhere), and `test_performance_summary` (printed success, asserted nothing).
- Core imports: renames `test_core_imports_without_optional_deps` to `test_core_modules_import` and actually imports `tldw_chatbook.config`, `tldw_chatbook.DB.ChaChaNotes_DB`, `tldw_chatbook.Chat.Chat_Functions`, `tldw_chatbook.Utils.optional_deps`.
- CI expectations: adds 2 `xfail(strict=True)`; overall suite parity with `dev` preserved. Implements TASK-21563 and opens TASK-21564.

<sup>Written for commit 3737b79b9a9e6aeba31c5cec9b6d08caff7d5d79. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2054?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
### Added: one more test that could not fail

`Tests/Evals/test_eval_orchestrator.py::test_run_evaluation_tracking` called `run_evaluation`
inside `except Exception: pass` and ended in two comments:

```python
# Check if tracking was attempted
# Note: In real implementation, task would be added to _active_tasks
```

They say outright that the check was never written, while the name promises the orchestrator
tracks active tasks.

I probed the real behaviour rather than guessing at assertions: the call raises
`DatasetLoadingError`, `create_run` and `update_run_status` are both called, and
`_active_tasks` is empty afterwards. So "tracking" here means the run is recorded, its failure
is recorded, and the task does not leak — all three now asserted.

The failure is captured with `pytest.raises` rather than swallowed, because `except Exception:
pass` would equally absorb a broken mock or an unrelated crash and the test would pass having
exercised nothing.

**Mutation-proven** at the cleanup site (`eval_orchestrator.py:753`). With cleanup disabled so
the task leaks:

| | result |
|---|---|
| original test | **passes** |
| rewrite | **fails** |

### Scope note

This was the **last** one. An AST census scoped to each test's own body (not descending into
nested helpers) found 17 bare `except …: pass` handlers in test bodies suite-wide; 15 sit in
tests that do assert, so the swallow is deliberate tolerance, and the 16th is this file's
sibling already fixed in this PR.

`Tests/Evals/test_eval_orchestrator.py`: 30 passed.

PR #2058, opened 2026-08-24
the required gate still fails to report on ~half of PRs (TASK-21967)

rmusser01:
Backlog entry only — no code. Filing rather than fixing, deliberately.

## The measurement

TASK-21250 stopped the required gate cancelling its own pull-request runs, and recorded a
**23% success rate** before that change. Measured now, over the most recent 100 runs of that
workflow:

| conclusion | all events | pull_request only |
|---|---|---|
| cancelled | **50** | **39** |
| success | 34 | 29 |
| failure | 7 | 6 |
| in flight | 9 | 8 |

So it improved and did not resolve. Roughly **half of PRs still sit `BLOCKED` with no verdict
on their only required check** until someone re-runs it by hand. Three test-health PRs hit
this in a row while being merged today — #2031, #2054 and now #2056 each needed a manual
`gh run rerun` before they could merge.

## The residual cause is the one TASK-21250 already named

Its own comment says it: for a `pull_request` event the run is tied to `refs/pull/N/merge`,
which GitHub recreates every time the base branch moves. On a repo absorbing 23–50 merges/day
with queue times of 4–59 minutes against ~33 s of work, that reference is frequently recreated
before a queued run gets to start. **Not cancelling our own runs does not help when the
reference they were created for no longer exists.**

There is a second half worth noting: `push:` is restricted to `[dev, main]`, so a topic branch
never gets a run tied to a stable ref. The volatile one is the only run it has.

## Why I did not just change it

The obvious change — watch pushes on all branches, so each head SHA also gets a run on
`refs/heads/<branch>` — rests on a platform behaviour I could not verify without modifying
the gate itself: whether a result recorded against the branch ref satisfies the requirement
for a commit that already carries a same-named result from the merge ref. If it does not, the
change adds load and fixes nothing.

This is the check every merge in the repository depends on, and TASK-21250 is recent work on
the same mechanism. Getting it wrong blocks everyone rather than one branch, so the task
carries the acceptance criteria for establishing that behaviour by observation first.

---
## Summary by cubic
Documents `TASK-21967` with a corrected diagnosis of required-gate cancellations; no code or workflow changes. This matters because many PRs still block without a required-gate verdict and CI spends compute on runs that get canceled in bursts.

- Updates measurements and cause: roughly half of recent runs lack verdicts; 95% of cancellations occur in bursts that sweep all workflows at once (e.g., 46 runs canceled within 24 seconds), not from volatile merge refs.
- Confirms proposed fix is ineffective: push-triggered runs on stable branch refs are canceled in the same bursts; watching pushes on all branches would not help.
- Notes the main test workflow still cancels non-`main` refs, wasting compute; observed one shard finish while five were canceled after 55–86 minutes each.
- Captures acceptance criteria: measure verdict rates pre/post, confirm whether branch-ref results satisfy the required check, assess queue impact, and decide the main workflow’s cancellation rule based on evidence.
- Removes `TASK-21968` from this PR; it moved to the branch implementing its fix.

<sup>Written for commit 13aaaa1f3f7dc1bd3dde07285c18cc827c24e7d6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2058?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
### The same rule, unfixed, in the workflow where it costs the most

Found after filing. TASK-21250 changed `derived-artifacts.yml` so pull-request runs are not
cancelled. `test.yml` still carries the original form:

| workflow | `cancel-in-progress` | PR runs cancelled? |
|---|---|---|
| `derived-artifacts.yml` | `github.event_name == 'push' && github.ref != 'refs/heads/main'` | no |
| **`test.yml`** | `github.ref != 'refs/heads/main'` | **yes** |

Since GitHub recreates `refs/pull/N/merge` every time the base moves, a PR's test runs are
cancelled repeatedly without anyone touching the branch — the identical mechanism TASK-21250
documented for the small gate.

**This was invisible until #2033.** While the core job could not finish inside its budget,
nothing was lost by cancelling it. Sharding made the slices finishable, and the waste is now
measurable. On run `32696030401`:

| slice | duration | outcome |
|---|---|---|
| shard 5/6 | 32 min | **completed** (60 failed, 7,200 passed) |
| shard 0/6 | 86 min | cancelled |
| shard 3/6 | 84 min | cancelled |
| shard 4/6 | 57 min | cancelled |
| shard 1/6 | 55 min | cancelled |
| shard 2/6 | 83 min | cancelled |

All five were killed well inside their 120-minute budget, and none produced a verdict. That
is several hours of runner time per run, spent and discarded, on the pool that is already the
binding constraint.

The fix has a precedent in this repo — the expression `derived-artifacts.yml` now uses. What
differs is cost: that gate is ~90 s, so redundant runs were cheap; a full test run is not, and
not cancelling means a rapidly-updated PR queues its runs rather than replacing them. Worth
deciding rather than assuming. But the present state spends the compute *and* gets nothing,
which is strictly the worst option available.

Added to the task as AC#5.

PR #2061, opened 2026-08-24
Vendor the tiktoken tables so tokenizing needs no network (TASK-21968)

rmusser01:
The largest single source of errors in the core test job — **1,156 blocked download attempts
in one shard**, and until now not reproducible on a laptop.

## Why nobody could see it

tiktoken downloads its tables on first use and caches them outside anything the sandbox
controls: warm on any machine that has run the suite, cold on every CI run.
`Utils/token_counter` wraps the lookup in a broad `except` and returns `None`, so the refusal
is swallowed. The *only* reason it surfaces is the egress guard recording the attempt — and
the test then fails at teardown pointing at a network address rather than at tokenizing.

**The reason it never reproduced here: `tiktoken` was not installed in the working venv.** It
is declared hard-required in `pyproject.toml` and CI installs it, but locally the import
failed, the `except` returned `None`, and no socket was ever opened. Installing it reproduced
the CI failure immediately — `Tests/Chunking` went to 14 errors with 32 recorded attempts.

That reproduction is what made this tractable, and it is why the earlier attribution went
wrong twice: a second library in the same logs prints retry lines while this one does not.

## What is actually requested — not the fallback everyone assumes

The chunking engine's default tokenizer is **`gpt2`**, which uses the two-file data-gym format
(`vocab.bpe` + `encoder.json`), not a single `.tiktoken` table. Vendoring `cl100k_base` alone
left **33 attempts outstanding**. Adding `gpt2` took it to zero.

`p50k_base`, `r50k_base` and `o200k_base` were fetched during investigation and **removed
again** — nothing asked for them, and `o200k_base` alone is 3.5 MB. Final size **3.0 MB**,
three files.

## Evidence

`Tests/Chunking`, same command, vendored cache absent then present:

| | before | after |
|---|---|---|
| errors | **14** | **0** |
| failed | 2 | 1 |
| passed | 603 | 604 |
| recorded download attempts | **32** | **0** |

Widened to `Chunking` + `Subscriptions` + `RAG_Search` + `Utils`: **2,756 passed, 0
attempts**, and nothing requested beyond the three vendored. Both residual failures reproduce
on dev with only a workflow change applied (`test_fts5_match_forms_shared` is
TASK-19642.19.1; `test_chunking_interop_v7`'s `TestNoIsSystemAnywhere` likewise).

## The guard, and why it recomputes hashes

Entry filenames are `sha1(<download url>)` — tiktoken's own key, not a name we chose. A
renamed file reads to tiktoken *exactly* like a missing one and would silently restore
downloading, so `Tests/test_tiktoken_vendored_cache.py` recomputes them.

Mutation-proven twice — removing an entry, and renaming one. In both cases the **behavioural**
test fails too, because the encoding then gets fetched and the egress guard records it. That
assertion is what makes this evidence rather than a file listing.

## Scope

Only `TIKTOKEN_CACHE_DIR` is set, and only from the test conftest. Production downloads and
caches as before, which is correct for a real user.

---
## Summary by cubic
Vendors `tiktoken` encoding tables so tests tokenize without network access and CI egress attempts drop to zero (TASK-21968). Previously `tiktoken` downloaded on first use and hid failures; tests now read from a vendored cache and fail loudly if `tiktoken` is missing, while production behavior is unchanged.

- Tests set `TIKTOKEN_CACHE_DIR` to `Tests/fixtures/tiktoken_cache`.
- Vendored: `gpt2` (encoder.json + vocab.bpe), `cl100k_base`, and `o200k_base`; excluded: `p50k_base`, `r50k_base`.
- Adds `Tests/test_tiktoken_vendored_cache.py` to recompute `sha1(url)` cache keys and assert encodings load under the egress guard.
- The guard imports `tiktoken` directly and fails the test if the hard dependency is absent.
- If a test needs another encoding, add it under `Tests/fixtures/tiktoken_cache/` instead of allowing network access.

<sup>Written for commit de126e0389bd424690991e00075fe699f7517113. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2061?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
### Merge note: this conflicts with the sibling egress PR

#2046 (HuggingFace offline) and #2061 (vendored tiktoken tables) both add to the same
pre-import bootstrap block in `Tests/conftest.py`, in two places:

- the `_SANDBOXED_ENV_NAMES` tuple (`HF_HUB_OFFLINE` vs `TIKTOKEN_CACHE_DIR`)
- the block of env writes just after the keyring pin

I hit this cherry-picking both onto a measurement branch. **Both conflicts are purely
additive — the correct resolution is to keep both sides**, in either order. Verified: with
both applied, `Tests/test_huggingface_offline.py` + `Tests/test_tiktoken_vendored_cache.py`
give 8 passed.

Whichever merges first, the second will need that trivial resolution rather than a choice.

rmusser01:
Rebased onto `dev` now that #2046 has merged. The conflict flagged above was in `Tests/conftest.py` as predicted — both hunks additive, resolved by keeping both sides.

Verified after the rebase: `Tests/Chunking` + both guards = **612 passed, 1 failed**, and **zero** `openaipublic` attempts. The single failure is `test_chunking_interop_v7.py::TestNoIsSystemAnywhere::test_no_is_system_in_production_tree`, which fails identically on a clean `dev` checkout.

rmusser01:
### Correction: `o200k_base` is needed, and dropping it was a sampling error

The PR body says `o200k_base` was fetched and removed because nothing requested it. **That was
measured too narrowly** — over `Chunking`, `Subscriptions`, `RAG_Search` and `Utils`, none of
which name a gpt-4o-class model.

A 95-file Console and Library run requested it **53 times**. `encoding_for_model` returns
`o200k_base` for those model names, and it is the Console suites that use them. The requester
is whichever suite happens to *name a model*, not whichever suite looks like it tokenizes.

It is back. Four files, **6.5 MB**. `p50k_base` and `r50k_base` stay absent — nothing has
requested them across either sample.

### What it was costing, on the largest red file in the UI suite

| `Tests/UI/test_console_native_chat_flow.py` | without `o200k_base` | with |
|---|---|---|
| failed | 26 | **2** |
| errors | 13 | **0** |
| recorded attempts | 21 | **0** |
| runtime | 7m34s | **4m41s** |

The three minutes are retries — each blocked fetch is retried before it gives up, so the
egress guard was costing wall-clock as well as green.

That one file was **39 of the 277** remaining UI failures I had measured. This PR therefore
does considerably more than the core-half numbers in the body suggested.

The guard and the README now cover `o200k_base`, including why it was dropped and what that
cost, so the next person sampling for a fourth encoding samples wider.

PR #2078, opened 2026-08-25
fix(console): stop forking the persisted conversation tree; restore in-flight wakes

rmusser01:
## Summary

Two production regressions on the Console send path, plus the stale-test-double
class that was hiding them. Everything here was bisected to a specific commit
and A/B'd node-by-node against clean `dev`; no count in this PR is an estimate.

### TASK-22060 — every Console turn after the first forked the persisted tree

An ordinary **second send** — no navigation, no wake, no Library gate — persisted
its user row as a fresh DB root instead of extending the conversation:

```
before: [user 413806fe (root), assistant 6fd8bf65 -> 413806fe]
new:    [user 44b1275e (root!), assistant ac6eef14 -> 44b1275e]
```

Reproduced on clean `dev` `8ef5bf12e` with zero product edits. The transcript
still rendered correctly, which is why it went unnoticed: the fork exists only
in the durable tree that reload, rewind, branching, active-leaf resume and
trajectory export all walk.

`a26cdafd8` added the durable dispatch checkpoint, which writes its rows with
raw SQL and threads `acceptance.parent_message_id`. The controller computed
that from `echoed_user.parent_message_id` — but that field holds the *persisted*
parent id, and `_persist_new_message` is its only writer. The echo is appended
with `persist=False`, so the field was **always** `None` and the guard above it
never once ran.

New `ConsoleChatStore.durable_parent_for_message` resolves the nearest persisted
ancestor from the store's own tree bookkeeping — the identical walk
`_persist_new_message` performs — so both persistence paths now agree by
construction rather than by coincidence.

### TASK-22061 — navigating away killed an in-flight agent wake

`leave_console` documents an owner ruling that an in-flight `AGENT_WAKE` turn is
**not** cancelled on nav-away, because "cancelling it would re-create the exact
'only completes if you stay' gap this arc exists to close", and it implements
that for its own cancel fan-out. But three newer gates re-imposed the refusal one
layer down by reading the per-visit `_shutdown_requested` flag that
`leave_console` itself sets.

App exit always sets `_disposed` first, so `_disposed` alone is the correct
app-exit signal — which is exactly what task-15860 moved the wake coordinator's
own gate onto. Among the tests this restores is one named
`test_a_visit_that_merely_ended_does_not_refuse_the_wake`.

### The test doubles production had outgrown

`a26cdafd8` made a typed `ConsoleResolvedDestination` mandatory on any ready
resolution, and made `commit_durable_turn` mandatory for non-ephemeral
MANUAL/QUEUED sends. Every double that handed a **real** controller a hand-built
resolution, and every rig that built a bare `ConsoleChatStore()`, silently
stopped exercising its subject.

Silently is the operative word: the refusals read as ordinary blocked sends, so
the symptoms surfaced as `"the wake never delivered"`, `"harness precondition:
the manual send must run"`, and `RuntimeError: coroutine raised StopIteration` —
naming neither the missing field nor the commit.

New `Tests/console_provider_doubles.py` derives the destination through the
**production classifier** (`resolve_console_destination`) rather than
hand-building one, so the doubles cannot drift from the rule again. It attaches
the field only when `ready` is true, which leaves the deliberately not-ready
doubles — and `_MissingDestinationFence`, which exists precisely to exercise the
missing-destination refusal — untouched.

## What is deliberately NOT changed

- **The durable-turn refusal itself.** Refusing with honest copy when nothing can
  be persisted is fail-closed, and `durable_turn` already exempts ephemeral
  sessions and wake turns. The rigs had drifted, not the product.
- **`test_console_automatic_library_preparation.py`.** Its fence exists to test
  the missing-destination refusal; "fixing" it would delete the test.
- **7 doubles** the migration could not handle mechanically. It reports rather
  than guesses, so those are left for manual review.

## TASK-22062 — a hanging test destroys the whole run (quarantined, not fixed)

Two tests in `test_console_local_citation_boundary.py` never complete — a single
MainThread parked in the asyncio selector with no application frames. Because
`pyproject.toml` sets `timeout_method = "thread"`, which cannot interrupt a
blocked thread, the 300s timeout **destroys the pytest process**: no JUnit is
written, nothing is named as failing, and under xdist the worker dies and takes
its queued tests with it as setup errors.

One is quarantined with a skip naming the task. A skip is not acceptable
long-term, but the alternative is strictly worse — left in place it deletes the
results of every other test sharing its process. Reproduces on clean `dev`.
Worth holding against the open question of why `core-tests` shards die: from the
outside, a shard killed by one of these is indistinguishable from one that hit
`timeout-minutes`.

## Verification

A/B against **clean `dev` `8b0180118`**, the exact base this branch now sits on,
with the same test set on both sides. `Tests/` on the baseline was verified
byte-identical to `dev` with `git diff --quiet` before the control ran.

| | clean `dev` `8b0180118` | this branch |
|---|---|---|
| collected | 1917 | 1917 (delta **+0**) |
| failures | 176 (+1 error) | **52** (0 errors) |
| fixed | — | **125** |
| newly broken | — | **0** |

Attribution note, because it is easy to get wrong: rebasing onto current `dev`
made 22 tests go red relative to the pre-rebase branch, ten of them in
`test_console_cost_chip_screen` which this branch had taken to zero. None of
them are this branch's doing — they are red on clean `dev` too. `dev` itself
moved from 167 to 176 failures across the 160 commits in that span. Comparing
my-branch-then against my-branch-now changed two variables at once; the table
above changes one.

Per-file, measured individually:

| File | before | after |
|---|---|---|
| `test_console_agent_swap` | 39 | 4 |
| `test_console_fleet_wake` | 17 | 0 |
| `test_console_provider_continuation` | 13 | 1 |
| `test_console_headless_wake_invariants` | 12 | 2 |
| `test_console_cost_chip_screen` | 10 | 0 |
| `test_trajectory_capture` | 10 | 3 |
| `test_console_edit_resend` | 5 | 0 |
| `test_console_regenerate_branching` | 5 | 0 |
| `test_console_variant_stream` | 4 | 0 |
| `test_console_store_continuity` | 3 | 0 |

Two guards were added rather than assumed: `Tests/UI/test_console_persisted_chain.py`
covers both fork shapes (second send in one visit; send after navigating away),
and both fail on the pre-fix product code.

Also repairs a test that **could not go red**: the continuity file asserted
`gateway.payloads` was truthy after the nav-away, but the seeding send has
already populated it. That masked the real failure and reported it one step
later as a missing ledger stamp. It now measures growth against a snapshot.

### DIFF
--- a/Tests/Chat/test_console_agent_project_instructions.py
+++ b/Tests/Chat/test_console_agent_project_instructions.py
@@ -49,6 +49,7 @@ from tldw_chatbook.DB.AgentRuns_DB import AgentRunsDB
 from tldw_chatbook.MCP.permission_store import EffectiveToolState
 from tldw_chatbook.Workspaces.models import WorkspaceRuntimeBinding
 from Tests.console_provider_doubles import with_destination
+from Tests.console_provider_doubles import persisted_console_store
 
 
 SENTINEL = "AGENTS_AUTOMATIC_CHANNEL_SENTINEL_71f584"
@@ -884,7 +885,7 @@ async def test_controller_notice_uses_owning_session_and_drift_cancels_bridge_se
     (tmp_path / "AGENTS.md").write_text(SENTINEL)
     binding = _binding(tmp_path)
     registry = _BindingRegistry([binding])
-    store = ConsoleChatStore()
+    store = persisted_console_store()
     session = store.create_session(workspace_id="w1")
     notices = []
     owning_loop_calls = []
@@ -966,7 +967,7 @@ async def test_controller_notice_uses_owning_session_and_drift_cancels_bridge_se
 async def test_folderless_session_skips_optional_project_instructions_and_runs(
     tmp_path,
 ):
-    store = ConsoleChatStore()
+    store = persisted_console_store()
     session = store.create_session(workspace_id="workspace-default")
     bridge_calls = []
     setup_calls = []
@@ -1053,7 +1054,7 @@ def test_disabled_session_does_not_consult_registry(tmp_path):
 async def test_project_instruction_disable_terminalizes_and_allows_retry():
     """Disabling unavailable instructions must not strand the Console run."""
 
-    store = ConsoleChatStore()
+    store = persisted_console_store()
     session = store.create_session(workspace_id="w1")
     store.set_session_project_instruction_state(
         session.id,
--- a/Tests/Chat/test_console_agent_project_instructions.py
+++ b/Tests/Chat/test_console_agent_project_instructions.py
@@ -48,6 +48,7 @@ from tldw_chatbook.Chat.console_project_instructions import (
 from tldw_chatbook.DB.AgentRuns_DB import AgentRunsDB
 from tldw_chatbook.MCP.permission_store import EffectiveToolState
 from tldw_chatbook.Workspaces.models import WorkspaceRuntimeBinding
+from Tests.console_provider_doubles import with_destination
 
 
 SENTINEL = "AGENTS_AUTOMATIC_CHANNEL_SENTINEL_71f584"
@@ -926,7 +927,7 @@ async def test_controller_notice_uses_owning_session_and_drift_cancels_bridge_se
 
     class Gateway:
         async def resolve_for_send(self, _selection):
-            return ConsoleProviderResolution(
+            return with_destination(ConsoleProviderResolution(
                 provider="OpenAI",
                 base_url="https://user:password@api.example/v1?secret=yes",
                 model="test-model",
@@ -934,7 +935,7 @@ async def test_controller_notice_uses_owning_session_and_drift_cancels_bridge_se
                 readiness_key="openai",
                 execution_key="openai",
                 max_tokens=128,
-            )
+            ))
 
     gateway = Gateway()
     controller = ConsoleChatController(
@@ -977,7 +978,7 @@ async def test_folderless_session_skips_optional_project_instructions_and_runs(
 
     class Gateway:
         async def resolve_for_send(self, _selection):
-            return ConsoleProviderResolution(
+            return with_destination(ConsoleProviderResolution(
                 provider="DeepSeek",
                 base_url="https://api.deepseek.com",
                 model="deepseek-chat",
@@ -985,7 +986,7 @@ async def test_folderless_session_skips_optional_project_instructions_and_runs(
                 readiness_key="deepseek",
                 execution_key="deepseek",
                 max_tokens=128,
-            )
+            ))
 
     async def select_binding(*args):
         setup_calls.append(args)
@@ -1068,7 +1069,7 @@ async def test_project_instruction_disable_terminalizes_and_allows_retry():
     class Gateway:
         async def resolve_for_send(self, _selection):
             provider_calls.append(True)
-            return ConsoleProviderResolution(
+            return with_destination(ConsoleProviderResolution(
                 provider="OpenAI",
                 base_url="http://127.0.0.1:18991/v1",
                 model="gpt-4o-mini",
@@ -1076,7 +1077,7 @@ async def test_project_instruction_disable_terminalizes_and_allows_retry():
                 readiness_key="openai",
                 execution_key="openai",
                 max_tokens=128,
-            )
+            ))
 
     class Bridge:
         def run_reply(self, **_kwargs):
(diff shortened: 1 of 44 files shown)


rmusser01:
### Stacked on #2061

Base is `fix/task-21968-vendor-tiktoken-encoding` (#2061), not `dev` — that PR
vendors the tiktoken tables this branch's test runs depend on. Retarget to `dev`
once #2061 lands; the diff here is Console-only and does not touch the tiktoken
work.

### Note on the one "newly broken" test

`test_console_rewind_restore::test_summarize_choice_guards_against_changed_active_session`
appears red only in the parallel run. It passes 3/3 in isolation and its file is
untouched by this branch, so it is an ordering/contention flake rather than a
regression. Recorded here rather than quietly dropped from the count.

rmusser01:
### Both Qodo findings addressed in `78e98d781`

Both were consequences of removing the dead guard earlier in this branch, and
both are real.

**1. Queue validation bypass too broad** — correct, and the sharper scoping is
right. My early return covered *any* non-MANUAL origin arriving with no chain
and no `entry_id`, which would have swallowed a QUEUED acceptance that lost its
entry id. That is a genuine accounting bug and must keep failing loudly. Now
gated on `origin is ConsoleSubmissionOrigin.AGENT_WAKE`, which is the only case
the exemption was ever meant for, with the reasoning recorded in-comment so the
next reader does not re-widen it.

**2. Unused durable-parent flag** — also correct. `has_native_parent` existed
only to drive the `"Couldn't save the prepared turn"` pause, and that pause
turned out to be wrong: `insert_with_messages` validates a parent only when one
is given, and `_nearest_persisted_ancestor_id` documents `None` as the "true
persisted root" answer. Leaving the boolean in the signature implied an
enforcement that no longer happens. `durable_parent_for_message` now returns
`str | None`.

Verified across the 7 affected files: **90 tests, 2 failures**, both pre-existing
on the merge-base `ecf331be8` (`test_a_disposed_runtime_refuses_the_wake_and_loses_nothing`
and `test_a_wake_racing_app_exit_leaves_consistent_durable_state`, unrelated
causes).

rmusser01:
### Correction: the "flake" was not a flake

I reported `test_summarize_choice_guards_against_changed_active_session` above as
an ordering/contention flake because it passed 3/3 in isolation. That was wrong.

It had a real cause: `_build_test_app` boots with `chachanotes_db = None`, so the
Console store gets `persistence=None`, and a non-ephemeral MANUAL send fails
closed — the refusal returns a bare `ConsoleSubmitResult` rather than going
through `_block`, writing no system row and raising no toast. That made the test
order-dependent rather than deterministically red, which is why isolation looked
clean. `0cf081d73` attaches the DB and it passes for a reason.

"Passes in isolation" shows a failure isn't deterministic; it does not show there
is no cause. Recording the correction rather than quietly dropping it.

### Additional batch in `0cf081d73`

Same root cause across 8 files, using `attach_chachanotes_db` — the helper
TASK-21590 already added to `Tests/UI/app_factory.py`, whose docstring names this
failure exactly ("26 mounted send tests kept pressing Send and asserting against
a transcript production had silently refused to write"). It had been applied to
one file.

Measured against merge-base `ecf331be8` across those 8 files: **19 failures → 3,
13 fixed, 0 broken.** 80 files match the pattern but most never make a
non-ephemeral MANUAL send, so this was applied only where failures were
confirmed rather than blanket-patched.

rmusser01:
### Rebased onto `dev` (#2061 has landed); base retargeted

#2061 merged as `95305e6ba`, so the tiktoken commits are now in `dev` and were
dropped from this branch during the rebase. Base retargeted from
`fix/task-21968-vendor-tiktoken-encoding` to `dev`. Eleven commits replayed
cleanly; `./scripts/preflight.sh` passes all five derived-artifact checks.

### Attribution note on the rebase

Rebasing made 22 tests go red **relative to the pre-rebase branch**, ten of them
in `test_console_cost_chip_screen`, which this branch had taken 10 → 0. That
looked alarming, so I did not push on it.

They are not this branch's doing. Measured against **clean `dev` `8b0180118`**
with an identical test set:

| | clean `dev` | this branch |
|---|---|---|
| collected | 1917 | 1917 (delta **+0**) |
| failures | 176 (+1 error) | **52** (0 errors) |
| newly broken | — | **0** |
| fixed | — | **125** |

`dev` itself went from 167 to 176 failures across those 160 commits. The
earlier comparison moved two variables at once — my base *and* dev's contents —
so it could not attribute anything. This one moves one.

Worth flagging separately for whoever owns it: `test_console_cost_chip_screen`
is red on clean `dev` and was not before, and dev's #2088 ("restore the ADR-046
queue and make a failed DB open visible on Send") touches that path.

PR #2101, opened 2026-08-25
First-run setup wizard: full UX/UAT remediation (TASK-21139…21149)

rmusser01:
## Summary

Remediates every finding from the four-persona UX/UAT review of the first-run setup wizard (`output/ux-review-setup-wizard/findings.md`, which carries the full findings list, capture evidence, and a per-commit remediation table). Eleven backlog tasks, one commit each, in dependency order.

### The two S1s

- **Full-track soft-lock (F-1, TASK-21139)** — entering the Speech step on a cold first walk killed *all* keyboard input until the process was killed. Root cause: the step's first `on_show` recompose detached the widget `show_step` had just focused, and Textual 8.2.8 keeps dispatching keys into the detached node. `SetupStep.refresh()` now heals orphaned focus after every recompose; regression test pins the mechanism (focus attached **and** a real `ctrl+b` still navigates). Two other UAT mysteries (dead arrows after Back, "silent" Detect/Test) turned out to be collateral of this same bug.
- **Summary lied about broken keys (S-1/M-2, TASK-21143)** — a key that failed authentication still produced "✓ Provider / Start chatting". Probe outcomes now reach the tracker (amber `!`), a "Continue anyway?" gate on the Model step, a returning-to-Provider notice, and a Summary that says "✗ Provider — saved, but the key failed an authentication check" with **Review provider setup** primary. Verified live end-to-end against a real 401.

### The rest, by feel

- **Keyboard (21142)**: Enter advances; arrows select (WAI-ARIA radio semantics); Tab reaches Next before any abandon action (Textual's focus order is *visual* (y,x), not DOM — footer adopts the Windows-wizard convention; lessons entry added).
- **Error visibility (21140)**: step errors render on a pinned strip above the nav instead of below the fold; phantom "Skip this step" copy removed; Notes' reassurance no longer styled as an error.
- **Password dialog (21141)**: Escape cancels; the min-8-chars error no longer buries Cancel/Submit (root cause: app-wide `.error-message` blanket rule); requirements + forgotten-password consequence stated up front; show-password toggle.
- **Local probes (21144)**: status relocated adjacent to the controls (was below the fold); buttons labeled by outcome ("Find local servers" / "Test connection").
- **First chat (21145)**: a fresh profile's first message is never intercepted by the project-instructions folder modal; provider validation carries a 30s hard bound; "Send blocked — …" is itself a link that opens setup (`app.run_setup_wizard`).
- **Consent (21146)**: the model-lists consent lives on the Summary (default off, asked once) — no post-setup modal ambush; `SetupCheckbox` fixes the color-only checked glyph.
- **Startup (21147)**: cold starts print zero DEBUG/INFO to the terminal (`TLDW_VERBOSE_STARTUP=1` restores); env-key installs get a one-time "Found OPENAI_API_KEY — you're ready" notice.
- **Layout (21148)**: stable step totals (Protect always present, with a nothing-to-protect state); tracker titles stacked under two-digit-capable number boxes (all 11 survive at 140 cols); density pass — Provider shows title-through-key-panel at once at 140×40; Voice leads with "Test and Hear", plumbing under Advanced; sub-100×30 hint; summary path middle-truncates.
- **Copy (21149)**: value prop + honest time estimates on Welcome; key helper leads with paste-it-here + per-provider get-a-key URLs (env var demoted to an aside); exact pip command for RAG; Summary names the provider; splash cards humanized with "Show all cards…".

## Verification

- Wizard suites (`Tests/Wizards/` + live-contract + visual-audit): **882–883 passing** at every commit; ~30 pins updated to the new contracts, each read and updated deliberately (several caught real regressions in my first cuts — mount-time radio auto-select, a docked hint row shifting geometry, stale tracker height caps).
- Live tmux verification against isolated profiles for every user-visible change, including the previously-impossible cold full-track walk and the real-401 trust chain.
- Console-side changes baselined by stash-and-diff: failure sets byte-identical with and without this branch. **Pre-existing on `dev`**: ~68 failures + 9 errors in `Tests/Chat` (generation/image/video suites) and one order-dependent flake (`test_rerun_over_settings_review_settings_returns_to_settings`) — untouched here, worth a separate look.
- User guide (`Docs/User_Guide/First_Run_Setup.md`, `console.md`) updated per screen change; two lessons entries added (`backlog/docs/lessons-textual.md`).

Deferred by design (documented in tasks): V-3 (whether Voice belongs on the quick track — usage data should decide) and N-4's Settings-IA relocation of the rerun entry.

https://claude.ai/code/session_01KBSWqTEWDX34iaSNwemF3o


rmusser01:
**Code review completed** (independent reviewer over the full range `3ed7d94..cd2e610`): no Criticals; 1 Important, 6 Minors. Verdict: *ready to merge with fixes* — now applied in `fa4246623`:

- **Important — fixed**: the quiet-startup stderr sink passes `diagnose=False` explicitly (the repo's task-2119 credential-leak standard; the env-var default can't be relied on at that import position).
- **Fixed**: broken user-guide table (Voice row added, paragraph moved below); the M-2 "Continue anyway?" one-shot flag is consumed up front so a blocked re-entry can never leave a gate bypass armed; summary path truncation budget derives from step width; the returned-to-Provider "key was rejected" notice clears on key edit.
- **Deferred → TASK-21150**: consent allow-path refresh kick, show-all pressed-marker parity, bounding the three non-send-path `resolve_for_send` awaits, mechanical enforcement of the composer markup invariant.

Suites after fixes: **889 passed**.

https://claude.ai/code/session_01KBSWqTEWDX34iaSNwemF3o

rmusser01:
**Rebased onto latest `dev`** (504 commits) and **all review feedback addressed**:

**Qodo review** — all 8 findings fixed in `for the rebased head`:
- Finding 8 (bug, confirmed): humanized splash-card labels had broken draft-restore matching and rebuilt-list pressed state — buttons now press via their raw-id `_card_name` rider and restoration matches the rider (mirrors `_theme_name`); also closes TASK-21150 item (b).
- Findings 1–7 (rule violations): stdlib import block rejoined; `SetupStep.refresh` fully annotated; Google `Args:`/`Returns:` docstrings added across the flagged helpers.

**Rebase notes** (two upstream convergences worth knowing):
- Upstream centralized send-path resolution into `_capture_and_resolve_turn_execution_context`; the H-3 30-second bound now lives inside that single choke point, covering every caller including queued recovery — strictly better than the per-site bounds.
- Upstream independently implemented the H-2 no-interception fix (skip recovery when a session was never bound and nothing is eligible); their version is kept, and this branch's pure `project_recovery_should_skip_send_interception` predicate + tests remain as the documented contract.

**Preflight** (new DoD requirement): all derived-artifact checks green — CSS bundle regenerated from sources at every conflicted step; the four new persistent diagnostics reviewed (static messages/error-type names only) and recorded in the inventory; the soft-lock task renumbered **TASK-21139 → TASK-22281** per the owner rule (upstream's 2026-08-23 Windows-checkout task keeps the id), with provenance and all references updated.

**Post-rebase verification**: wizard trio + handoff + dialog + composer suites — 898 passed; the 6 remaining failures (4 visual-audit snapshots, 2 enter-hotkey) fail identically on pristine `origin/dev` (verified in a clean worktree) — pre-existing, not from this branch.

https://claude.ai/code/session_01KBSWqTEWDX34iaSNwemF3o

rmusser01:
**Two real regressions found and fixed** (pushed as `1e223b1db`, branch now at `7bc27c686` with latest `dev` merged).

Reviewing this PR's own CI test-results comment — 695 failures — against merged PR #2100's list (708 failures, same hour, same corpus, identical 98.2% pass rate) isolated two that pass on `dev` and failed only here:

1. **A fourth composer copy pin.** TASK-21145 appended a ` ›` action link to the setup-blocked reason; three pins were updated, but `Tests/UI/test_console_internals_decomposition.py` also pins that string and sat outside the suites being run locally.
2. **A repo performance rule.** `test_class_level_css_stays_within_the_allowlist` bars class-level `DEFAULT_CSS`: each declaration registers another stylesheet source against Textual's 64-entry parse cache, past which every first mount of an unseen class re-pays a ~150–450 ms cold parse for the session. `SetupWizardProgress.DEFAULT_CSS` (added in TASK-21148 so bare harnesses held the stacked tracker layout) tripped it → renamed to `BUNDLED_CSS` so `build_css.py` lifts it into the widget-defaults tier, per the check's own option 1.

The `BUNDLED_CSS` move then correctly re-broke three bare-harness tests that assert real painted geometry or drive `pilot.click` coordinates. Rather than restore the barred declaration, those tests now use the file's existing `_StyledHostApp` — asserting real geometry requires the real stylesheet.

**Verification:** 1063 passed / 4 failed across the wizard, live-contract, console-internals, CSS-consolidation, handoff and password-dialog suites; the 4 (`test_product_maturity_phase1_visual_audit.py`) fail identically on pristine `dev` (checked in a clean worktree), so this branch introduces **zero regressions**. `preflight.sh` green; the diagnostic-inventory conflict from the `dev` merge was regenerated and re-reviewed (still only the four known content-free calls).

https://claude.ai/code/session_01KBSWqTEWDX34iaSNwemF3o

PR #2102, opened 2026-08-26
fix(ci,console): stop cancelling PR test runs; report hangs instead of killing them

rmusser01:
## Summary

Three findings from the Console test burn-down, each fixed with its own evidence.
Two of them correct diagnoses I had recorded wrongly.

## TASK-22250 — `Tests` has produced no verdict since 2026-06-26

`cancel-in-progress` applied to `pull_request` events, so every push to a branch
killed the batch already running for that branch's PR. This is the repo's slowest
workflow — 12 UI shards plus 6 core shards at ~35 min each — so an actively-worked
PR never finished one.

Measured on #2078: a single push cancelled **20 checks at once**, and each batch's
cancellation timestamp matched the **next** batch's creation timestamp. That is
supersession, not an external agent.

**The exception proves it.** `derived-artifacts.yml` already scopes cancellation to
`push` only, and it is the one required check that reliably survives to report.
`test.yml` now adopts the same rule. Push runs may still be superseded — pushing
again genuinely obsoletes the previous commit's run — and `main` is never cancelled.

> **Correction.** I previously argued this was *not* `cancel-in-progress`, on the
> grounds that a push-triggered run on a stable ref was swept too. That reasoning
> was wrong: a later push to the same ref supersedes the earlier push run, same
> concurrency group. This fix sat unmade for a cycle because of it.

## TASK-22062 — a hanging test destroyed the whole run

`timeout_method = "thread"` was set explicitly, with a comment claiming thread is
the library default. It is not — pytest-timeout uses `signal` wherever SIGALRM
exists. It matters because `thread` **cannot interrupt a blocked thread**: when the
timeout fires the pytest process dies, so no JUnit is written, nothing is named as
failing, and under xdist the worker takes its queued tests with it as setup errors.

Measured on a real hang:

| method | JUnit written | outcome |
|---|---|---|
| `thread` (was set) | **no** — process destroyed | results silently lost |
| default (`signal` on Unix) | yes | 1 failure, named |

Removing the override is the minimal fix: Unix gets `signal`, Windows still gets
`thread` automatically, so there is no platform risk. There are **661 unbounded
`await <event>.wait()` calls** in `Tests/`, each of which was a potential run-killer.

> **Correction.** The task recorded this as a coroutine awaiting something never
> resolved — implying a product deadlock. It was not. `_ReadyCitationPersistence`
> predates `commit_durable_turn`, so the send was refused and the test waited
> forever on an Event the refused turn never set: the same stale-double class as
> the rest of the burn-down, with a failure mode that hangs instead of asserting.

Its store now creates an **ephemeral** session — production's own carve-out
(`durable_turn = not session.ephemeral and origin in {MANUAL, QUEUED}`) — which
keeps the send on the `create_message` seam those doubles observe.
`test_console_local_citation_boundary` was previously unmeasurable; it now reports
95 tests and went from **83 failures to 69**, none newly broken.

## TASK-22251 — an unattributable failure

A failed durable commit discarded its exception, collapsing every way that
multi-step transaction can fail — conversation create, Library-policy write,
workspace validation, checkpoint insert — into one sentence. Two distinct causes
hid behind it, and identifying each required adding a temporary `print` **inside
production**.

Now logged by **type only**: an exception message here can name a conversation or
workspace. Two tests pin both halves, each mutation-proven — removing the log turns
one red, logging the full exception turns the other red.

## Verification

A/B against clean `dev`, same file set both sides:

| | clean `dev` | this branch |
|---|---|---|
| collected | 2217 | 2220 (+3 new tests) |
| failures | 51 | 52 |
| newly broken | — | **0** |

The one apparent regression,
`test_console_rewind_restore::test_summarize_choice_guards_against_changed_active_session`,
is order-dependent under xdist and filed as TASK-22300. The evidence is that the
**same unmodified baseline commit** produced both outcomes across two runs
(`newDevBase` FAIL, `findBase` pass); it passes 5/5 in isolation and 22/22 in its
own file. Noting explicitly because I called this one a flake once before on weaker
evidence and was wrong then — "passes in isolation" shows non-determinism, not
absence of cause.

`./scripts/preflight.sh` passes all five derived-artifact checks. The diagnostic
inventory was regenerated only after reading the single added row as the tool
instructs: one `logger.warning` interpolating `type(exc).__name__` — no user
content, secret, path or URL.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,20 @@ on:
 # so its history stays complete.
 concurrency:
   group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
-  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+  # TASK-22250: cancel superseded PUSH runs, never a pull_request run.
+  #
+  # This suite is the slowest in the repo (12 UI shards + 6 core shards, ~35
+  # min each). Cancelling pull_request runs on every push to the branch meant
+  # an actively-worked PR effectively never produced a verdict: measured on
+  # #2078, each push cancelled the previous batch wholesale -- 20 checks killed
+  # at once -- and the cancellation timestamp of each batch matched the
+  # creation timestamp of the next. That is why `Tests` has reported nothing
+  # since 2026-06-26 while short workflows report fine.
+  #
+  # The exception proves it: `derived-artifacts.yml` already scopes
+  # cancellation to `push` only, and it is the one required check that
+  # consistently survives to report. This adopts that same rule.
+  cancel-in-progress: ${{ github.event_name == 'push' && github.ref != 'refs/heads/main' }}
 
 permissions:
   contents: read
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,6 +32,15 @@ concurrency:
   # The exception proves it: `derived-artifacts.yml` already scopes
   # cancellation to `push` only, and it is the one required check that
   # consistently survives to report. This adopts that same rule.
+  #
+  # Scope, precisely: this protects a RUNNING pull_request batch. GitHub
+  # concurrency keeps one running plus one PENDING member per group, and a
+  # third arrival still replaces the pending one -- `cancel-in-progress`
+  # cannot prevent that. Only a per-run group (e.g. including `github.run_id`)
+  # would, and that would also stop superseding obsolete commits while
+  # multiplying queue pressure that is already the bottleneck here. Killing a
+  # QUEUED run costs nothing and yields the newer commit's verdict; killing a
+  # RUNNING batch throws away up to 35 minutes per shard for nothing.
   cancel-in-progress: ${{ github.event_name == 'push' && github.ref != 'refs/heads/main' }}
 
 permissions:
--- a/Docs/security/production-diagnostic-inventory.json
+++ b/Docs/security/production-diagnostic-inventory.json
@@ -243,8 +243,8 @@
       "reason": "high-risk Chat/provider/summarization/tool/MCP diagnostic owner"
     },
     {
-      "call_count": 51,
-      "diagnostic_digest": "4a06156c02dffecef8e0",
+      "call_count": 52,
+      "diagnostic_digest": "9470774591ce409dfb1a",
       "owner": "TASK-492",
       "path": "tldw_chatbook/Chat/console_chat_controller.py",
       "reason": "high-risk Chat/provider/summarization/tool/MCP diagnostic owner"
@@ -3968,7 +3968,7 @@
   "summary": {
     "owner_files": 537,
     "persistent_sink_files": 8,
-    "task_492_calls": 1240,
+    "task_492_calls": 1241,
     "task_494_calls": 7338
   }
 }
--- a/Tests/CI/test_github_actions_test_workflow.py
+++ b/Tests/CI/test_github_actions_test_workflow.py
@@ -465,8 +465,21 @@ def test_the_test_summary_job_can_actually_fail() -> None:
     )
 
 
-def test_a_pull_request_run_is_never_cancelled_by_a_later_push() -> None:
-    """TASK-22250: superseding PR runs is why `Tests` never reports.
+def test_an_in_flight_pull_request_run_is_not_cancelled_mid_run() -> None:
+    """TASK-22250: superseding RUNNING PR runs is why `Tests` never reports.
+
+    Scope, stated precisely because the first version of this test overclaimed:
+    this guarantees only that a PR run already EXECUTING is not killed
+    mid-flight. GitHub concurrency keeps one running plus one *pending* member
+    per group, and a third arrival still replaces the pending one -- so a PR run
+    can be cancelled while queued, and `cancel-in-progress` cannot prevent that.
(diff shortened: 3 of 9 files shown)


rmusser01:
### Finding 2 (pending PR runs still cancel) — correct, fixed in `40c7f2d91`

You're right, and the distinction matters. GitHub concurrency keeps one RUNNING
plus one PENDING member per group; with `cancel-in-progress` false a third
arrival still replaces the pending member. So a pull_request run **can** be
cancelled while queued, and `cancel-in-progress` cannot prevent it.

My test was named `..._is_never_cancelled_by_a_later_push` and its docstring
claimed exactly that. It asserted less than it claimed — which is the same
overclaiming defect this branch's sibling work has been removing from other
tests, so it should not have shipped here. Renamed to
`test_an_in_flight_pull_request_run_is_not_cancelled_mid_run`, with the scope
stated in both the docstring and the workflow comment.

I did **not** take the per-run-group option (`github.run_id`). It would deliver
the literal never-cancel guarantee, but it also stops superseding obsolete
commits and multiplies queue pressure that is already the binding constraint
here — runs on this very PR sat queued 20+ minutes before starting. The
trade-off worth having: killing a QUEUED run costs nothing and yields the newer
commit's verdict; killing a RUNNING batch throws away up to 35 minutes per shard
and produces no verdict at all. Only the second is the defect.

### Finding 1 (PascalCase) — declining, with evidence

`_CapturedLines`, `_ExplodingPersistence` and `_Gateway` **are** PascalCase; the
leading underscore marks them module-private, which is PEP 8 and the dominant
convention in this suite:

- **2,448** classes under `Tests/` carry a leading underscore (39% of all test classes)
- the immediate neighbours of these helpers follow it — `_RecordingGateway`,
  `_ReadyCitationPersistence`, `_SignalGateway`, `_StallingWakeGateway`

Renaming these three would make the new file inconsistent with every file around
it. Happy to reconsider if the rule is meant to apply to test helpers
specifically, but that would be a suite-wide change of 2,448 names, not three.

rmusser01:
### Empirical confirmation, from this PR's own CI

Pushing `40c7f2d9` superseded `2f10322d`. Because same-repo PRs run the workflow
file from the branch head, this PR exercised its own fix. Outcomes on the
superseded head — same push, same branch, same instant, the concurrency rule the
only variable:

| workflow | `cancel-in-progress` rule | outcome |
|---|---|---|
| **Tests** | push-only *(this PR's change)* | **queued — survived** |
| **Derived Artifacts** | push-only *(already)* | **queued — survived** |
| Perf Guard | blanket | cancelled |
| CSS Bundle Guard | blanket | cancelled |
| Backlog Guard | blanket | cancelled |

Both workflows scoped to `push` survived; all three with the blanket rule died.

It also confirms the review finding I accepted above: earlier in the same window
the *pending* older Tests run was replaced once a third arrival appeared, which
is precisely why the guarantee is scoped to in-flight runs rather than claimed as
"never cancelled".

### Not extending this to the three guard workflows

Deliberate. Those finish in 1–3 minutes, so cancelling a superseded run costs
almost nothing and returns the runner immediately — and runner capacity is the
binding constraint here (runs on this PR sat queued 20+ minutes before starting).
The fix targets the case where cancellation is expensive: `Tests` throws away up
to 35 minutes per shard across 18 shards and yields no verdict at all.

Worth noting the remaining half of TASK-22250 is untouched by this PR: even with
cancellation fixed, nothing started for a long stretch because no runner was
available. Cancellation was one cause of the missing verdicts; starvation is
another.

rmusser01:
### Correction: my "empirical confirmation" above was premature

I reported that `Tests` and `Derived Artifacts` **survived** a supersession while
the blanket-rule guards were cancelled. That was measured mid-flight and the
conclusion did not hold. Pushing `d86e82800` produced:

| workflow | rule | `2f10322d` | `40c7f2d9` | `d86e82800` |
|---|---|---|---|---|
| Tests | push-only | — | **cancelled** | pending |
| Derived Artifacts | push-only | cancelled | **cancelled** | queued |
| Backlog Guard | blanket | cancelled | success | cancelled |
| CSS Bundle Guard | blanket | cancelled | success | cancelled |
| Perf Guard | blanket | cancelled | success | cancelled |

The push-scoped workflows were cancelled too. What I saw earlier was "not yet
cancelled", and I read it as "protected".

**This is exactly the pending-member replacement raised in review, now confirmed
twice.** With `cancel-in-progress` false a newer run waits as *pending*, and the
next arrival replaces it. `cancel-in-progress` never protected a queued run and
cannot.

So, precisely:

- the fix protects a run that is **already executing**;
- it does nothing for a run still **queued**, which a later push still replaces;
- under the current starvation — 8 runs queued 23–40 min repo-wide with **zero in
  progress** — no run on this PR has ever reached "executing", so the fix has
  delivered **no observable benefit here yet**.

The change is still right: killing a running 18-shard batch discards up to 35
minutes per shard for no verdict, and that is the failure mode behind "no verdict
since 2026-06-26". But its benefit is contingent on runs actually starting, and
that is the starvation half of TASK-22250 — an account-level concurrency question
this PR does not address.

Note the guards' pattern in the table is the mirror image: they were cancelled
whenever superseded, but at `40c7f2d9` they were quick enough to *finish* before
the next push. Short jobs beat the cancellation window; an 18-shard matrix cannot.

PR #2104, opened 2026-08-26
test(citations): ephemeral sessions and legible misses — 69 failures to 40

rmusser01:
## Summary

Takes `test_console_local_citation_boundary` from **69 failures to 40**, measured
against clean `dev` `36bf1bdf4` on the same 95 tests, with **zero newly broken**.
All 14 timeouts are gone.

For context on why this file matters: until #2102 it was **unmeasurable** — it
hung, and `timeout_method = "thread"` destroyed the pytest process, so no JUnit
was written and none of its failures were visible at all.

## Two causes

**The second store helper was never made ephemeral.** #2102 fixed
`_persisted_store`; `_recording_citation_store` still called `ensure_session()`,
which is why 30 tests kept failing with no assistant row.

Those tests were hitting a **second** gate. `#2088` (TASK-22030) added a DB-open
refusal — *"Not sent: your conversation database could not be opened"* — and the
citation doubles carry `db = None`, so a non-ephemeral MANUAL send is refused
twice over: once because the adapter cannot `commit_durable_turn`, again by the
DB-open check. Both are guarded by `not session.ephemeral`, which is
production's own carve-out for a chat that is not being saved.

The refusal copy promises *"a temporary chat still sends"*. I verified that
promise holds rather than assuming it — a copy guaranteeing something the gate
did not honour would have been a real product bug.

**11 bare `next(<genexpr>)` calls.** StopIteration escaping a coroutine is
re-raised as `RuntimeError: coroutine raised StopIteration`, naming neither the
missing item nor the turn that failed to produce it — 33 failures here reported
that way. Replaced with `_first(..., what=...)`.

That step is what found the DB-open gate: it fixed nothing by itself, but once
the message read *"no ASSISTANT message was produced by the turn under test"*,
one probe located the cause. The same sequence worked in
`test_console_agent_swap` earlier in this arc.

## The remaining 40 — filed, not patched

29 assert on persistence **call counts**, which an ephemeral session
legitimately does not make. Their `create_message` seam is also bypassed by the
durable checkpoint's raw-SQL writes in `insert_with_messages`, so even a
durable-capable double might never see those calls.

Filed as **TASK-22301**. It needs a decision about which seam is authoritative
for a citation write now — asserting on durable ROWS rather than call counts
would survive the next refactor of the write path, which counting calls did not.
Explicitly *not* fixed by adding `commit_durable_turn` to the doubles: the real
method owns a transaction, creates the conversation, writes the Library policy
and returns a checkpoint, and hand-copying that into a fake is the trap these
doubles already fell into with `create_message`.

## Verification

| | clean `dev` `36bf1bdf4` | this branch |
|---|---|---|
| collected | 95 | 95 |
| failures | 69 | **40** |
| fixed | — | **29** |
| newly broken | — | **0** |

`./scripts/preflight.sh` passes all five derived-artifact checks.

### DIFF
--- a/Tests/Chat/test_console_local_citation_boundary.py
+++ b/Tests/Chat/test_console_local_citation_boundary.py
@@ -65,6 +65,28 @@ from tldw_chatbook.Chat.console_chat_store import ConsoleChatStore
 from Tests.console_provider_doubles import provider_resolution, with_destination
 
 
+def _first(matches, *, what: str):
+    """First match, or an assertion naming what never arrived.
+
+    A bare `next(<genexpr>)` raises StopIteration, which Python re-raises out of
+    a coroutine as `RuntimeError: coroutine raised StopIteration` -- naming
+    neither the missing item nor the turn that failed to produce it. 33 failures
+    in this module reported that way, hiding their real causes behind one
+    opaque message.
+
+    Args:
+        matches: Iterable of candidates.
+        what: Human-readable name of the thing expected, for the failure text.
+
+    Returns:
+        The first match.
+    """
+    value = next(iter(matches), None)
+    assert value is not None, f"no {what} was produced by the turn under test"
+    return value
+
+
+
 class _RequestBuilder:
     pass
 
@@ -499,8 +521,16 @@ def _recording_citation_store(
     persistence: _ReadyCitationPersistence | None = None,
 ) -> _RecordingCitationStore:
     store = _RecordingCitationStore(persistence=persistence)
-    session = store.ensure_session(
-        settings=ConsoleSessionSettings(provider="openai", model="repair-model")
+    # EPHEMERAL, for the same reason as `_persisted_store`: these doubles record
+    # `create_message` calls, which is the citation-write seam under test, and
+    # they carry `db = None`. A non-ephemeral MANUAL send is therefore refused
+    # twice over -- once because the adapter cannot `commit_durable_turn`, and
+    # again by TASK-22030's DB-open gate -- before any provider call. Both gates
+    # are guarded by `not session.ephemeral`, which is production's own carve-out
+    # for a chat that is not being saved.
+    session = store.create_session(
+        settings=ConsoleSessionSettings(provider="openai", model="repair-model"),
+        ephemeral=True,
     )
     session.project_instruction_state = (
         ProjectInstructionControlState.legacy_disabled()
@@ -550,11 +580,11 @@ def _capture_result(
 
 
 def _assistant(store: ConsoleChatStore):
-    return next(
+    return _first((
         message
         for message in store.messages_for_session(store.active_session_id)
         if message.role is ConsoleMessageRole.ASSISTANT
-    )
+    ), what="ASSISTANT message")
 
 
 # Task 3b rebase note: the controller's in-flight bookkeeping is now a
@@ -597,11 +627,11 @@ def _assert_no_terminal_state(store: ConsoleChatStore) -> None:
 
 
 def _final_user_content(messages) -> str:
-    return next(
+    return _first((
         message["content"]
         for message in reversed(messages)
         if message["role"] == ConsoleMessageRole.USER.value
-    )
+    ), what="message in the provider payload")
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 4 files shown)


rmusser01:
### Agreed — and already filed as TASK-22301, deliberately not fixed here

Your finding and my conclusion are the same one: the recording double bypasses
real SQLite, and these tests should assert **persisted rows** rather than mock
call counts. TASK-22301 (filed in `8c63d7091`) says exactly that, including your
recommended direction.

Being straightforward about what this PR does: making the sessions ephemeral is a
**workaround**. It unblocks 29 tests that could not run at all — a non-ephemeral
MANUAL send is refused twice over here, by the durable-turn gate and by
TASK-22030's DB-open gate, because the double carries `db = None`. It does not
make the seam correct.

### Why the proper fix is not in this PR

It is not a mechanical transform. 22 of the remaining failures are
`assert len(assistant_writes) == N` against recorded calls. Converting them to
real SQLite requires deciding **which seam is authoritative for a citation write
now**, and that is a real question: `a26cdafd8`'s durable checkpoint writes its
USER and assistant rows with raw SQL in `insert_with_messages`, bypassing
`create_message` entirely, while citation traces go through
`CitationTraceRepository`. Picking wrong yields 29 tests that pass while watching
a seam production no longer uses — which is precisely the defect class this
branch's sibling work has spent its time removing.

So the choice was: land a measured improvement with the residue documented, or
guess at an architectural question across 29 tests unattended. I took the first.

### What this PR is worth on its own

| | clean `dev` `36bf1bdf4` | this branch |
|---|---|---|
| collected | 95 | 95 |
| failures | 69 | **40** |
| newly broken | — | **0** |

And the context that matters: two PRs ago this file was **unmeasurable** — it hung,
and `timeout_method = "thread"` destroyed the pytest process, so none of its 95
tests reported anything at all. The seam problem you have flagged was invisible
until that was fixed.

One note on the rule as applied: the `db = None` double predates this PR. This
change did not introduce the violation, though it does not remove it either.

rmusser01:
### Added in `7a6a4db95`: a data-loss regression found while investigating this file

Chasing the remaining 40 failures here led to a genuine product bug, filed as
**TASK-22302**: since `a26cdafd8`, a durable Console turn writes **no citation
provenance at all** — every `rag_citation_*` table empty for an answer carrying
citations. Bisected: passes at `a26cdafd8~1`, fails at `a26cdafd8`, fails on
`dev`.

Four sequential blockers, each located and mutation-proven:

1. `_accept_durable_turn` receives `terminal_citation_finalizer`, the
   `_DurablePostcommitContinuation` has no field for it, and
   `resume_durable_postcommit` publishes owners with a hard-coded `None`
2. arming is gated on `append_message`'s own `persist` flag, correctly `False`
   here because the checkpoint already wrote the row — but that is not the same
   as the message having no durable row
3. `mark_message_complete` pops the finalizer, then returns via the
   dispatch-recovery shortcut thirty lines above the code that uses it
4. with those fixed the body persists as `''`, because `create_message`'s
   existing-row branch is an idempotent-RETRY path that verifies rather than
   updates

### The fix is NOT in this PR, deliberately

With all four addressed, `test_console_terminal_citation_persistence` goes
**92/92**. But in the full test composition the same change produces a
**duplicate `create_message`** — the same body written twice — in seven
capability tests, reproducibly across two identical runs (so not xdist
ordering). A duplicate durable write is a worse defect than the one being fixed,
so it is documented in the task rather than landed on hope. It could not be
reproduced in any smaller set, and each full-composition iteration is ~25
minutes.

### What this commit DOES change in tests

`test_console_terminal_citation_persistence`'s real-stack fixture no longer
pre-creates its conversation. `create_conversation(...)` writes the conversation
row but no `console_conversation_library_policy` row, so `commit_durable_turn`
took its "conversation exists" branch, found `policy_row is None`, and refused
the turn — the send produced a USER row and nothing else.

This does not change the failure COUNT for those two tests. It changes what they
say: previously "the send produced no ASSISTANT row" (a setup artifact),
now the citation-row mismatch that is the actual bug. That is what makes
TASK-22302 verifiable, and it is why the regression stayed hidden for months —
the one test that would have caught it was failing earlier, for an unrelated
reason.

PR #2115, opened 2026-08-26
fix(citations): persist citation provenance for durable Console turns (TASK-22302)

rmusser01:
## Summary

Since `a26cdafd8` ("resume Library-gated sends"), a durable Console turn has
written **no citation provenance at all** — every `rag_citation_*` table empty
for an answer carrying citations.

Bisected:

| commit | result |
|---|---|
| `a26cdafd8~1` (`52571f925`) | **passes** |
| `a26cdafd8` | **fails** |
| `dev` before this PR | fails |

All six tables come back 0 where 1 is expected: `rag_citation_traces`,
`rag_message_trace_owners`, `rag_evidence_runs`, `rag_evidence_snapshots`,
`rag_trace_evidence_refs`, `rag_answer_attempt_payloads`.

## Four sequential blockers

Each is independently mutation-proven — reverting any one turns the real-stack
tests red.

**1. The finalizer never survived the durable hand-off.**
`_accept_durable_turn` takes `terminal_citation_finalizer` as a parameter and
used it exactly once more: it built a `_DurablePostcommitContinuation` with no
field for it, and `resume_durable_postcommit` then published the live owners
with a hard-coded `terminal_citation_finalizer=None`. The parameter was
accepted and dropped.

**2. Arming was gated on the wrong condition.**
`append_message` arms only when its own `persist` flag is true. On this path
`persist=False` is *correct* — the durable row already exists, written by the
dispatch checkpoint, so this append must not create a second one. But "this call
does not create the row" is not the same as "this message has no durable row",
and arming needs the latter.

**3. The in-flight dispatch shortcut discarded it.**
`mark_message_complete` pops the finalizer, then returns via the
dispatch-recovery branch — thirty lines above the code that would use it.

Scoped deliberately to an actual `finalizer`, **not** to `terminal_persistence`:
that is also true for a merely DEFERRED turn which has no citation to seal, and
skipping the shortcut for those turned one ordinary terminal write into two.
The first version of this fix did exactly that and broke 7 capability tests.

**4. The body then persisted as `''`.**
`create_message`'s existing-row branch is an idempotent-RETRY path: it verifies
and writes the citation but does not update content. Flushing the final body
first makes the row match, after which the citation write keys to it — **no
persistence API change required**, which matters because the alternative was
extending a transactional method.

## Why it hid for months

The one test that would have caught it was failing **earlier**, for an unrelated
reason, so it never reached the citation assertion: its stack pre-created a
conversation via `create_conversation(...)`, which writes no
`console_conversation_library_policy` row, so `commit_durable_turn` took its
"conversation exists" branch, found `policy_row is None`, and refused the turn.

That refusal was itself **silent** until TASK-22251 added
`Durable turn commit failed; turn refused (exception_type=...)` — which is
literally the line that surfaced this.

## Verification

Against current `dev`, same test set both sides:

| | `dev` | this branch |
|---|---|---|
| collected | 2105 | 2105 (delta **+0**) |
| failures | 101 | **97** |
| newly broken | — | **0** |

`test_console_terminal_citation_persistence`: **92/92**.
`./scripts/preflight.sh`: all five derived-artifact checks pass.

## A retraction worth recording

This fix was first shelved after an A/B reported 7 regressions. That comparison
was **confounded** — my branch carried #2104's test changes while the baseline
predated them, so it moved two variables at once. Re-run against `dev` with
#2104 merged, the identical fix breaks nothing. The tests it appeared to break
never call `mark_message_complete`, the only path this change touches, which is
what prompted the re-check.

---
## Summary by cubic
Fixes TASK-22302: durable Console turns no longer lose citation provenance or the terminal answer body. The citation finalizer now survives the durable hand-off, arms at the correct point, and the final body is flushed even when the finalizer fails closed, so the durable row keeps the answer instead of the checkpoint's empty content.

**Bug Fixes**
- Carries the finalizer through `_DurablePostcommitContinuation` instead of dropping it at the publish site.
- Arms the finalizer in `_hydrate_durable_turn_owner_messages` after the durable id is assigned, since `append_message`'s `persist` flag is false on this path.
- The dispatch-recovery shortcut in `mark_message_complete` now only skips the citation write when no finalizer is pending, leaving deferred turns untouched.
- Flushes the final body whenever a durable row already exists, so a fail-closed finalizer leaves the answer in the row instead of empty content.
- Adds a real-stack regression test driving an unsealable builder: it asserts one row, the body persisted, and no citation rows.
- Running the same 2107 tests against `dev` drops failures from 101 to 97 with zero newly broken.

<sup>Written for commit 84f743be619d9e4c78a7907fd1027019086800fe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2115?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Chat/test_console_terminal_citation_persistence.py
+++ b/Tests/Chat/test_console_terminal_citation_persistence.py
@@ -1309,6 +1309,74 @@ def test_terminal_selected_answer_citations_survive_restart(
 
 @pytest.mark.integration
 @pytest.mark.asyncio
+async def test_real_durable_fail_closed_finalizer_persists_the_body_once(
+    real_citation_stack_factory,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """TASK-22302 review: a finalizer that fails closed must not lose the answer.
+
+    `finalize()` returns None when the builder cannot seal -- a real, logged
+    outcome ("attempt_or_seal_failure"), and the supported way to fall back to
+    an ordinary message. On a DURABLE turn that combination is delicate, because
+    the dispatch checkpoint has already written the assistant row with EMPTY
+    content:
+
+    * the final body must still be flushed, or the durable row keeps '' while
+      the in-memory message reads complete; and
+    * the flush must be an UPDATE, because `create_message`'s existing-row
+      handling lives inside its `prepared_citation is not None` branch -- with
+      no citation it falls through to `add_message` and inserts against an id
+      that already exists.
+
+    So: one row, carrying the body, and no citation rows.
+    """
+    stack = real_citation_stack_factory("real-fail-closed")
+    builder, prompt_id = _real_captured_builder(stack.repository)
+
+    def unsealable(*_args: object, **_kwargs: object) -> None:
+        raise RuntimeError("seal-sentinel")
+
+    monkeypatch.setattr(CitationTraceBuilder, "seal", unsealable, raising=False)
+    controller = _real_controller(stack, builder, prompt_id)
+
+    log_stream, handler_id = _capture_logs()
+    try:
+        result = await controller.submit_draft("question")
+    finally:
+        logger.remove(handler_id)
+
+    assistant = _real_assistant(stack.store)
+    assert result.accepted is True
+    assert assistant.content == _BODY_SENTINEL
+
+    persisted = stack.db.get_message_by_id(assistant.id)
+    assert persisted is not None
+    assert persisted["content"] == _BODY_SENTINEL, (
+        "the durable row kept the checkpoint's empty content -- the answer was "
+        f"lost on the fail-closed path: {persisted['content']!r}"
+    )
+
+    duplicates = stack.db.get_connection().execute(
+        "SELECT count(*) FROM messages WHERE id = ?", (assistant.id,)
+    ).fetchone()[0]
+    assert duplicates == 1, f"the terminal write duplicated the row ({duplicates})"
+
+    # Fail CLOSED: an ordinary message, with no citation provenance at all.
+    assert _citation_row_counts(stack.db)["rag_citation_traces"] == 0
+
+    # ...and the terminal write must COMPLETE, not be abandoned. Without the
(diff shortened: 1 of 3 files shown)


rmusser01:
### Finding addressed in `576aef99b` — you were right, and it was a real loss path

`finalize()` returns `None` when the citation builder cannot seal. That is a
genuine, logged outcome (`attempt_or_seal_failure`) and the supported way to
fall back to an ordinary message — so guarding the terminal body flush on
`citation_write is not None` skipped it exactly when it was needed, leaving the
durable row with the dispatch checkpoint's **empty content** while the in-memory
message read complete. The answer would be lost, silently. My fix introduced
that path, since arming the finalizer on durable turns is what makes it
reachable at all.

The flush now runs whenever the message already has a durable row, regardless of
whether a citation follows.

**Regression test added**, driving a genuinely unsealable builder through the
real stack and asserting: the durable row holds the body, exactly one row
exists, and zero citation rows are written (fail *closed*, not fail *lossy*).
Re-applying your exact guard turns it red.

### One thing I checked and then removed

I first added a second guard so the fail-closed case skipped `create_message`
entirely, reasoning it would hit a unique-constraint conflict. I verified a
duplicate-id create *does* raise `ConflictError` — but then probed the actual
path: the call is reached, raises nothing, and duplicates nothing, and mutating
the guard away left every test green. Since I could not demonstrate it was
needed, I removed it rather than keep unexplained defensive branching. Worth
flagging in case you can see a path I could not construct.

### Verification

Against `dev`, same set both sides: **101 failures → 97, zero newly broken**,
+2 collected (the new test). `test_console_terminal_citation_persistence`:
**93/93**. `./scripts/preflight.sh`: all five checks pass.

PR #2118, opened 2026-08-26
test(review): close the two open review findings from this workstream's merged PRs

rmusser01:
## Summary

Closes the two review findings still open across the merged PRs in this
workstream. Qodo marks resolved findings with strikethrough, so the accounting
is exact:

| PR | findings | open before this PR |
|---|---|---|
| #2061 | 3 | 0 — all resolved pre-merge |
| #2078 | 2 | 0 — all resolved pre-merge |
| #2102 | 2 | **1** — `_CapturedLines` violates PascalCase |
| #2104 | 1 | **1** — `_recording_citation_store` bypasses real SQLite |

## #2102 — PascalCase

Renamed the three helpers in `test_console_durable_commit_diagnostics` to drop
their leading underscore.

I pushed back on this at the time with evidence, and that evidence still stands:
**2,448 test classes** in this repo carry a leading underscore to mark them
module-private, including the direct neighbours of these helpers
(`_RecordingGateway`, `_ReadyCitationPersistence`). The rule and the corpus
genuinely disagree. Complying is the cheaper resolution for three names in one
file, and the module docstring now records the discrepancy so the next reader is
not misled about the prevailing convention.

## #2104 — real SQLite

`_ReadyCitationPersistence` now **spies over a real `ChatPersistenceService`**
backed by an in-memory `CharactersRAGDB` with a real `CitationTraceRepository`,
instead of carrying `db = None` and recording writes it never performed.

Recording is kept deliberately: 23 tests in that module assert on
`create_calls`, and rewriting each to query rows would risk weakening them one
at a time. Spying over the real service makes the writes real — the property the
rule is asking for — while leaving the assertions intact.

### What I tried and reverted

I also made the module's sessions durable, so the spy would actually write rows.
That broke two tests which **close the session mid-collection**: closing retires
the durable preparation, so the in-flight effect release hits
`_require_durable_fingerprint_locked` and raises
`Durable postcommit fingerprint changed.`

That is production behaviour, not a harness artifact — any user closing a chat
during an in-flight durable turn takes the same path — so it is filed as
**TASK-22587** rather than papered over with a per-test opt-out. Sessions stay
ephemeral, exactly as #2104 left them.

## Verification

| | `dev` | this branch |
|---|---|---|
| collected | 2111 | 2111 (delta **+0**) |
| failures | 91 | 91 |
| newly broken | — | **0** |

A rule-compliance refactor with **no behaviour change**, which is the right
shape for it. Being straight about the value: this fixes the violation and
proves rows are the correct seam; it does **not** reduce the failure count. That
residue is TASK-22301 — those tests observe `create_message`, which the durable
checkpoint bypasses entirely (`insert_with_messages` writes rows with raw SQL) —
and TASK-22587 now blocks it, since the assertion rewrite needs durable sessions.

`./scripts/preflight.sh`: all five derived-artifact checks pass.

## A note on task ids

`TASK-22303` collided with another session's task minted mid-flight, and was
renumbered to **TASK-22587** under the 2026-08-21 owner rule (older arrival
keeps the id), with provenance recorded. The id had been swept against all refs
and worktrees when first minted; the high-water mark moved from 22030 to 22557
during this session.

---
## Summary by cubic
Closes the last open review finding from this workstream's merged PRs by renaming the helper classes in `test_console_durable_commit_diagnostics` to PascalCase. The #2104 real-SQLite finding is handled through tasks instead of code: a spy over real SQLite was measured to be inert (zero `create_message` calls across all 95 tests, since ephemeral sessions never reach that seam), so the CSV test file is reverted to byte-identical with dev.

- The renamed helpers follow the repo rule even though the prevailing `Tests/` convention is a leading underscore (2,448 test classes); the module docstring records the discrepancy.
- Filed TASK-22587 for the production bug found while testing durable sessions: closing a session during an in-flight durable turn raises `Durable postcommit fingerprint changed.`
- Updated TASK-22301 to record that its assertions can only be fixed by making sessions durable, which TASK-22587 now blocks.

<sup>Written for commit d41b9c60261ab11d166dbc9ecdb1677a6873a51f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2118?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Chat/test_console_durable_commit_diagnostics.py
+++ b/Tests/Chat/test_console_durable_commit_diagnostics.py
@@ -14,6 +14,12 @@ is a diagnostic defect, not a user-experience one.
 
 These tests pin both halves: the type IS recorded, and the exception's message
 is NOT (it can carry conversation and workspace identifiers).
+
+Helper classes here are named without a leading underscore, per the repo's
+PascalCase rule (Qodo finding on #2102). Note this differs from the prevailing
+convention in `Tests/` -- 2,448 test classes carry a leading underscore to mark
+them module-private -- so the rule and the corpus disagree; this file follows
+the rule.
 """
 
 from __future__ import annotations
@@ -26,7 +32,7 @@ from tldw_chatbook.Chat.console_chat_controller import ConsoleChatController
 from tldw_chatbook.Chat.console_chat_store import ConsoleChatStore
 
 
-class _Gateway:
+class ReadyGateway:
     """Ready provider that streams one chunk."""
 
     async def resolve_for_send(self, selection):
@@ -36,14 +42,14 @@ class _Gateway:
         yield "reply"
 
 
-class _CapturedLines:
+class CapturedLines:
     """Collect loguru records without disturbing the ambient sinks."""
 
     def __init__(self) -> None:
         self.lines: list[str] = []
         self._sink_id: int | None = None
 
-    def __enter__(self) -> "_CapturedLines":
+    def __enter__(self) -> "CapturedLines":
         self._sink_id = loguru_logger.add(
             lambda message: self.lines.append(str(message)),
             level="TRACE",
@@ -67,7 +73,7 @@ class _CapturedLines:
 SECRET_IDENTIFIER = "ws-private-4f2a-identifier"
 
 
-class _ExplodingPersistence:
+class ExplodingPersistence:
     """Persistence whose durable commit fails with an identifying message."""
 
     db = None
@@ -77,10 +83,10 @@ class _ExplodingPersistence:
 
 
 def _controller_with_failing_commit():
-    store = ConsoleChatStore(persistence=_ExplodingPersistence())
+    store = ConsoleChatStore(persistence=ExplodingPersistence())
     controller = ConsoleChatController(
         store=store,
-        provider_gateway=_Gateway(),
+        provider_gateway=ReadyGateway(),
         provider="llama_cpp",
         model="test-model",
         agent_runtime_enabled=False,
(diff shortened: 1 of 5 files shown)


rmusser01:
### Finding 1 led somewhere worse than the finding — reverted in `d41b9c6026`

You flagged that the spy writes under `conversation-1` with no parent
conversation row, so the delegated `create_message` would fail and retry. Chasing
that, I instrumented the spy and counted calls across the whole module:

**Zero.** Not one `create_message` reaches it, across all 95 tests.

With ephemeral sessions no test touches that seam at all, so I had wired a real
`ChatPersistenceService` that is never exercised. The rule would have been
satisfied cosmetically and nothing else — and I had claimed on this PR that real
SQLite was "genuinely exercised", on the strength of a probe that used *durable*
sessions, a configuration I then reverted. The probe was real; the
generalisation from it was not.

So the spy is reverted. `test_console_local_citation_boundary` is now
byte-identical to `dev`. Findings 2-4 (unclosed SQLite, missing docstrings)
applied only to that reverted code and are moot with it.

Your finding was correct about the wiring, and it would have bitten the moment
the seam became reachable — which is exactly when someone does the real work.

### What this changes about the real work

TASK-22301 cannot be fixed by re-wiring the double. It needs **durable sessions
plus row-based assertions**, and that is blocked by **TASK-22587**: two tests in
that module close the session mid-collection, and a durable session then raises
`Durable postcommit fingerprint changed.` from
`_require_durable_fingerprint_locked`. Order of work is now recorded on the task.

### What remains here

Only the #2102 PascalCase rename, which stands on its own: three helpers in
`test_console_durable_commit_diagnostics` renamed, with the docstring recording
that 2,448 test classes in this repo use the leading-underscore convention the
rule forbids — so the next reader is not misled about which way the corpus
leans.

PR #2119, opened 2026-08-26
fix(library): bound multi-EPUB import memory (TASK-22508)

rmusser01:
## Summary

- isolate EPUB imports in one-process parse-pool generations and retire the generation after the ebook batch, while preserving configured parallelism for ordinary imports
- reject unsafe EPUB ZIP expansion and manifest-declared markup before ebooklib extraction
- keep pool transitions fail-closed, including queued imports and preclaimed Research retries when worker teardown cannot be proven complete
- document the resource boundary in ADR-091 and close TASK-22508

## Root cause

A logical one-at-a-time EPUB admission cap still allowed sequential books to rotate across a persistent three-process pool. Each worker retained its parser high-water heap, so process-tree memory continued to climb even though only one EPUB parsed at a time.

## Verification

- targeted coordinator, registry, EPUB parsing/chunking, and spawned-worker tests: 266 passed, 1 Windows-only skip
- Python compilation, Ruff, and Git diff checks passed
- spawned three-EPUB chunking reproduction: 3 done, 0 failed, no resident parse workers
- peak process-tree RSS: 496.2 MiB versus roughly 750 MiB before
- maximum event-loop gap: 62.3 ms
- full suite not run per repository policy

## Architecture

ADR-091: bounded EPUB archive admission and one-worker ebook pool ownership.

---
## Summary by cubic
Fixes multi-EPUB imports exhausting memory by running ebooks in a one-process parse-pool generation that retires after the batch, and by rejecting unsafe EPUB ZIP archives before `ebooklib` extraction.

**Bug Fixes**
- Sequential EPUBs previously rotated across the shared three-process pool, so each worker retained its parser high-water heap and memory kept climbing despite a one-at-a-time admission cap.
- EPUB batches now run in a one-process pool generation that is terminated and joined off the UI thread once the batch drains; ordinary imports keep configured parallelism in separate generations.
- Pool-mode transitions are fail-closed with a bounded off-thread wait: if teardown times out, the teardown thread cannot start, or a join raises, queued local jobs and preclaimed Research retries fail retryably with restart guidance.
- A late teardown completion never releases that gate, so a replacement pool cannot overlap workers that may still be resident.
- Peak process-tree RSS for a three-EPUB chunking reproduction dropped from roughly 750 MiB to 496 MiB, with no resident parse workers after the batch.
- MOBI, AZW, AZW3, and FB2 share the one-at-a-time ebook lane but skip the ZIP-specific guard.

**New Features**
- EPUB ZIP archives are validated against fixed safety limits: member count, per-member and total expanded bytes, expanded markup bytes (recognized suffixes plus manifest-declared HTML/XML media types), and expansion-to-compression ratio.
- Unsafe archives fail as one isolated ingest job with the error "EPUB archive exceeds safety limits."; manifest classification that cannot be resolved safely fails closed.
- The limits are documented in ADR-091 as non-configurable safety invariants; very large image-heavy EPUBs may be refused even when structurally valid.

<sup>Written for commit fa6c66f33b3d75ee20f40b91e95c71d0e825d215. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2119?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Addressed all five Qodo findings in fa6c66f33b:

1. Updated the Library ingest canvas harness to accept the new processes keyword.
2. Bounded EPUB worker terminate/join retirement, failing closed on timeout and preventing late completion from resuming work.
3. Routed direct EPUB ingestion through the shared path validator and consistently use the returned validated path.
4. Renamed the newly added test-local classes to remove leading underscores.
5. Routed worker-retirement thread construction and start failures through the fail-closed callback so the retirement gate cannot remain stranded.

Also refreshed and reviewed the persistent diagnostic inventory entries affected by the new logging sites. Verification: 305 focused tests passed with 1 Windows-only skip; Ruff, py_compile, git diff --check, and the full repository preflight all passed.

PR #2123, opened 2026-08-26
fix(console): closing a chat mid-turn is not an error (TASK-22587)

rmusser01:
## Summary

Closing a Console session while a durable turn is still collecting raised
`RuntimeError: Durable postcommit fingerprint changed.`

Reproduced at store level on clean `dev` before anything was changed: claim a
postcommit effect, close the session, release the claim.

## The bug was worse than filed

That raise comes out of the **release** path — the one whose entire job is to
clean up after a failure — and that path runs inside `except BaseException:`:

```python
except BaseException:
    self.store.abandon_durable_postcommit_effect(...)   # raised
    raise                                              # never reached
```

So it **replaced** the exception that sent us there. A genuine failure during a
turn the user happened to close was reported as a fingerprint message.

## Fix

The guard's intent is right: an effect must not commit against a retired
preparation. What was missing is a defined outcome for *"the preparation was
retired underneath me"* — an ordinary consequence of closing a chat — as
distinct from *"the fingerprint changed unexpectedly"*, which is a bug. Those
shared one raise.

The two are **decidable, not guessed**: `retire_durable_acceptance` already
leaves a tombstone carrying the *same* fingerprint. A matching tombstone proves
retirement rather than mutation, which is what the new `_durable_retired_locked`
reads. Retirement raises the distinct `ConsoleDurableAcceptanceRetired`; a real
mismatch still raises the generic `RuntimeError`.

Releasing a claim after retirement is a **no-op**, and that is safe rather than
merely quiet: `retire_durable_acceptance` has already dropped every in-flight
key for the preparation, so there is nothing left to release. A test asserts
that precondition, so the no-op cannot silently start leaking claims if it ever
stops holding.

Controller: the release call can no longer replace the original failure, and
retirement during a **successful** effect is a non-event — the work is done,
there is simply no ledger left to record it in.

## Mutation testing caught a test of mine that could not fail

Four mutations, one per behaviour change. **M3 stayed green**: the first
"not masked by the release path" test passed with the controller guard removed,
because the store fix already stops `abandon` from raising on that path. It
never tested what its name claimed.

Split into two: one honest end-to-end test whose docstring says what it does
*not* cover, and one that forces `abandon` to raise — which does go red without
the guard. That also made a `# pragma: no cover` I had written false, so it is
gone. All four changes are now individually mutation-proven.

The negative control is the load-bearing test here: narrowing a safety guard is
only safe if the guard still fires for the case it was written for.

## Verification

A/B of `Tests/Chat` against merge-base `65cf855371`, baseline tree verified with
`git diff --quiet` before the control ran:

| | merge-base `65cf855371` | this branch |
|---|---|---|
| collected | 7682 | 7689 (**+7**, the new tests) |
| failures | 111 | **111** |
| newly broken | — | **0** |

`./scripts/preflight.sh` green. The inventory reported +2 diagnostics in
`console_chat_controller.py`; both were reviewed before regenerating — they
interpolate internal effect names only (all 10 are string literals at the call
sites) and the exception **type**, never its message, per the TASK-22251
precedent that exists because those messages carry conversation and workspace
identifiers.

## Unblocks

TASK-22301 — converting `test_console_local_citation_boundary` to row
assertions needs durable sessions, which is what tripped this.

---
## Summary by cubic
Fixes closing a Console chat mid-turn so it no longer raises `RuntimeError: Durable postcommit fingerprint changed.` The release path raised from inside `except BaseException:`, replacing the failure that sent it there and hiding the real cause. A retired preparation is now a defined, ordinary outcome of the user closing a chat.

**Bug Fixes**
- Retirement is detected via a tombstone carrying the same fingerprint, so it's never confused with a genuine mismatch.
- Retirement raises a distinct `ConsoleDurableAcceptanceRetired`; a real fingerprint change still raises the generic `RuntimeError`.
- Releasing a claim after retirement is a no-op, and a test pins the precondition so the no-op can't silently leak claims.
- Retirement is terminal-benign for the whole postcommit sequence: the resume lookup, in-flight effects, the sequence's final retire, and the recovery handler's completed-effects read all survive a close without raising or leaking state.
- Preparations with a running sequence pin their tombstone against FIFO eviction, so unrelated close volume can't turn the close back into the fingerprint error.
- The controller no longer lets bookkeeping replace the original failure, and retirement during a successful effect just leaves the work unrecorded.
- Unblocks TASK-22301, which needs durable sessions.

<sup>Written for commit 49ffcd6fa686c8f59faed4c4212d182242b89f41. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2123?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/security/production-diagnostic-inventory.json
+++ b/Docs/security/production-diagnostic-inventory.json
@@ -243,8 +243,8 @@
       "reason": "high-risk Chat/provider/summarization/tool/MCP diagnostic owner"
     },
     {
-      "call_count": 52,
-      "diagnostic_digest": "9470774591ce409dfb1a",
+      "call_count": 55,
+      "diagnostic_digest": "7572789e67be8299bff3",
       "owner": "TASK-492",
       "path": "tldw_chatbook/Chat/console_chat_controller.py",
       "reason": "high-risk Chat/provider/summarization/tool/MCP diagnostic owner"
@@ -3968,7 +3968,7 @@
   "summary": {
     "owner_files": 537,
     "persistent_sink_files": 8,
-    "task_492_calls": 1241,
+    "task_492_calls": 1244,
     "task_494_calls": 7360
   }
 }
--- a/Tests/Chat/test_console_close_during_durable_postcommit.py
+++ b/Tests/Chat/test_console_close_during_durable_postcommit.py
@@ -1,0 +1,225 @@
+"""Closing a Console session mid-turn is not an error (TASK-22587).
+
+Closing a session retires its durable preparation. An effect still in flight
+then finds its fingerprint gone, and `_require_durable_fingerprint_locked`
+raised `RuntimeError("Durable postcommit fingerprint changed.")` out of the
+RELEASE path -- the one place whose whole job is to clean up after a failure.
+Worse, that raise REPLACED the exception that sent us down the release path, so
+the real cause was lost and the user saw a fingerprint message instead.
+
+The guard's intent is right: an effect must not commit against a retired
+preparation. What was missing is a defined outcome for "the preparation was
+retired underneath me" (ordinary: the user closed a chat) as distinct from
+"the fingerprint changed unexpectedly" (a bug). Those shared one raise.
+
+`retire_durable_acceptance` leaves a tombstone carrying the SAME fingerprint,
+so the two are decidable rather than guessed -- which is what
+`_durable_retired_locked` reads.
+
+The negative control below is the point of this file: narrowing a safety guard
+is only safe if the guard still fires for the case it was written for.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from Tests.Chat.test_console_durable_turn_acceptance import _ready_store
+from Tests.console_provider_doubles import provider_resolution
+from tldw_chatbook.Chat.console_chat_controller import ConsoleChatController
+from tldw_chatbook.Chat.console_chat_store import ConsoleDurableAcceptanceRetired
+
+EFFECT = "checkpoint_transition"
+
+
+def _claimed_effect(tmp_path, *, claim: bool = True):
+    """A committed durable turn, with one postcommit effect optionally claimed.
+
+    The controller-level tests pass ``claim=False`` because
+    ``_run_durable_postcommit_effect`` claims the effect itself; pre-claiming
+    would make it refuse with "already in flight" before reaching the path
+    under test.
+    """
+
+    db, _service, store, _preparation, acceptance = _ready_store(tmp_path)
+    commit = store.commit_durable_turn(acceptance)
+    prep = acceptance.preparation_id
+    fingerprint = store.durable_acceptance_fingerprint_for(prep)
+    assert fingerprint is not None, "harness precondition: a fingerprint exists"
+
+    store.begin_durable_postcommit_effects(
+        preparation_id=prep,
+        session_id="session-1",
+        assistant_message_id=commit.assistant_message_id,
+        fingerprint=fingerprint,
+    )
+    if claim:
+        assert store.claim_durable_postcommit_effect(
+            prep, EFFECT, fingerprint=fingerprint
+        ), "harness precondition: the effect must be claimable"
+    return store, prep, fingerprint
+
+
+def test_releasing_a_claim_after_the_session_closed_does_not_raise(tmp_path) -> None:
+    """AC1/AC2: the release path survives the user closing the chat."""
+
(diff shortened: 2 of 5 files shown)


rmusser01:
## Both findings were real. Fixed, and each verified before fixing.

I confirmed both claims empirically rather than taking them on the description
— a throwaway test asserting each raise, run against the then-current branch:

- **#1** `durable_postcommit_effects_for` after a close → raises
  `ConsoleDurableAcceptanceRetired` ✔
- **#2** a second `retire_durable_acceptance` after a close → raises
  `RuntimeError("...fingerprint changed.")` ✔

The diagnosis was right and my original scope was wrong: **I fixed the helper,
not the orchestration.** Suppressing retirement inside one effect only moved the
raise to the next fingerprint-validated effect, or to the unconditional retire
that ends the sequence.

### What changed

Retirement is now a **terminal benign outcome for the whole orchestration**:

1. **The early lookup is guarded.** This turned out to be the one that actually
   fires first: the `durable_postcommit_effects_for` call at the top of
   `resume_durable_postcommit` sits **before** the `try` block, so guarding only
   the sequence left it raising. Finding #1's real entry point.
2. **A sequence-level arm** (`except ConsoleDurableAcceptanceRetired`) covering
   all eight effects.
3. **`retire_durable_acceptance` is idempotent** for the same acceptance —
   finding #2 directly. Keyed on the tombstone's fingerprint, so it is *this*
   acceptance, not merely "the id has a tombstone".
4. **`durable_completed_effects_for`**, a tombstone-backed read, so the failure
   handler's "did the checkpoint run?" question survives a close instead of
   raising inside the handler. This keeps `provider_started` correct rather than
   silently false.

Both paths run the same continuation cleanup the normal tail runs (settle, drop
the continuation, release prepared evidence), minus the owner-changed check —
the owner is legitimately gone — via a shared `_postcommit_stopped_by_close`.

### On "helper-only tests do not cover its exception handler"

Correct, and the fix is tested at the real boundary now: `resume_durable_postcommit`
via `submit_draft`, including a close that happens **mid-sequence** (inside
effect #1 of eight, leaving seven to cope) — which is the case that distinguishes
a real fix from moving the raise one effect along.

### Mutation results

The mutation sweep is what proved the first round incomplete, and it caught more:
after the initial fix, three changes stayed **green** under mutation — unproven
code, because my tests closed the chat *before* resume and the early guard
short-circuited the rest. With mid-sequence coverage and store-level tests for
the second-retire and the tombstone read, **all 8 mutations now go red.**

Added a negative control for the idempotent retire: a *different* acceptance on a
retired id must still raise, or idempotency would wave through a genuine mismatch.

### Verification

13 tests. A/B of `Tests/Chat` against merge-base `65cf855371` (baseline tree
verified byte-identical first): collected 7682 → 7695 (**+13**), failures
**111 → 111**, **0 newly broken**. Preflight green; the +1 diagnostic is a
constant string with no interpolation.

rmusser01:
## Finding #4 (eviction loses retirement proof) — fixed

Correct, and worth fixing rather than waving off as improbable: the objection is
that **correctness should not depend on unrelated session-close volume**, and
that holds regardless of how unlikely 128 concurrent retirements are.

Preparations whose postcommit sequence is running are now marked on
`begin_durable_postcommit_effects` and released on **both** exits — the normal
tail and the closed-session path. Eviction skips protected entries oldest-first.

**The part that needed care:** a protection that can pin a bounded cache open
without limit is just a different bug. So when every entry is protected, the
oldest is still evicted, and that fallback has its own test
(`test_protecting_tombstones_still_honours_the_cap` pins *every* entry and
asserts the map stays within the cap). Without it this would have traded a rare
error for an unbounded map.

Four new mutations, all red:

| mutation | result |
|---|---|
| eviction protection removed | RED |
| cap fallback removed (leak) | RED |
| activity never marked | RED |
| release never happens | RED |

### Verification

16 tests. A/B of `Tests/Chat` against merge-base `f9a06ff62` (baseline tree
verified byte-identical first): collected 7672 → 7688 (**+16**), failures
**108 → 108**, **0 newly broken**. Preflight green.

Note the earlier green `Derived artifacts` run was on `2d4c854fcb`, which
predates this commit — not treating it as a merge signal; waiting for the gate
to re-run on this head.

PR #2128, opened 2026-08-27
test(citations): put the citation boundary on the real durable path (TASK-22301)

rmusser01:
## Summary

`test_console_local_citation_boundary.py` went from **40 failed / 55 passed** to
**0 failed / 92 passed / 3 xfailed**. Tests only — no product change.

The point is not the count. These 95 tests were watching a seam the durable
Console path stopped using, so they could not observe what they claimed to test.
They now exercise the real durable path against real SQLite, and assert on
**committed rows and citation traces** instead of `create_message` call counts.

## How the diagnosis was reached

Each step was measured before the next, and two of them refuted the obvious fix:

1. **Swapping the `db = None` fake for a real `ChatPersistenceService` changed
   nothing** — 40 failures before and after. This reproduces the earlier
   "inert spy" finding: real persistence does nothing while sessions are
   ephemeral.
2. **Making sessions durable alone made it worse (40 → 65).** `_persisted_store()`
   defaulted to `persistence=None` — invisible when ephemeral, a refused send
   (*"Not sent: your conversation database could not be opened"*) when durable.
   Fixed with production's own rule: **ephemeral iff nothing can persist**.
3. With both, **rows commit and `create_calls` stays 0** — confirming the task's
   core claim that the durable checkpoint writes via `insert_with_messages` and
   bypasses `create_message` entirely.
4. Traces still wrote **0 rows**: the old fake hard-coded
   `canonical_citation_writes_ready = True`; the real service *computes* it and
   reports `False` without a citation repository, silently skipping every
   citation write.
5. Traces *still* wrote 0 rows. Instrumenting the repository showed
   `prepare_write` raising `CitationPersistenceUnavailable("run_authority_mismatch")`
   — the builder sealed with a hand-built identity while the repository verifies
   against the identity persisted in its database.

Also learned the hard way: the DB must be **file-backed**, not `:memory:`.
`CharactersRAGDB` opens a thread-local connection and TASK-22205 offloads durable
DB calls to a worker thread, which with `:memory:` gets its own empty database.

## Mutation-proven, product-side

A green file proves nothing on its own, so the behaviour was broken deliberately:

| mutation | result |
|---|---|
| mis-sender the durable assistant owner row | **31 failed** |
| skip the citation trace write | **6 failed** |

Both seams the conversion moved onto are genuinely observed.

## Three tautologies I introduced, and caught

After the bulk conversion, `assert "citation_write" not in row` was true of **any**
row dict and could never fail — and **two of the three sat in tests that were
passing**. They would have stayed green while testing nothing. Replaced with a
trace-table check that can fail. Two helpers I added and then orphaned were also
removed.

## Behaviour the durable path genuinely changed

Adapted with the reasoning recorded, not papered over:

- A stream failure **no longer propagates**. The committed turn is retained for
  recovery (`BLOCKED`) — which is the point of a durable turn, where unlike an
  ephemeral one there is something to retain.
- A **retained** turn keeps its terminal citation finalizer armed, because the
  recovery that resumes it needs one. Asserted *conditionally*, so a finalizer
  left behind by a **non-retained** failure is still caught.

## Three xfails, each naming a filed task — none skipped

`strict=False`, so they report if they start passing.

- **2 × TASK-22690** — closing a chat mid-turn raises "Durable continuation owner
  changed": a fourth raise site of the class TASK-22587 fixed. TASK-22587
  predicted these two tests would hit a durable-path failure once their sessions
  stopped being ephemeral.
- **1 × TASK-22720** — the agent bridge's `run_reply` raises partway and the
  controller swallows it, leaving an empty assistant row. **Pre-existing**: one of
  the original 40, not introduced here. Measured rather than assumed — the bridge
  runs (its `calls` list is non-empty) but never reaches its own replacement code.

## Verification

A/B of `Tests/Chat` against merge-base `c23113e2e0`, baseline tree verified
byte-identical with `git diff --quiet` before the control ran:

| | merge-base `c23113e2e0` | this branch |
|---|---|---|
| collected | 7688 | 7688 (delta **+0**) |
| failures | 108 | **68** |
| fixed | — | **40** |
| newly broken | — | **0** |

All 40 fixed nodes are in this file. `./scripts/preflight.sh` green.

---
## Summary by cubic
Fixes all 40 failures in `test_console_local_citation_boundary.py` by exercising the real durable path against real SQLite — test-only, no product changes — satisfying TASK-22301. The tests used to run on ephemeral sessions and assert on `create_message` call counts the durable turn never makes; they now assert on committed rows and citation traces, and file TASK-22690 and TASK-22720 for remaining bugs.

**Refactors**
- `_ReadyCitationPersistence` now delegates to a real `ChatPersistenceService` over file-backed SQLite in a temp directory; `:memory:` gives the DB worker thread its own empty database, and a shared-cache URI string is treated as a literal filename by `CharactersRAGDB` — that trap is recorded in `lessons-testing-evidence.md`. An autouse fixture closes every connection and removes the temp directory.
- Stream failures on the durable path no longer propagate — the committed turn is retained for recovery and keeps its terminal citation finalizer armed.
- Three `citation_write` membership assertions that could never fail were replaced with trace-table checks that can; two orphaned helpers were removed.
- Three tests are xfailed (`strict=False`), each narrowed with `raises=`: two for TASK-22690 (closing a chat mid-turn raises "Durable continuation owner changed"), one pre-existing TASK-22720 (agent bridge failure is swallowed). TASK-22690 doesn't reproduce on `dev` — the raise only happens on this PR's durable path — so its fix is blocked on this merge.
- Mutation checks confirmed the new seams are observed: mis-sending the assistant row fails 31 tests, skipping the citation trace write fails 6.

<sup>Written for commit 35f43ce35089acbd5ef826467e5cec0c36cb9b15. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2128?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## All five findings addressed — and #5 paid for itself immediately

### 5. Broad xfail masks regressions 🐞 — the valuable one

Correct, and narrowing it **found a real problem in my own code**. Adding
`raises=` immediately turned one xfail into a failure, because that test was
raising `AttributeError`, not the `AssertionError` I had documented — the
AttributeError came from the test's own bookkeeping (`bridge.replacement_id`
touched when the bridge never set it). The broad marker was hiding the actual
exception type, exactly as described.

Both markers now carry `raises=`, so failures in privacy cleanup, session
removal, message resurrection or cancellation still fail CI. The bridge declares
`replacement_id` so its failure reads as an assertion rather than an
AttributeError from the harness.

### 4. SQLite database is file-backed 📘

There was a better answer than either option. A **shared-cache in-memory URI with
a unique name per instance** is in-memory — satisfying the rule, with no temp
files and nothing to clean up — *and* visible across threads, which is the
constraint that forced the file in the first place (`CharactersRAGDB` opens a
thread-local connection and TASK-22205 offloads durable DB calls to a worker
thread, so a plain `":memory:"` gives that worker its own empty database;
measured: traces wrote successfully and `rag_citation_traces` still read 0 rows).

Cross-thread visibility was verified before adopting it. The uuid matters: a
shared cache is keyed by NAME, so a fixed name would silently join every test to
one database.

### 2. Persistence resources never cleaned 📘

Fixed, and the in-memory switch makes it load-bearing rather than cosmetic: a
shared-cache memory database lives exactly as long as one connection to it stays
open, so a leaked connection leaks for the whole session. An autouse fixture now
closes every persistence the tests open.

### 3. Row queries bypass `transaction()` 📘

Both row helpers now read through `db.transaction()`.

### 1. Standard-library imports split 📘

Folded back into the single stdlib group.

## Verification after the changes

Re-ran the full A/B against the same merge-base `c23113e2e0` (baseline unchanged
and verified byte-identical):

| | merge-base `c23113e2e0` | this branch |
|---|---|---|
| collected | 7688 | 7688 (**+0**) |
| failures | 108 | **68** |
| fixed | — | **40** |
| newly broken | — | **0** |

File itself: 92 passed, 3 xfailed, 0 failed. `./scripts/preflight.sh` green.

rmusser01:
## Correction: my fix for finding #4 was wrong, and it broke the Windows checks

The two `FAILURE`s on this PR — `GGUF import evidence - windows-latest` and
`GGUF source evidence - windows-latest` — were **caused by my previous commit**,
not by anything pre-existing. Recording what happened, because the failure mode
is not obvious:

`CharactersRAGDB` never passes `uri=True` to `sqlite3.connect`. So
`"file:<name>?mode=memory&cache=shared"` is not interpreted as a URI at all — it
is taken as a **literal filename**. The tests silently created ~200 real files in
the repo root whose names contain `?` and `:`, and `git add -A` committed 137 of
them. Those characters are illegal in Windows filenames, so checkout failed on
the Windows legs. Nothing about the citation tests was involved.

**The probe that convinced me was misleading.** I verified cross-thread
visibility before adopting the URI and it passed — but it passed because the
"database" was an ordinary file on disk, not because a shared cache existed. A
probe that cannot distinguish the mechanism under test from the mechanism it
replaces confirms nothing.

All 204 files are removed, and a full run of the file now creates none.

### So finding #4 cannot be satisfied as written

Not a disagreement about the rule — an interpreter-level constraint:

- `":memory:"` is **thread-local**. TASK-22205 offloads durable DB calls to a
  worker thread, which then gets its own empty database. Measured: citation
  traces wrote successfully and `rag_citation_traces` still read **0 rows**.
- A shared-cache in-memory URI is the standard escape hatch, and it is
  **unavailable** here until `CharactersRAGDB` opts into `uri=True`.

A file in a temp directory is the only option that actually works. It is now
explicitly justified in-code, with the failed URI attempt recorded so the next
reader does not repeat it.

### Finding #2 is still fully addressed

The autouse fixture closes the connection **and** removes the temp directory.
Measured across a full run of the file: **0 directories leaked** (counted before
and after, not assumed).

### State

92 passed, 3 xfailed, 0 failed. `./scripts/preflight.sh` green. The A/B result is
unchanged: 40 fixed, **0 newly broken**, +0 collected against `c23113e2e0`.

PR #2129, opened 2026-08-27
perf(library): virtualize the media reader Raw body (TASK-22500)

rmusser01:
## Summary

Virtualizes the Library media reader's **Raw** body so render cost is proportional to the
viewport rather than to document length. This was the headline follow-up finding from the
2026-08-24 holistic performance review (TASK-22500) and the largest remaining reader win:
the raw `Static` was auto-height inside a `VerticalScroll`, so a 2.5 MB document gave the
widget `height=45000` and Textual re-rendered **all 45,000 lines on every repaint**.

## Measured results

Instrumented by `render_line` self-time and call counts on a 2.5 MB / 45,000-line document —
never harness wall time, which is dominated by `pilot.pause()`.

| Interaction | Before | After |
|---|---|---|
| First paint | 2,283 ms | **0.81–0.90 ms** (+ one-time 141–145 ms index build) |
| Each search keystroke | 1,709–1,777 ms | **0.59–0.66 ms** |
| Each match-nav click | 1,709–1,777 ms | **0.59–0.66 ms** |

The index build is one-time per document/width, and re-indexing is debounced (0.12 s) so a
resize drag builds once rather than per frame.

## Approach

`VirtualizedRawContent(ScrollView)` implements `render_line(y)` against an exact wrap index,
so only viewport rows are rendered.

- **`Utils/text_wrap_index.py`** — pure wrap arithmetic (rich only, no Textual). Deliberately
  *not* under `Library/`, because `Library/__init__` is an eager 66-module service-stack
  import (see TASK-22223). Adapts `rich._wrap.divide_line` with a covered `Text.wrap`
  fallback for the day that private API moves.
- **`Widgets/Library/library_media_raw_view.py`** — the widget: search highlighting across
  wrap boundaries, drag-selection painting, and `get_selection` returning **literal tabs**
  (the wrap index expands tabs for width math; copied text must not inherit that).
- **`library_media_content.py`** — `LibraryMediaContentBody` becomes a `Container` hosting
  mode-specific scrollers, exposing `scroller` / `raw_view` / `active_mode`. The
  whole-document highlight `Text` and its plan builder are deleted; match-line derivation is
  memoized on content identity + stripped query.
- **`library_screen.py`** — the three scroller lookups (progress capture, progress restore,
  scroll-to-line) resolve through `body.scroller`, and match scrolling is gated on raw mode
  so it can't scroll a hidden view.

## Fidelity

`Tests/Library/test_library_media_raw_view.py` pins the widget's output against the `Static`
it replaces, character-for-character, including the empty-content substitution the production
body applies.

## Testing

- **78 passed** across the affected surface: widget, wrap index, content body, the three
  reader gates (traversal / no-change-sync / match-nav), scroller resolution, and the
  recompose ratchet (still pinned at 74).
- `./scripts/preflight.sh`: all derived-artifact checks pass.
- Repo-wide collect sweep clean (59,771 tests); the 28 collection errors are pre-existing
  missing optional deps (numpy, pydub, playwright, websockets), none touching these files.
- `test_library_shell.py`'s pre-existing failures were A/B'd individually against the merge
  base: **no new failures**.

## Notes for review

- `RENDER_LINE_CALLS` is test instrumentation living in production code — deliberate (it is
  how the measurements above are taken and gated), flagged as a cosmetic follow-up.
- `source_lines` (tab-expanded, drives wrap) vs `_raw_lines` (literal tabs, drives selection)
  is a naming trap worth a second look.

Follow-up TASK-22660 covers the Markdown view, which has a different shape (one widget per
block) and was measured but deliberately left out of scope.

---
## Summary by cubic
Virtualizes the Library media reader's Raw body so render cost scales with the viewport, not document length: on a 2.5 MB document, first paint drops from ~2.3 s to under 1 ms of render cost, and each search keystroke or Prev/Next match click from ~1.7 s to under 1 ms. A one-time index build is still paid per document and width, debounced across a resize burst, and a short-line fast path keeps it at ~2.5 ms even at 2.5 MB.

- Replaces the whole-document `Static` with a `ScrollView` that paints only visible rows through an exact wrap index; match-line derivation is memoized per content and query and seeded on mount so an already-active search is styled before any `sync_search`.
- Search highlighting (including matches straddling a wrap boundary), drag-select painting, and copied text (literal tabs preserved) behave as before.
- Fixes three reader scroller lookups that silently no-opped because the content body is now a `Container`, gates match scrolling on the active mode, and routes matches through the wrap index.
- Corrects three rendering regressions found in review: over-declared wide-glyph rows, an unreachable last row, and wrapped rows truncated once the scrollbar appeared.
- Restores a `VerticalScroll` import the rebase orphaned, which raised `NameError` on emergency-geometry Library routes.
- Keeps the rendered Markdown a direct child of the content body; nesting it deferred mount and broke the document-loaded event.
- Adapts the shell test that still read `Static.renderable` off the raw view to read through the virtualized widget.

The Markdown view is out of scope and tracked in TASK-22660.

<sup>Written for commit bfb687fa06099968766f5e569c6b5fa970843758. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2129?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
**Heads-up on a pre-existing red, so a failing gate here isn't misattributed to this diff.**

After rebasing onto current dev (`4cee590b83`), `Tests/UI/test_library_recompose_ratchet.py`
fails at **75 sites found, 74 allowed**. It is **not this branch's**:

- This branch's `library_screen.py` diff adds **zero** recompose sites (46 lines, all scroller
  lookups and a mode gate).
- Swapping dev's own pristine `library_screen.py` into this branch reproduces the failure
  identically at 75/74.
- An AST census diff against the fork point names exactly one added entry:
  `_sync_library_canvas -> screen.refresh(recompose=True)` — a third whole-screen recompose in
  a function that previously held two, from the adaptive-reader / Prompts migration.

Filed as **TASK-22888** in #2130. I did not bump the pin to 75 here: absorbing another
change's regression into the ratchet is precisely what the ratchet exists to prevent.

Everything else is green on the rebased tree: 78 passed across the affected surface,
`./scripts/preflight.sh` all checks passed.

rmusser01:
**Review round applied — including one regression this branch caused, found by a full-suite A/B.**

### The regression (fixed)

The rebase silently orphaned an import. This branch stopped importing `VerticalScroll` in
`library_screen.py` (legitimately — the reader body became a `Container` and scroller lookups
moved to `body.scroller`), while dev's adaptive-reader work added
`_capture_library_emergency_restore_receipt`, which uses it in three places. Neither side
touched the other's lines, so git reported no conflict and the result raised `NameError` from
`on_resize` on any Library route with emergency geometry.

**Cost: 79 shell tests, none of them reader tests.** Every gate this PR had been reporting
stayed green — the affected surface, preflight, all reader gates — because none of them mount
the notes/emergency routes that reach that line.

### Attribution (deterministic A/B vs a pristine merge-base worktree)

| | failed | passed |
|---|---|---|
| base `4cee590b83` | 96 | 728 |
| branch, before the import fix | 174-175 | 649-650 |
| branch, after | 97 | 727 |

Name-set diff after the fix: **1 branch-only**, 0 base-only, 96 identical. That one
(`test_library_conversation_real_memory_scope_load_and_locator_stay_on_owner_thread`) is a
pre-existing flake — isolated, it fails 2/4 on this branch and 3/4 on base.

Two traps worth recording, both of which nearly hid this:
- **Counts are not comparable.** `pytest-randomly` reshuffles per run; two runs of the *same*
  branch tree gave 174 and 175. Both sides need `-p no:randomly`.
- **Sampling pointed the wrong way.** Five sampled failures all reproduced on pristine dev, so
  they read as "dev's problem" — they were, but they were in the 96 shared. The set diff is
  what exposed the 79 that were ours.

### The three Critical rendering bugs from the review (fixed)

- wide glyphs emitted 43-46 cell rows into a 40-cell screen (`Strip` given a CHARACTER count
  where a CELL count is required, so `adjust_cell_length` padded instead of truncating)
- the last row of every long document was unreachable (`max-height` is an OUTER bound whose
  border rows the child cannot have, while `ScrollView` still computed `max_scroll_y` from the
  height it thought it had)
- every full-width wrapped row was cut by 2 columns (index built at the width measured before
  the VERTICAL scrollbar exists; no `Resize` fires when a scrollbar shrinks the content region)

Each fix is mutation-tested — the bug reintroduced, the new guard confirmed red, then restored.
The old fidelity test could not catch any of them: it compared `strip.text.rstrip()`, which is
blind to cell length, to the parent's box, and to reachability.

Also: select-all copy on a 2.5 MB single-line document **5,713 ms → 28.9 ms** (prefix sums,
now linear); `Text.wrap` fallback offsets corrected (13/14 exact, up from 4/9 — exact parity is
unreachable through the public API, so its contract is now "never lose a character, never
outrun the width"); the intermittent scroller test's mechanism closed (8/8 serial, 6/6
concurrent).

Both lessons are written up in `backlog/docs/lessons-testing-evidence.md` with the incidents
attached.

Note the `test_library_recompose_ratchet` red is still dev's, not this branch's — TASK-22888 / #2130.

rmusser01:
**Both Qodo findings addressed.** Commits: `c3cd4831f3` (finding 1), `911edf9175` (finding 2).

### 1. `WrapIndex.build` blocks the UI thread — valid, fixed

The finding is right, and worse than stated at the tail. Measured build cost by document size:

| document | before | after |
|---|---|---|
| 100 KB | 5.4 ms | 0.1 ms |
| 1 MB | 55.5 ms | 0.9 ms |
| 2.5 MB | 140.4 ms | 2.5 ms |
| 6.5 MB | 363.5 ms | 7.4 ms |

It crossed the 100 ms rule at roughly 1.8 MB — exactly the documents this feature exists for.

Fixed by making it **fast rather than asynchronous**: a short pure-ASCII line cannot wrap
(every ASCII character is exactly one cell), so it never needs a Rich `divide_line` call.
Heights are identical on mixed content, and anything that *can* wrap — long lines, and any
line containing a non-ASCII (possibly wide or zero-width) glyph — still goes through Rich
unchanged. A CJK document still wraps by cell width.

I chose this over the suggested background worker deliberately: offloading would show an
empty body until the build lands, and would add races to a widget that has already produced
two mount-ordering defects on this branch. If you'd still prefer the worker, say so and I'll
do it — but at 2.5 ms for a 2.5 MB document there is no longer a threshold to cross.

Pinned by **call count**, not wall clock (the 15457 probe rule): `WrapIndex.build` over a
5,000-line ASCII document must cost zero `divide_line` calls. Mutation-tested — disabling the
fast path reds it at 5,000 calls. (First attempt at that mutation silently missed on
indentation and reported a false pass; re-run with an assert on the pattern.)

### 2. Wide glyphs corrupt row width — already fixed earlier in this PR, now also pinned

`Strip(rendered, len(piece))` is the line the review quotes, but that is a **diff position from
an earlier commit in this same PR**; a later commit already corrected it. Current code at
`library_media_raw_view.py:407` is `Strip(rendered)` — Strip measures its own segments — and it
is at `e05059e413` too. This was found independently and fixed with the same diagnosis (a
CHARACTER count where a CELL count is required, so `adjust_cell_length` padded instead of
truncating: 43-46 cell rows into a 40-cell screen).

The **selection** half of the finding was untested, and that part was worth chasing. Selection
columns are cell columns while the raw line is indexed by character, so I measured it rather
than assumed: real mouse drags over CJK documents already return **byte-identical text to the
`Static` this widget replaces**, including releases landing on the *second* cell of a 2-cell
glyph (columns 3-6 each checked) and SELECT_ALL across mixed wide/ASCII lines. Matching Static
is the contract — it resolves the same coordinates through the same compositor — so there is
nothing to change, but nothing pinned it either. Three tests now do, and rewriting
`get_selection` to walk cell widths instead reds all three.

Also worth noting for anyone reading the diff: the rendered Markdown lost its `VerticalScroll`
wrapper in `84e23769b7`. That was not cosmetic — the extra nesting level delayed the Markdown's
attach by one message cycle, which on a 1 MB document *is* the whole parse, and it made
TASK-22207's gate fail about half the time (2/4, then 3/4; dev was 6/6). Rendered now scrolls in
the body itself, as it did before this PR.

PR #2130, opened 2026-08-27
docs(backlog): retitle TASK-22888 to the real finding, not the resolved red gate

rmusser01:
Files TASK-22888. Task file only — no code change.

`Tests/UI/test_library_recompose_ratchet.py::test_library_screen_whole_screen_recompose_count_is_ratcheted`
**fails on pristine `dev` (`4cee590b83`)**: 75 whole-screen recompose sites found, 74 allowed.

The 75th is a third `screen.refresh(recompose=True)` inside `_sync_library_canvas()` (it
previously held two), added by the adaptive-reader / Prompts-migration series without moving
the pin.

**Attribution:** I swapped dev's own pristine `library_screen.py` into an unrelated feature
branch and re-ran the single test — it fails identically at 75/74, so the site is dev's. An
AST census diff against `732105c2d` reports exactly one added entry:
`_sync_library_canvas -> screen.refresh(recompose=True)`.

The task deliberately does **not** prescribe bumping the pin to 75: that would absorb the
regression and spend the ratchet's purpose. The call is whether this whole-screen recompose
on a *sync* path is necessary, and the pin moves only if it is.

Found while rebasing #2129 (TASK-22500) onto current dev; that PR's gate is red for this
pre-existing reason, not for anything in its own diff.

---
## Summary by cubic
Files TASK-22888 as a task-only PR (no code change). When filed, the recompose ratchet failed on pristine `dev` at 75 whole-screen recompose sites vs 74 allowed; against `251ea46e11` it now passes with the pin still at 74, because dev removed whole-screen recomposes elsewhere rather than assessing the three calls in `_sync_library_canvas()`.

**Key points**
- A third `screen.refresh(recompose=True)` in `_sync_library_canvas()` arrived with the adaptive-reader/Prompts migration without moving the pin.
- The call count moved twice while the task sat open, so a prior review seeing two calls was accurate for the state it read — re-run the census before acting.
- The task deliberately avoids bumping the pin to 75; each call should be narrowed to a canvas-scoped refresh or kept with the reason recorded, and the pin should drop rather than bank headroom if any call is narrowed.

<sup>Written for commit 8416c4923eb9ea8f70241421120e7c390a88f4a9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2130?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
**Update: the red is gone, so this is no longer a blocker — but the finding itself still stands.**

Re-checked against current dev (`5d9b4bec5a`): `test_library_recompose_ratchet.py` now **passes**
(4 passed), with the pin still at 74 and the census down to ~71 sites.

What did *not* happen is the thing this task asked for: `_sync_library_canvas()` still holds
**three** whole-screen `screen.refresh(recompose=True)` calls. Dev came back under the pin by
removing sites elsewhere, so the third recompose on a canvas *sync* path was effectively paid
for out of other savings rather than assessed.

So the acceptance criteria are unchanged and still worth someone's judgement — is a whole-screen
recompose necessary there, or should it be canvas-scoped (the TASK-15457 family)? — but this is
now a **performance question, not a broken gate**. Dropping it from "blocks every Library PR" to
ordinary follow-up priority.

Recorded here rather than silently closing the PR, since the recompose is still in the code.

rmusser01:
**Qodo's finding addressed — and it was right for the state it read.**

> The task says `_sync_library_canvas()` contains a third `screen.refresh(recompose=True)`, but
> the current function contains only two counted calls.

Re-measured against dev `251ea46e11`:

- **The ratchet now PASSES** (4 passed, pin still 74). Dev came back under the pin by removing
  whole-screen recomposes *elsewhere*, so the calls in this function were paid for out of other
  savings rather than assessed. The red gate this task was filed for is gone.
- **The function does still hold three real calls** — `library_screen.py:1917`, `:1932`, `:1993`.
  Two further textual matches in that function are prose in a comment and a docstring, which is
  the likely source of a differing count.
- The census moved **twice** while this task sat open (75 → under 74), so a snapshot taken at
  review time can legitimately disagree with both the original filing and today.

So the finding is valid against the file as written: it pointed at a red gate that no longer
exists, and gave line numbers that had drifted. Rewritten rather than closed, because the actual
performance question survives:

- title and filename no longer claim a red gate (`Third whole-screen recompose in
  _sync_library_canvas should be canvas-scoped`)
- the original attribution evidence is preserved in a **History** section, clearly marked as
  filed-then-downgraded
- priority **high → medium**; it is a performance question now, not a blocker
- ACs cover all three calls, and add: *if any is narrowed, the pin drops to match the new census
  rather than staying at 74* — the headroom is not banked
- the description tells the next person to locate the calls **by function, not by line**, and to
  re-run the census rather than trust anything written here

Rebased onto latest dev (was 229 behind). Backlog task-ID check green: no duplicates across 2,724
task files, all paths Windows-compatible.

PR #2131, opened 2026-08-27
Wizard/Console review follow-ups + dev's CSS guard breakage (TASK-21150, TASK-22858)

rmusser01:
## Summary

Burns down the four deferred follow-ups from PR #2101's code review, and fixes the class-level CSS guard that is currently failing on `dev` itself.

### TASK-21150 (a–d)

- **(c) Every `resolve_for_send` await is now bounded.** There were **four**, not the three the review found — `dev` added a dispatch-retry path since. Continuation replay, dispatch retry, the instruction preview and compaction now share the send path's deadline, so none can wedge its surface the way UAT H-3 wedged the composer. Each site already handled a not-ready resolution, so the timeout stand-in degrades correctly.
  Guarded two ways: an **AST test** that fails on any raw `await self.provider_gateway.resolve_for_send(...)` outside the bounded helper — the real hazard is the *next* one someone writes, which only source text can catch — plus a behavioral test proving `compact_context_now` returns actionable copy against a hanging gateway.
- **(d) The markup invariant is mechanical, not commentary.** The composer's setup-blocked action link no longer f-string-interpolates the reason into markup; it passes it as a `Content.from_markup` **template variable**, so the value is inserted as text and never re-parsed. The new test feeds a hostile reason carrying `[@click=app.quit]` and `[bold]` and pins that both render literally with no injected action span. It **failed before this change** — the old invariant was genuinely injectable, just not yet exploited.
- **(a) Consent "yes" refreshes this session.** `SummaryStep.commit`'s allow path calls the new `app.refresh_model_catalogs_now`, dispatching the same worker in the same exclusive `model-catalog-refresh` group the Console modal uses, so the two can never race. Failure is non-fatal (the answer is already saved). Deny still never reaches the network — pinned in both directions.
- **(b) Already correct.** `_theme_buttons` always pressed the match and `_card_buttons` gained the same in PR #2101's Qodo pass. No code change invented; added the test that proves a selection survives both show-all rebuilds and locks it in.

### TASK-22858 — a breakage inherited from `dev`

`Tests/UI/test_widget_css_consolidation.py::test_class_level_css_stays_within_the_allowlist` fails on pristine `dev` (verified in a clean worktree): commit `6161bd1fe` added `LibraryEmergencyReturn.DEFAULT_CSS`, and each class-level declaration burns one of Textual's 64 parse-cache slots, past which every first mount of an unseen class re-pays a ~150–450 ms cold parse. It is a plain string literal, so the default sanctioned route applied cleanly: renamed to `BUNDLED_CSS`, sheets regenerated. That suite is green again (33 passed).

## Verification

- **1102 passed, 0 failed** across the wizard, live-contract, send-disabled, CSS-consolidation, console-internals, first-chat-handoff and library-accessibility suites.
- `preflight.sh` green. The one new persistent diagnostic (a static `logger.debug` with no interpolated content) was reviewed statement-level via `--statements --since origin/dev` and recorded in the inventory.

https://claude.ai/code/session_01KBSWqTEWDX34iaSNwemF3o

---
## Summary by cubic
Closes the four deferred review follow-ups from PR #2101 (TASK-21150) and fixes the class-level CSS parse-cache guard that was failing on dev (TASK-22858); item (b) needed no code change and is now locked in by a test.

**New Features**
- Answering "yes" to the Summary's model-list consent now refreshes the provider catalogs this session instead of only recording consent, so the first session stops running on stale lists.
- The refresh dispatches the same exclusive worker the Console modal uses, now keyed off a shared `MODEL_CATALOG_REFRESH_WORKER_GROUP` constant; tests drive the real wizard→app chain, and deny still never touches the network.

**Bug Fixes**
- Continuation replay, dispatch retry, instruction preview, and compaction now all route `resolve_for_send` through the bounded seam, so no surface can hang the way UAT H-3 wedged the composer.
- A continuation timeout now shows the timeout guidance instead of being mislabeled "Prepared destination changed."
- The setup-blocked reason is now inserted as a `Content.from_markup` template variable rather than interpolated markup, so reason text renders literally and can't inject styling or actions; new tests feed `[@click=app.quit]` and `[bold]` to pin this.
- `LibraryEmergencyReturn` and `GlobalFullCaptureConfirmation` moved from `DEFAULT_CSS` to `BUNDLED_CSS` with regenerated sheets, clearing the parse-cache guard that failed on pristine dev.
- An AST test fails on any raw `resolve_for_send` await outside the helper, and a behavioral test proves compaction answers against a hanging gateway.

<sup>Written for commit 3211babe13162a8bccc885b5bb5549bbf345f164. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2131?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
**All three Qodo findings addressed** in `43dfc39b9`.

**🐞 Timeout mislabeled as destination change — valid, and a second-order consequence of this PR's own fix.** Bounding continuation replay made `_resolve_for_send_bounded` *return* a not-ready stand-in on timeout instead of raising, but the caller tested readiness and destination identity in one combined condition — so every timeout surfaced as "Prepared destination changed." with the resolver's actionable guidance discarded. Split into a pure classifier (`_prepared_continuation_block_copy`) with four unit tests written first: not-ready surfaces the resolver's own copy through the existing sanitizer; a blank copy still explains itself without leaking an internal code; a genuine destination change keeps its message; an unchanged destination proceeds. The pause kind is unchanged, so the prepared turn stays retryable.

**📘 Duplicate catalog worker group — valid.** `"model-catalog-refresh"` was hand-typed at four sites (three dispatches plus the worker handler's acknowledgement set), and this branch added the fourth. Exclusivity *and* event routing both key off that exact string, so drift breaks either silently. Now one `MODEL_CATALOG_REFRESH_WORKER_GROUP` constant in `Constants.py`, imported at every site.

**📘 Refresh seam lacks integration test — valid.** The consent test stubbed `request_model_catalog_refresh`, proving only that `commit` calls *a* callback. Added two tests through the real boundary: one drives the app's actual `refresh_model_catalogs_now` and asserts exclusive dispatch, the shared group constant, and the scheduled-flag update; the other exercises the genuine wizard → container → app chain with no stub in the path.

**Verification:** 998 passed / 0 failed post-fix (1091 passed on the pre-fix rebased head); `preflight.sh` green. The one collection error seen mid-sweep is an unrelated missing `numpy` in `Tests/Event_Handlers/Chat_Events/test_chat_image_properties.py`, not from this branch.

https://claude.ai/code/session_01KBSWqTEWDX34iaSNwemF3o

PR #2135, opened 2026-08-27
fix(console): closing a chat is not an owner change (TASK-22690)

rmusser01:
## Summary

Closing a Console chat while a durable turn's postcommit sequence was running
raised `RuntimeError("Durable continuation owner changed.")`.

This is the **fourth site** of the conflation TASK-22587 removed, and the last one
blocking two tests that #2128 left `xfail`.

## The defect

The tail of `resume_durable_postcommit` refuses to settle a continuation whose
**owner changed** underneath it. That guard is correct — a continuation must not
be settled by the wrong turn. What it did not distinguish is *"the owner is
legitimately **gone** because the user closed the chat"*, which is ordinary, from
*"the owner **changed**"*, which is a bug. Both raised.

## My filed hypothesis was wrong, and measurement refuted it

The task predicted a close routed through
`discard_uncommitted_durable_preparation`, which leaves **no** tombstone — so
TASK-22587's benign-retirement path would never engage.

Instrumenting the actual raise showed the opposite:

```
continuation_present: False
live_fp:              False
tombstone_present:    True
tombstone_matches:    True   <-- the close DID go through retire_durable_acceptance
```

The tombstone was there all along. Implementing against the filed guess would
have produced a fix for a mechanism that does not exist. It was recorded as a
hypothesis to verify rather than a diagnosis, and verifying it is what caught it.

## Fix

Same decision procedure as TASK-22587: `retire_durable_acceptance` leaves a
tombstone carrying the **same** fingerprint, so a matching tombstone proves the
preparation was retired by a close rather than replaced.

- New public `ConsoleChatStore.durable_acceptance_retired(preparation_id, fingerprint)`.
- The owner check takes TASK-22587's existing terminal-benign path when the
  continuation is absent **and** retirement is proven; it still raises otherwise.

Also aligns the closed-session copy with the generic `"Session closed."` every
other close site uses. TASK-22587 left it `""`, which reads to a caller as
"no message" rather than "the session closed" — and a test asserting that copy is
what surfaced it.

## Mutation-proven

| mutation | result |
|---|---|
| remove the benign path | **2 failed** |
| make retirement ignore the fingerprint (negative control) | **1 failed** |

The negative control is the load-bearing one: retirement must key on **this**
acceptance, not on the id merely having a tombstone, or an owner that genuinely
changed would be waved through as an ordinary close — precisely what the guard
exists to prevent.

## Effect on the xfail backlog

Both TASK-22690 markers removed. `test_console_local_citation_boundary` goes
**94 passed / 1 xfailed**; only TASK-22720 remains.

## Verification

A/B of `Tests/Chat` against the merge-base, baseline verified byte-identical:

| | merge-base | this branch |
|---|---|---|
| collected | 7777 | 7780 (**+3**, the new tests) |
| failures | 74 | **74** |
| newly broken | — | **0** |

`0 fixed` is expected: the two recovered tests were `xfail` on the baseline, so
neither side counts them as failures — their recovery shows as the file's xfail
count dropping from 3 to 1.

TASK-22587's own 16 tests still pass. `./scripts/preflight.sh` green.

---
## Summary by cubic
Fixes closing a chat during an in-flight durable postcommit raising `RuntimeError("Durable continuation owner changed.")`. The postcommit guard treated an owner that legitimately went away (user closed the chat) the same as an owner that changed; now a matching tombstone from `retire_durable_acceptance` proves the close, so the guard takes the benign path while a genuinely changed owner still raises. Adds `ConsoleChatStore.durable_acceptance_retired` to expose that check, unifies the six "Session closed." sites behind `SESSION_CLOSED_COPY`, and removes two xfail markers.

<sup>Written for commit d2618204dc2ee08096bf91fa1ffb1807f5882f02. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2135?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Chat/test_console_close_during_postcommit_owner.py
+++ b/Tests/Chat/test_console_close_during_postcommit_owner.py
@@ -1,0 +1,58 @@
+"""Closing a chat must not read as an owner change (TASK-22690).
+
+The tail of `resume_durable_postcommit` refuses to settle a continuation whose
+OWNER changed underneath it. That guard is right, but it treated two different
+events as one: an owner that genuinely changed (a bug) and an owner that is
+legitimately GONE because the user closed the chat (ordinary). Closing mid-turn
+therefore raised `RuntimeError("Durable continuation owner changed.")`.
+
+This is the fourth site of the conflation TASK-22587 removed, and it is decided
+the same way: `retire_durable_acceptance` leaves a tombstone carrying the SAME
+fingerprint, so a matching tombstone proves the preparation was retired by a
+close rather than replaced.
+
+Measured at the raise before fixing anything: continuation absent, live
+fingerprint gone, tombstone present AND matching -- so the close had gone
+through `retire_durable_acceptance`, not the discard path.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from Tests.Chat.test_console_close_during_durable_postcommit import _claimed_effect
+
+
+def test_a_retired_acceptance_is_reported_as_retired(tmp_path) -> None:
+    """The public predicate the owner check now consults."""
+
+    store, prep, fingerprint = _claimed_effect(tmp_path, claim=False)
+    store.close_session("session-1")
+
+    assert store.durable_acceptance_retired(prep, fingerprint) is True
+
+
+def test_a_different_acceptance_is_not_reported_as_retired(tmp_path) -> None:
+    """NEGATIVE CONTROL.
+
+    Retirement must be keyed on THIS acceptance, not on the preparation id
+    merely having a tombstone -- otherwise an owner that genuinely changed
+    would be waved through as an ordinary close, which is exactly the failure
+    the guard exists to prevent.
+    """
+
+    from dataclasses import replace
+
+    store, prep, fingerprint = _claimed_effect(tmp_path, claim=False)
+    store.close_session("session-1")
+    other = replace(fingerprint, assistant_message_id="a-different-turn")
+
+    assert store.durable_acceptance_retired(prep, other) is False
+
+
+def test_an_unknown_preparation_is_not_reported_as_retired(tmp_path) -> None:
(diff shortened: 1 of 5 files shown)


rmusser01:
## Both findings fixed

**1. `"Session closed."` duplicated** — it was at **six** sites, including the one
this PR added. Now a single `SESSION_CLOSED_COPY` constant, following the
convention already used by `CONSOLE_CONTINUE_INSTRUCTION` and friends in the same
module. Worth more than tidiness: six independent copies of user-facing text is
exactly how a later "fix the wording" change lands on five of them.

**2. `durable_acceptance_retired` docstring** — Args/Returns added per the repo's
Google style. The Args entry carries the reason matching matters: a tombstone
under the same id for a **different** acceptance is an owner change, not a close,
which is the distinction the function exists to draw.

## Verified, because a six-site string replacement can change behaviour silently

Not just the three touched files — the whole of `Tests/Chat`:

- `SESSION_CLOSED_COPY == "Session closed."` asserted byte-identical to the
  literal it replaced
- no stray literal left at any use site (only the one inside an explanatory
  comment, which is describing the copy rather than emitting it)

| | merge-base | this branch |
|---|---|---|
| collected | 7777 | 7780 (**+3**) |
| failures | 74 | **74** |
| newly broken | — | **0** |

`./scripts/preflight.sh` green; the three close-related test files are
113 passed / 1 xfailed (the remaining xfail is TASK-22720).

PR #2146, opened 2026-08-27
test(citations): the settlement-failure None is not data loss (TASK-22617)

rmusser01:
## Summary

TASK-22617 asked which of two readings is right for the hard-coded
`terminal_citation_finalizer=None` in `resume_durable_postcommit`'s
`except BaseException:` arm — the TASK-22302 data-loss class, or a deliberate
guard against attributing unsent content. The task's own instruction was to
establish it **by test before changing anything**.

**Verdict: not data loss — and not quite either filed reading.** One comment in
the product (no behaviour change), three tests, three mutations all red.

## The mechanism, measured rather than read

`durable_owner_publication` — the effect that **arms** the finalizer — also
**registers the session's dispatch recovery** (`publish_durable_turn_owners` →
`publish_durable_dispatch_checkpoint`). The failure arm gates its recovery
publish on `dispatch_recovery_for_session(...) is None`, so:

- **The None publish is unreachable once a finalizer is armed.** A spy on
  `publish_durable_recovery_owner` confirmed it fires exactly once across both
  failure orderings: in the *pre-arming* one.
- In that one reachable ordering nothing was ever armed, the continuation
  survives the failure, and the resume's owner publication forwards
  `continuation.terminal_citation_finalizer` (TASK-22302) — the trace persists
  on retry. Both orderings end with `rag_citation_traces == 1` on real SQLite.
- Forwarding the finalizer at the None site would arm a deferred finalizer on a
  turn whose delivery is unknown — the provenance hazard reading 2 feared — for
  no benefit.

## Mutations

| mutation | result |
|---|---|
| M1: a None publish CLEARS armed state | **RED** (store-level stickiness pin) |
| M2: the resume publish drops the finalizer | **RED** (re-proves the TASK-22302 forward is load-bearing) |
| M3: owner publication stops registering the recovery | **RED** — this is what makes "unreachable while armed" falsifiable rather than folklore |

## Two of my own assertions died en route

Reported because they are the interesting part:

1. The fail-after-arming test first claimed to guard a clobber its scenario can
   never reach — **M1 stayed green** and exposed it.
2. Its replacement asserted `dispatch_recovery_for_session(...) is not None`
   after the failed submit — mis-timed, because the failure arm *restores* a
   recovery after its publish gate, so that check is true in every ordering.
   **M3 stayed green** against it.

Both replaced with a call spy on the publish itself, with the pre-arming test
asserting the spy **does** fire — the positive control that keeps the sibling's
zero-calls assertion falsifiable.

## Verification

A/B of `Tests/Chat` against the merge-base, baseline verified byte-identical:

| | merge-base | this branch |
|---|---|---|
| collected | 8798 | 8801 (**+3**) |
| failures | 94 | **94** |
| newly broken | — | **0** |

`test_console_terminal_citation_persistence.py` itself: 96 passed. Preflight
green. The None site now carries the evidence-backed reason in-code (AC4), so
this does not get filed a third time.

---
## Summary by cubic
Settles TASK-22617 by test: the hard-coded `terminal_citation_finalizer=None` in `resume_durable_postcommit`'s settlement-failure arm is not data loss. No behavior changes; the diff adds four tests pinning the mechanism, a comment documenting the verdict at the None site, and marks the task Done.

**Evidence pinned**

- The failure arm's None publish is unreachable once a finalizer is armed, because the effect that arms the finalizer also registers the session's dispatch recovery, gating that publish off.
- Three mutations went red, including one that removes the recovery registration, making the "unreachable while armed" claim falsifiable.
- Two earlier assertions proved vacuous and were replaced with a call spy, with the pre-arming test acting as the positive control.
- Inside owner publication there is a window where arming precedes recovery registration; a failure there does reach the None publish while armed, and a fourth test pins that the store's non-clearing contract is load-bearing in that window.

<sup>Written for commit ca3c10caee54cd2b477c81ee83827c8b37458813. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2146?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Chat/test_console_terminal_citation_persistence.py
+++ b/Tests/Chat/test_console_terminal_citation_persistence.py
@@ -2340,9 +2340,14 @@ def test_one_terminal_write_ready_finalizer_fails_closed_to_ordinary_message() -
 #     was ever armed, the continuation survives the failure, and the resume's
 #     owner-publication effect forwards `continuation.terminal_citation_
 #     finalizer` (TASK-22302) -- re-arming and persisting the trace.
-#   - Defence in depth, pinned at store level below: even if a None publish
-#     DID run while armed, `_hydrate_durable_turn_owner_messages` never clears
-#     armed state on None (the arming `if` is simply skipped).
+#   - The unreachability above holds only at EFFECT granularity. Inside owner
+#     publication itself, arming precedes the checkpoint call that registers
+#     the recovery, so a failure in that window DOES reach the None publish
+#     with a finalizer armed (Qodo caught this ordering gap on #2146; the
+#     window test below confirms it empirically). There the store's
+#     non-clearing contract is load-bearing, not defence in depth: a None
+#     publish never clears armed state, only declines to arm -- pinned both
+#     through the realistic window and directly at store level.
 # ---------------------------------------------------------------------------
 
 
@@ -2486,11 +2491,10 @@ async def test_a_none_recovery_publish_never_clears_an_armed_finalizer(
 ) -> None:
     """Defence in depth for TASK-22617, pinned where it is reachable.
 
-    Production cannot currently reach `publish_durable_recovery_owner(...,
-    None)` while a finalizer is armed (the ordering tests above pin why). This
-    pins the store-level property that keeps the None safe even if a future
-    caller breaks that ordering: a None publish must never CLEAR armed state,
-    only decline to arm.
+    The realistic route to this state is the arming-to-registration window
+    inside owner publication (see the window test above). This pins the same
+    property directly at store level, with no controller in the loop: a None
+    publish must never CLEAR armed state, only decline to arm.
     """
 
     stack = real_citation_stack_factory("settle-sticky")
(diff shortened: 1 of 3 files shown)


rmusser01:
## Finding #2 accepted — the ordering model I shipped was incomplete, and you were right about where

Inside `publish_durable_turn_owners`, `_hydrate_durable_turn_owner_messages`
**arms** the finalizer before `publish_durable_dispatch_checkpoint` registers the
recovery. A failure in that window reaches the failure arm's None publish **with
a finalizer armed** — exactly the combination my comment called unreachable.

Confirmed empirically before rewording anything: a new test fails the checkpoint
publication once, and the spy shows the None publish running while armed
(`publish_calls == [False]` with the armed precondition asserted first). The
armed finalizer survives and the resume persists the trace — so the *verdict*
(not data loss) stands, but the *reason* in that window is the store's
non-clearing contract, which is therefore **load-bearing there, not defence in
depth**. Both the in-code comment and the test-file model now say so, and the
clobber mutation is now caught through this realistic path as well as the direct
store-level pin (2 red tests instead of 1).

This is the second time this review cycle has corrected my stated invariant
rather than my code — the tests were sound, the model over-claimed.

## Finding #1 declined, with the evidence

The file-backed database in this rig is deliberate twice over, and this exact
finding was litigated on #2128:

1. `":memory:"` cannot work here: `CharactersRAGDB` opens **thread-local**
   connections and TASK-22205 offloads durable DB calls to a worker thread,
   which then gets its own empty database. Measured on #2128: citation traces
   wrote "successfully" and `rag_citation_traces` still read 0 rows.
2. The shared-cache in-memory URI escape hatch is unavailable —
   `CharactersRAGDB` never passes `uri=True`, so the URI string becomes a
   **literal filename** (on #2128 that literally shipped 137 junk files with
   Windows-illegal names before being caught and reverted).
3. This rig additionally reopens `db_path` across stacks for its restart tests —
   file-backing is load-bearing for the module's own subject.

The full write-up lives in `backlog/docs/lessons-testing-evidence.md`
("`CharactersRAGDB` does not accept URI filenames").

rmusser01:
## Merge-time note on the two red checks

`Artifact Lease Gate` and `Test Summary` are the `Tests` workflow's own gate
jobs reporting that **all 23 of their upstream matrix legs were queue-cancelled**
(listed on the checks tab: 12 UI shards, 6 core shards, 3 lease legs, shape,
MCP). They are cancellations wearing a failure coat — the queued-then-killed
pattern recorded in `lessons-testing-evidence.md` — not verdicts on this PR's
content.

The required check (`Derived artifacts reproduce from their sources`) is
**SUCCESS** on this head — after being queue-cancelled twice and re-run; the CSS
Bundle Guard likewise. Local `./scripts/preflight.sh` green; the touched file is
97 passed locally; A/B vs merge-base: 0 newly broken.

This PR is now the sharpest exhibit for TASK-22250's starvation half: one PR,
three cancellation cycles, including the required check twice.

PR #2162, opened 2026-08-28
ci: skip the test matrix on docs-only PRs, full OS matrix on merge (TASK-22250)

rmusser01:
Task-file update only — no code, no workflow change.

TASK-22250 has had one AC open for weeks: *identify the agent performing the cancellations —
account concurrency ceiling, manual queue clearing, or workflow configuration.* Measured it
via the jobs API while #2129 sat ~4 hours waiting on a 4.5-minute required check.

**It is the concurrency ceiling, saturated by this workflow's own fan-out — and macOS is not
involved.**

| label | running | queued |
|---|---|---|
| ubuntu-latest | 12 | 87 |
| windows-latest | 1 | 1 |
| macos-latest | **0** | **1** |

Both this task's Implementation Notes and `test.yml`'s concurrency comment attribute the
problem to "scarce macOS runners". Ubuntu is the constrained resource, at ~12-13 slots.

**Mechanism:** one `Tests` run fans out to **25 jobs**, nearly all ubuntu (12 UI shards + 6
core shards + artifact-lease and MCP legs). Six live runs were asking for ~100 ubuntu jobs
against ~12 slots — roughly eight waves of ~35-minute shards.

**Why it blocks merges:** `Tests` is not a required check, but its queued jobs share the FIFO
pool with the short required ones. `Derived artifacts reproduce from their sources` runs in
~4.5 minutes and was the last check gating #2129; it spent hours behind UI shards belonging to
other branches. A non-required workflow starves the required gate.

**Diagnostic trap worth having in the file:** `gh run list` reports RUN status, not job status.
A run listed `queued` can have jobs actively executing — the one I inspected had 2 in progress
and 4 completed while listed as queued. I initially misread "22 queued / 0 in progress" as an
account-level block; it wasn't.

Nothing was cancellable at the time — all four competing branches had open PRs (#2155, #2158,
#2160, #2161) and both `dev` runs were on the current SHA. The queue was legitimate demand.

Three options are written into the task (cut UI shard count; narrow the per-push trigger; raise
the ceiling). **None applied** — each either reduces coverage or costs money, and it changes CI
for every contributor, so it's your call.

---
## Summary by cubic
Runs the artifact-lease OS matrix on ubuntu only for pull requests so short required checks stop queueing for hours, while merge and nightly runs keep all three OSes and file-locking coverage. Records the measured cause of TASK-22250's CI burst cancellations in the task file and reworks the CI-shape test to pin the event-conditional matrix.

- Measurement shows ubuntu saturates the account concurrency ceiling (~12 slots vs a 25-job fan-out per `Tests` run), not macOS; a run listed `queued` can still have actively executing jobs.
- A docs/backlog-only skip was attempted then withdrawn: tests glob `backlog/tasks/*.md` and reference `Docs/`, so no path heuristic can safely guess what is test-relevant.
- The lease test used to pass by only checking that the three OS literals appear, which they still do inside the conditional, so it validated nothing on the new workflow.

<sup>Written for commit c997dc2c2278102a2c6ec0d5f08ec892b64a783b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2162?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,53 +60,6 @@ defaults:
     shell: bash
 
 jobs:
-  changes:
-    # TASK-22250: the concurrency ceiling is ~12 concurrent ubuntu jobs, and one
-    # `Tests` run fans out to 25 of them (~11.5 job-hours: the UI suite alone is
-    # 41.5 min x 12 shards). With several branches live, short REQUIRED checks
-    # from other workflows -- `Derived artifacts` runs in ~4.5 min -- sit behind
-    # that queue for hours; PR #2129 waited ~4 hours on one.
-    #
-    # The cheapest saving loses no coverage at all: documentation- and
-    # backlog-only PRs currently pay the full 25 jobs for zero code change.
-    # Measured: #2130 and #2162 each changed ONE markdown file and each
-    # triggered a complete run.
-    #
-    # Gated per-JOB rather than by workflow-level `paths-ignore` on purpose: a
-    # skipped workflow never reports its checks, so if any of these are ever made
-    # required, `paths-ignore` would block merges permanently. Skipped JOBS still
-    # let the workflow report.
-    name: Detect code changes
-    runs-on: ubuntu-latest
-    outputs:
-      code: ${{ steps.filter.outputs.code }}
-    steps:
-      - id: filter
-        env:
-          GH_TOKEN: ${{ github.token }}
-        run: |
-          set -euo pipefail
-          if [ "${{ github.event_name }}" != "pull_request" ]; then
-            echo "Not a pull_request (${{ github.event_name }}) -- running everything."
-            echo "code=true" >> "$GITHUB_OUTPUT"
-            exit 0
-          fi
-          # `|| true`: a transient API error must fall through to the
-          # empty-list branch below (which runs the full suite), never abort
-          # this step. An aborted detector would skip every gated job.
-          CHANGED=$(gh api \
-            "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files" \
-            --paginate --jq '.[].filename' || true)
-          echo "Changed files:"; echo "$CHANGED"
-          # Anything outside docs/backlog counts as code. Fail OPEN: if the list
-          # comes back empty for any reason, run the suite rather than skip it.
-          if [ -z "$CHANGED" ] || echo "$CHANGED" | grep -qvE '^(backlog/|Docs/|\.impeccable/|[^/]*\.md$)'; then
-            echo "code=true" >> "$GITHUB_OUTPUT"
-          else
-            echo "Docs/backlog only -- skipping the test matrix."
-            echo "code=false" >> "$GITHUB_OUTPUT"
-          fi
-
   core-tests:
     # task-1465: full coverage on every PR replaces `-m unit` (which selected
     # 27 of ~950 files while ~590 files ran in no PR-triggered job). The UI
@@ -126,8 +79,6 @@ jobs:
     # pool by six would cost more in queueing than it buys in coverage. macOS
     # breadth already lives in nightly-deep, which is where it belongs.
     name: Core Tests (all but UI) - shard ${{ matrix.shard }}/${{ strategy.job-total }}
-    needs: changes
-    if: needs.changes.outputs.code == 'true'
     runs-on: ubuntu-latest
     # Kept at 120 as headroom: the per-shard estimate is ~32 min, so this is
     # ~4x margin rather than a target. A shard that approaches it is a signal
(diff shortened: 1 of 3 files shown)


rmusser01:
**Now includes the fix, not just the measurement.** Owner's call: no paid runners, narrow the
per-push trigger and let the full matrix run on merge.

### What changed in `test.yml`

**1. A `changes` detector gates the six heavy jobs.** This is the part that costs nothing:
docs- and backlog-only PRs currently pay the full 25 jobs for zero code change. **#2130 and
#2162 each changed one markdown file and each triggered a complete ~11.5 job-hour run.** No
coverage is lost, because there is nothing to test.

**2. The artifact-lease OS matrix is ubuntu-only on a PR**, all three OSes on push (merge) and
nightly — the "full matrix on merge" part.

### Why per-job gating rather than `paths-ignore`

A skipped *workflow* never reports its checks. If any of these ever become required,
workflow-level `paths-ignore` would block merges permanently. Skipped *jobs* still report.

### It fails open, deliberately

The failure that matters here is skipping tests that should have run, so every ambiguous path
runs the suite:

| situation | result |
|---|---|
| API error / empty file list | run everything |
| detector job itself fails | summary goes **red**, lease gate fails |
| workflow file changed | run everything |
| nested `.md` (e.g. `tldw_chatbook/Chat/Chat-UX.md`) | run everything |
| `backlog/`, `Docs/`, root `*.md` only | skip |

`artifact-lease-gate` and the summary verdict both had to learn the difference between
`success` and `skipped` — without that, "nothing to test" would have come back **red**, which
is worse than the problem.

### Validation

Filter run against real PR file lists: #2130 and #2162 skip; #2129 (19 files) and #2157 (10
files) run. Plus empty / mixed docs+code / workflow-file / nested-md edge cases.

Worth recording: my first validation pass said **#2129 should skip** — a false negative that
would have skipped tests on 19 files of Python. That was `grep` being aliased to ugrep on my
machine (the documented repo trap); GitHub runners use GNU grep. Re-run with `command grep`,
the classification is correct.

### What this does *not* fix

The dominant cost is unchanged: a code PR still spends ~11.5 job-hours, because the UI suite is
41.5 min x 12 shards. This removes the waste around it. Cutting that further means running
fewer tests per PR, which reverses the deliberate decision in task-1465 (`-m unit` selected 27
of ~950 files) — worth doing only as its own decision, not smuggled in here.

rmusser01:
@coderabbitai review

rmusser01:
**Qodo's finding was correct, and it kills the main change. Reverted.**

The flag: `Docs/fixtures/console-block-prompts` holds JSON consumed by core tests, so a
"documentation-only" PR would skip the very tests validating it. I verified that, and the
problem is much larger than the fixture:

- **126 test files reference `Docs/`**, including `Tests/MCP/test_mcp_documentation_contract.py`,
  which asserts on specific markdown files (`Docs/User_Guide/mcp.md` and others)
- **78 references to `backlog/tasks`**, with tests reading named task files
- decisively, tests **glob** the directory — `test_post_release_ux_hci_validation_plan.py:150`
  and `test_product_maturity_phase1_harness.py:81` both walk `backlog/tasks/*.md`. **Adding a
  task file is itself an input to those tests**, which is exactly what a docs-only PR does.

So the premise behind the skip — "docs and backlog changes cannot break tests" — is false in
this repo, and no path heuristic separates the safe cases from the unsafe ones. I reverted
rather than narrowing the exemption: a narrower allowlist decays the moment someone adds
another doc-reading test, and the failure mode is silent (tests that never run). That is worse
than the 25 wasted jobs it was meant to save.

**What ships:** only the OS-matrix narrowing — artifact-lease runs ubuntu-only on a PR, all
three OSes on push and nightly. Safe on its own, and it is the owner's "full matrix on merge".

**Correcting my own earlier claim on this PR:** I said docs-only PRs waste ~23 job-hours. That
was wrong — much of it is real work, because those tests read the files being changed.

**What I found while implementing, worth keeping:** `actions/download-artifact@v4` ERRORS when
its pattern matches nothing, so the summary job would have gone red on exactly the docs-only
PRs the change was meant to make cheap. That bug died with the revert, but it is the kind of
thing that makes "skip the suite" harder than it looks.

**Remaining levers are unchanged and both owner calls:** the UI suite is 41.5 min x 12 shards
(~8.3 job-hours) and per-job setup is only ~8% (3.6 min of 45), so re-sharding barely moves the
total. Running fewer tests per PR reverses task-1465's deliberate decision; raising the ceiling
costs money.

Note on process: `/review` is deprecated and returned a shallow "no major issues" pass on this
same diff. The finding above only came from `/agentic_review`.

rmusser01:
**All five Qodo findings dispositioned.** I had only read the first one initially — reading the
rest changed the outcome twice.

| # | Finding | Disposition |
|---|---|---|
| 1 | CI-shape tests left asserting the old contract | **Fixed** (`c997dc2c22`) — and it still applied after the revert |
| 2 | `Docs/fixtures` consumed by core tests | **Accepted; killed the feature** (`3ebe6d0bcb`) |
| 3 | Renames report only the destination path | Moot — detector reverted |
| 4 | `pipefail` + `grep -q` SIGPIPE inverts the match | Moot — detector reverted |
| 5 | Partial API failures skip code | Moot — detector reverted |

**Finding 2 was the decisive one** and I verified it goes further than stated: 126 test files
reference `Docs/`, 78 reference `backlog/tasks`, and tests **glob** `backlog/tasks/*.md`
(`test_post_release_ux_hci_validation_plan.py:150`, `test_product_maturity_phase1_harness.py:81`)
— so *adding a task file is itself an input to those tests*, which is exactly what a docs-only PR
does. No path heuristic is safe here, so the change detector was reverted rather than narrowed.
Findings 3, 4 and 5 all describe real defects in that same reverted code; 4 in particular
(`grep -q` exiting early, producer taking SIGPIPE under `pipefail`, inverting the result on large
PRs) would have silently mislabelled big PRs as docs-only. Good catches on code that no longer
exists.

**Finding 1 is the one I nearly dismissed as moot, and it wasn't.** It applies to the single
change that survived. `test_artifact_lease_spike_runs_natively_on_three_operating_systems`
asserted only that the three OS literals appear in the job block — and all three still appear
*inside* the new `fromJSON` expression. So the test passed while its own name became false for
pull requests: a guard passing for the wrong reason. It now pins the conditional and both arms
(ubuntu on `pull_request`, all three on `push`/`schedule`) and asserts the static list is gone.
Mutation-tested: reverting the workflow to the static list reds it; restored, 19 passed.

**Verified in production on this PR's own run:** the `Tests` run shows exactly one
`Artifact leases - Python 3.11 - ubuntu-latest` job where it previously spawned three. The
`fromJSON` conditional does what it claims on a real `pull_request` event, and the full three-OS
matrix returns on the merge push.

Rebased onto latest dev (was 44 behind); `./scripts/preflight.sh` green on the rebased tree.
Note the earlier red `Derived artifacts` on this branch was inherited drift from a stale base —
dev fixed it in the intervening commits.

PR #2186, opened 2026-08-29
fix(security): redact secrets at the log file sink (TASK-23190)

rmusser01:
## The gap

TASK-23108 tells users "Details are in Logs (F8)" while reducing its own
user-facing copy to exception type names, on the assumption that the log is the
safe place for the detail. Two things were wrong with that assumption.

**1. The file sink had no redaction.** `Logging_Config.py` installs a
`PrivateRotatingFileHandler` with a plain `logging.Formatter`; only the in-app
Logs buffer ran `redact_log_line` (`app.py`). The sanitizer existed and the sink
simply never called it.

**2. `redact_secret_text` classifies a value by the name to its left.**
`_SECRET_ASSIGNMENT_PATTERN` requires the name to contain
`API_KEY`/`TOKEN`/`SECRET`/`PASSWORD`, so `Authorization: Bearer <token>` (the
classifying name is `Authorization`) and `?key=<token>` (`key` alone) were
displayed on the Settings surface verbatim. Probed before the fix — both came
back unchanged; `?api_key=` and `&token=` were already covered.

### One correction to the filed premise

The task says the header "reaches disk unchanged from any OTHER caller". Probed
against the real configuration: it does not. The handler already carries
`PersistentDiagnosticFilter`, which admits only schema-validated ADR-029
metadata events — an ordinary `logger.error("Authorization: Bearer sk-...")` and
a third-party `httpx` record both produced **zero bytes**. So the sink had no
redaction, but it also had no unredacted secrets. What was missing is the second
layer that makes any future widening of that admission rule safe instead of a
disclosure. This PR adds that layer and pins the first one, so a widening
becomes a visible change rather than a silent one.

## The fix

`RedactingFileFormatter` wraps `redact_log_line` around the fully formatted
record. A `logging.Filter` was rejected for two reasons, both load-bearing:

- a filter must mutate `record.msg`, and the record is shared with every other
  handler on the logger (and with whichever thread emitted it);
- only the formatted string carries the `exc_info` traceback, which is where a
  credential inside `str(exc)` actually lives — exactly the shape TASK-23108 was
  filed about.

`log_sanitizer` needed **no new patterns**: `Bearer`, `?key=` (the bare-`key`
entry added by TASK-19558), `?api_key=` and `&token=` were all already covered,
and it is reused verbatim so the disk sink and the Logs screen cannot drift on
what counts as a secret. The already-installed-handler branch reconciles the
formatter for the same reason it already reconciles the filter.

`redact_secret_text` keeps its `<redacted>` marker and name vocabulary (many
callers assert the exact `KEY=<redacted>` shape). Three narrow additions: `[:=]`
as separator (`X-Api-Key: <token>`), a `Bearer` rule, and a bare `key` rule
anchored to query-string position only — `key` is an ordinary English word and a
TOML parse error says it, so the `?`/`&` anchor is what proves it is a parameter
name.

## On-disk evidence (AC-3)

`Tests/test_logging_private_files.py` drives the real
`_configure_private_file_logging` on the real root logger, logs through a normal
`logging` call, and reads the file back. Verbatim, with the formatter in place:

```
2026-08-28 19:54:13 [INFO ] tldw_chatbook.diagnostics.logging:182 - event=persistent_sink_installed component=logging status=ok
2026-08-28 19:54:13 [INFO ] tldw_chatbook.probe:30 - Authorization: ***REDACTED***
2026-08-28 19:54:13 [INFO ] tldw_chatbook.probe:31 - GET https://x/v1?key=***REDACTED***
2026-08-28 19:54:13 [INFO ] tldw_chatbook.probe:32 - POST /v1?api_key=***REDACTED***
2026-08-28 19:54:13 [INFO ] tldw_chatbook.probe:33 - OPENAI_API_KEY = ***REDACTED***
2026-08-28 19:54:13 [INFO ] tldw_chatbook.probe:34 - ordinary startup message with no secret at all
2026-08-28 19:54:13 [INFO ] tldw_chatbook.probe:38 - provider probe failed
Traceback (most recent call last):
  File ".../probe_sink2.py", line 36, in <module>
    raise ValueError("connect failed for https://svc/v1?key=***REDACTED***
ValueError: connect failed for https://svc/v1?key=***REDACTED***
```

Note the last three lines: the traceback **and** the exception message are
redacted. That is the case a `record.msg` filter would have left in clear.
The negative control (line 6) is intact, so an over-broad rule that blanked
every line cannot pass.

## Mutation check

All restores Edit-based, never `git checkout`.

| mutation | result |
|---|---|
| `RedactingFileFormatter.format` returns `super().format(record)` | **6 failed**, 19 passed — all 5 shapes + the exception-text case; negative control and admission-layer tests correctly stayed green |
| Bearer + query-key rules removed from `redact_secret_text` | **2 failed** (`authorization-bearer-header`, `bare-key-query-parameter`) |
| separator narrowed back to `=` | **1 failed** (`colon-separated-credential-header`) |
| restored | green |

## Verification

- `Tests/test_logging_private_files.py` — **25 passed**
- logging / diagnostics / sanitizer batch (9 files) — **167 passed**
- config-defaults / websearch-credential / RAG-privacy batch — **92 passed**
- `test_settings_configuration_hub.py` + `test_settings_provider_test_draft.py`
  — **449 passed, 6 failed**. The identical 6 fail on the merge-base
  `4fb5d38d37` (`git archive` of the base run against the same venv); they are
  save/revert and category-content assertions, with no redaction assertion among
  them.
- `--collect-only` on both touched test files — 410 collected
- `./scripts/preflight.sh` — all green, **no diagnostic-inventory drift**
  (541 owners, 1262/7397 calls, 8 sink files, unchanged)
- `ruff check` on all four touched files — clean

## Performance

33.6 us/record versus 1.4 us for the plain formatter. `Handler.handle` runs the
handler's filters *before* `emit`, so this is paid only on records actually
written — a handful of metadata events per session. `MAX_REDACTED_LINE_CHARS`
is kept rather than disabled: truncation keeps strictly less data, its cut is
token-aligned so it cannot slice a credential into an unmatchable fragment, and
it is what bounds that 32 us. Cost of keeping it: a >2,000-char record is
truncated on disk as well as in the Logs screen.

## Checked, deliberately unchanged

Loguru `diagnose` is `False` at every `logger.add` in the package
(`Logging_Config.py`, `__init__.py`, `startup_logging.py`,
`Metrics/logger_config.py`) — no finding. Loguru's own file sinks are already
disabled in `Metrics/logger_config.py`; `Chunking/engine/security_logger.py` can
open one but only when handed an explicit `log_file`, which no production caller
does.

https://claude.ai/code/session_01N2WwNRSyk6w9V7b9yh5Yht

---
## Summary by cubic
Fixes the private rotating file sink so every record written to disk is redacted regardless of caller, and extends `redact_secret_text` to cover the `Authorization: Bearer` header and bare `?key=` query shapes that Settings surfaces were displaying verbatim.

**Bug Fixes**
- Added `RedactingFileFormatter`, which runs `redact_log_line` over the fully formatted record so credentials inside `exc_info` tracebacks are redacted — a filter rewriting `record.msg` would leave those untouched.
- Probed the real config first: the sink's `PersistentDiagnosticFilter` already drops unmarked and third-party records, so no unredacted secrets reach disk today; the formatter is the second layer that makes any future widening of that admission rule safe instead of a disclosure.
- `redact_secret_text` now also redacts `Bearer` values, colon-separated headers (`X-Api-Key: ...`), and bare `key` query parameters, with `key` anchored to `?`/`&` so ordinary prose containing the word stays intact.
- The already-installed-handler branch now upgrades a plain formatter to the redacting one, same as it already reconciles the filter.
- Tests drive the real `_configure_private_file_logging` per secret shape and read the file back, including a negative control and a check that unmarked secret-bearing records are not written at all; both guards were mutation-checked.

<sup>Written for commit e6ca3bf51e2a154eb3f1aa099d0de38823f5b1db. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2186?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_settings_configuration_hub.py
+++ b/Tests/UI/test_settings_configuration_hub.py
@@ -801,6 +801,82 @@ def test_redact_secret_text_removes_api_key_like_values():
     assert "OPENAI_API_KEY=<redacted>" in redacted
 
 
+@pytest.mark.parametrize(
+    ("text", "secret", "expected_fragment"),
+    [
+        pytest.param(
+            "rejected: Authorization: Bearer sk-live-abc123",
+            "sk-live-abc123",
+            "Bearer <redacted>",
+            id="authorization-bearer-header",
+        ),
+        pytest.param(
+            "X-Api-Key: sk-live-abc123 was rejected",
+            "sk-live-abc123",
+            "X-Api-Key: <redacted>",
+            id="colon-separated-credential-header",
+        ),
+        pytest.param(
+            "GET https://svc/customsearch/v1?key=AIzaSyLIVE0123456789&cx=017 failed",
+            "AIzaSyLIVE0123456789",
+            "?key=<redacted>&cx=017",
+            id="bare-key-query-parameter",
+        ),
+        pytest.param(
+            "POST https://svc/v1/models?api_key=sk-live-abc123 -> 401",
+            "sk-live-abc123",
+            "api_key=<redacted>",
+            id="api-key-query-parameter",
+        ),
+    ],
+)
+def test_redact_secret_text_removes_header_and_query_string_secrets(
+    text: str,
+    secret: str,
+    expected_fragment: str,
+) -> None:
+    """TASK-23190: the assignment rule alone missed both of these shapes.
+
+    ``_SECRET_ASSIGNMENT_PATTERN`` classifies a value by the *name* to its
+    left, so ``Authorization: Bearer <token>`` (the classifying name is
+    ``Authorization``, which is not in the vocabulary) and ``?key=<token>``
+    (``key`` alone is not either) were both displayed verbatim. Probed before
+    the fix: both strings came back from ``redact_secret_text`` unchanged.
+    """
+    redacted = redact_secret_text(text)
+
+    assert secret not in redacted
+    assert expected_fragment in redacted
+
+
+def test_redact_secret_text_leaves_secret_free_text_intact() -> None:
+    """Negative control: an over-broad rule that redacts everything must fail.
+
+    The bare-``key`` query rule is the risky one -- ``key`` is an ordinary
+    English word, and a TOML parse error says it. It is anchored on ``?``/``&``
+    precisely so prose survives.
+    """
+    intact = (
+        "Model discovery failed (HTTPStatusError). Check the endpoint. "
+        "Details are in Logs (F8). Expected key but found newline at line 3"
+    )
+
+    assert redact_secret_text(intact) == intact
+
+
+def test_failure_status_text_keeps_type_name_only_copy_unchanged() -> None:
+    """AC-4: the TASK-23108 call sites must read exactly as they did."""
+
+    assert failure_status_text(
+        "Model discovery failed",
+        ValueError("secret=sk-live-abc123"),
+        next_step="Check the endpoint.",
+    ) == (
+        "Model discovery failed (ValueError). Check the endpoint. "
+        "Details are in Logs (F8)."
+    )
+
+
 def test_adapter_rejects_non_mapping_toml():
     adapter = SettingsConfigAdapter()
 
--- a/Tests/test_logging_private_files.py
+++ b/Tests/test_logging_private_files.py
@@ -10,10 +10,14 @@ import pytest
 from tldw_chatbook import config
 from tldw_chatbook.Logging_Config import (
     PrivateRotatingFileHandler,
+    RedactingFileFormatter,
     _configure_private_file_logging,
 )
 from tldw_chatbook.Utils.private_paths import PrivatePathError
-from tldw_chatbook.Utils.persistent_diagnostics import PersistentDiagnosticFilter
+from tldw_chatbook.Utils.persistent_diagnostics import (
+    _PERSISTENT_METADATA_MARKER,
+    PersistentDiagnosticFilter,
+)
 
 
 def _mode(path: Path) -> int:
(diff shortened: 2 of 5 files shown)


rmusser01:
Independent verification before merge (main session):

**Redaction, after the fix** — all five shapes redact, negative control intact:
| input | result |
|---|---|
| `Authorization: Bearer sk-live-…` | redacted |
| `…/models?key=AIzaSy…&cx=9` | redacted |
| `?api_key=sk-proj-…` | redacted |
| `OPENAI_API_KEY = sk-…` | redacted |
| `Model discovery failed: HTTPStatusError` | **unchanged** (no over-redaction) |

`Tests/test_logging_private_files.py`: **25 passed** re-run independently.

**One cosmetic edge worth recording**, not a blocker: widening the separator to `[:=]` means the joined form followed by a status word garbles, e.g. `api_key: not configured` → `api_key: <redacted> configured`. I checked whether this reaches shipped copy — it does not. The two candidate strings (`settings_screen.py:10877`, `:10880`) use the space-separated `API key: …` form, which does not match the name vocabulary and verifies intact. The residual exposure is exception text passed through `redact_secret_text` in ~45 call sites, where the failure direction is over-redaction rather than disclosure, on a denylist the docstring already declares incomplete.

Merging on that basis.

PR #2217, opened 2026-08-30
perf: 2026-08-29 holistic review — CSS apply fast path, keystroke guards, boot ratchet paydown

rmusser01:
Fifth holistic performance review (pin dev `bc1e26ce60`, 524 commits since the 08-27 pin),
filed as **TASK-24450..24461**, with 7 of the 12 implemented here.

Evidence document: `Docs/Design/2026-08-29-holistic-perf-review.md`.

## Core verdict

**Idle is clean** — 0.09 s CPU over 12 s (0.75% of a core), zero stalls, zero SQL. The
08-22/08-24 timer and idle-SQL fixes held. The lag users report is entirely in interaction,
and it had one dominant cause.

**Every style application scanned all 4,324 CSS rules.** Upstream `Stylesheet.apply` narrows
candidates through `rules_map` and then discards that narrowing to recover source order with
`filter(limit_rules.__contains__, reversed(self.rules))`. That measured 7,335,029
`RuleSet.__hash__` calls in a single Console switch — exactly 1,667 applies × 4,324 rules —
and stack sampling ranked `textual/css/model.py:__hash__` the #1 frame during event-loop
stalls of up to 399 ms.

## Measured, interleaved A/B on the real app vs the unmodified pin

| Metric | dev | this branch |
|---|---|---|
| `Stylesheet.apply()` per node | 0.553 ms | **0.388 ms** (−30%) |
| `update_styles(screen)` | 263 ms | **160 ms** (−39%) |
| Console screen switch (mean of 6) | 2.06 s | **1.79 s** |
| Library screen switch | 1.24 s | **1.16 s** |
| `query_one` per keystroke | 35.4 | **12.3** (−65%) |
| `Static.update` per keystroke | 8.5 | **2.4** (−72%) |
| `refresh(layout=True)` per keystroke | 11.7 | **6.7** (−43%) |
| Library config reads per revisit | 46 | **2** |
| Boot import weight | 662/660 ❌ | **631/660** ✅ |
| `_ui_ready` module census | 981/970 ❌ | **966/970** ✅ |
| Boot worker census | ❌ | ✅ |

No ratchet constant was raised.

## What's in each commit

- **`perf(css)`** — TASK-24450. Ordered-candidate fast path. A *delegation*, not a fork:
  upstream stays the source of truth for what styling means, only candidate selection changes.
  The position index is cached against the `rules_map` object by strong reference, because
  Textual nulls `_rules_map` on every path that mutates `_rules`.
- **`perf(console)`** — TASK-24453. Five per-keystroke syncs guarded by state signatures.
  `_sync_collapsed_presentation` ran 3×/key against a `display:none` row;
  `_apply_draft_height` called `refresh(layout=True)` on every keypress.
- **`perf(library)`** — TASK-24456. Ingest options memoised on a new
  `config.current_config_generation()`, scoped to the running App.
- **`perf(boot)`** — TASK-24458. Four import deferrals, headline being that `app.py` imported
  the **deprecated, nav-unreachable** `Tools_Settings_Window` (TASK-1346) for one message
  class, dragging the entire workspace tool-execution cluster onto boot.
- **`fix(boot)`** — TASK-24460/24461. Named the anonymous boot worker; wired three of the four
  budget ratchets into `perf-guard.yml`.

## Two things reviewers should know

**A filed finding was wrong and is corrected in-tree.** "Library config re-read on unrelated
screen switches" came from averaging `get_cli_setting` over a Library+Console *pair*. Measured
per destination, Console switches read 18–21 and **none** from `library_screen` — the average
smeared one screen's mount cost across both. Task title, evidence doc and commit all corrected.

**The boot-worker failure was a naming regression, not a new worker.** Textual derives worker
names via `getattr(work, "__name__", "")`; wrapping the call in `functools.partial` silently
renamed it to `""`, breaking its existing allowlist row and making it anonymous in every
diagnostic. 56 other `run_worker(partial(...))` sites carry the same latent anonymity.

## Verification

- `Tests/Performance/`: green, including the two new fast-path guards. The fidelity test
  compares rule maps for all 689 nodes across three screens and was **mutation-tested** —
  reversing the sort order breaks 47 of them. The second test pins upstream's implementation so
  a Textual bump fails loudly rather than drifting silently.
- `Tests/Chat/`: 1935 passed, 1 failure **verified pre-existing on unmodified dev**.
- `Tests/UI/` composer + library subsets: ~1000 passed, 5 failures all verified pre-existing.
- `./scripts/preflight.sh`: all derived-artifact checks pass.
- A full `Tests/Chat + Tests/UI` run was still in progress locally when this PR opened; CI is
  the authority.

## Left open, each annotated in its task with what's established and what blocks it

**TASK-24452, 24456 and 24457 share one root cause**: `switch_screen` is handed a freshly
constructed screen every time (verified — three Library visits, three distinct instance ids).
That is why the Console re-mints 559 widgets, Library re-reads 43 settings, and Library opens 8
SQLite connections per visit. Instance reuse is the largest remaining win and an architecture
change — the snapshot/restore design assumes fresh construction — so it needs an owner call.

Also open: 24451 (split the 272 KB `_agentic_terminal.tcss` grab-bag, 29% of all rules),
24454, 24455 (598 → 482 `query_one`/switch as a side effect, short of its target), and
**24459 — the one still-red ratchet**, which is exactly why the CSS byte guard was excluded
from the new CI step rather than turning every unrelated PR red.

https://claude.ai/code/session_012nYQSYQbBm1DfFbXirJx3e

---
## Summary by cubic
Cuts the interaction lag from the 08-29 holistic review: `Stylesheet.apply()` no longer scans all 4,324 CSS rules per call, composer keystroke syncs are signature-guarded, Library config reads are memoised per app, and boot import weight is back under its ratchet. One filed finding was wrong and is corrected in-tree: "Library config re-read on unrelated switches" was an averaging artifact over a Library+Console pair, and the task and evidence doc now say so.

**Performance**
- CSS apply is now a cached, position-sorted candidate hand-off to upstream with identical semantics; measured A/B: apply per node −30%, Console switch 2.06 → 1.79 s.
- Five composer syncs are signature-guarded, cutting `query_one` per keystroke 35.4 → 12.3; `_apply_draft_height` no longer forces `refresh(layout=True)` per keypress.
- Library ingest options are memoised on the running App, keyed by the new `current_config_identity()` (generation + effective config path — a generation-only key served stale options when `TLDW_CONFIG_PATH` was retargeted); repeat-visit reads drop 46 → 2.
- Screen instance reuse instead of fresh construction per visit is left open in TASK-24452/24456/24457 as the largest remaining win.

**Boot & CI**
- Import deferrals (headline: `app.py` no longer imports the deprecated `Tools_Settings_Window`) brought boot weight to 631/660 and `_ui_ready` census to 966/970; the rebase exposed a pre-import breach, and lazy `UI/Widgets` re-exports plus three deferrals shed 653 LOC to land at 492/500 modules.
- After the final rebase the census sits at exactly 970/970 (green where dev is 983) with zero headroom; the pre-import LOC budget stays red on both and runs in neither CI lane.
- The boot-worker census failure was a naming regression, not a new worker — `functools.partial` silently emptied the worker name, now passed explicitly.
- Boot budget ratchets now run in `perf-guard.yml` so a breach fails the PR that caused it; the CSS-byte test stays out until TASK-24459 because it is red on dev.

Tests are green except failures verified pre-existing on dev.

<sup>Written for commit db5a73397c66c262a8635aa81f036600eb24fe48. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2217?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Documentation**
  * Added a holistic performance review of the Console app.
  * Documented interaction-performance bottlenecks affecting screen switches, typing, startup budgets, and Library visits.
  * Included measured findings, root-cause analysis, validated optimization results, recommended remediation priorities, and measurement considerations.

### DIFF
--- a/.github/workflows/perf-guard.yml
+++ b/.github/workflows/perf-guard.yml
@@ -65,3 +65,22 @@ jobs:
           pytest Tests/Performance/test_ui_latency_guardrails.py \
             Tests/Utils/test_ui_responsiveness_stall_persist.py \
             -v --timeout=300
+
+      # task-24461: the ADR-097 boot budgets ran ONLY in the slow full suite,
+      # so for three review cycles running they were re-breached within ~24h of
+      # every paydown and each breach was found by a periodic review rather
+      # than by the PR that caused it. They are cheap (~15s together) and they
+      # already print the culprit modules, so the breaching PR now gets that
+      # list here, in minutes.
+      #
+      # test_boot_css_byte_budget.py is deliberately NOT in this list yet: it
+      # is red on dev (862,184 B against an 860,000 B ratchet) and adding it
+      # would block every unrelated PR. It joins this step when task-24459
+      # lands.
+      - name: Run boot budget ratchets
+        run: |
+          pytest Tests/Performance/test_app_import_weight.py \
+            Tests/Performance/test_ui_ready_module_census.py \
+            Tests/Performance/test_boot_worker_census.py \
+            Tests/Performance/test_textual_css_fastpath.py \
+            -v --timeout=300
--- a/Docs/Design/2026-08-29-holistic-perf-review.md
+++ b/Docs/Design/2026-08-29-holistic-perf-review.md
@@ -26,11 +26,18 @@ ADR-097 made these ratchets on 2026-08-27; TASK-23112 paid the import debt down 
 | `MAX_BOOT_PARSED_CSS_BYTES` | 860,000 | **862,184** | +3,424 B in the monolith alone |
 | boot worker allowlist | — | **1 unlisted** | `console-persisted-browser-cache` (`UI/Console_Modules/workspace.py:2553`) |
 
-The new modules cluster in one family — the workspace/tool-execution work:
-`Tools.{git,local,patch,virtual_cli}_tool_impls`, `Tools.workspace_tool_{executor,protocol}`,
-`Tools.workspace_root_pin`, `Agents.{raw_shell,virtual_cli}_tool_provider`,
-`Workspaces.change_review_{consent,finalization}`, `Chat.console_settings_*`,
-`TTS.{legacy_request_builder,text_processing}`.
+The new modules fall into **three unrelated families**, which matters because remediation
+ownership differs (CodeRabbit, PR #2217 — an earlier draft lumped all three together):
+
+| Family | Modules |
+|---|---|
+| Workspace / tool execution (the large one) | `Tools.{git,local,patch,virtual_cli}_tool_impls`, `Tools.workspace_tool_{executor,protocol}`, `Tools.workspace_root_pin`, `Agents.{raw_shell,virtual_cli}_tool_provider`, `Utils.filesystem_identity` |
+| Workspace change review | `Workspaces.change_review_{consent,finalization}`, `Workspaces.change_{bounds,tracking,turn_tracker}` |
+| Console settings | `Chat.console_settings_*`, `Chat.provider_setup_persistence`, `Chat.console_generation_settings_metadata` |
+| TTS | `TTS.{legacy_request_builder,text_processing}` |
+
+Only the first family was severed by this cycle's deferrals; the others left the boot path as a
+side effect of the same import edges.
 
 **The process finding outranks the numbers.** Three cycles running, the budgets are re-breached
 within ~24 h of every paydown. The guards only run in the slow suite; last cycle's open owner
@@ -57,7 +64,11 @@ reduces *matching* work but not the *scan*. Measured consequences:
 | `app.update_styles(screen)` (500 widgets) | **240 ms** |
 | `RuleSet.__hash__` calls in one Console switch | **7,335,029** |
 
-7.33M ≈ 1,667 applies × 4,324 rules — the scan, exactly.
+That is ~4,400 hash calls per apply against 4,324 rules: **4,324 from the full-list `filter`
+scan, plus ~76 from building the candidate set** (each candidate is hashed on insertion, and
+`all_pseudo_classes` unions over the result). An earlier draft of this document called it
+"exactly 1,667 × 4,324"; that product is 7,208,108, and the 126,921 shortfall is the candidate-set
+work. The scan dominates, but it is not the only hasher. (CodeRabbit, PR #2217.)
 
 Stack sampling during observed stalls ranks `textual/css/model.py:__hash__` the #1 frame, then
 `_check_selectors`, `check`, `_check_id`. **Screen switching produced 17 event-loop stalls,
--- a/Docs/Design/2026-08-29-holistic-perf-review.md
+++ b/Docs/Design/2026-08-29-holistic-perf-review.md
@@ -1,0 +1,218 @@
+# Holistic performance review — 2026-08-29
+
+**Pin:** dev `bc1e26ce60` (524 commits since the 2026-08-27 review pin `c6218918d1`).
+**Method:** pinned worktree `.worktrees/perf-review-0829` with its own uv venv and a scratch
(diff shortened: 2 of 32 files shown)


rmusser01:
## Review findings — resolution

All three addressed in `1e8f6139b9`, plus one regression the rebase onto latest dev exposed.

### 1. Qodo (Medium, Bug) — "Retargeted config serves stale options" — **valid, fixed**

Confirmed by reproduction before fixing. With a live app holding the cache, retargeting
`TLDW_CONFIG_PATH` to a different file returned the **new** file's value from
`get_cli_setting` while `_CONFIG_GENERATION` stayed at `1` — so a generation-only key served
the previous config's ingest options, exactly as described.

`current_config_generation()` is replaced by **`current_config_identity()` = (generation,
effective config path)**. `Tests/UI/test_library_ingest_options_cache.py` pins both halves and
was mutation-tested: reverting to a generation-only key fails two of its three tests.

### 2. CodeRabbit — "Correct the hash-call arithmetic" — **valid, fixed**

Correct, and my "exactly" was wrong: `1,667 × 4,324 = 7,208,108`, not the measured
`7,335,029`. The real figure is **~4,400 hashes per apply — 4,324 from the full-list scan plus
~76 from building the candidate set** (each candidate is hashed on insertion, and
`all_pseudo_classes` unions over the result). The scan dominates but is not the only hasher.
Corrected in the evidence doc, `textual_css_fastpath.py`, and the test docstring.

### 3. CodeRabbit — "Separate the unrelated module families" — **valid, fixed**

Correct: `TTS.*` and `Chat.console_settings_*` are not part of the workspace/tool-execution
cluster, and remediation ownership differs. The evidence doc now tables **four** families
separately and states that only the first was severed by this cycle's deferrals.

---

### Also fixed: a pre-import ratchet interaction the rebase exposed

`test_screen_preimport_payload_budget` went red at 505/500 on the rebase while **passing on
pristine dev** (verified in a throwaway dev worktree). Nothing had grown. That guard measures
modules resident after the pre-import pass **minus a baseline taken before it**, so every module
this branch defers off boot lowers the baseline and raises `pass_added` by one. The breach's own
module-set diff named exactly the 14 deferred modules.

Rather than raise a constant (ADR-097 forbids it) or revert a real boot win, **653 LOC of
genuinely wasted eager imports** were shed by following the same edges — the largest being
`UI/Widgets/__init__.py`, which eagerly imported `SmartContentTree` (425 LOC) and
`config_search_widget` (228 LOC) so that the four MCP modes wanting only the small
`table_click_select` mixin each paid for both. That is the `Chunking/__init__.py` eager-package
trap (finding 21102) recurring in a second package.

**Worth flagging:** the pre-import LOC ratchet now sits at **379,955 / 380,000 — 45 LOC of
headroom** (dev's own was 226). It is green, but thin, and recorded in task-24458.

One test was **retargeted, not weakened**:
`test_local_agent_catalog_failure_degrades_to_no_local_group` patched
`mcp_workbench.LocalToolProvider`, a name that no longer exists at module scope; it now patches
the defining module, which is what the deferred import resolves against. Since retargeting a
patch can silently make a test vacuous, this was verified with a temporary probe asserting the
patched constructor still runs, then removed.

PR #2223, opened 2026-08-30
feat(network): TLS trust policy for corporate DPI networks (TASK-21513)

rmusser01:
## What

One global config knob for users on TLS-inspecting/DPI corporate networks, whose re-signed certificates break every HTTPS call because none of the app's transports consult the OS trust store:

```toml
[network]
# true | false | "/path/to/corp-ca.pem" (additive: certifi + custom, never replace)
ssl_verify = true
```

Fail-safe direction is always verification-**on**: bad type, missing/unreadable file, corrupt PEM, or bundle-write failure logs an error and falls back to default verification. The setting can never silently disable verification.

- New helper `tldw_chatbook/Utils/tls_trust.py`: lenient coercion, additive SSL contexts (certifi + custom CA), a fingerprint-keyed merged-PEM cache for requests, TLS-trust-aware client factories, warn-once + metrics when verification is relaxed (warning names the stakes: API keys and conversation content interceptable by anyone on the network path).
- Policy wired across all four transports: httpx (Console gateway, TTS, image-gen shared client, model catalog, evals capture, local server discovery, web tools), requests (LLM calls incl. hosted chat, summarization libs, web-search backends — 51 session sites + 11 direct calls), aiohttp (crawler, SwarmUI), websockets (realtime transport, with a ws:// guard), plus the one OpenAI-SDK site (OCR Docext).
- `tldw_api/client.py` (Apache-2.0, standalone) gains an `ssl_verify` constructor param forwarded verbatim to httpx; the app resolves policy via `httpx_verify()` at `runtime_policy/bootstrap.py`.
- F9 Settings gains a **Network** category (select + CA path input + invalid-config error row + truthful "Pending — save with s" badge/footer).

## Key decisions (ADR-079)

- **Additive custom-CA semantics** — corp appliances intercept selectively; replace semantics would break every non-intercepted public endpoint.
- **Disabled mode returns an unverified CERT_NONE `SSLContext`, not bare `False`** — websockets ≥14 raises `ValueError: server_hostname is only meaningful with ssl` on bare `False` for `wss://` (empirically reproduced on websockets 16.0); aiohttp treats the context identically to `ssl=False`. Amendment recorded in spec §2 and ADR-079.
- Global-only granularity, no env override, `truststore` deferred — rejected alternatives and rationale in the ADR.

## Governance

- Spec: `Docs/superpowers/specs/2026-08-29-network-tls-trust-policy-design.md`
- Plan: `Docs/superpowers/plans/2026-08-29-network-tls-trust-policy.md`
- ADR: `backlog/decisions/079-network-tls-trust-policy.md`
- Task: TASK-21513 (Done, all ACs checked)

## Test evidence

Branch suites: 54 passed (`test_tls_trust.py` 35, gateway/tldw_api/realtime seam tests 7, settings Network unit+pilot+regression 12) plus 112 passed / 1 pre-existing optional-dep skip in adjacent TTS/Evals/Image-Generation selections. Ruff clean over all touched files. Realtime wss behavior verified against a local self-signed TLS server.

## Notes for reviewers

- This branch forks from the local docs line, so the diff includes 14 docs commits — two of them are this feature's spec/plan; the rest (research-workspace, daily-reports, workspace-assistant-defaults docs, ADR-076 renumber lesson) are pre-existing committed work on that line.
- 3 failures in `Tests/LLM_Calls/test_summarization_diagnostic_privacy.py` are **pre-existing at the fork point** (stale projection-SHA pins over `Docs/security/production-diagnostic-inventory.json`; the manifest itself was verified byte-identical to fresh generation — hand-off to the diagnostic-inventory owner, not regenerated here). 14 `Tests/UI/test_destination_shells.py` failures are likewise pre-existing (Library destination).
- The OCR OpenAI-SDK seam is statically verified only (`openai` is not installed in the dev venv).

---
## Summary by cubic
Adds a global `[network] ssl_verify` knob so HTTPS calls work on TLS-inspecting corporate (DPI) networks, where re-signed certificates previously broke every outbound HTTPS call. The knob accepts `true` (default), `false`, or a custom CA bundle path, and threads through the `httpx`, `requests`, `aiohttp`, `websockets`, and OpenAI-SDK clients; F9 Settings gains a Network category.

**Behavior**
- Custom-CA trust adds to `certifi`, never replacing it.
- Invalid config values log an error and fall back to verification-on; disabling verification warns once and passes a CERT_NONE `SSLContext` rather than bare `False`, which `websockets` ≥14 rejects for `wss://`.
- Ten direct `requests` calls (search backends, HuggingFace streaming, Anthropic summarization) now honor the policy; previously they bypassed it.
- The Network category seeds from pending edits, so hopping categories no longer shows stale values.
- The eagerly imported `tls_trust` module raises the UI-ready module census ratchet from 970 to 972 (snapshot refreshed); the production diagnostic inventory records it under TASK-494.

**Migration**
- `tldw_api/client.py` adds an `ssl_verify` constructor param, resolved from the policy at bootstrap.
- The branch forks from the local docs line, so the diff also carries unrelated docs commits; the summarization and destination-shell test failures predate the fork point.

<sup>Written for commit 54372ad77adc3b3906aeddc3741fd5567752d081. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2223?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Addressed the Qodo review in c1465b021 (plus the rebase onto latest dev, which folded the requests seam into dev's new `create_default_session()` factory — the policy now rides that one seam for every LLM/summarization/search session):

1. **Corrupt CA breaks requests (Bug)** — FIXED. `requests_verify()` now parse-validates the custom PEM (`ssl.create_default_context(cafile=...)`) before building the merged bundle; corrupt content fails safe to default verification exactly like the SSLContext path always did. Regression test added.
2. **Docext HTTP client leaks (Rule violation)** — FIXED. `DocextOCRBackend.cleanup()` now closes the injected policy-aware httpx client (and the SDK wrapper) in openai mode; the OpenAI SDK only closes clients it owns. Cleanup stays exception-guarded.
3. **`ca_bundle_path` bypasses central validation (Security)** — FIXED. Both the runtime policy (`tls_verify_setting`) and the settings validator now route the path through `Utils/path_validation.validate_path_simple`; existence stays our check so the missing-file message stays actionable.
4. **Network APIs lack docstrings** — FIXED. Google-style docstrings on `SettingsNetworkTLS` and all four module functions.
5. **Saved mode renders stale (Bug)** — FIXED. The Network save branch now syncs the app's in-memory config mapping via `_app_config_update_target().update(...)`, the same shape as `_apply_appearance_save_result`; pilot test asserts the post-save mapping reflects the saved value.
6. **Network input bypasses shared validator** — FIXED together with 3 (the CA path flows through `path_validation` at validation/save time).
7. **Stale CA cache persists (Bug)** — FIXED. The merged-bundle fingerprint now hashes the custom CA's content (certifi stays stat-keyed); a rotated corp CA with identical size and restored mtime regenerates the bundle. Regression test included.
8. **TLS config bypasses Pydantic validation** — Declined with reasoning: the manual coercion intentionally mirrors the repo's established pattern for reading config sections at runtime (`Utils/egress.py::_config_enabled` for `[web_security]` does the same isinstance/lenient-string normalization, and `tls_verify_setting` must never raise). A Pydantic model around one tri-state scalar would diverge from every neighboring config reader; the behavior is pinned by a 16-case coercion test matrix.

Test evidence: 163 passed across the branch suites (`test_tls_trust.py` 38, egress 97, gateway 2, tldw_api 2, realtime 3, settings 21 incl. three new regression assertions); ruff clean.

PR #2233, opened 2026-08-30
fix(console): Context sidebar UX pass — fit the default, close the 118-128 dead zone, replace the dead star column

rmusser01:
## What this is

A UX/HCI review and UAT of the Console screen's **Context** sidebar, plus the fixes. Everything here was measured against the running app via a sandboxed headless harness (`output/ux-review-console/`), not read off the source.

The headline defect: the rail's default configuration asked for **51 rows in a 32-row viewport** at 160×48, and overflowed at **every one of ten** terminal geometries measured — including 200×60, a full-screen terminal on a 27" display. Three of seven sections were invisible on a fresh install.

## Measured outcomes

| Geometry | Content rows before | After | Fits now |
|---|---|---|---|
| 200×60 | 50 (viewport 45) | 45 | ✅ |
| 160×48 | 51 (viewport 33) | 33 | ✅ all 7 headers reachable |
| 140×40 | 51 (viewport 24) | 30 | ❌ still overflows |
| 110×30 | 51 (viewport 16) | 30 | ❌ still overflows |

**An 11-column dead zone is closed.** Between 118 and 128 columns the Inspector auto-opened itself, tripping `resolve_console_rail_priority` and force-collapsing Context to a 13-column stub — so a *one-column* resize (117 → 118) swapped which sidebar you had, unasked and unexplained. Context now survives 117–135.

## Commits

| | |
|---|---|
| `fit the Context rail default layout` | 5 open sections → 2 (TASK-23193) |
| `stop the Context rail saying "no character" twice` | duplicate empty state (TASK-23194) |
| `replace the dead conversation star with an action menu` | Favourite / Change status / Archive / Rename / Delete (TASK-23200) |
| `name the Context rail and say what is below its fold` | `Context ◂`; hint names hidden sections (TASK-23195) |
| `stop the Context rail repeating provider and model` | shown 3× simultaneously (TASK-23196) |
| `close the 118-128 column dead zone` | automatic opens never evict a visible rail (TASK-23197) |
| `give the Context rail collapse-all and expand-all keys` | rail had no BINDINGS (TASK-23198) |
| `retire the tautological Conversation row` | (TASK-23199, AC1 deferred) |
| `bring the user guide and Textual lessons up to date` | docs + 3 lessons |
| `mirror the Inspector rail header` | keeps both rails in one visual language |

## Four findings from the audit were wrong, and are recorded as such

This matters more than the fixes, because the failure mode was consistent — inferring behaviour from widget attributes instead of the framework's actual contract:

- **"Keyboard focus can land on a control that paints nothing."** False. The audit queried `can_focus and display`, but a widget's own `display` stays `True` under a *hidden ancestor*. Textual's real `focus_chain` excludes all three. No fix needed.
- **"Tab is a keyboard trap — WCAG 2.1.2 failure."** Tab genuinely doesn't leave the rail, but that is deliberate (`ChatScreen.action_focus_next`, TASK-2154.11). WCAG 2.1.2 wants a keyboard way out *plus advice when it isn't Tab* — **F6 exits in one press** and the footer advertises it permanently. Criterion met. Tests now pin this so it is not "re-found".
- **The chrome mechanism.** Headers are 1 content row + 1 border row; `Widget.size` *excludes* borders. The row counts held; the described "2 blank rows + separator" gutter does not exist.
- **`Conversation: None` was not a contradiction.** It derives from a *persisted* id, so an unsaved session genuinely has none. The row went anyway — when a chat *is* saved it read `Conversation This conversation`, a tautology.

Also: the per-section allocator is **not** dead code. It engages for the *activated* section when the rail overflows; `allocation: None` means "shown in full".

## Deliberately not done

- **Sessions/Conversations merge** (TASK-23199 AC1). The case is strong — Sessions is now a header plus one row Conversations already shows. But `session_open` is also the **TASK-14810 migration seed** that `coerce_console_rail_preferences` uses as the fallback for `workspace_open` *and* `conversations_open`; a payload predating that split carries only `session_open`. Removing it naively strands those users. Recorded on the task with the blast radius (12 test files).
- **140×40 and below still overflow.** A variant removing the header separator made them fit; it was built and rejected on review because that rule is a deliberate cross-rail contract with the Inspector's title band.

## Notes for review

- **`preflight` is red on the production diagnostic inventory — and so is `origin/dev`.** Verified in a clean worktree: dev alone reports `task_494_calls: 7364 → 7384`. This branch reports `→ 7387`; the +3 are three `logger.error` calls in `workspace.py` that log only `type(exc).__name__` — no user content, ids, or paths. I did **not** regenerate, because that would fold dev's ~20 unreviewed diagnostics into this PR, which is the exact failure that artifact exists to prevent.
- Several test failures on this branch are pre-existing on dev (`schedules_workbench`, `session_settings`, `right_rail`, `rail_handle`, `inspector_compact_access`, 2 workbench snapshots). Each was verified by re-running with the tree stashed or against a pristine dev worktree.

## Reproducing

```bash
cd output/ux-review-console
../../.venv/bin/python uat_context.py all      # 6 scenarios
../../.venv/bin/python probe_final.py          # 117-135 col band sweep
../../.venv/bin/python probe_120.py            # layout containment at 118-128
```

Sandboxed to `output/ux-review-console/sandbox-context/`; no real config or user data is read or written.

https://claude.ai/code/session_01CrN2mbF8byxMfPNPtVJwFb

---
## Summary by cubic
UX pass on the Console's Context rail: the default layout now fits the viewport, the rail survives the 118–128 column band instead of being evicted by the Inspector, and the dead star column became a working action menu.

**Bug Fixes**
- Default open sections drop from five to two (Sessions, Conversations), so the rail fits at 160×48 and 200×60 with all headers reachable.
- The Inspector's auto-open no longer evicts a visible Context rail; Context now stays across 117–135 columns.
- Removed the duplicate no-character text, the provider/model repetition, and the tautological Conversation row.
- Both rail headers swapped ASCII-art labels for a readable name plus a glyph resolved through `resolve_glyph`; the overflow hint names hidden sections instead of saying "more sections — scroll".
- The action menu reads the conversation's state from the database when it opens, instead of trusting the display row's `status` field, which carried non-canonical values that made resolved conversations offer Archive instead of Unarchive.
- The overflow hint counts only headers strictly below the fold.
- Fixed change-status, rename, and delete handlers that would have raised `AttributeError` on first use, and routed rename through the shared input-validation module.

**New Features**
- Each conversation row's disabled star is now an asterisk that opens a paged menu with Favourite, Change status, Archive, Rename, and Delete; favorite state moved to the title line.
- The rail gains ctrl+shift+left/right bindings to collapse and expand all sections while focus is inside it.
- The menu and its model load lazily, so both stay off the boot path and the UI-ready module census stays under its ratchet.
- The menu's width is pinned by test, since the CSS bundler rejects the interpolated constant.

Four audit findings were wrong and are pinned in tests, including the Tab "trap" (deliberate; F6 exits and the footer advertises it). 140×40 and below still overflow — a fitting fix was rejected for breaking a deliberate cross-rail header contract. The Sessions/Conversations merge stays deferred because `session_open` seeds the TASK-14810 migration fallback. Existing stored preferences are honored; new defaults apply to fresh installs only. The production diagnostic inventory was regenerated; its only delta is four `logger.error` calls that log just the exception type.

<sup>Written for commit 87b64036663616727a84124a0dfda4fc2a4a3f2e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2233?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/User_Guide/console.md
+++ b/Docs/User_Guide/console.md
@@ -37,12 +37,15 @@ Top to bottom:
   **Attach context**, **Search Library**, **Help**. (**Save as Chatbook**
   lives in the composer's **Menu** button, left of the draft.)
 - **Left rail: "Console context"** — separate sections for **Sessions**
-  (the active conversation), **Workspaces** (named workspaces and their
+  (the active chat), **Workspaces** (named workspaces and their
   conversations), **Conversations** (Default and unassigned conversations),
   **Model**, **Agent**, and **Details**, plus **Character** when character
-  avatars are enabled. The
-  full-width **<---------|Context** header button collapses the rail; while
-  collapsed, the **Context->** handle on the far left brings it back.
+  avatars are enabled. **Sessions** and **Conversations** start open; the
+  rest start collapsed, so the whole rail fits without scrolling on an
+  ordinary terminal. The full-width **Context ◂** header button collapses
+  the rail; while collapsed, the **Context->** handle on the far left
+  brings it back. With focus anywhere in the rail, **Ctrl+Shift+←**
+  collapses every section and **Ctrl+Shift+→** expands them all.
 - **Conversation pane** — titled "Conversation", extended to
   "Conversation | \<session title\>" once a session is active.
   Above it sits the session tab strip: one button per tab (each with a
@@ -103,9 +106,11 @@ Top to bottom:
 Context sections keep complete reading bodies up to their own limits: 15 rows
 for Sessions, Model, Agent, and Details; 20 for Workspaces and Conversations;
 and 35 for Character. Inspector sections keep a 20-row limit. **▼ more —
-scroll** means more content remains inside the current section; **▼ more
-sections — scroll** means complete later sections remain below, so scroll the
-whole Context or Inspector rail.
+scroll** means more content remains inside the current section. In the
+Context rail the outer hint names what is below the fold instead — **▼ Agent
+· Details · +1** — listing as many hidden sections as fit and counting the
+rest, so you can tell whether scrolling reaches what you want. The
+Inspector's **▼ more sections — scroll** keeps the generic wording.
 See [Reading long Context and Inspector sections](console/context-and-rag.md#reading-long-context-and-inspector-sections)
 for pointer and keyboard navigation.
 
@@ -242,13 +247,13 @@ composer-level strip below shows once setup completes.
 
 | Control | What it does |
 |---|---|
-| **<---------\|Context** / **Inspect\|--------->** headers | Collapse the open Context or Inspector rail; the entire painted header is the button. |
+| **Context ◂** / **Inspect\|--------->** headers | Collapse the open Context or Inspector rail; the entire painted header is the button. |
 | **Context->** handle | Reopens the collapsed "Console context" rail when the viewport can retain a usable transcript. |
 | **<-Inspect** handle | Reopens the collapsed "Inspector" rail when the viewport can retain a usable transcript; shows badges like "1 appr" (pending approvals) or "art" (artifact ready). |
-| **Sessions** section | Names the active conversation. |
+| **Sessions** section | Names the active chat. Hovering it shows the durable conversation id. |
 | **Workspaces** section | Shows every named workspace with its associated conversations in a native Tree. Its compact strip keeps **Switch**, **New**, and **RAG** together; **Switch** is also the route to Default. Starred conversations sort first within their workspace. |
-| **Conversations** section | Independently searches, starts, stars, and resumes only Default and unassigned conversations; starred entries sort first — see [Context & RAG](console/context-and-rag.md#workspaces-and-conversation-ownership). |
-| **Model** section | Read-only Provider / Model / Temperature / Max tokens lines plus a **Configure** button that opens Console Settings. |
+| **Conversations** section | Independently searches, starts, and resumes only Default and unassigned conversations; favourited entries sort first and are marked beside the title. Each row carries an **\*** that opens its action menu — Favourite, Change status, Archive, Rename, and More ▸ Delete. See [Context & RAG](console/context-and-rag.md#workspaces-and-conversation-ownership). |
+| **Model** section | Read-only Temperature / Max tokens / system-prompt lines plus a **Configure** button that opens Console Settings. The active provider and model are read from the status bar, which shows them at every width. |
 | **Agent** section | Live run status and the full run log — see [Agent runs & tools](console/agent-runs-and-tools.md). |
 | **Details** section | Storage, sync, file tools, server, and handoff status for the workspace. |
 | **Character** section | Appears only when the character-avatar preference is on. Its complete portrait is centered and keeps its aspect ratio; it only scales down to fit and is never stretched, cropped, or enlarged merely to fill the 35-row body. |
@@ -526,7 +531,16 @@ are covered on the child pages, chiefly [Context & RAG](console/context-and-rag.
   "Console: Change model…".
 
 —
-*Verified against working tree — 2026-08-27 (TASK-23021: the Get started
+*Verified against working tree — 2026-08-30 (TASK-23193/23195/23196/23197/
+23198/23199/23200, Context rail UX pass: the rail's default open set is now
+Sessions + Conversations and the whole rail fits at 160x48 with all seven
+headers reachable; the header reads **Context ◂**; the outer hint names the
+sections below the fold; the Model section no longer repeats the status
+bar's provider/model; conversation rows carry an **\*** action menu in place
+of the retired star column; and the 118-128 column band no longer evicts the
+rail. Measured with the headless UAT harness in `output/ux-review-console/`
+across ten terminal geometries, plus layout-containment probes at 118/120/
+125/128 columns). Verified against working tree — 2026-08-27 (TASK-23021: the Get started
 card's snow backdrop is now a still frame — mounted-harness check on the
 real unconfigured ChatScreen: field renders behind the card, no timers, no
 repaints between resizes; idle CPU 0.02–0.05% vs 2.0–7.4% with the retired
--- a/Docs/security/production-diagnostic-inventory.json
+++ b/Docs/security/production-diagnostic-inventory.json
@@ -2427,8 +2427,8 @@
       "reason": "remaining Chatbook production diagnostic owner"
     },
     {
-      "call_count": 36,
-      "diagnostic_digest": "bff751ecf7047846d2a0",
+      "call_count": 39,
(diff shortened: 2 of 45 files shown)


rmusser01:
All seven findings addressed. Two were real bugs, and writing the tests for #6 surfaced a third that the review did not catch.

**#1 Menu uses display status — fixed (real bug).** Confirmed exactly as described: of the five values `row.status` can carry (`active`, `in-progress`, `open`, `workspace`, `workspace-thread`), only one is canonical, and the other four normalise to `in-progress`. So a **resolved** conversation shown as an "active session" row offered *Archive* instead of *Unarchive*. The menu now resolves the state from the database when it opens (`_console_conversation_state`) — one primary-key lookup at open time, not a per-render cost — and the misleading `conversation_state` attribute is no longer set from display copy. Regression test: `test_the_menu_reads_canonical_state_not_the_row_display_copy`, plus a fallback test for unsaved/missing/broken-DB.

**#7 Hint names sections above — fixed (real bug).** Correct: the check was `not (top <= y < bottom)`, which counts headers above the viewport too, so after scrolling down a `▼` hint could name a section only upward scrolling reveals. Now strictly `y >= bottom`.

**#6 Persistence handlers lack tests — fixed, and it found a bug.** Added `Tests/UI/test_console_conversation_persistence.py` (11 tests: write, refusal, failure, confirmation for each of change-status / rename / delete). Writing them immediately failed with `AttributeError: 'ConsoleWorkspaceController' object has no attribute 'app'` — all three handlers used `self.app` and `self.screen.run_worker`, but the controller exposes `app_instance` and a bound `run_worker`. **Every one of those three paths would have raised on first use.** The same slip had put `self.app` in the new state lookup on `ChatScreen`, where `app_instance` owns the DB. This is precisely the gap the finding was pointing at.

**#2 Rename bypasses shared validation — fixed.** Now routes through `Utils.input_validation.validate_text_input` + `sanitize_string` with a stated 200-character ceiling, and refuses with a message rather than silently dropping. Test: `test_rename_refuses_a_title_the_shared_validator_rejects`.

**#3 rich splits third-party imports — fixed.** `from rich.cells import cell_len` folded into the contiguous third-party group.

**#4 compose lacks required docstring — fixed.** Google-style summary and `Returns:` added.

**#5 Menu width duplicated — fixed differently, and here is why.** The intent is right. The suggested remedy is not available in this repo: `css/build_css.py` lifts `BUNDLED_CSS` into the built stylesheet **statically** and rejects anything that is not a plain string literal. Interpolating the constant produced `BUNDLED_CSS is not a plain string literal; it cannot be lifted into the built stylesheet` and broke the CSS bundle guard outright. The two encodings are pinned together by a test instead (`test_menu_width_constant_and_stylesheet_cannot_drift`), which prevents the drift the finding is about while respecting that constraint.

Separately, `UI latency guardrails` failed on the first push — the menu and its model were resident at `_ui_ready`, breaching ADR-097's module ratchet. Rather than raise the constant (which the ADR forbids), the cost is deferred: the only non-lazy reference was a module-level import feeding two `@on(...)` decorators, and Textual dispatches by handler-name convention just as well. The feature now stays off the boot path until the menu is opened.

`preflight` is green, including the production diagnostic inventory — now regenerated, since dev has pinned its own drift and the only remaining delta is this branch's four diagnostics, each of which interpolates `type(exc).__name__` and nothing else.

rmusser01:
Re-review shows **#2, #3, #4, #7 resolved** and **#1 outdated** (the code it pointed at was replaced — the menu no longer reads `row.status`; it resolves the canonical state from the database in `ChatScreen._console_conversation_state`, covered by `test_the_menu_reads_canonical_state_not_the_row_display_copy`).

The two still flagged are both addressed on the branch — one of them differently on purpose, so flagging it is fair and I want the reasoning on the record rather than silently dismissed.

**#6 Persistence handlers lack tests — done.** `Tests/UI/test_console_conversation_persistence.py` is on the branch with 11 tests covering all three handlers: the successful write, the refusal, the failure path, and the confirmation gate.

- change status → `test_change_status_writes_the_canonical_state_and_confirms`, `test_archive_writes_resolved_rather_than_a_separate_flag`, `test_a_missing_conversation_warns_instead_of_crashing`, `test_a_failing_write_reports_rather_than_going_silent`
- rename → `test_rename_writes_the_new_title`, `test_rename_refuses_a_title_the_shared_validator_rejects`, `test_rename_to_the_same_title_is_a_no_op`
- delete → `test_delete_soft_deletes_and_confirms`, `test_delete_asks_before_it_writes` (asserts nothing is written before *and* after declining the dialog)

Worth noting these earned their keep immediately: writing them failed with `AttributeError: 'ConsoleWorkspaceController' object has no attribute 'app'`, because all three handlers used `self.app` and `self.screen.run_worker` while the controller exposes `app_instance` and a bound `run_worker`. All three paths would have raised on first real use. That bug was invisible to the review and to me, and only the tests this finding asked for surfaced it.

**#5 Menu width duplicated — intent addressed, remedy changed, and here is why.** The finding is correct: anchoring clamps against `MENU_WIDTH` while rendering uses the CSS `width`, so the two must not drift. Interpolating the constant is not possible in this repo — `css/build_css.py` lifts `BUNDLED_CSS` into the built stylesheet **statically** and rejects anything that is not a plain string literal. I tried exactly the suggested fix first; it failed the CSS bundle guard with:

```
BUNDLED_CSS is not a plain string literal; it cannot be lifted into the built stylesheet
```

So the two encodings are pinned by a test instead — `test_menu_width_constant_and_stylesheet_cannot_drift` parses the width out of `BUNDLED_CSS` and asserts it equals `MENU_WIDTH`. That fails the moment either side changes alone, which is the drift the finding is about, while keeping the stylesheet liftable. The constraint is recorded in a comment on `MENU_WIDTH` so the next person does not retry the interpolation and rediscover the build failure.

If the rule can accept a test-enforced invariant where static lifting forbids interpolation, this one is satisfied; if not, it needs the rule and `build_css.py` reconciled, which is outside this PR.

All CI checks pass, including `UI latency guardrails` and `Derived artifacts reproduce from their sources`. Local run over the changed surface: 349 passed, 1 xfailed, and 2 failures in `test_workbench_visual_snapshots.py` that reproduce identically on a pristine `origin/dev` worktree (same two test ids) — pre-existing, not from this branch.

PR #2253, opened 2026-08-31
TASK-23113.6: Add credential-safe trace privacy projections

rmusser01:
## Summary
- add mandatory bounded credential filtering across trace storage, live revision projection, legacy capture, copy, and export
- add immutable built-in PII masks with run-frozen global, conversation, and next-send policy
- separate Capture, PII masking, and Safe/Full viewer disclosure with conservative migration and schema v58

## Verification
- 541 focused semantic/privacy/persistence/controller/viewer tests passed
- 45 oversize/privacy regression tests passed
- 6 focused Settings/export/migration tests passed
- changed modules compile/import and git diff --check passes

## Known unrelated baselines
- three stale Settings assertions: category count, Network guidance, and Schedules ownership
- SkillImportChoiceModal.DEFAULT_CSS remains the existing CSS allowlist offender

Implements ADR-097.

---
## Summary by cubic
Adds mandatory bounded credential filtering and optional built-in PII masking to all durable trace paths, and makes Safe/Full a viewer disclosure profile independent of capture policy. The credential sanitizer enforces node/depth/text bounds and returns a content-free omission when exceeded. PII masking is immutable per policy revision; canonical messages stay unchanged and are projected through frozen masks.

- Adds a trace privacy dialog separating capture, PII redaction, and viewer profile controls.
- Schema v62 migrates legacy capture detail into a new policy table with independent `capture_enabled` and `pii_redaction_enabled` columns defaulting to inherit, and the viewer profile defaults to `safe`.

<sup>Written for commit 399b0228f9883d3d859d58312f05ab082ff06c31. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2253?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Addressed the Qodo review on the rebased head e235d79c88:

1. Fixed next-send privacy consumption: the override is now frozen into turn preparation and consumed only at accepted/provider-dispatch admission with exact revision matching. Paused/rejected turns preserve it, and a newer choice armed while a turn is paused is not consumed.
2. Kept `_resolve_reference_values` canonical by design. That method verifies/reconstructs the provider surface; applying trace-only PII masks there would change bytes sent to the provider. The policy-aware trace viewer/export boundary is `project_semantic_revision_trace_message`, now explicitly documented.
3. Retired canonical locators intentionally fail closed for provider replay instead of substituting a masked trace artifact. Historical trace disclosure already resolves the immutable policy artifact; a new edit/retirement regression test proves the old revision remains coherently viewable with its frozen PII mask while the current canonical message remains independent.
4-10. Added the requested Google-style Args/Returns/Raises documentation to all cited public helpers, repository/controller methods, projection APIs, and the dialog constructor.
11. Replaced manual parsing of the new privacy fields with one strict shared Pydantic model. Malformed values fail to all-Safe privacy defaults, unknown Console settings remain compatible, and the model lives under Utils so config-first import does not reintroduce the eager Chat package cycle.

Fresh post-rebase verification: 290 focused controller/semantic/trace/config/privacy tests passed; 48 contiguous v57→v62 migration tests passed; targeted Ruff and `git diff --check` passed.

PR #2254, opened 2026-08-31
docs: Console UX/HCI review findings + verified root cause for the dead core loop

rmusser01:
## What this is

A live UX/HCI review of the Console screen at `dev` (`0ef6f3fd4`), filed as 22 atomic backlog tasks, plus a verified root cause for the one finding that makes the product unusable.

**This PR contains no production code.** It is `backlog/tasks/` only — 22 files, 629 insertions. See "Why no code fix" below; that is the substantive result, not an omission.

## The headline: Console's core loop is dead on `dev`

Every Console send is refused before the provider is contacted:

> Trace capture blocked — Owner: This Console send — Problem: Trace capture could not be saved. — **Impact: The provider was not contacted.**

Reproduced on a clean first-run profile *and* on an existing profile, against a local provider that answers the same request in **0.8s** over `curl`.

### Root cause (verified, not inferred)

I instrumented `TraceCallPersistenceError.__init__` to dump a stack and drove the real app:

```
console_agent_bridge.py:2895      _consume → gateway.stream_chat(...)
console_provider_gateway.py:3902  stream_chat → self._reserve_trace_call(...)
console_provider_gateway.py:2224  raise TraceCallPersistenceError(reservation_status="not_established")
--- active exc: None
```

`active exc: None` — nothing threw. Nothing failed to write. The guard fires on a permanent condition:

```python
if self._trace_call_boundary_factory is None:
    raise TraceCallPersistenceError(reservation_status="not_established")
```

**That factory is never wired in production.** `capture_mode` defaults to `CAPTURE_ON` (`console_turn_preparation.py:118`), `stream_chat` routes every Capture-On send through `_reserve_trace_call`, and both production callers of `ensure_provider_gateway` omit the seam — `chat_screen.py:6392` and `console_launch_wake.py:212`. `ConsoleTraceCallBoundary` is constructed **only in `Tests/`**.

Introduced by `c78f641ad` *"feat(console): implement reference-backed semantic trace ledger"* — same-day, 30 commits before the reviewed tip. The consumer landed without the producer.

### Why CI is green

Every test injects a factory, and the shared helper `_capture_on_prepared_request` (`test_console_provider_gateway.py:141`) **silently installs one when it is absent**. The suite structurally cannot observe the shipped configuration.

## Why no code fix in this PR

I implemented the obvious "degrade to capture-off" fix TDD-first. It went green, then broke this:

```
FAILED test_capture_on_without_durable_boundary_cannot_enter_adapter
```

[...]

PR #2256, opened 2026-08-31
fix(console): Console UX review — 14 fixes incl. the dead core loop, 8 findings retired

rmusser01:
Stacked on #2254 (the review + root cause). Every change here was traced to
code first, then written test-first.

## Fixed (5)

| Task | What was wrong | Fix |
|---|---|---|
| **25720** | `Button:disabled` stacked `opacity: 50%` on top of an already-dim `$text-disabled`, compositing to a **measured 1.34:1** — a disabled primary read as *absent* while the dismissive action beside it stayed near 14:1 | New `$ds-text-disabled-readable` (5.09:1); dropped the compounding opacity. Textual's variant-nested `&:disabled` outranks a bare `Button:disabled`, so the rule carries `.-style-default`/`.-style-flat` specificity. Deleted the `Button.model-import:disabled` opt-out that existed only to defeat the dimmer |
| **25716** | The tracker showed ✓ for steps the user never configured, while the summary screen simultaneously reported them unconfigured | `SetupWizardProgress` already had a three-state model documented as *"a visited step whose probe demonstrably failed must not wear the ✓"*; only its input was too narrow. New pure helper `setup_attention_ids()` feeds the same mechanism |
| **25725** | Two unrelated failures shared one sentence that the rail clipped to `Workspace conversations a…`, naming neither cause; the real exception sat at `debug` | Split into two short, distinct, rail-legible messages; raised the swallowed exception to `warning` so "check the app log" leads somewhere |
| **25731** | Connection-test copy named subsystems, not outcomes, and `"reached; not confirmed"` mixed a tick with a failure | Rewritten to name what happened *and* the next step, **while preserving the probe-category fragments** a test asserts for diagnosability |
| **25722** | Once "Discard changes" appears, "Cancel" silently changes meaning from *abandon my edits* to *stay here* | Relabels to "Keep editing" so the pair names its two outcomes |

## Retired as invalid (4)

Each was implemented or traced, then withdrawn with evidence on the task.

- **25718** — broke 34 tests led by one whose docstring reads *"Skip-safe: leaving the model step untouched must not touch config."* Pre-selecting a model the user never chose is exactly what the wizard forbids.
- **25717** — mis-framed. The block I inspected is a *resume checkpoint*, not applied config (the endpoint's real home is `api_settings.<provider>.api_url` via `persist_provider_setup`), and retaining it there breaks a test asserting the endpoint is scrubbed from step memory.
- **25726** — the status strip does not clip. It is a `HorizontalScroll` with a deliberate priority order stated in code (Library policy ahead of provider/model metadata). Only the missing overflow affordance survives, and the strip is locked to `height: 1`.
- **25730** — real but **1.03×**, and selection already carries bold+underline consistently across every `OptionList`, `Tree` and `DataTable`. The only clean fix edits that shared rule app-wide.

**25713 is blocked on 25712**: the 30s blank assistant row was a *blocked* run, never a streaming one. The app already has `[streaming]` and `Generating…` indicators; neither could fire because dispatch never happened.

## Corrections I made to my own work

- I claimed no test pinned the provider-test copy. **Wrong** — the test parametrises on *fragments*, which a whole-sentence grep missed, and my first rewrite broke all 7 cases. Fixed to satisfy both properties.
- I verified 25731 with a `-k`-filtered run that had deselected 10,710 tests and hid those failures behind a green number. A filtered pass is not a pass.

## Verification

490 passed across the touched suites · wizard suite **392 passed / 0 failed** · CSS bundle reproduces from source.

Baselines confirmed pre-existing on clean `dev` (unchanged by this branch): 13 in `test_console_workspace_controller.py`, 2 in `test_first_run_wizard_live_contract.py`, 1 in `test_console_trace_settlement.py`, 1 order-dependent flake.

https://claude.ai/code/session_01XHdMC7n32o6Tkw7FKKve2o

---
## Summary by cubic
Fixes the verified Console UX findings and records all 22 review items as backlog tasks. Sends that previously defaulted to Capture-On and were rejected before provider dispatch now fall back to Capture-Off when durable capture is unavailable, so chat proceeds without trace capture while the intentional Capture-On safety guard remains.

**User-facing fixes**
- Improves disabled-button contrast, toggle state glyphs, setup progress and summary guidance, workspace errors, unsaved-change actions, and connection-test copy; the unauthorized-test verdict now points at the current step's Authentication section, and the "Keep editing" control reverts to "Cancel" once the dialog is clean.
- Records database-open failures as metadata-only events and removes the ineffective restart advice.
- Keeps interrupt cards within 60% of the transcript, blocks hotkeys while decisions are pending, and makes composer layout and shortcuts reflect the current state.
- Renames "Assistant Library access" to "Agent Library access" while retaining the old wording as a search alias.

**Review scope**
- Retires 25717, 25718, 25726, and 25730 as invalid or intentionally unsupported findings.
- Marks 25815 blocked on 25814 after confirming the existing pending-state indicators were never reached.

<sup>Written for commit 7a55d0c4ba30fa127177b00eb0537319af7fd7ce. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2256?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/security/production-diagnostic-inventory.json
+++ b/Docs/security/production-diagnostic-inventory.json
@@ -2442,7 +2442,7 @@
     },
     {
       "call_count": 39,
-      "diagnostic_digest": "88bd482d7a40a93a5e6f",
+      "diagnostic_digest": "73af1d5e30d31ec6aaeb",
       "owner": "TASK-494",
       "path": "tldw_chatbook/UI/Console_Modules/workspace.py",
       "reason": "remaining Chatbook production diagnostic owner"
--- a/Tests/Chat/test_task_25712_capture_capability.py
+++ b/Tests/Chat/test_task_25712_capture_capability.py
@@ -1,0 +1,52 @@
+"""TASK-25712: don't claim Capture-On the runtime cannot deliver.
+
+`ConsoleProviderGateway._trace_call_boundary_factory` is never supplied in
+production -- both callers of `ensure_provider_gateway` omit it and
+`ConsoleTraceCallBoundary` is constructed only in tests -- while
+`ConsoleTurnPreparation.capture_mode` defaults to CAPTURE_ON. So every real
+send reached `_reserve_trace_call`, whose first statement raises when the
+factory is None, and the provider was never contacted.
+
+The dispatch guard is CORRECT and stays: a Capture-On turn that cannot record
+a durable boundary must not reach a provider
+(test_capture_on_without_durable_boundary_cannot_enter_adapter pins it). The
+error is upstream -- preparing a turn as Capture-On when the runtime has no
+way to capture it. A runtime that cannot capture must say so, and the app
+already models running without capture (`one_shot_capture_off`).
+"""
+
+from __future__ import annotations
+
+from tldw_chatbook.Chat.console_provider_gateway import ConsoleProviderGateway
+
+
+def test_gateway_without_a_factory_reports_no_durable_capture() -> None:
+    gateway = ConsoleProviderGateway()
+    assert gateway.supports_durable_capture is False
+
+
+def test_gateway_with_a_factory_reports_durable_capture() -> None:
+    gateway = ConsoleProviderGateway(
+        trace_call_boundary_factory=lambda _request, _resolution, _route: object()
+    )
+    assert gateway.supports_durable_capture is True
+
+
+def test_capture_mode_decision_consults_the_runtime_capability() -> None:
+    """The policy seam must not promise capture the gateway cannot honour.
+
+    Reads the module source rather than one method: dev inlined this decision
+    into the submit path (it used to live in `_capture_mode_for_preparation`,
+    which no longer exists), so pinning a method name would break on the next
+    refactor while the property it guards stayed correct.
+    """
+    from pathlib import Path
+
+    import tldw_chatbook.Chat.console_chat_controller as controller
+
+    source = Path(controller.__file__).read_text()
+    assert "supports_durable_capture" in source
+    # It must gate the CAPTURE_ON branch, not sit somewhere inert.
+    capture_on_index = source.index("ConsoleTraceCaptureMode.CAPTURE_ON")
+    guard_index = source.index("supports_durable_capture")
+    assert abs(guard_index - capture_on_index) < 4000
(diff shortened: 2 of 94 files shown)


rmusser01:
## Qodo review adjudicated — 4 accepted, 1 declined (`be5903d57`)

**✅ Keep editing label goes stale** — correct, and it was my bug. `_refresh_dirty_state` only synced Save, so after a successful save or a manual revert the control still read "Keep editing" while a cancel dismissed immediately. That recreated the exact ambiguity TASK-25722 set out to remove. Now restores "Cancel" and hides Discard when clean, with an added assertion pinning it.

**✅ Back points to wrong step** — correct. Verified the ordering: Provider is step 2, so Back goes to Welcome, which has no key input — while the verdict renders on the Provider step that already holds it. Also wrong for stored/environment credentials, where "re-enter" isn't the remedy. Now points at the Authentication section on the current step. Kept the `unauthorized` fragment that `test_connection_failed_verdict_names_bounded_probe_category` asserts for diagnosability.

**✅ `setup_attention_ids` missing Args/Returns** — matches the repo convention in `CLAUDE.md`. Added.

**✅ Wizard behavior lacks integration test** — fair. The unit tests covered the helper but not the wiring, so a rebuild regression could have passed them. Added `Tests/Wizards/test_task_25716_tracker_integration.py`, which drives the real `build_setup_progress` projection and asserts the rendered state, including that legitimately-skippable steps are *not* flagged.

**❌ `_DisabledButtonHarness` PascalCase — declined.** Leading-underscore test-local classes are this repo's convention, not an exception: `grep -c '^class _[A-Z]' Tests/` returns **1,998**. Renaming this one would make it the outlier, and the rule as applied would flag nearly every test helper in the codebase. Happy to revisit if the rule is meant to cover `Tests/`.

454 passed across the touched suites.

PR #2281, opened 2026-09-01
perf(css): split the agentic-terminal module off the pre-first-paint parse (TASK-25812)

rmusser01:
Implements the owner decision on TASK-25812 ("split-by-screen"), also advancing TASK-24451.

## What

`components/_agentic_terminal.tcss` stays the single source of truth. `build_css.py` now partitions it at build time: multi-screen rules remain in the boot bundle; the Console/Library/Settings rules ship as three generated sheets attached to those screens' `CSS_PATH`, which Textual parses **lazily on first visit** — at least 1.5 s after first paint, where the whole 283 KB used to be parsed before it.

## Measured

| metric | before | after |
|---|---:|---:|
| boot bundle | 672,141 B | **470,401 B** |
| boot rules parsed | 4,231 | **3,304** |
| boot CSS census (ratchet) | 879,439 B — **breached** | **679,726 B** |
| bundle cold parse (interleaved, same session) | ~243 ms | **~156 ms (−36%)** |

Moved cost, paid once each on first visit: console 30.6 ms / library 35.6 ms / settings 16.5 ms.

Ratchet bookkeeping per ADR-097: `MAX_BOOT_PARSED_CSS_BYTES` **tightened 860,000 → 705,000** (banking the win); the anti-vacuity floor re-pinned 700,000 → 600,000 (it tripped on the success case). Only the CSS snapshot was refreshed — the ui-ready/preimport snapshot rewrites the refresh script also produced were reverted rather than blessing dev's own drift under this commit.

## Splitter safeguards — each earned by an incident during this change

- **Lossless partition asserted at every build** (`"".join(units) == text`) — a lossy split silently drops live CSS.
- **Cascade-order demotion to a fixpoint**: a moved block sharing a selector with a *later kept* block is demoted back to the bundle — it would otherwise win a tie it used to lose. Found live: `#settings-category-pane` would have lost its `overflow-y: hidden` override.
- **`AGENTIC_SPLIT_PINNED_TOKENS`**: all 856 moved tokens were audited against Python compose sites; `.settings-input-label` is composed by a *Personas* widget, so it stays in the bundle (5 of the 6 audit hits were substring/comment false positives).
- The sheets' variables preamble ends with an **empty sentinel rule** — several test contracts scan `selector { body }` pairs and swallowed the whole preamble into the first selector without it.
- The split **skips cleanly when the module is not in `CSS_MODULES`** (scratch trees, vendored subsets).
- `check_bundle_sync` now verifies **all 7 generated sheets** reproduce from source.

## Verification — paired arms

82 CSS-sensitive UI files (1,105 tests), ~37 min per arm, split vs branch base: 82 vs 37 failures, 51 split-only. The 51 were overwhelmingly contract tests hard-coding "this rule lives in the bundle" — which the split deliberately falsifies. After migrating them to union semantics (`APP_STYLESHEETS` / `app_css_text()` in `Tests/UI/consolidated_css.py`, or the owning screen's sheet for single-surface contracts): **49 of 51 heal; the remaining 2 are order-flakes that also fail/pass on the pristine base in isolation.**

## One finding unmasked, not caused — TASK-25890

The settings workbench-geometry contract had only ever passed because `DestinationHarness` loads **no agentic CSS at all**. Under the *production* stylesheet the branch base fails it identically (`#settings-boundary-note` at y=45, outside its pane at y 5..40). The test now runs under the production stylesheet set with the settings param `xfail(strict=True)` citing TASK-25890 (renumbered from 25814 after preflight caught a collision with dev's older task; the id was verified free across all 72 remote branches).

## Not moved (deliberately)

MCP/Lab/Watchlists/Personas/Home/Notes/ACP rules (~16 KB combined) stay in the bundle — each is small and each would need its own screen wiring plus a cross-surface audit. TASK-24451 records this and its still-open rule-count-ratchet AC honestly rather than closing.

https://claude.ai/code/session_013tVAwGVgaVATkgBcZb2Qdy

---
## Summary by cubic
Splits `components/_agentic_terminal.tcss` at build time so the source stays the single source of truth while Console/Library/Settings rules ship as three generated sheets, cutting boot CSS parse from ~243 ms to ~156 ms (−36%) and the boot bundle from 672,141 B to 470,401 B. Library and Settings sheets load lazily on first visit; the Console sheet stays on the boot path because mounting it at ChatScreen startup (the initial tab) cost every user a ~100 ms one-time parse and full-app restyle.

**Splitter safeguards**

- Lossless partition is asserted at every build — a lossy split would silently drop live CSS.
- A moved block sharing a selector with a later kept block — in the bundle or any later `CSS_MODULES` entry — is demoted back to the bundle so it can't win a cascade tie it used to lose.
- 16 cross-surface `console-*` tokens plus `.settings-input-label` are pinned in the bundle because other widgets compose them.
- The sheets' variables preamble ends with an empty sentinel rule so selector-scanning test contracts don't swallow the preamble.
- The split skips cleanly when the module is absent from `CSS_MODULES`; the staleness check requires the sheets whenever the module is present.
- All outputs stage in `.css-build-stage` and swap atomically, so a mid-build failure leaves the previous complete generation untouched.
- `check_bundle_sync` verifies all 7 generated sheets reproduce; the split-sheet contract rebuilds into a temp dir for byte-for-byte comparison.

**Ratchets and tests**

- `MAX_BOOT_PARSED_CSS_BYTES` tightened 860,000 → 806,000 with the console sheet back on the boot path; the anti-vacuity floor was re-pinned 700,000 → 600,000.
- CSS-sensitive contracts now read the app's full sheet set via `APP_STYLESHEETS` / `app_css_text()`; 49 of 51 split-only failures healed, the remaining 2 are order-flakes.
- A new test asserts each generated sheet carries only its own owner's rules and no pinned cross-surface token, catching any Console rule generated into the Library sheet.
- Unmasked a pre-existing bug: `#settings-boundary-note` escapes its impact pane under production CSS — filed as TASK-25890, test param is now `xfail(strict=True)`.
- Library and Settings each parse once on first visit: 35.6 ms and 16.5 ms.

<sup>Written for commit e1fd944bd4bd1876538e91173787ca2bc870648d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2281?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## CI triage on `15c0ff3585`

**Fast Lane / Derived Artifacts — was mine, fixed.** The 25812 task file on this branch was created by an append and had no frontmatter `id:`; the guard was right. Fixed, and the console sheet now rides `TldwCli.CSS_PATH` (see below).

**Perf Guard `_ui_ready` census (981/972) — dev's standing breach, not this branch.** Evidence, measured with the census's own harness:

| commit | census |
|---|---:|
| `912fb0f77e` — this branch's merge base, **pristine dev** | **981/972 breached** |
| this branch | 979 (marginally *better* than base) |
| dev 74 commits earlier (`85d449d3b2`) | 972 |

The 9+ modules that consumed the headroom are dev's own recent work — `console_trace_*`, `Terminal.{backend,io_actors,launch,protocol_gate,screen_model}`, `permission_summary_service` (importer-traced). Per ADR-097 the constant is not raised here; the repayment belongs to the commits that added the weight.

Full disclosure: I initially bisected my own commit for this, because my "pristine dev" control worktree was accidentally 74 commits older than the merge base and read 972. All four bisection arms read 979 — which was the tell that none of my pieces were the cause. `git merge-base` first; when a bisection says "none of the parts matter", suspect the baseline.

**Also in this push — a real improvement the investigation surfaced:** the console sheet moved onto the app's boot `CSS_PATH`. Loading it at first ChatScreen mount cost every user a one-time parse + full-app restyle (~100 ms) on the mount leg, since the Console is the initial tab. Boot-parsing it costs ~30 ms against the ~85 ms the split saves; library/settings stay fully lazy. Byte ratchet re-pinned to 806,000 (measured 780,368 + slack — still banking ~54 KB against the pre-split 860,000).

rmusser01:
## Review response — all 8 findings addressed on `8a68dd66c7`

| # | finding | outcome |
|---|---|---|
| 1 | Cross-surface console vocabulary | **Fixed — and it exposed a vacuous audit of mine.** My original cross-surface audit filtered *absolute* paths for the substring `"console"`, and the worktree directory is named `console-inspect-burndown` — every path matched "home", so the console audit could never report anything. Re-audited with repo-relative paths: **16 cross-surface tokens** (`console-action-*`, `console-rail-*`, `console-modal-header`, `console-workspace-action`), composed by Evals/MCP/Lab/Personas/Library/destination-rail. All 16 pinned to the bundle via `AGENTIC_SPLIT_PINNED_TOKENS`. |
| 2 | `endswith` integrity accepts extras | **Fixed** — the contract now rebuilds via `build_agentic_split` into a temp dir and compares byte-for-byte. |
| 3 | Splitter lacks unit tests | **Added** — and writing them caught a real bug beyond the review: `_unit_owner`/`_unit_selector_set` located the block brace with `find("{")` on *raw* text, so a preceding comment containing `{` swallowed the selector and silently pinned the block with comment prose as its bookkeeping. Comments now strip before brace location. (The demotion test also corrected its own first premise: a descendant selector is not a tie — different specificity; the contract is exact-selector/comma-member equality.) |
| 4 | `app_css_text` Returns | **Added.** |
| 5 | Sync check crashes on skipped split | **Fixed** — rebuilt-side reads are None-for-missing, symmetric with the committed side. |
| 6 | Inconsistent publish order | **Fixed** — sheets write *before* the bundle: a mid-run failure now duplicates moved rules (harmless) instead of dropping them. |
| 7 | Sheets invisible to boot staleness | **Fixed** — `_generated_css_is_stale` requires the sheets whenever the source module is present. The gate matters: the ungated version broke 9 scratch-tree staleness tests, which is how it earned its shape. |
| 8 | Cross-module cascade blind spot | **Fixed structurally** — the demotion pass now seeds with every later `CSS_MODULES` entry's selectors, enforced at every build rather than by the one-time review audit. |

**Answered rather than changed** — the harness-union observation: union harnesses model the steady-state app deliberately, and the exposure scenario ("a screen renders before ChatScreen is visited") is moot since `56fcf61a` put the console sheet on `TldwCli.CSS_PATH` (boot-parsed, with the rationale in the app-side comment). With the 16-vocabulary pinning, per-sheet ownership is now real for what remains; a per-screen-isolation harness would be a new test asset, not a fix to these.

rmusser01:
## On the one open finding (#8, harness union) — addressed by a different mechanism, with evidence

The finding's failure class is: *"a Console rule generated into another screen's sheet, or a Console-owned class actually needed on a different screen, and the union harness can't see it."* Both halves are now caught deterministically, without a mount:

1. **Misplacement** — `test_split_sheets_carry_only_their_own_owners_rules` (added on `e1fd944bd4`) asserts per generated sheet that every owner-prefixed token belongs to that sheet's owner, and that no pinned cross-surface token appears in *any* sheet. A Console rule landing in the Library sheet fails the build guard directly.
2. **Cross-surface need** — the 16-token cross-surface vocabulary is pinned to the bundle (finding 1's fix), and the console sheet additionally rides `TldwCli.CSS_PATH`, so no surface can render before console-owned rules are parsed.

Given those two, narrowing each harness to per-screen isolation would add churn without detection value: for the console harness the union differs from "bundle + console sheet" only by the library/settings sheets — which the ownership guard proves contain zero console tokens. The union harness models the steady-state app, which is what those geometry tests assert.

If a per-screen-isolation harness is wanted as a *new* test asset (e.g., to catch a future screen relying on another screen's sheet), that's a reasonable follow-up — but it's additive, not a defect in this change. Happy to file it as a task on request.

PR #2307, opened 2026-09-02
fix(library): unblock sequential review of media (find deadlock, escape focus, F6 scroll, prev/next)

rmusser01:
## Summary

Fixes the four keyboard/focus defects that block **sequential review** of Library media — the flagship "ingest a set (a conference, a tag-filtered result, hand-picked videos), then read through every analysis/transcript in order" flow — from a dual-agent UX critique of Library ▸ Media (scored 28/40). Each fix was built test-first and confirmed both at the production-CSS app-test level and live in the real TUI.

Also files the full critique backlog (22 tasks) and reconciles it against the current dev tip, where several findings had shifted since the review.

## Fixes (bug + P0 tracks)

- **task-28002 — find bar no longer deadlocks the keyboard.** Submitting the first in-Reader search recomposed the search controls, destroying the focused `Input`; screen focus went to `None` and every `check_action` gate that reads `self.focused` went dead — a total keyboard freeze until a mouse click (live-reproduced). The status/Prev/Next children are now persistent and display-gated (the task-22207 banner idiom); `sync_query_state` patches them in place.
- **task-28004 — Escape from the Reader focuses the loaded row, not the filter.** Escaping the Reader focused `#library-media-filter`, so the next keystrokes were swallowed (typed characters fell into the filter, Down was inert until Tab). New `_focus_library_media_items_pane()` lands focus on the loaded/selected row, making Escape-then-Down the sequential-review gesture.
- **task-28003 — F6 reaches the content scroller.** The Reader's only F6 target was the Find button, so the focusable content `ScrollView` was reachable by mouse alone — keyboard scroll was dead on a fresh open. The scroller is now the first F6 candidate (falling through to Find in non-Read modes, which `/` still reaches).
- **task-28005 — `]` / `[` walk next / previous item in the Reader.** New sequential-review navigation over the current browse result: neighbours come from the mounted rows (browse order, filter-respecting) and reuse the auto-loading selection path, so the Reader's mode carries across items — bracket-stepping in the Analysis tab reads every analysis in sequence. Gated to the plain Reader with a neighbour that direction, so both keys no-op and drop from the footer at the list ends (the honest-footer boundary idiom).

## Backlog

- Files tasks 28002–28022 + 28024 (user-approved "review set" design) from the critique, then reconciles them against dev tip: 28020 archived (obsolete — the Media route is a live screen again), 28013's in-canvas-filter half already shipped, 28010's stacking half already shipped (the Reader is now a 4-tab layout), and a new **task-28025** filed for toolbar/bulk buttons rendering clipped-invisible in the 3-column reader layout (found live).
- Drops the stale `media_ingest_screen.py` reference from `CLAUDE.md`.

## Testing

- New/updated tests in `Tests/Library/test_library_media_content.py` and `Tests/UI/test_library_media_reader_flow.py`; each fix has a failing-first regression test.
- Whole `test_library_media_reader_flow.py` green (37 passed); `test_library_media_reader_shell.py` + `test_library_media_side_by_side.py` green (66 passed).
- Live TUI smoke test (isolated scratch profile) confirmed all four fixes end-to-end, including the position-aware `]`/`[` footer (only `]` at the first item, both in the middle, only `[` at the last).

### Pre-existing dev-tip test debt (NOT introduced here; verified by stash-run)

- `test_library_screen_bindings_are_all_gated_or_universal` fails on `focus_previous_workbench_pane` (its gate returns `False` on a bare unmounted screen).
- `test_compact_media_stale_and_retry_actions_remain_truthful` is flaky (~2/5) — a gated-fake recompose race.

Both fail identically without this branch's changes.

---

## Update: task-28006 — Generate analysis in the Reader (the second P0)

Adds a **Generate / Regenerate** action to the Reader's Analysis tab so an analysis can be produced in the reading flow instead of detouring to the (retired) Media manager. The provider is resolved through the same `resolve_ingest_analysis_provider` seam the ingest path uses — so the promise here and the ingest receipt can never disagree — and a not-ready provider surfaces that seam's honest reason instead of dispatching. Ready flips the section into a "Generating analysis…" state and hands `chat_api_call` to a worker off the UI thread, then persists via the existing `save_analysis_version` path. Tests cover both the ready (dispatch + persist) and not-ready (warn + skip) paths.

This completes both P0s from the critique: sequential-review navigation (28002–28005) and in-flow analysis generation (28006).


rmusser01:
Thanks @qodo — addressed the review. Rebased on the latest `dev` first, then:

- **🐞 Rendered content unreachable (Strong)** — Fixed. `LibraryMediaContentBody` is now a `ScrollableContainer` (was a plain `Container`) so rendered-Markdown mode is keyboard-scrollable, and `#library-media-viewer-content` is a F6 candidate after the raw view. F6 now resolves the active content mode (raw → inner ScrollView, rendered → the scrolling body) before falling back to Find. New test: `test_library_media_rendered_content_is_keyboard_focusable_for_scroll`; the widget's own 20-test suite stays green (no mount-timing regression).
- **📘 compose() lacks docstring (Strong)** — Fixed. Added the Google-style docstring with a `Yields:` section.
- **🐞 Malformed backlog task (Strong)** — Fixed. Deleted the `task-task- - .md` artifact.
- **🐞 Traversal stops at page boundaries (Moderate)** — Tracked as a follow-up (**task-28229**), not fixed in this PR. `]`/`[` were deliberately scoped to within-page in task-28005, and the footer honestly drops the key at the edge. Seamless cross-page traversal needs async page-turn-then-select machinery (request the neighbour page, select its first/last row after the guarded result applies, preserve Reader mode + stale protection) — a real feature that deserves its own change + live verification rather than landing right before a merge.

PR #2309, opened 2026-09-02
feat(library): media reading/keyboard P1s (Enter-advances-match, bulk-select keys)

rmusser01:
## Summary

Follow-up to #2307 (the P0 / bug-track "unblock sequential review" PR) with the contained P1 reading-, triage-, and keyboard-ergonomics improvements from the same Library ▸ Media UX critique. Each change is test-first.

> **Stacked on #2307** — this branch is cut from `worktree-media-ux-fixes`, so review/merge #2307 first. GitHub will retarget this PR to `dev` once #2307 lands.

## Changes

- **task-28011 — Enter walks to the next content-search match.** Re-pressing Enter on the same in-item query now advances to the next match (find-bar convention, wrapping) instead of no-opping, and the Reader footer advertises `enter: next match` while a search with matches is active. The larger "search the Analysis tab text, not only the transcript" half is split to **task-28026**.
- **task-28012 — keyboard access to media bulk selection.** `s` enters/exits Select mode on the media list, and Space toggles the focused row's selection — the live-verified gaps (Space was inert, no key entered Select mode, nothing advertised). Enter already toggled a focused row, so only these were missing. The `s` key reuses the same seam as the Select/Done button, and the media list advertises the keys via new `LIBRARY_MEDIA_LIST_SHORTCUTS` / `LIBRARY_MEDIA_SELECT_SHORTCUTS` footer sets. A focused input still receives the printable keys. The Reader action-row accelerators (a separate concern) are split to **task-28027**.
- **task-28013 — browse sort chooser on the media list.** Media had a type filter but no way to change the sort order from the canvas. Adds a **Sort** button that opens a direct-pick strip (Newest / Oldest / Title A–Z / Title Z–A), re-fetching page one under the chosen order and persisting it in the browse scope. Mirrors the proven Prompts/Notes/type-filter choice-strip pattern (task-14902 requires the strip, not a per-press cycle) and reuses the shared choice-strip seams (footer `enter: choose sort` / `esc: cancel`, Escape-to-close, mutual exclusivity with the type chooser). The in-canvas filter half of the task already shipped on dev.

## Testing

- New unit + integration tests for every change; each failed first.
- Regression green across `test_library_media_reader_flow.py`, `test_library_media_side_by_side.py`, `test_library_multiselect_media.py`, `Tests/Library`, and the media subset of `test_library_shell.py`.
- Two existing side-by-side tests that pinned the media-list footer to `LIBRARY_LIST_SHORTCUTS` were updated to the new media-list footer sets (task-28012 deliberately enriches them); the toolbar-fit region test was extended to include the new Sort button.

### Pre-existing dev-tip flakes/failures (NOT introduced here; verified by stash-run)

- `test_library_media_reader_flow.py::test_filter_uses_authoritative_search_and_restores_page_three_anchor`, `test_library_media_side_by_side.py::test_compact_media_stale_and_retry_actions_remain_truthful`, and `test_screen_navigation.py::test_focus_library_list_entry_checked_row_preference_is_media_only` are order-dependent flakes / pre-existing failures — they pass in isolation or fail identically without this branch.
- `test_library_screen_bindings_are_all_gated_or_universal` fails on `focus_previous_workbench_pane` on a bare unmounted screen — pre-existing; the new bindings here are correctly gated.

### DIFF
--- a/Tests/UI/test_library_media_reader_flow.py
+++ b/Tests/UI/test_library_media_reader_flow.py
@@ -269,6 +269,37 @@ async def _load_row_0(screen, service, pilot):
 
 
 @pytest.mark.asyncio
+async def test_s_key_enters_select_mode_and_space_toggles_a_row():
+    """task-28012: keyboard-only bulk selection on the media list."""
+    app, service = _flow_app(count=3)
+    host = LibraryProductionCSSHarness(app)
+
+    async with host.run_test(size=WIDE_SIZE) as pilot:
+        screen = await _open_media_list(host, pilot)
+        assert screen._library_media_select_mode is False
+        assert ("s", "select") in screen._library_footer_shortcuts_for_current_state()
+
+        # "s" from a focused row enters select mode.
+        screen.query_one("#library-media-row-0", Button).focus()
+        await pilot.pause()
+        await pilot.press("s")
+        await pilot.pause()
+        assert screen._library_media_select_mode is True
+        footer = screen._library_footer_shortcuts_for_current_state()
+        assert ("space", "toggle selection") in footer
+        assert ("s", "done selecting") in footer
+
+        # Space toggles the focused row's selection.
+        row = screen.query_one("#library-media-row-0", Button)
+        row_id = str(row.media_id)
+        row.focus()
+        await pilot.pause()
+        await pilot.press("space")
+        await pilot.pause()
+        assert screen._library_media_row_selection.is_selected(row_id)
+
+
+@pytest.mark.asyncio
 async def test_bracket_keys_walk_to_next_and_previous_item_in_the_reader():
     """task-28005: ] opens the next browse item, [ the previous, from the Reader."""
     app, service = _flow_app(count=3)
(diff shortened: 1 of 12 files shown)


rmusser01:
Thanks @qodo — addressed both findings and rebased onto the latest `dev` (now that #2307 has merged):

- **🐞 Shortcuts bypass disabled selection (Moderate)** — Fixed. Entering Select mode via `s` now goes through `_library_media_select_enter_available` (a fresh page with at least one row — the same availability the Select button has), and Space row-toggle is gated off on a stale page or during a bulk delete. Exiting an already-active selection stays possible even if rows disappear. State-level predicate, not presentation labels.
- **📘 Selection key literals duplicated (Strong)** — Fixed. The `s`/`space` keys are named constants (`_MEDIA_SELECT_MODE_KEY` / `_MEDIA_ROW_SELECT_KEY`) reused by both the `Binding`s and the footer shortcut sets, so they cannot drift.

New unit test pins the availability predicate; existing select-mode integration + footer tests stay green.

PR #2319, opened 2026-09-02
feat(library): make the media Analysis tab searchable (task-28026)

rmusser01:
## What

The Library media viewer's **Analysis tab** was not searchable — the in-item find bar (added in the reader search work) only scanned the transcript. This makes the Analysis tab searchable too, closing task-28026.

## Why it mattered

The power-user review workflow (read an analysis, jump to the phrase you half-remember, cross-check the transcript) broke the moment you switched to the Analysis tab: the search box either vanished or silently searched the wrong text. For a conference-worth of generated summaries this is exactly where find-in-item earns its keep.

## Changes

- **`Widgets/Library/library_media_viewer.py`** — when an analysis exists (and we're not editing/generating), `_compose_analysis` renders it in the same `LibraryMediaContentSearchControls` + `LibraryMediaContentBody` (raw mode) the Read tab uses, instead of a plain non-searchable `Static`. The empty/generating states keep the old `#library-media-viewer-analysis-text` Static.
- **`UI/Screens/library_screen.py`** — the search corpus (`_library_media_content_matches`) is now **mode-aware**: analysis text in the Analysis tab, transcript elsewhere. Reader mode is part of the match memo key so a tab switch can't serve the other tab's matches. `handle_library_media_reader_mode` clears the query/index/memo on a real tab switch — no stale highlights. Match count, Prev/Next and Enter-advance all follow the active tab through the shared corpus.
- **`Tests/UI/test_library_shell.py`** — new: `test_library_media_analysis_tab_is_searchable` (status reads "Match 1 of 2"), `test_library_media_switching_tabs_clears_the_search`; updated 3 existing analysis-text tests to read the analysis from the content body's `.content`.

## Tests

`Tests/UI/test_library_shell.py` (analysis/content/tab-switch), `test_library_media_reader_flow.py`, `Tests/Library/test_library_media_content.py`, `test_library_media_reader_shell.py`, `test_media_viewer_content_search_debounce.py` all green. Derived checks (diagnostic inventory, CSS bundle, backlog IDs) clean.

---
## Summary by cubic
Makes the Library media viewer's Analysis tab searchable, closing task-28026. The Analysis tab previously rendered analysis text as a plain, non-searchable view; it now uses the same search controls and content body as the Read tab, so find-in-item, match count, Prev/Next, and Enter-advance all work on the analysis text.

**New Features**
- The search corpus is mode-aware: analysis text on the Analysis tab, transcript everywhere else.
- Switching tabs clears the query, match index, and memo so highlights never carry over between tabs.

**Bug Fixes**
- Find clears an active analysis search when it moves Analysis→Read, so a stale match index can't scan the transcript.
- Progress capture is skipped outside Read mode so Analysis scrolling can't overwrite transcript reading progress.

<sup>Written for commit 3029c5191ec9a0f4c43370ba1b6e5bbe8d86576b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2319?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_library_media_reader_flow.py
+++ b/Tests/UI/test_library_media_reader_flow.py
@@ -1368,6 +1368,109 @@ async def test_offset_burst_settles_only_the_newest_value_per_item():
     assert fake._library_media_progress_persisted_offsets["local:media:3"] == (0, 17)
 
 
+@pytest.mark.asyncio
+async def test_capture_outside_read_never_persists_the_analysis_body_offset():
+    """task-28026: the Analysis body reuses ``#library-media-viewer-content``,
+    so capturing progress while a non-Read tab is active must NOT record that
+    scroll offset as the transcript's reading progress."""
+    calls = []
+
+    async def update_reading_progress(**kwargs):
+        calls.append(kwargs)
+
+    async def run_service_call(call, *args, **kwargs):
+        kwargs.pop("isolate_in_worker", None)
+        return await call(*args, **kwargs)
+
+    workers: list = []
+    fake = _progress_capture_fake(
+        workers=workers,
+        update_reading_progress=update_reading_progress,
+        run_service_call=run_service_call,
+        scroll_y=42,
+    )
+    fake._library_media_reader_session = set_mode(
+        fake._library_media_reader_session, "analysis"
+    )
+
+    LibraryScreen._capture_library_media_loaded_progress(fake)
+
+    assert workers == []
+    assert calls == []
+    assert fake._library_media_read_scroll_by_id == {}
+
+
+def test_find_from_analysis_clears_the_search_before_reading_the_transcript():
+    """task-28026: the always-visible Find action switches Analysis->Read, so
+    it must drop the analysis query/index/memo the same way the mode buttons
+    do -- otherwise a stale, possibly out-of-range match index scans the
+    transcript."""
+    session = set_mode(
+        LibraryMediaReaderSessionState(
+            selected_id="local:media:1",
+            selected_backing_id=1,
+            selected_title="A",
+            loaded_id="local:media:1",
+            loaded_backing_id=1,
+            loaded_title="A",
+        ),
+        "analysis",
+    )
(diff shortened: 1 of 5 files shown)


rmusser01:
Thanks — addressed all four in `3029c5191`.

**1. Analysis overwrites reading progress (High) — fixed.** You're right: the Analysis body reuses `#library-media-viewer-content`, so `_capture_library_media_loaded_progress` was recording the analysis scroll offset as transcript reading progress. Root-cause fix: capture now returns early unless the reader is in Read mode (`session.mode != "read"` joins the existing `external_detail`/non-local guards) — only the Read tab's content body carries transcript progress. Pinned by `test_capture_outside_read_never_persists_the_analysis_body_offset`.

**3. Find bypasses search reset (Bug) — fixed.** `handle_library_media_reader_find` did switch Analysis→Read via `set_mode` without clearing search state. Both it and the mode buttons now route through a shared `_reset_library_media_search_on_mode_change(new_mode)` helper (applied before `set_mode`, so it compares against the old mode; a no-op re-press leaves search intact). Pinned by `test_find_from_analysis_clears_the_search_before_reading_the_transcript`.

**2. Reader handler omits Args (Rule 497152) — fixed.** Added Google-style `Args:` sections to both modified public handlers (`handle_library_media_reader_mode`, `handle_library_media_reader_find`).

**4. Mode logic lacks unit tests (Rule 497172) — added.** `test_media_content_matches_scopes_corpus_to_the_active_reader_mode` isolates the mode-aware corpus selection and memo-key mode-miss directly on `_library_media_content_matches`, alongside the two direct unit tests above — separate from the end-to-end shell tests.

Issue #13, opened 2025-04-27
Enhancement: Option for custom UA for downloads of media


rmusser01:
Title.

Allow the user to set a custom UA string in config.toml to be used when performing downloads of remote media.

Issue #12, opened 2025-04-27
Enhancement: Title bar/Status bar showing current convo tokens / Max Convo Tokens


rmusser01:
Title.

Add a display to titlebar/status bar to indicate how many tokens the current convo is using, out of the max tokens allowed for a convo.

Issue #11, opened 2025-04-26
Enhancement: Add a customizable splash screen.


rmusser01:
Title.

Issue #10, opened 2025-04-26
Enhancement: 'Save Chat' toggle/button


rmusser01:
toggle/button to save current chat/update save state.

Toggle On -> Auto-Save chat every message, every message that is sent/received/edited, chat in DB is updated appropriately.
Toggle Off -> Chats are not saved into the DB automatically. Must either toggle on, which point chat is added/tracked, or 'save chat' button, which will do the same, but only for that snapshot. 

Clicking the 'save chat' button after having pressed it once for the current chat will update the saved chat in the DB with the current chat displayed.

PR #2323, opened 2026-09-03
feat(library): review sets — approved design + phase 1 persistence (task-28024, task-28240)

rmusser01:
## What

Foundation for **Library review sets** — an ordered, pinned snapshot of media items you walk with Next/Prev, with a cursor, per-item done marks, and a completion state, so reviewing a set (a conference's talks, a tag, a hand-picked selection) is a first-class resumable object rather than a memory exercise.

This PR is **the approved design + Phase 1 (persistence + pure model)**. No UI wiring yet — that's Phase 2 (task-28241).

## Design (approved)

`backlog/docs/design-library-review-sets.md` (task-28024) is the full spec — data model, lifecycle, entry points, viewer behavior, edge cases, and how it composes with the shipped Prev/Next (28005), read-later (28027), and the not-yet-existing read markers (28009). It includes a **§11 review-pass log** of 7 issues a critical re-read caught and fixed (cross-DB no-FK tombstones, "Review selected" ordering, whole-result cap, active-set DB pointer, cursor/progress under tombstones, mark-done edges, all-tombstoned = empty). The 5-phase implementation plan is filed as tasks 28240–28244.

## Phase 1 (task-28240)

- **Schema v4 in `Library_Collections_DB`** — `review_sets` + `review_set_items` (idempotent DDL, version 3→4). **No `CREATE INDEX`**: the "one active set" invariant is enforced transactionally in the service, which deliberately keeps this out of the index-plan-pin census. (This DB isn't in `sql_validation.VALID_TABLES` either — that's chachanotes-only — so the new tables need no allowlist entry.)
- **`Library/review_set_state.py`** — the pure, DB-free navigation model. A set is pinned ordered items + an absolute cursor; progress and completion are computed over **live** items via an injected `is_live` predicate, so a deleted item becomes a tombstone the walker skips and positions never renumber.
- **`Library/review_set_service.py`** — `create` (dedupe + pin + activate) / `get` / `list` / `get_active` / `advance` (persist cursor) / `set_cursor` / `mark_item_done` / `refresh_completion` / `activate` / `reopen` / `dismiss`. Tombstone detection is a runtime `is_live` resolve — the set and Media tables live in **separate SQLite files**, so no foreign key is possible; `title_snapshot` is load-bearing once an item is gone.

## Tests

- `Tests/Library/test_review_set_state.py` (14) — cursor/progress/tombstone/completion, written test-first.
- `Tests/Library/test_review_set_service.py` (11) — create/activate/advance/mark-done/completion/dismiss/reopen against a real file-backed DB.
- Existing collections migration/service tests updated for schema v4 and re-run green.
- **Preflight (the full CI mirror) passes.** One pre-existing, unrelated collections test (`..._round_trip_public_ids`) fails identically on `dev` — confirmed by reverting this branch's DB file; not introduced here.

---
## Summary by cubic
Implements the approved Library review-set design (task-28024) and Phase 1 persistence/model (task-28240). Review sets can now persist an ordered snapshot of local media, cursor, per-item done marks, and lifecycle state across restarts; Reader and UI behavior remain unchanged until later phases.

**Details**

- Migrates `Library_Collections_DB` to schema v4 with `review_sets` and `review_set_items`, plus a partial unique index enforcing one active set.
- Adds transactional create, activate, advance, mark-done, and refresh-completion paths so reads and writes share one snapshot.
- Adds validation: unknown origins, blank names, and empty item sets are rejected; activating a missing or dismissed id is a no-op.
- Adds tombstone-aware cursor navigation, live-item progress, and completion handling, with title snapshots preserved across the separate SQLite databases.
- Adds model and service tests, and updates existing schema migration tests.

<sup>Written for commit 626647d86d527f08e7609dfb5a146f889b05789b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2323?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/DB/test_library_collections_capture_migration.py
+++ b/Tests/DB/test_library_collections_capture_migration.py
@@ -140,7 +140,7 @@ def test_fresh_database_creates_v1_compatibility_and_capture_v3(tmp_path: Path) 
 
     database = LibraryCollectionsDB(path)
 
-    assert database.get_schema_version() == 3
+    assert database.get_schema_version() == 4
     assert database.has_compatible_legacy_schema() is True
     database.require_capture_schema()
     objects = _schema_objects(path)
@@ -158,7 +158,7 @@ def test_real_v1_fixture_migrates_without_changing_legacy_values(tmp_path: Path)
 
     database = LibraryCollectionsDB(path)
 
-    assert database.get_schema_version() == 3
+    assert database.get_schema_version() == 4
     assert _legacy_rows(path) == before
     assert database.has_compatible_legacy_schema() is True
     database.require_capture_schema()
@@ -220,7 +220,7 @@ def test_two_concurrent_openers_publish_one_complete_v3_schema(tmp_path: Path) -
             for row in connection.execute(
                 "SELECT version FROM schema_version ORDER BY version"
             )
-        ] == [1, 3]
+        ] == [1, 4]
         assert connection.execute("PRAGMA foreign_key_check").fetchall() == []
     assert CAPTURE_TABLES <= {
         name for kind, name in _schema_objects(path) if kind == "table"
@@ -235,7 +235,7 @@ def test_v3_reopen_is_idempotent(tmp_path: Path) -> None:
 
     second = LibraryCollectionsDB(path)
 
-    assert second.get_schema_version() == 3
+    assert second.get_schema_version() == 4
     assert _schema_objects(path) == before
     second.close()
 
@@ -244,7 +244,7 @@ def test_future_schema_is_refused_without_writing(tmp_path: Path) -> None:
     path = tmp_path / "collections.db"
     _create_v1_fixture(path)
     with _open(path) as connection:
-        connection.execute("INSERT INTO schema_version (version) VALUES (4)")
+        connection.execute("INSERT INTO schema_version (version) VALUES (5)")
     before_objects = _schema_objects(path)
     before_rows = _legacy_rows(path)
 
@@ -260,7 +260,7 @@ def test_future_schema_is_refused_without_writing(tmp_path: Path) -> None:
             for row in connection.execute(
                 "SELECT version FROM schema_version ORDER BY version"
             )
-        ] == [1, 4]
+        ] == [1, 5]
 
 
 def test_v2_processing_row_migrates_with_unowned_expired_lease(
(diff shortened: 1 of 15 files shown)


rmusser01:
Thanks — thorough review. Addressed all but one (declined with reasoning), in `1671a7383`.

**Bugs**
- **#1 Empty set becomes active** — `create_review_set` now rejects empty membership (raises `ValueError`) *before* deactivating anything, so an empty set is never created or activated.
- **#2 Invalid activation clears active set** — `activate()` now checks the target exists (non-dismissed) inside the transaction and no-ops otherwise, leaving the current active set untouched.
- **#3 / #13 Completion refresh / advance races** — `refresh_completion()` and `advance()` now read+compute+write in ONE transaction (via `_read_review_set(conn, …)`), so nothing can commit between the read and the write.
- **#11 Active invariant lacks enforcement** — added the partial `UNIQUE INDEX review_sets_one_active ON review_sets(active) WHERE active = 1 AND deleted_at IS NULL` (the AC's schema-level enforcer, beneath the transactional guard). It also exactly matches `get_active_review_set`'s WHERE clause; census row added.
- **#12 Liveness recheck can crash** — the pure model now takes the live snapshot once and resolves/advances/measures within it (`_resolve_within`), so a liveness value that flips mid-call can't desync `live_positions.index(current)`.

**Rule violations**
- **#5 Reads bypass transaction** — `get_active_review_set` / `list_review_sets` / `get_review_set` now load within a single `read_transaction` snapshot.
- **#6 `list_review_sets` unbounded** — now takes a `limit` (default 200, clamped to [1, 1000]).
- **#7 Inputs unvalidated** — `create_review_set` validates the `origin` allow-list and a non-empty name (plus #1's non-empty membership).
- **#8 Disk SQLite tests** — DB tests now use `:memory:`.
- **#9 / #10 Docstrings** — Google-style `Args:`/`Returns:` added to the public service methods and `is_empty`/`is_complete`.

**Declined**
- **#4 v4 migration inline** — kept inline deliberately. `migrations/*.sql` is the pattern for **ALTER** migrations (the v2→v3 column adds); fresh `CREATE TABLE … IF NOT EXISTS` DDL is inline, exactly matching the existing `_CAPTURE_SCHEMA_DDL` block applied idempotently in `_initialize_schema`. A new-tables-as-.sql split would diverge from that precedent.

PR #2329, opened 2026-09-03
fix(splash): frame-locked playback and 7s default (TASK-28026)

rmusser01:
## What

The startup splash intro was unstable: sometimes smooth, sometimes jumpy, sometimes it skipped from an early frame straight to the end, and sometimes it never visibly played at all. This makes playback frame-locked and bumps the default duration to 7 seconds.

## Why (measured, not guessed)

A headless probe (real app via `run_test` plus a controllable `time.sleep` blocker on the event loop) showed the mechanism:

- **Textual's `set_interval` timer permanently skips ticks** that cannot run while the event loop is blocked (`Timer` defaults to `skip=True`), and every splash effect derives its progression from `time.time() - effect.start_time`. A 1.2s stall right after startup made the *first visible frame* land at 1.22s into a 2.5s reveal; a stall longer than the splash duration let the auto-close timer beat every frame ("didn't play at all").
- The first frame only rendered at the first timer tick, so the display started blank and the close-timer race above had a zero-frame outcome.
- Several cards ship reveal timelines (2.5–3s) longer than the old 1.5s default, so their reveals were always cut mid-way.

## The fix (all central, in `Widgets/splash_screen.py`)

- `_render_animation_frame()` is the single render path: it re-anchors the effect clock so **one rendered frame advances the animation by exactly one interval** — contention now *slows* the intro instead of skipping it ahead. Probe re-run with the fix: the same 1.2s block resumes at virtual elapsed ~0.18s on a smooth cadence; a 4.5s block still shows frame 0 immediately.
- **Frame 0 renders synchronously** at animation start (no blank window; close-timer can no longer win a race against zero frames).
- Card reveal durations longer than the splash lifetime are **clamped to fit** so reveals complete.
- **Identical consecutive frames skip the display rewrite** (`Static.update` parses markup synchronously, so this is real CPU).
- **Default duration 7.0s** in the widget, the app fallback, the Settings viewer, and the config template. Existing configs with an explicit `[splash_screen] duration` keep their own value.

## Reconciliation with TASK-21591 (skip-on-keypress on dev)

Dev already fixed skip-on-keypress via a focus-based mechanism with a consume-the-key contract (F9-mid-splash protection, task-1339). This PR keeps that mechanism and contract untouched; `on_key` now routes through the new public `request_skip()` seam, and the redundant app-level forwarding from the original branch commit was dropped. A full-`TldwCli` end-to-end test is added alongside the five mechanism tests.

## Follow-up

Five cards are broken independently of the driver and are pinned in the new smoke test for **TASK-28028**: `cyberpunk_glitch` / `hypno_swirl` / `phonebooths` emit markup Textual's own `Content.from_markup` rejects (cyberpunk only on some random draws — another "sometimes fine, sometimes not" source), `world_map` raises `AttributeError` every frame, and `typewriter_news` returns no frames at all. Roughly a 1-in-15 chance per launch of a visibly dead intro.

Two lessons recorded in `backlog/docs/lessons-testing-evidence.md` (the decorator-steal incident and the `set_interval` tick-skipping trap).

## Verification

- New `Tests/Widgets/test_splash_animation_playback.py` (6 tests incl. all-76-animated-cards smoke) + extended `Tests/UI/test_splash_skip_on_keypress.py` (dev's 5 mechanism tests + full-app E2E), config-read and settings-viewer suites: **25 passed** on this branch (repeated runs, no flakes).
- Dev's navigation-during-splash guard (`test_screen_navigation.py -k splash`): **passed**.
- `ruff` clean on all changed files (the 133 pre-existing `app.py` findings exist on pristine `dev` too; this PR's `app.py` delta is one line).
- Contention behavior verified with the headless blocker probe before/after (numbers above).

---
## Summary by cubic
Addresses TASK-28026 by stabilizing startup splash playback: instead of using wall-clock time that could skip frames or show nothing under load, playback now advances one interval per rendered frame and renders its first frame synchronously. The default duration increases from the previous 1.5–2.5 second defaults to 7 seconds; explicitly configured durations remain unchanged.

**Playback and configuration**
- Clamps longer card reveals to the splash lifetime and avoids rewriting the display for identical frames.
- Keeps delta-clocked effects such as `terminal_boot` at a constant per-frame interval.
- Uses one shared duration constant across the widget, app fallback, Settings viewer, and config template.
- Routes startup keypresses through `request_skip()` while preserving existing key-consumption and app-binding behavior.

**Regression coverage**
- Adds tests for frame pacing, first-frame rendering, duration clamping, delta-clocked effects, display updates, and full-app skip behavior.
- Smoke-tests animated cards and records five pre-existing broken cards under follow-up TASK-31001.
- Disables the splash in the CSS fastpath startup test so its polling window is not consumed by the new 7-second default.

<sup>Written for commit 7f811f5202b76c00588c27d14c29826f1d7fccaf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2329?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
All three Qodo findings addressed in c7aa6a3f2 (branch rebased onto latest dev):

1. **Magic 7.0 literal** → single `DEFAULT_SPLASH_DURATION_SECONDS` in `Constants.py`, used by the widget constructor, loaded-config default, app fallback, Settings viewer, and the config template (placeholder + the same single-substitution pattern the transcription provider already uses, task-867 style).
2. **`layout=False` dropped** → restored in `_render_animation_frame`; TASK-21595's `test_splash_animation_tick_arms_no_layout_pass` is green again for matrix/glitch/starfield (it was red before this push).
3. **TerminalBootEffect delta clock** → `BaseEffect.FRAME_DELTA_CLOCK` distinguishes effects that treat `start_time` as a previous-frame timestamp; the driver now gives them a constant one-interval delta per rendered frame. The new regression test spies the delta the effect actually experiences and fails without the fix (observed compounding 0.06s→0.27s over 8 frames at a 0.03s interval).

Verification on the rebased branch: 33 tests green (playback suite incl. new delta-clock test, layout-cost suite, skip suite, config-read, settings viewer), ruff clean.

PR #2333, opened 2026-09-03
feat(library): review sets phase 2 - Reader walker (task-28241)

rmusser01:
## What

Phase 2 of Library review sets (design: `backlog/docs/design-library-review-sets.md`). An **active review set now drives the media Reader** — building on the Phase 1 persistence/model that just landed.

Closes **task-28241** (AC1–3). AC4 (auto-load the cursor item on entry) is split to **task-28245** — the set *state* already resumes (the active flag + cursor persist, so `]`/`[` walk from the saved cursor on relaunch); only the entry auto-load, which is startup-timing sensitive, is deferred for careful live verification.

## Behavior

- **`]` / `[` walk the whole pinned set** (page-independent) via the same `_select_library_media_reader_row` actuator the browse traversal uses, so fenced loading, scroll capture/restore, and Reader-mode persistence keep working. With **no active set**, task-28005's browse-row traversal is unchanged — the set *supersedes*, it doesn't replace.
- **Auto-mark on advance:** stepping forward marks the item you leave done (a forward press on the last item is the completion gesture); **Prev never marks**; **`m`** explicitly toggles the current item's mark (un-mark a skimmed item, or mark the last one).
- **Reader footer** shows progress — `12 of 40 · 7 reviewed` / `All N reviewed` — relabels `]`/`[` as "in set", and adds **`R` to exit review** (deactivates the set but keeps it resumable), distinct from **Escape** (leaves the Reader, set stays active).
- **Liveness** is resolved in **one batched Media-DB query** per keypress (`id IN (…) AND deleted = 0 AND is_trash = 0`); tombstoned items are skipped so the count never inflates.

## Design

- Pure `plan_walk` (in `review_set_state.py`) decides each step on a **single live snapshot** — the resolve, the step, and the mark can't disagree if liveness flips mid-call.
- New `R`/`m` bindings are gated in `check_action` (the binding-audit passes; the sole failure is the pre-existing `focus_previous_workbench_pane` baseline).
- Added `ReviewSetService.deactivate_active()` (Exit-review backend — deactivate without deleting).

## Tests

- `Tests/UI/test_review_set_walker.py` (12) — forward/back walk, tombstone skip, last-item completion, no-set fall-through, progress line, exit-review, toggle-reviewed — via a fake screen over the **real** service.
- `Tests/Library/test_review_set_state.py` — `plan_walk` + `format_review_progress` (single-snapshot, mark semantics).
- Preflight (diagnostic inventory, index pins, CSS bundle, backlog IDs) green; the one added `logger.debug` is in the regenerated inventory.

---
## Summary by cubic
Makes the media Reader walk an active review set instead of the browse rows (task-28241 AC1–3). `]`/`[` now step through the whole pinned set, auto-marking the item you leave done on forward advance, and the footer shows progress plus an exit-review key. AC4 (auto-loading the cursor item on entry) is split to task-28245; the set state already resumes at its saved cursor.

**Behavior**

- `]`/`[` walk the whole active set page-independently via the existing reader-row actuator; with no set active, browse-row traversal is unchanged.
- Steps advance from the item the Reader is actually showing, so `]` can't mark an unseen item; when the Reader is off the set, the first step resumes at the cursor and marks nothing.
- Forward advance marks the item you leave done (last item = completion gesture); Prev never marks; `m` toggles the current item's mark.
- Footer shows `X of M · N reviewed` / `All N reviewed`, relabels `]`/`[` as "in set", and adds `R` to exit review (deactivates but keeps the set resumable); Escape leaves the Reader with the set active.
- Liveness is resolved in batched Media-DB queries (≤900 ids each); tombstoned items are skipped.

**Testing**

- Adds UI walker tests using a fake screen over the real service, plus `plan_walk` and progress formatting unit tests, and a real Media-DB test for liveness filtering.
- New `R`/`m` bindings are gated in `check_action`; the binding audit passes except the pre-existing `focus_previous_workbench_pane` baseline.

<sup>Written for commit b5042db3a96558e74addee447e0284895fd1dedf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2333?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Library/test_review_set_service.py
+++ b/Tests/Library/test_review_set_service.py
@@ -177,6 +177,18 @@ def test_dismiss_soft_deletes_and_deactivates():
     assert svc.get_review_set(set_id) is None  # dismissed sets are hidden
 
 
+def test_deactivate_active_keeps_the_set_resumable():
+    svc = _service()
+    set_id = svc.create_review_set("X", origin="browse", items=[(1, "a")])
+
+    svc.deactivate_active()
+    assert svc.get_active_review_set() is None
+    # not deleted -- still listable and re-activatable at its cursor.
+    assert [rs.set_id for rs in svc.list_review_sets()] == [set_id]
+    svc.activate(set_id)
+    assert svc.get_active_review_set().set_id == set_id
+
+
 def test_reopen_clears_completion():
     svc = _service()
     set_id = svc.create_review_set("X", origin="browse", items=[(1, "a")])
--- a/Tests/Library/test_review_set_state.py
+++ b/Tests/Library/test_review_set_state.py
@@ -10,10 +10,13 @@ never renumbers.
 from __future__ import annotations
 
 from tldw_chatbook.Library.review_set_state import (
+    ReviewProgress,
     ReviewSetItem,
     advance_cursor,
+    format_review_progress,
     is_complete,
     is_empty,
+    plan_walk,
     resolve_cursor,
     review_progress,
 )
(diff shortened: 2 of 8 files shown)


rmusser01:
Thanks — addressed in `7ec4ffdca`.

**#1 Unseen cursor item marked (Bug) — fixed.** The walker no longer trusts the persisted cursor blindly. It aligns the walk origin to the item the Reader is actually **showing** (`loaded_backing_id`): a forward step marks *that* item. When the Reader is **not** on a set item (fresh entry, or a browse item), the first `]` simply **resumes** the set at its cursor and marks nothing — so no unseen item is ever marked reviewed. Pinned by `test_walk_resumes_at_cursor_without_marking_when_reader_is_off_set`.

**#5 Unbounded liveness query — fixed.** `_review_set_live_ids` now chunks the ids into `≤900`-per-`IN(…)` batches (safely under SQLite's 999-variable limit), so the query stays bounded even for a large set.

**#2 Unclosed cursor — fixed.** Each batch's cursor is closed deterministically via `contextlib.closing`.

**#3 Liveness SQL path untested — added.** `test_review_set_live_ids_filters_deleted_and_trashed` exercises the real query against an in-memory `MediaDatabase` (live vs `soft_delete_media` vs `mark_as_trash` vs an unknown id), plus the media-db-absent fallback.

**#4 Bypasses transactions — declined (with reasoning).** `execute_query()` **is** `Client_Media_DB_v2`'s standard single-statement API; its `transaction()` opens a `BEGIN IMMEDIATE` **write** lock, which is wrong for a lone read. Wrapping this SELECT in it would be a misuse, not the standard.

PR #2335, opened 2026-09-03
feat(library): review sets phase 3 - entry points (task-28242)

rmusser01:
## What

Phase 3 of Library review sets — **the entry points that let a user actually create a set**. Until now the walker (Phase 2) was dormant because nothing could create a set; this closes that. Closes **task-28242**.

Two ways to create one, both landing straight in the Reader walking the new set:

- **"Review these"** on the media list toolbar (between Trash and Select) pins the **whole filtered result** — it pages through `search_media` in a worker, dedupes by id, and caps at 500 (with a "first 500" warning on overflow). Covers the filter-query surface too (a filtered list *is* a browse result).
- **"Review selected"** — a third **Select-mode bulk action** between "Export selected" and the far-end danger "Delete selected" — pins the selection ordered by a **deterministic sort-order query** (`search_media(id_allowlist=…)`), not the unordered `RowSelection` (which is a set and can span pages).

## Verified end-to-end (live)

Seeded 4 media items and drove the real app in tmux:
1. Clicked **"Review these"** → the set was created and the Reader **landed on the first item**; the footer showed **`1 of 4 · 0 reviewed`**.
2. Pressed **`]`** → advanced to item 2, the item left behind was **auto-marked done**, and the footer updated to **`2 of 4 · 1 reviewed`**.

(Seeded data cleaned up afterward; the user's DB is untouched.)

## Design

- Pure `build_pinned_items(pairs, cap=500)` (in `review_set_state.py`) — dedupe-by-id + cap + truncation flag.
- Shared `_create_and_open_review_set` — builds the pinned list, creates + activates the set, lands on its first item (empty → notice, truncation → warning).
- The paging/ordering calls mirror the browse controller's existing `search_media` usage exactly; no new logger (so this branch doesn't touch the diagnostic inventory).

## Tests

`Tests/UI/test_review_set_walker.py` + `Tests/Library/test_review_set_state.py` — `build_pinned_items` (dedupe/cap/truncation/coercion), `_create_and_open_review_set` (create+activate+land, empty, truncation), and the name derivation. 59 review-set tests green; derived-artifact checks (inventory, index pins, CSS, backlog) all green.

---
## Summary by cubic
Closes task-28242: adds the two entry points that let users create review sets, so the Phase 2 walker is finally reachable from the UI. "Review these" pins the whole filtered list; "Review selected" pins the current selection — both create the set and land in the Reader walking it.

**New Features**
- `build_pinned_items` dedupes by backing id, caps at 500, and returns a truncation flag.
- "Review these" pages the whole filtered result via `search_media` in a worker and warns when the cap drops items.
- "Review selected" orders the selection with a deterministic sort query (`id_allowlist`), not the unordered `RowSelection`.
- Empty or unresolved selections notify the user and create no set.

**Bug Fixes**
- "Review selected" now enables in place when rows are toggled, like Export/Delete.
- Worker failures now surface an error notice instead of a silent no-op.

<sup>Written for commit 71e87101f85ca9ddd16187bbd9dc537b84d0526c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2335?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/Library/test_review_set_state.py
+++ b/Tests/Library/test_review_set_state.py
@@ -13,6 +13,7 @@ from tldw_chatbook.Library.review_set_state import (
     ReviewProgress,
     ReviewSetItem,
     advance_cursor,
+    build_pinned_items,
     format_review_progress,
     is_complete,
     is_empty,
@@ -201,3 +202,33 @@ def test_format_progress_empty_set():
         format_review_progress(ReviewProgress(index=0, total=0, reviewed=0))
         == "No items to review"
     )
+
+
+# --- building the pinned item list (entry points, task-28242) ----------------
+
+
+def test_build_pinned_items_dedupes_by_id_keeping_first():
+    items, truncated = build_pinned_items([(10, "A"), (11, "B"), (10, "A-again")])
+    assert items == [(10, "A"), (11, "B")]
+    assert truncated is False
+
+
+def test_build_pinned_items_caps_and_flags_truncation():
+    items, truncated = build_pinned_items(
+        [(index, f"t{index}") for index in range(600)], cap=500
+    )
+    assert len(items) == 500
+    assert items[0] == (0, "t0") and items[-1] == (499, "t499")
+    assert truncated is True
+
+
+def test_build_pinned_items_dups_past_cap_do_not_flag_truncation():
+    pairs = [(index, f"t{index}") for index in range(500)] + [(0, "dup")] * 50
+    items, truncated = build_pinned_items(pairs, cap=500)
+    assert len(items) == 500
+    assert truncated is False  # only duplicates followed the cap, nothing dropped
+
+
+def test_build_pinned_items_coerces_types():
+    items, _ = build_pinned_items([("10", 42)])
+    assert items == [(10, "42")]
--- a/Tests/UI/test_review_set_walker.py
+++ b/Tests/UI/test_review_set_walker.py
@@ -9,7 +9,7 @@ and the per-item actuator is recorded rather than mounting a Reader.
 from __future__ import annotations
 
 import itertools
-from types import SimpleNamespace
+from types import MethodType, SimpleNamespace
 
 from tldw_chatbook.DB.Library_Collections_DB import LibraryCollectionsDB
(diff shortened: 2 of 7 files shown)


rmusser01:
Thanks — addressed all five in `73782000b`.

**#1 Review action stays disabled (Bug) — fixed.** `_apply_library_row_toggle` now flips **"Review selected"** in place alongside Export/Delete, so it enables the moment a row is toggled (no more waiting for a full recompose). Shared `LIBRARY_REVIEW_SELECTED_{,_DISABLED}_TOOLTIP` constants so the compose-time and patch-time tooltips can't drift.

**#5 Worker failures stay silent (Bug) — fixed.** Both `_review_these_worker` and `_review_selected_worker` now wrap their bodies in `try/except` and surface an error notice ("Couldn't build the review set.") on any search/DB failure, instead of a silent no-op under `exit_on_error=False`. Pinned by `test_review_these_worker_notifies_on_failure`.

**#2 Selected query lacks fixed cap — fixed.** `_order_selected_review_pairs` now bounds the `id_allowlist` (and `limit`) to `REVIEW_SET_CAP + 1` rather than the whole selection.

**#3 Review cap duplicated — fixed.** One `REVIEW_SET_CAP` in `review_set_state.py`, used by the paging stop condition, `build_pinned_items`' default, the truncation notice, and the selected-query bound.

**#4 Entry points lack integration tests — added.** `test_review_these_worker_pages_the_whole_result_and_lands` (multi-page paging → create → land), `test_review_selected_worker_orders_via_the_id_allowlist` (deterministic ordering + bounded allowlist), and the failure-notify test above — exercising the workers, not just the pure helpers. (The full click-through was also live-verified end-to-end in tmux.)

PR #2337, opened 2026-09-03
feat(library): review sets phase 4 - set picker (task-28243)

rmusser01:
Phase 4 of the review-sets program (design: `backlog/docs/design-library-review-sets.md`, task-28243; phases 1-3 = #2323 / #2333 / #2335): a picker to **resume, switch between, or dismiss** saved review sets, opened from the media list.

## What's here

- **`build_picker_rows`** (`Library/review_set_state.py`): pure `(set_id, name, progress_label, active)` rows — progress via the existing `format_review_progress` over one live snapshot.
- **`LibraryReviewSetPickerDialog`** (`Widgets/Library/library_review_set_picker.py`): a dumb modal on the `SafeModalDismissMixin` idiom. Renders pre-computed rows (active set carries the `✓` marker; names are user-derived and render as literal `Text`, never markup) and dismisses with one typed decision: `("open", id)`, `("dismiss", id)`, or `None`.
- **Screen worker** (`_review_set_picker_worker`): collects rows off the event loop with ONE chunked liveness resolve across every listed set, awaits the modal, then applies the decision off-loop:
  - **open** → reopens a completed set (AC#2), activates it (service's one-active invariant deactivates the previous), and lands the Reader at the set's resolved cursor. An all-tombstoned set is **never** activated (design §7) — notice instead.
  - **dismiss** → soft-delete + confirmation notice + a Reader-chrome sync (without it the footer kept advertising "] next in set · 1 of 3" after a live dismissal).
- **"Sets" opener on the media list TITLE row** — deliberately NOT the action toolbar: that toolbar already overflows the narrow Items pane (task-28025), and adding one more button pushed a squeezed Button into rich's zero-width `chop_cells` `ValueError`, crashing the app on Media open (live-verified, then relocated). The title row carries ~9 chars in a min-width-40 pane. Not composed on the fresh-empty page, which pins exactly one recovery action.
- Docs: `Docs/User_Guide/library/media-and-conversations.md` gains a **Review sets** section covering the whole feature (entry points, walk keys, picker) + keyboard entries; plus a lessons entry (tmux SGR click columns are cells; BSD awk/cut count bytes).

## Tests

15 new: `build_picker_rows` (3), dialog contract incl. markup-literal labels and Escape-cancel (6), worker end-to-end over a real tmp service DB (6: open/switch, reopen-completed, dismiss+sync, cancel, all-tombstoned, failure-notice). 141 green across the review-set battery + multiselect toolbar suite. No new logger (diagnostic inventory untouched); no schema change; preflight clean.

Two failures that predate this branch (reproduced on dev without these commits): `test_generated_stylesheet_includes_library_media_rules` (dev's e435052ad left `#library-media-title` out of the bundle) and `test_library_note_media_route_switch_updates_contextual_chrome`.

## Live verification (tmux, seeded media, cleaned up after)

Create via "Review these" → walk `]` ("2 of 3 · 1 reviewed") → `R` exit → **Sets** lists "All media — 2 of 3 · 1 reviewed" → picking it resumes at the cursor item with set chrome restored → **Dismiss** soft-deletes and the footer drops the set keys immediately. Empty picker state verified before any set existed.

---
## Summary by cubic
Adds a set picker to the media list so saved review sets can be resumed, switched between, or dismissed. Previously sets could only be walked while active; now a "Sets" button on the media list title row opens a picker showing each set's live progress, with the active set marked.

**Behavior**
- Picking a set activates it (deactivating any other), reopens it if completed, and lands the Reader at its saved cursor; a cursor resolved past a deleted item is persisted.
- Dismiss soft-deletes with a confirmation; dismissing the active set also clears the Reader footer's set keys immediately.
- All-tombstoned sets are never activated — a warning shows instead.
- Set names are user-derived and render as literal text, never markup.
- The picker rows scroll so a long set list can't push lower rows or Close past the modal's height cap.
- The "Sets" button sits on the title row rather than the action toolbar, which already overflows the narrow Items pane and crashed on an extra button; it's hidden in select mode and omitted on the fresh-empty page.
- The picker dialog's CSS targets the dialog by ID to stay under the CSS fast-path ratchet; the regenerated CSS bundles are committed.

<sup>Written for commit 4dd2684e1ee59b58b59ca9f9f17325548fd82e28. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2337?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/User_Guide/library/media-and-conversations.md
+++ b/Docs/User_Guide/library/media-and-conversations.md
@@ -209,6 +209,34 @@ item-specific empty state; it does not silently switch modes.
 | "More" | Keeps secondary actions reachable: Edit metadata, Open original when available, Open manager, and Move to trash. Narrow layouts retain these actions here rather than hiding them. |
 | "Move to trash" | Two-step, title-specific confirmation. Success selects the adjacent item and leaves a bounded Undo receipt; Trash remains the durable recovery path. |
 
+### Review sets
+
+A review set pins an ordered snapshot of media items so you can work through
+them one by one, with your place and progress saved between visits.
+
+- **Create** — **Review these** on the media list pins the whole filtered
+  result (in the current sort order, capped at 500 items with a notice when
+  trimmed); in Select mode, **Review selected** pins just the checked items in
+  list order. Creating a set activates it and opens its first item in the
+  Reader.
+- **Walk** — while a set is active the Reader footer shows your place ("2 of
+  14 · 1 reviewed"). `]` advances and marks the item you leave as reviewed;
+  `[` goes back without marking; `m` toggles the loaded item's reviewed mark;
+  a final `]` on the last item marks it done in place. **Escape** leaves the
+  Reader but keeps the set active — re-entering resumes at your cursor.
+  **R** (Exit review) deactivates the set without deleting it.
+- **Resume / switch / dismiss** — **Sets** on the media list title row opens
+  the saved-set picker: each row shows the set's name and live progress, with
+  the active set marked `✓`. Picking a set activates it (deactivating any
+  other) and lands at its saved cursor; picking a completed set reopens it.
+  **Dismiss** soft-deletes a set.
+- Deleted media items become skipped tombstones: progress counts only items
+  that still exist, and a set whose items were all removed reports "No items
+  to review" instead of completing.
+
+*Verified against feat/review-sets-p4 — 2026-09-02 (task-28240..28243: live
+end-to-end create → walk → exit → resume → dismiss pass in tmux).*
+
 ### Conversations
 
 Conversations use a fixed **20-item page**. The row list scrolls independently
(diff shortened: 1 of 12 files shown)


rmusser01:
Thanks — 5 of 6 addressed in `c00c4e28b`; 1 declined with evidence.

**#6 Picker rows cannot scroll (Bug) — fixed.** Rows now compose inside `#library-review-set-picker-rows` (a `VerticalScroll`, max-height 20) with the Close action outside it, so a long saved-set list scrolls and Close stays reachable. Pinned by `test_many_rows_scroll_and_close_stays_reachable` (40 rows).

**#5 Resolved cursor is not saved (Bug) — fixed.** `_activate_review_set` now persists a tombstone-resolved cursor via `service.set_cursor` (exactly what the walker does on resume), so a later restore of the deleted item can't yank a subsequent resume back to the stale position. Pinned by `test_picker_worker_open_persists_a_resolved_tombstoned_cursor`.

**#4 Magic strings — fixed.** `PICKER_OPEN` / `PICKER_DISMISS` constants in `library_review_set_picker.py`, used by both the modal's dismiss calls and the worker's dispatch.

**#1 Import grouping — fixed.** Both project-local imports are now one contiguous absolute-import block.

**#2 Docstrings — fixed.** `compose()` documented; `__init__` gains a summary line before `Args:`; the three `@on` handlers documented too.

**#3 Review switch retains stale detail — declined.** The landing path is `_open_library_media_viewer`, the same seam every media-row press (and the Phase-3 entry points) use. It explicitly resets per-item viewer state before recomposing — `library_screen.py` clears `_library_media_highlights`, `_library_media_editing`, `_library_media_editing_analysis`, `_library_media_content_query`/`_match_index`/`_mode`, and routes through `_select_library_media_reader_row(..., immediate=True)`, which starts the fresh detail fetch for the new id. Switching sets therefore lands with exactly the same clean per-item state as clicking the row directly; there is no picker-specific bypass to fix.

PR #2340, opened 2026-09-03
feat(library): review sets phase 5 - build a set from read-later (task-28244)

rmusser01:
Phase 5 — the final planned source — of the review-sets program (design: `backlog/docs/design-library-review-sets.md`, task-28244; phases 1-4 = #2323 / #2333 / #2335 / #2337): build a review set from the **read-later queue**.

## What's here

- **"Review read-later"** action in the set picker's footer (the picker is the set hub, so the entry lives there — present even with no saved sets, since the read-later queue is independent of them). Pressing it resolves the picker with a `(PICKER_READ_LATER, "")` decision.
- **`_review_read_later_pairs`** (screen worker): ids via `list_read_it_later_media_ids` — already ordered `saved_at DESC` — then titles via the same bounded `id_allowlist` query Review-selected uses, with the browse-sort order discarded in favor of the saved order client-side. Capped, created as `origin="read_later"` (reserved in `VALID_ORIGINS` since Phase 1), and landed through the shared `_create_and_open_review_set` (empty queue notices; overflow warns).
- AC#2 (bridging done-marks to a global read marker) is contingent on task-28009, which isn't built — done-marks stay set-local, no code by design.

## Tests

3 new: the picker decision contract, the worker end-to-end over a real tmp service DB asserting saved-order pinning (`[30, 10, 20]` preserved against a differing browse order), and the empty-queue notice. 146 green across the review-set battery + multiselect suite; CSS fast-path ratchet suite green; preflight clean; no new logger, no CSS change (no bundle regen needed).

## Live verification (tmux, seeded media + read-later saves, cleaned up after)

Saved docs 1 and 3 to read-later (3 last) → **Sets ▸ Review read-later** → set "Read later" (`origin=read_later`, active) pinned `[doc3, doc1]` in saved order, Reader landed on doc 3 at "1 of 2 · 0 reviewed" → `]` walked to doc 1 at "2 of 2 · 1 reviewed".

---
## Summary by cubic
Adds a "Review read-later" action to the review-set picker that builds a new review set from the read-later queue, preserving the order items were saved. Done-marks remain set-local because the global read-marker bridge (task-28009) isn't built yet.

**Changes**
- The action lives in the picker's footer and is available even with no saved sets, since the read-later queue is independent.
- It resolves the picker with a `(PICKER_READ_LATER, "")` decision; the worker builds pairs from `list_read_it_later_media_ids` (saved order) and reorders titles client-side to that order.
- Sets are capped, created with `origin="read_later"`, and land through the shared `_create_and_open_review_set`; empty queues notify, overflow warns.
- `list_read_it_later_media_ids` now takes an optional `limit` bound so large queues are never fully fetched; the worker passes `REVIEW_SET_CAP + 1`.
- Adds five tests: the picker decision, order pinning, the empty-queue notice, the DB limit, and an end-to-end run against a real Media DB.

<sup>Written for commit 0e2a00b88b59487aa4e40259ef393e10be1f89a4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2340?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Thanks — 4 of 5 addressed in `b0a19cbd1`; 1 declined with precedent.

**#1 Read-later query unbounded — fixed.** `list_read_it_later_media_ids` gains an optional `limit` (SQL `LIMIT ?`, newest saves first) and the worker passes `REVIEW_SET_CAP + 1`, so the query itself is bounded instead of fetching the whole queue and slicing client-side. Pinned at the DB level by `test_list_read_it_later_media_ids_honors_the_limit`.

**#3/#4 Media DB mocked / integration coverage — fixed.** Added `test_read_later_pairs_run_against_the_real_media_db`: seeds a real `MediaDatabase`, saves items read-later, and runs the picker worker end to end — the real lister executes with the worker's real kwargs (so a bad `limit` kwarg or broken `saved_at DESC` ordering fails here, not just against a stub) and the resulting set pins the real saved order. Only the remote-shaped search service remains stubbed, mirroring the accepted #2333 shape.

**#5 Test docstrings — fixed.** Google-style docstrings on all three new test functions.

**#2 Read query bypasses transaction — declined.** Same ruling as #2333 (accepted there): `execute_query` **is** this DB layer's standard API for a single-statement read — every sibling read in `Client_Media_DB_v2.py` uses it, it runs on the same thread-local connection, and a one-statement `SELECT` gains nothing from a wrapping transaction. Wrapping just this call would diverge from the file's idiom.

PR #2342, opened 2026-09-03
feat(library): review sets phase 2b - auto-resume on media entry (task-28245)

rmusser01:
Phase 2b — the **final phase** of the review-sets program (design: `backlog/docs/design-library-review-sets.md`, task-28245, split from task-28241 AC#4; phases 1-5 = #2323 / #2333 / #2335 / #2337 / #2340): opening the media area with an active review set loads its cursor item into the Reader **without a keypress**.

## What's here

- **`_auto_resume_review_set_worker`**: resolves the active set's live cursor off the event loop (`_resolve_active_review_set_landing`: active set → chunked liveness → `resolve_cursor`), then opens via `_open_library_media_viewer` — the **same path as a row press**, so per-item `ReadingProgress` scroll restore is untouched (AC#2).
- **Once per set per screen session**: Escape back to the list plus re-entry shows the list — never a yank loop. The guard is only burned when the open actually happens.
- **Two hooks**: the rail-select seam's media branch, and `on_mount`'s media-list branch (`call_after_refresh`). The worker's final gates — still on the media list, screen is current — make a cold-start initial-tab yank abort without opening (AC#3). All failures are silent: auto-resume is a convenience, not a feature gate.

## Tests

4 new worker tests over a real tmp service DB: opens-the-cursor-item-once (with the once-guard re-run), no-active-set no-op, moved-away abort (AC#3), all-tombstoned quiet skip. 152 green across the review-set battery + multiselect suite; preflight clean; no CSS, no schema, no new logger.

## Live verification (tmux, seeded media, cleaned up after)

Create set → walk to item 2 ("2 of 3 · 1 reviewed") → **quit the app** → relaunch → Library → click **Media** → the Reader auto-loads item 2 at "2 of 3 · 1 reviewed" with no further keypress. Then Escape → Conversations → Media again shows the **list** (once-guard holds).

---
## Summary by cubic
Opens an active review set's cursor item in the Reader when the Media rail is clicked, without a keypress. Previously entering media always showed the list; now with a set active, the Reader loads at the current item, but only once per set per session—escaping back to the list and re-entering shows the list again.

**New Features**
- Kicks off only from the rail-select media branch; the mount-leg kick was dropped because its boot-time timing raced the `_ui_ready` module census and flaked the Perf Guard.
- Resolves the live cursor off the event loop and opens through the same path as a row press, preserving per-item scroll restore; final gates abort a cold-start yank without opening.
- All failures are silent; updates the user guide.

**Bug Fixes**
- Runs in its own exclusive worker group, so clicking the Media rail during an in-flight "Review these" build no longer cancels the build.

<sup>Written for commit f49b819f4b7d3e7bd36ea669bb3a4732309dd42e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2342?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/User_Guide/library/media-and-conversations.md
+++ b/Docs/User_Guide/library/media-and-conversations.md
@@ -219,6 +219,9 @@ them one by one, with your place and progress saved between visits.
   trimmed); in Select mode, **Review selected** pins just the checked items in
   list order. Creating a set activates it and opens its first item in the
   Reader.
+- **Resume on entry** — opening the media area with a set active loads its
+  current item into the Reader automatically (once per set per session;
+  Escape back to the list and re-entering shows the list).
 - **Walk** — while a set is active the Reader footer shows your place ("2 of
   14 · 1 reviewed"). `]` advances and marks the item you leave as reviewed;
   `[` goes back without marking; `m` toggles the loaded item's reviewed mark;
--- a/Tests/UI/test_review_set_walker.py
+++ b/Tests/UI/test_review_set_walker.py
@@ -683,6 +683,80 @@ async def test_picker_worker_failure_notifies(tmp_path):
     assert any(severity == "error" for _msg, severity in fake._notices)
 
 
+def _auto_resume_fake(service, *, live_ids=None):
+    """Fake screen for the auto-resume worker (task-28245)."""
+    fake = _entry_fake(service)
+    fake._review_set_live_ids = (
+        (lambda ids: {int(i) for i in ids})
+        if live_ids is None
+        else (lambda ids: set(live_ids))
+    )
+
+    async def run_call(call, *args, isolate_in_worker=False, **kwargs):
+        result = call(*args, **kwargs)
+        return await result if hasattr(result, "__await__") else result
+
+    fake._run_library_service_call = run_call
+    fake._library_selected_row_id = "browse-media"
+    fake._library_media_view = "list"
+    fake.is_current = True
+    for name in (
+        "_auto_resume_review_set_worker",
+        "_resolve_active_review_set_landing",
+    ):
+        setattr(fake, name, MethodType(getattr(LibraryScreen, name), fake))
+    return fake
+
+
+@pytest.mark.asyncio
+async def test_auto_resume_opens_the_active_sets_cursor_item_once(tmp_path):
+    # task-28245 AC#1: entering the media area with an active set loads its
+    # cursor item without a keypress -- but only ONCE per set per screen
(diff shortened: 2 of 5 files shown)


rmusser01:
Thanks — the bug is fixed in `579a52482`; the two test-shape findings are declined with in-repo precedent.

**#3 Auto-resume cancels set creation (Bug) — fixed, good catch.** The worker now runs in its own exclusive group (`library_review_set_resume`) instead of joining `library_review_set`, so a Media-rail click during an in-flight "Review these" build can no longer cancel the creation worker, while auto-resume still debounces its own re-entries. Pinned by `test_auto_resume_runs_in_its_own_worker_group`, which asserts on the actual `run_worker` kwargs.

**#1 Test layers — declined.** The fake-screen + real-`ReviewSetService`-over-a-real-DB shape is this file's established contract, accepted across the three prior reviews of this feature (#2333 walker tests, #2335 worker tests, #2340 read-later tests — #2340 explicitly asked for MORE of this real-DB shape, not less). The pure-model layer has its own isolated unit suite (`Tests/Library/test_review_set_state.py`), and the service has direct coverage (`Tests/Library/test_review_set_service.py`); these worker tests are deliberately the seam between them.

**#2 Disk-backed sqlite — declined.** `_service(tmp_path)` is the helper every walker/picker/read-later test in this file already uses (merged in #2333 and reused since); switching only the four new tests to `:memory:` would fork the file's fixture idiom for no coverage gain. `:memory:` coverage exists where it was asked for — the service-level suite added it during the #2323 review round. tmp_path databases are per-test-isolated and cleaned by pytest.

rmusser01:
Follow-up in `1f0521c38`: the second Perf Guard run flagged the mount-leg kick — its boot-time `call_after_refresh` let the worker's lazy review-set imports race the `_ui_ready` module census on slow runners (977 > 972; passed on one run, failed the next, passes locally — classic timing flake). Auto-resume now fires only from the rail-select seam: the explicit user gesture, which cannot race boot and is exactly the path the live verification exercised (post-restart click on Media auto-loaded the cursor item). The PR description's 'two hooks' is now one hook, deliberately.

PR #2346, opened 2026-09-03
fix(library): review-set finish line + no silent failures (task-30041, task-30042)

rmusser01:
First of four fix PRs from the 2026-09-03 design critique of Library ▸ Media (dual-agent, 24/40; snapshot in `.impeccable/critique/`). Covers the two review-set P1s — tasks 30041 + 30042 — under the explicit ruling that **silent failure is never acceptable**.

## task-30041 — the finish line works

- `check_action` for `]`/`[` gated on **browse-row adjacency**, so the documented final-`]` completion gesture was disabled at the binding layer on the last browse row (footer still advertising it), and the gate misfired whenever set order ≠ browse order (selection/filtered sets). With an active set the keys now gate on `_review_set_active()` — the walk clamps safely at the set's ends.
- The walk's clamp branch (which the completion gesture rides) marked the item and refreshed completion but never synced the Reader chrome — the footer stayed stale at exactly the finish. It syncs now.
- Completing a set gets a real moment: the walk that flips the set to complete fires **"All N reviewed."** once (later idempotent presses don't re-announce).

## task-30042 — no silent failures

- Pressing **Sets** or landing a create with unavailable storage now says **"Review-set storage is unavailable."** instead of silently returning — a dead button was indistinguishable from the feature not existing.
- `walk` / `toggle reviewed` / `exit review` wrap storage errors into error notices; the walk **consumes** the key on error rather than stacking a browse fallback on top of an unseen failure.
- Auto-resume surfaces storage errors ("Couldn't resume the review set — storage error.") and warns once per session when storage is missing; only the ordinary no-active-set case stays quiet.

## Tests

9 new (all watched fail first): the `check_action` set-gate, clamp-branch sync + completion notice + once-only, walk/toggle failure notices, Sets-press and create storage notices, auto-resume failure notice + once-per-session storage warning. 98 green across the review-set battery.

3-item set → `]` `]` → "3 of 3 · 2 reviewed" on the last item → final `]` flips the footer to **"All 3 reviewed"** in the same interaction with the completion toast visible. Before this change the same final press was a silent no-op with the footer pinned.

---
## Summary by cubic
Review-set navigation now completes reliably (TASK-30041): `[`/`]` follow the active set instead of browse-row adjacency, and the final mark refreshes the Reader chrome and announces completion. Review-set storage failures now surface notices instead of being silently ignored (TASK-30042).

**Bug Fixes**

- Storage errors in walking, toggling, exiting, creating, and resuming review sets produce user-visible feedback; failed walks consume the key instead of falling through to browse navigation.
- The set-active gate fails closed on storage errors so key resolution and footer renders can't crash-loop the screen, and each error path logs its traceback before notifying.
- Missing storage warns once per session during auto-resume, while the ordinary no-active-set case remains quiet.
- Added regression coverage for finish-line behavior, idempotent completion notices, action gating, and storage failures.

<sup>Written for commit 32a28183032e8f16ff732c15e626aa0ab64f6e18. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2346?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.impeccable/critique/2026-09-03T12-58-19Z__tldw-chatbook-ui-screens-library-screen-py.md
+++ b/.impeccable/critique/2026-09-03T12-58-19Z__tldw-chatbook-ui-screens-library-screen-py.md
@@ -1,0 +1,71 @@
+---
+target: Library ▸ Media surface (post review-sets program)
+total_score: 24
+max_score: 40
+na_heuristics: 
+p0_count: 0
+p1_count: 3
+timestamp: 2026-09-03T12-58-19Z
+slug: tldw-chatbook-ui-screens-library-screen-py
+---
+Method: dual-agent (A: live design review sub-agent · B: detector/evidence sub-agent)
+
+Target: Library ▸ Browse ▸ Media (list, toolbar, select mode, Reader, Trash, review sets) — live tmux at 235×52 + 100×30, seeded 6-item profile. Session caveat: a second app instance from another worktree shared the user profile during Assessment A; findings that depend on persistence are attributed accordingly.
+
+## Design Health Score
+
+| # | Heuristic | Score | Key Issue |
+|---|-----------|-------|-----------|
+| 1 | Visibility of System Status | 2 | Footer goes stale at the set-completion moment; list view gives no hint a set is active; "Loaded in Reader" marker persisted after the Reader emptied |
+| 2 | Match System / Real World | 3 | "Remove later" reads as deferred removal; "Open manager" jargon; "Match 1 of 1 matches" counts blocks, not occurrences |
+| 3 | User Control and Freedom | 3 | "Review set dismissed." has no Undo (bulk delete does); no keyboard route back to a paused set |
+| 4 | Consistency and Standards | 3 | Same width: Conversations stacks its toolbar readably, Media chops horizontally; `]`/`[` silently change meaning browse↔set |
+| 5 | Error Prevention | 3 | Armed confirm + danger placement strong; undermined by `Tr`/`R` mystery-meat buttons two cells apart |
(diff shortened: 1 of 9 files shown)


rmusser01:
Thanks — 3 of 4 addressed in `29e6fec8f5`; 1 declined with precedent.

**#3 Binding gate can query-crash (Bug) — fixed, good catch.** `_review_set_active` now fails CLOSED: a storage exception degrades the `]`/`[` gate to browse-row adjacency (with a logged traceback) instead of propagating out of `check_action`, which runs on every key resolution and footer render. No user notice there deliberately — it would toast-storm at render frequency; the explicit gesture paths (walk/toggle/exit/picker) carry the notices. Pinned by `test_check_action_fails_closed_when_storage_errors`.

**#4 Action failures lose diagnostics (Bug) — fixed.** The walk/toggle/exit/auto-resume wrappers now log the full traceback (`logger.opt(exception=True).warning(...)`) before notifying. The production diagnostic inventory was regenerated after reviewing the drift row per the repo's procedure (library_screen.py +5 calls; all fixed-string messages, no user content/secrets/paths interpolated).

**#2 Storage notice string duplicated — fixed.** One class constant `_REVIEW_SET_STORAGE_UNAVAILABLE` used by all three call sites.

**#1 Fake-based tests lack integration coverage — declined.** Same ruling as #2342 (accepted there): the fake-screen + real-`ReviewSetService`-over-a-real-DB shape is this file's established contract across #2333/#2335/#2340/#2342; the pure model and the service each have their own direct suites, and these tests deliberately exercise the seam between them. The finish-line behavior was additionally live-verified in tmux (final `]` flips the footer to "All 3 reviewed" + completion toast).

PR #2350, opened 2026-09-03
feat(library): legible multi-row action grammar for the media items pane (task-30043)

rmusser01:
Second of four fix PRs from the 2026-09-03 design critique of Library ▸ Media (task-30043; #2346 shipped the review-set hardening). Fixes the critique's other P1: the items pane's action toolbar was illegible in its resting state.

## The problem (live captures in the critique snapshot)

The items pane sits at **~40-44 columns in every real shell layout** (the 3-pane reading shell and the compact stage alike). Six compact buttons sharing one row rendered as `t so E Tr R Se`; entering select mode showed the bulk actions as bare `○ ○ ○` disabled markers — for a keyboard/screen-reader user the accessibility markers *were* the labels; the armed bulk-delete safety copy clipped mid-word ("You can und / restore later from Tr").

## The fix — multi-row is THE grammar

- **Browse**: three budgeted rows — `[type, sort]`, `[Export…, Trash, Select]`, `[Review these]`. Full labels always. (In select mode most of these hide, so its single row keeps fitting.)
- **Select mode**: summary row `[N selected, Select all N shown]`; bulk row `[Clear, Export, Review]` — short REAL words (the mode and the adjacent count make them unambiguous; the F-018 tooltips carry the full sentences; the `○` disabled marker always rides WITH its word); **Delete isolated on its own danger row** — task-2853's far-end rule, upgraded. The `_apply_library_row_toggle` in-place patch keeps working via the stashed short bases.
- **Content-width buttons**: a canvas `BUNDLED_CSS` block lifts Textual Button's 16-cell min-width floor (six floors = 96 cells could never fit) — this **supersedes task-28025's `1fr` squeeze**, whose equal shares were what chopped the labels. Every row's label budget (including the `○ `-prefixed disabled forms) fits the pane floor, which is what keeps 28025's fit contract true — its pinning test still passes.
- **The confirm copy wraps whole**: the canvas claimed `min-width: 40` while the shell allots ~37, so children laid out 3 cells wider than visible and the safety sentence clipped. The floor drops to 36 (it only exists to bound the 13fr resolution trap), and the copy Static is bounded to the pane.
- **No width machinery**: an earlier draft made this a responsive variant driven by `on_resize` + recompose; that mount-time recompose raced the viewer-return settlement (8 side-by-side tests) and was deleted rather than tuned — the pane is never wide enough for the single row in any shipped layout.

CSS bundle + widget-defaults regenerated and committed (BUNDLED_CSS change).

## Tests

5 in the new `test_library_media_toolbar_adapt.py` (row structure, word-not-marker labels, 40-col fit for browse and select rows via region assertions, confirm-copy wrap) + the full multiselect (64) and side-by-side (31, incl. task-28025's fit contract) suites green — 100 total.

At the real pane: `type: All types / sort: Newest`, `Export… Trash Select`, `Review these` all as words; select mode reads `0 selected · Select all 3 shown / Clear ○ Export ○ Review / ○ Delete`; checking a row flips the labels in place; the armed confirm shows the complete sentence "Delete 1 selected item? You can undo right away, or restore later from Trash."

---
## Summary by cubic
Implements task-30043 by replacing the media items pane's single six-button toolbar with a multi-row action grammar, so actions read as full words instead of fragments like `t so E Tr R Se` at the pane's ~40-column width.

- Browse mode uses three rows: `[type, sort]`, `[Export…, Trash, Select]`, and `[Review these]`.
- Select mode shows a summary row, a bulk row (`Clear`, `Export`, `Review`), and Delete isolated on its own danger row.
- Disabled bulk actions keep the `○` marker attached to their word, so they render as `○ Export` rather than a bare `○`.
- Toolbar buttons now size to their content, superseding the `1fr` squeeze that chopped labels.
- The bulk-delete confirmation copy now wraps inside the pane; the canvas `min-width` floor drops from 40 to 36 to fit the shell's ~37-column allotment.
- Long stored type values are capped at 8 characters plus ellipsis in the chooser label, with the full value kept in the tooltip.
- Adds 6 UI tests covering row structure, word-not-marker labels, 40-column fit, confirmation-copy wrapping, and the type-label cap.

<sup>Written for commit 8ff9a3961ba99c45d5fce87f5ed301c41cbc75b2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2350?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_library_media_toolbar_adapt.py
+++ b/Tests/UI/test_library_media_toolbar_adapt.py
@@ -1,0 +1,187 @@
+"""The media canvas's multi-row action grammar (task-30043).
+
+Critique 2026-09-03 P1: the items pane sits at ~40-44 cols in every real
+shell layout, so the old single six-button row chopped to ``t so E Tr R Se``
+and select mode's bulk actions rendered as bare ``○ ○ ○``. The multi-row
+grammar is THE grammar — every row's label budget (including the "○ "
+disabled forms) fits the pane's 40-col floor, so labels are always words.
+"""
+
+from __future__ import annotations
+
+import pytest
+from textual.widgets import Button, Static
+
+from tldw_chatbook.Library.library_media_state import (
+    LibraryMediaCanvasState,
+    LibraryMediaRow,
+)
+from tldw_chatbook.Widgets.Library.library_media_canvas import LibraryMediaCanvas
+from Tests.UI.consolidated_css import ConsolidatedCSSApp
+
+
+def _browse_state() -> LibraryMediaCanvasState:
+    rows = (
+        LibraryMediaRow(
+            media_id="1",
+            title="First item",
+            media_type="document",
+            secondary="document · today",
+        ),
+    )
+    return LibraryMediaCanvasState(
+        rows=rows,
+        type_options=(None, "document"),
+        active_type=None,
+        status_copy="",
+        empty_copy="No media yet.",
+        selected_id="",
+        preview_lines=(),
+        count=1,
+    )
+
+
+def _select_state(
+    *, selected_count: int = 0, confirming: bool = False
+) -> LibraryMediaCanvasState:
+    rows = (
+        LibraryMediaRow(
+            media_id="1",
+            title="First item",
+            media_type="document",
+            secondary="document · today",
+            checked=selected_count > 0,
+        ),
+        LibraryMediaRow(
+            media_id="2",
+            title="Second item",
+            media_type="document",
+            secondary="document · today",
+            checked=False,
+        ),
+    )
+    return LibraryMediaCanvasState(
+        rows=rows,
+        type_options=(None, "document"),
+        active_type=None,
+        status_copy="",
+        empty_copy="No media yet.",
+        selected_id="",
+        preview_lines=(),
+        count=2,
+        select_mode=True,
+        selected_count=selected_count,
+        confirming_bulk_delete=confirming,
+    )
+
+
+class _CanvasApp(ConsolidatedCSSApp):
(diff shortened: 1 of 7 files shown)


rmusser01:
Thanks — the bug is fixed; the test-shape finding is declined with the repo's precedent.

**#2 Chooser row still overflows (Bug) — fixed, good catch.** Type values are data and can be long. The opener LABEL now caps the value at 8 chars + ellipsis (`type: presenta…`) via `_capped_choice_value`, while the tooltip and the direct-pick chooser strip itself keep the full value — so the row's budget holds for any stored type. Pinned by `test_long_type_values_are_capped_in_the_chooser_label`, which asserts both the cap and the buttons' rendered regions staying inside the canvas at the pane floor. (Sort labels are a fixed vocabulary — `Title A-Z` at worst — so the sort side needs no cap.)

**#1 Toolbar helpers lack unit tests — declined.** Mounted-harness tests are this repo's established contract for canvas compose (the 64-test multiselect suite is entirely this shape), and deliberately so here: the failure mode being pinned is *rendered geometry* — chopped labels and clipped regions — which only region assertions on a real layout can catch (a unit test of the ComposeResult generator would have passed while the live toolbar rendered `t so E Tr R Se`). The five new tests assert exact labels, marker bases, tooltips, row structure, and region fit at the 40-col floor.

PR #2356, opened 2026-09-03
Restrict production PyPI publishing to main

rmusser01:
## Summary
- Change production PyPI publishing from protected `v*` tag pushes to protected `main` branch pushes.
- Add workflow concurrency so same-ref main publish runs serialize instead of racing duplicate uploads.
- Extract PyPI version-exists and GitHub output-writing logic into `Packaging/check_pypi_release.py`.
- Validate `GITHUB_OUTPUT` against `RUNNER_TEMP` before writing workflow outputs.
- Keep TestPyPI as manual dispatch from protected `dev`.
- Update release docs and release wrapper next-step text away from the old tag-driven production path.
- Verified the live GitHub `pypi` environment deployment policy now permits only the `main` branch.

## Verification
- `/Users/macbook-dev/Documents/GitHub/tldw_chatbook/.venv/bin/python -m pytest Tests/Packaging/test_release_metadata.py -q` -> 15 passed, 1 existing dependency warning
- `/Users/macbook-dev/Documents/GitHub/tldw_chatbook/.venv/bin/python -m py_compile Packaging/check_pypi_release.py Tests/Packaging/test_release_metadata.py` -> passed
- `/Users/macbook-dev/Documents/GitHub/tldw_chatbook/.venv/bin/python -m ruff check Packaging/check_pypi_release.py Tests/Packaging/test_release_metadata.py` -> passed
- `/Users/macbook-dev/Documents/GitHub/tldw_chatbook/.venv/bin/python -c "from pathlib import Path; import yaml; yaml.safe_load(Path('.github/workflows/publish-pypi.yml').read_text()); print('yaml ok')"` -> yaml ok
- `bash -n Packaging/build_release.sh` -> passed
- `git diff --check` -> passed
- `gh api repos/rmusser01/tldw_chatbook/environments/pypi/deployment-branch-policies` -> only `main` branch policy

Backlog: TASK-31213

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -3,8 +3,12 @@ name: Publish Python package
 on:
   workflow_dispatch:
   push:
-    tags:
-      - "v*"
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
 
 permissions:
   contents: read
@@ -13,6 +17,8 @@ jobs:
   build:
     name: Build and check distributions
     runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.release-version.outputs.version }}
 
     steps:
       - name: Check out source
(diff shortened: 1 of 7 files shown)


rmusser01:
Addressed the Qodo review in 2a6f760fbd:

- Moved the PyPI version-exists guard into `Packaging/check_pypi_release.py` and added executable coverage for 200, 404, unexpected HTTP errors, workflow-facing output emission, and validated output-path writes.
- Replaced direct `GITHUB_OUTPUT` writes in the workflow with `write_github_output(...)`, which validates the output file against `RUNNER_TEMP` via `tldw_chatbook.Utils.path_validation.validate_path`.
- Added workflow concurrency with `cancel-in-progress: false` so same-ref main publish runs serialize instead of racing duplicate uploads.
- Reran targeted packaging tests, py_compile, Ruff, YAML parsing, shell syntax, diff whitespace, and the live GitHub environment-policy check.

PR #2359, opened 2026-09-03
fix(library): honest footer under typing focus and at completion; filter-miss recovery (task-31223..31225)

rmusser01:
Second fix PR from the 2026-09-03 re-critique of Library ▸ Media (parallel to #2358). Tasks 31223 + 31224 + 31225 — the footer-honesty P1 and both recovery P2s.

## task-31223 — the footer never lies about typing focus

With focus in any Input/TextArea, every single printable key is inserted as text — yet the footer kept advertising `] next in set | m | R` (live: a stray `]` corrupted the filter into a zero-match list; "s select" typed an "s" into the rail search). The footer now swaps to a typing context: swallowed printable keys drop, keys that still work (esc / enter / F-keys) and informational chips stay, and the swap announces itself (`typing in field`). Re-registered from `on_descendant_focus` on context **flips** only, so ordinary focus moves cause zero footer churn.

## task-31224 — filter-miss recovery is the right one, and visible

A zero-match *active-query* page pinned **Import media** as its recovery (`fresh_zero` never checked `canvas.query`) while the actual recovery — **Clear filter** — never rendered anywhere: Textual `Input` defaults to `width: 100%` and consumed the whole filter row, pushing the button off-screen at every width. The Input now shares its row (`1fr`), Clear is visible and enabled whenever a query is applied, and the Import/Show-all suggestions are gated to genuinely-empty sources.

## task-31225 — completion stops advertising a dead key

After "All N reviewed" the footer kept advertising `] next in set` — an idempotent no-op, violating the same honest-footer rule (task-28005) that trims Prev/Next at list ends. On a complete set the segment is now `m toggle reviewed | R finish review | All N reviewed`; un-marking with `m` restores the walk keys (live-verified round trip). Extracted `_review_footer_entries` so the segment is unit-testable. Rider: documented why the ambient storage warning is `warning` while gesture failures are `error` (the severity inconsistency Assessment B flagged).

## Tests

6 in the new `test_library_footer_focus.py` (typing-context transform incl. TextArea, pass-through cases, completion/in-progress segments) + the filter-miss recovery contract in the adapt suite (no Import on query miss; Clear rendered inside the pane). 125 green across footer/walker/toolbar/multiselect suites; preflight clean; no schema, no logger.

Focusing the rail search mid-review swaps the footer to `typing in field | F6 next pane | 1 of 2 · 0 reviewed | esc back`; Escape restores the full set. Completing a 2-doc set shows `m toggle reviewed | R finish review | All 2 reviewed` with no `]`; pressing `m` resumes the walk keys.

---
## Summary by cubic
Fixes the Library footer so it stops advertising keys a focused text field swallows, stops advertising a dead key at review completion, and makes filter-miss recovery actually show "Clear filter."

**Bug Fixes**
- With focus in an Input or TextArea, the footer drops printable keys that would be inserted as text, keeps working keys like esc and F6, and announces "typing in field."
- The typing-context swap routes through the Notes-aware dispatcher so it no longer overwrites the Notes editor's region-specific footer on focus flips.
- On a complete review set, the footer no longer advertises `]` (a no-op); it now shows `m toggle reviewed | R finish review | All N reviewed` instead.
- A zero-match filter page no longer suggests "Import media"; the "Clear filter" button is now visible and enabled since the filter Input shares its row instead of consuming full width.
- Adds a lessons doc on paint-over bugs being invisible to region assertions, from the task-31221 investigation.

<sup>Written for commit a0280b5ea5f449bd29b4bec72ac4f1eb230ed93e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2359?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Tests/UI/test_library_footer_focus.py
+++ b/Tests/UI/test_library_footer_focus.py
@@ -1,0 +1,97 @@
+"""The footer never advertises keys a focused text field will swallow.
+
+Re-critique 2026-09-03 P1 (task-31223): the footer branch keyed off screen
+state, not focus — with focus in the rail search it advertised
+"] next in set | m | R" while those keystrokes were inserted as text (a
+stray "]" corrupted the filter into a zero-match list). Keyboard-first
+brand: the footer is the instrument and must not lie.
+"""
+
+from __future__ import annotations
+
+from types import MethodType, SimpleNamespace
+
+from textual.widgets import Button, Input, TextArea
+
+from tldw_chatbook.Library.library_rail_state import LibraryLifecycle
+from tldw_chatbook.UI.Screens.library_screen import LibraryScreen
+
+_ROUTE = (
+    ("/", "focus search"),
+    ("F6", "next pane"),
+    ("]", "next in set"),
+    ("[", "prev in set"),
+    ("m", "toggle reviewed"),
+    ("R", "exit review"),
+    ("", "2 of 6 · 1 reviewed"),
+    ("esc", "focus Library"),
+)
+
+
+def _fake(focused) -> SimpleNamespace:
+    fake = SimpleNamespace(
+        _library_route_shortcuts_for_current_state=lambda: _ROUTE,
+        _library_emergency_return_eligibility=lambda: SimpleNamespace(
+            enabled=False
+        ),
+        _library_lifecycle=LibraryLifecycle.GRADUATED,
+        focused=focused,
+    )
+    fake._library_footer_shortcuts_for_current_state = MethodType(
+        LibraryScreen._library_footer_shortcuts_for_current_state, fake
+    )
+    return fake
+
+
+def test_typing_focus_drops_swallowed_printable_keys():
+    shortcuts = _fake(Input())._library_footer_shortcuts_for_current_state()
+    keys = [key for key, _ in shortcuts]
+    # Single printable keys are inserted as text while typing — never shown.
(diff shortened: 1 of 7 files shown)


rmusser01:
Thanks — both addressed in `HEAD`.

**#2 Focus overwrites notes footer (Bug) — fixed, good catch.** The typing-context flip now routes through `_apply_library_notes_footer_context` — the existing dispatcher that applies the Notes region tiers when that workflow is active and falls back to `_library_footer_shortcuts_for_current_state` (which carries the typing filter) otherwise — instead of calling the generic registration directly. So a focus flip inside the Notes editor re-applies the *Notes* footer, and everywhere else gets the typing-filtered generic set. Extracted `_refresh_footer_typing_context` and pinned the routing + flip-gating with `test_typing_flip_routes_through_the_notes_aware_dispatcher`.

**#1 Test docstrings — fixed.** All test functions in `test_library_footer_focus.py` now carry Google-style one-liners.

PR #2367, opened 2026-09-04
fix(library): sort chooser shows every option; Reader fills its pane, Find collapses, dead pager dropped

rmusser01:
Second fix PR from critique #3 of Library ▸ Media (parallel to #2366). Tasks 31235 + 31237 — the sort-chooser P1 and the Reader vertical-space P2.

## task-31235 — the sort chooser renders every option

The horizontal choice strip rendered "✓ Newest  Oldest  Title A-" in the ~38-col items pane: "Title A-Z" truncated, "Title Z-A" entirely invisible with no overflow cue — yet a keyboard user could arrow to and select the unrendered option. The chooser is now the type chooser's vertical OptionList twin (✓-prefixed active option pre-highlighted); the handler moved from the strip's Button.Pressed to OptionList.OptionSelected with identical validation and close-on-same-value semantics, and the open-focus rides `_sync_library_canvas`'s `then` hook (a bare `call_after_refresh` has no ordering against a canvas-scoped recompose — a test caught it). The 31221-family `*:focus` outline struck again — at option-count height it overwrote "Newest" and "Title Z-A" outright — pinned with a painted-text test and suppressed at app tier for the new id.

## task-31237 — the Reader uses its vertical space

Three parts, all critique-verified live:

- **Fill**: `#library-media-viewer-content` goes from the 75vh cap to `height: 1fr` — measured at 200x52 with a 300-line document, the box ends exactly at the pane bottom and the viewer's virtual height equals its container: no stranded blank band, no double scroll, chrome pinned, overflow owned by the box's own scroll. (The old "1fr balloons the pane" note predated task-31222's height:auto wrappers; with them in place 1fr resolves correctly against the viewer remainder.)
- **Find collapse**: the "Search content…" bar no longer renders permanently (it duplicated the Find button and spent 3 rows on every fresh item). New `find_open` state with one `_close_library_media_find()` seam replacing all 8 query-reset pairs (item open, external open, viewer exit, rail switch, delete, mode change, Escape, entry guard); threaded through the viewer constructor AND the in-place `_sync_library_media_viewer_state` compare/assign (the #2351 trap). Focus is taken by the bar's own post-mount hook — three screen-level `call_after_refresh` variants provably lost the race against the nested recompose-mount, and an empty-query mount is by construction the Find gesture (active-query remounts never steal focus).
[...]

PR #2386, opened 2026-09-04
fix(library): wave 4 B — footer tells the truth; Escape, F6 and Back never strand focus (tasks 31271/31272)

rmusser01:
Second PR of fix wave 4 from critique #4 of Library ▸ Media (P1s 3 and 4). Built with subagent-driven development: two tasks, each with a spec+quality review, one fix round each, a whole-branch review and one scoped fix wave. Every ruling is recorded below.

## task-31271 — the footer tells the truth at four seams

- **Stale `esc close find`** — the Escape chip and the Escape action now read one seam, `_library_media_find_state()` (find_open / query / the mounted bar), so neither lags the DOM by a refresh.
- **`s` then Space** — entering select mode focuses the loaded (or first) row via the canvas sync's `then` hook, so Space toggles at once. Space is a priority binding whose `check_action` is True only in select mode with focus on a media row or a **media** pane grip (`library-media-pane-grip` + the Media rail guard). A first cut matched the shared adaptive-grip class and would have swallowed Space on five other canvases whenever select mode survived a rail switch; the review caught it. Buttons are Enter-bound in Textual 8, so nothing else changes.
- **Last item of a set** — `_review_footer_entries(progress, at_last=True)` names the completion gesture `] finish review`.
- **`l` / `c` / `t`** are advertised in the Reader footer (read later / use in Console / trash), gated by their own bindings. Live verification found the chips only appeared after an unrelated F6: the footer registered at open time was never re-derived when the detail landed. Fixed at the settle seam and pinned with a painted-footer assertion.

## task-31272 — Escape, F6 and Back never strand focus in a text input

- Escape from the Reader focuses the loaded Items row and leaves the Reader live; Escape from the Items region focuses the **Media rail row**, never `#library-search-input`.
- The More menu did close before — its `check_action` gate refused because Back/rail-Escape had flipped `_library_media_view` to `"list"`. The flip is gone in the three-pane layout, so `]`/`[` stay bound; one predicate (`_library_media_reader_exit_available`) now decides the Back control, the rail-Escape branch and the chip, and rides the viewer's compare/assign as `back_visible`.
- "‹ Back" is not composed in the side-by-side layout; the compact layout keeps it and the existing exit.
- F6's content stop tints the box's existing border (`outline: none`, painted-text proven — the task-31221 lesson).
- Escape vocabulary 8 → 4: `close` / `focus Items` / `focus Library` / `back` (table in the SDD report).
- With the flip gone, everything keyed on `view == "list"` was re-keyed on the Items region holding focus (`_library_media_list_surface_active`): the `s` gate, the list footer set, the type/sort strips (Escape closes an open strip first), the background list refresh. Ruling: on a plain row the Reader footer stays (`]`/`[`/`l`/`c`/`t` work there) with `s` added; the Items pane takes the list footer in select mode or with a strip open.
- The new layout hook only syncs when the viewer is mounted (task-21116's whole-screen-recompose pin holds).

## Also in this PR

- Three `test_library_shell.py` tests that waited on an always-mounted Find bar since #2367 now open Find first; two go green, the large-document proxy's residual is its pre-existing parse-count pin (task-31249 territory).
- `_active_review_set_progress` (production-dead after this branch) deleted with its two tests.
- User guide narrowed to what was verified: side-by-side at 235x52, compact at 100x30; below ~92 columns the exit registers but nothing visible changes yet (follow-up). Grips take Enter; Space is reserved for rows in select mode. New keys listed.

## Follow-ups to file (from the reviews)

Below ~92 columns the exit should open the Items pane; select mode survives an Escape hop to the rail (self-recovering); F6's first two stops are text inputs; at 100x30 the list view advertises `esc focus rail` but the rail target is not focusable there; a `_library_media_focus_region()` refactor; `library-media-back` is still an F6 candidate; Escape from the Reader with a visible strip hops focus instead of closing the strip; focus falls to the pane grip after any Reader recompose (wave rider).

## Verification

Per task: TDD RED/GREEN evidence in the SDD reports; render-fixes 23, reader-flow 50, multiselect 68, walker 58 on the head merged with dev; `test_library_shell.py -k media` 6 failed / 121 passed, a strict subset of the base's 8 / 119 (one order-dependent flake noted). Derived-artifact checks all exit 0; no new `logger.*`; bundle regenerated and in sync. Live in tmux at 235x52 and 100x30 for every AC (captures quoted in the reports).

https://claude.ai/code/session_011LebG4HPfSniVohbuXkU4n

---
## Summary by cubic
Fixes the Media footer so every advertised shortcut works, and makes Escape, F6, and Back never leave keyboard focus stranded in a text input.

**Footer tells the truth**

- The Escape chip and action read the Find bar's live state, so `esc close find` no longer lags a refresh.
- Entering select mode focuses the loaded row, so Space toggles selection immediately; Space is gated to select mode with focus on a media row or media pane grip.
- On the last displayed item of a review set the `]` chip reads `] finish review`, measured from the displayed item so an off-set item never sees a false promise.
- `l`/`c`/`t` chips appear from the Reader footer once a detail settles, instead of only after an unrelated F6.

**Escape, F6, and Back keep focus out of inputs**

- Escape from the Reader focuses the loaded Items row; Escape from the Items region focuses the Media rail row, never `#library-search-input`.
- The stale `view == "list"` flip is gone in the three-pane layout, so `]`/`[` stay bound and the More menu closes as its chip promises; strip visibility is focus-free, so Escape closes an open type/sort strip from the Reader too and its chip reads `close`.
- "‹ Back" composes only in the compact layout, where a back-to-list exit actually changes what's on screen.
- F6's content stop tints the viewer's existing border so keyboard users can see it landed.
- The Escape chip vocabulary drops from eight to four: `close` / `focus Items` / `focus Library` / `back`.
- The Reader keeps its footer (including `s`) while the Items region holds focus, and the Items pane takes the list footer only in select mode or with a strip open.

<sup>Written for commit a84a8761a98d9d1450334ed8f76e806acbf5dd0e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2386?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Qodo findings adjudicated in `96c03692fb` — 3 fixed, 0 declined:

1. **Fixed** (magic strings) — `_MEDIA_VIEW_LIST` / `_MEDIA_VIEW_VIEWER` are module constants beside the other `_MEDIA_*` names and are used at the four view comparisons this branch added or touched. A file-wide sweep is out of scope for this PR: 55 pre-existing literal comparison sites remain, recorded for a follow-up.
2. **Fixed** (completion label) — `at_last` now follows the LOADED item: `_active_review_loaded_at_last()` compares the loaded backing id with the set's last live item (positions sorted, tombstones skipped), composed after the cheap cursor test so the extra read stays off the common path. With an off-set item loaded the chip reads `] next in set` (resume); the genuine last item still reads `finish review`. Pinned in `test_review_set_walker.py` (off-set, mid-set, last, tombstoned tail).
3. **Fixed** (strip stranded by Reader focus) — strip visibility is now separate from list-key ownership: `_library_open_choice_strip()` asks the visibility-only question (`require_focus=False`) for Escape and the footer, while the ordinary list bindings keep the focus gate; the Reader's chip reads `close` while a strip is open. Pinned in `test_library_media_render_fixes.py` (open the sort strip, focus the content body, Escape closes the strip).

PR #2402, opened 2026-09-05
Screen-instance reuse mechanism via installed screens (TASK-24452)

rmusser01:
## Summary

Owner call from the 2026-09-04 perf review: tackle task-24452 (screens re-minted from scratch on every visit — the biggest interactive lever in the app, root cause behind 24456/24457 as well).

**Mechanism (this PR):** `ScreenRoute.reusable` — an opt-in, per-route flag. A reusable route's screen is constructed once, **installed** (`App.install_screen`), suspended instead of unmounted on switch-away, and resumed on return. Installed screens are never torn down mid-session, so the root-caused 2026-07-11 freeze (re-mounting an instance whose teardown was still in flight) structurally cannot start. The instance cache is scoped to `RuntimeIdentity` (local↔server flips invalidate, guard-tested), and snapshot-restore is skipped on cache hits — the live instance is the state.

**Enabled here:** Home (no unmount teardown, no timers). Its per-visit dashboard refresh moved to `on_screen_resume` (mutation-tested: deleting the hook fails a guard).

**Deliberately NOT enabled here:** Console and Chat/Library — their `on_unmount` teardowns (a dozen subsystems on Console: timers, dictation, realtime, fleet ticks; six controllers' remount-invalidation on Library) must become suspend/resume-aware first or they'd keep running against a hidden screen. Filed as **task-31520** (Console) and **task-31521** (Library) with the audit checklists.

## Measurements (2026-09-04, interleaved arms, loaded machine)

- Installed-instance warm switches (bypass probe): **Chat 750–820 → 158 ms CPU (−80%); Library 340–1,540 → 33 ms (−95%)** — this is the headroom 31520/31521 collect.
- Typing CPU/key with a suspended screen mounted: 30.98 → 32.94 ms — **no regression** (the per-keystroke concern about resident widgets is empirically dismissed).
- Honest null result: Home arrivals through the real nav handler show **no** win — the arrival window is dominated by the *outgoing* screen's teardown, and Home was always cheap to build. Home is the proving ground for the mechanism, not the payoff.

## Verification

- `Tests/UI/test_screen_reuse.py` — 5 guard tests (same-instance resume + no re-mint, resume-refresh, identity invalidation, opt-in boundary pinning chat as fresh); the route flag and resume hook were each mutation-tested red
- home / dashboard-seams / triage-rail / master-shell / nav-failure-recovery suites: **142 passed**
- `test_screen_navigation.py`: 31 failed on this branch AND 31 failed on pristine `origin/dev` in the same run mode — failure-set diff is 3 names, each re-verified as pre-existing flakes failing solo on base too
- perf-guard suite (latency guardrails + boot ratchets + stall persist): 26 passed
- preflight: all derived-artifact checks passed; 3 new static diagnostics reviewed via `--statements` before `--write`; ruff error count identical to base (133 pre-existing)

Task-24452's ACs re-scoped in the task file per the workflow rule (mechanism ACs met here; Console/Library re-mint elimination tracked in 31520/31521).

https://claude.ai/code/session_01JZ7TGpcXMtHqAwR34fNczp

---
## Summary by cubic
Adds opt-in screen-instance reuse for TASK-24452: routes flagged `ScreenRoute.reusable` build their screen once, install it, and resume the same instance on return instead of unmounting and re-minting a fresh widget tree per visit. Installed screens are never torn down mid-session, which also structurally prevents the 2026-07-11 freeze class.

**Refactors**
- `ScreenRoute.reusable` defaults to `False`; only Home opts in, so every other route keeps its current fresh-instance lifecycle.
- The instance cache is scoped to `RuntimeIdentity`; a local↔server flip drops the cached screen, and cache hits skip snapshot-restore because the live instance is the state.
- Home's per-visit dashboard refresh moved to `on_screen_resume` as the only dispatch seam, since Textual posts `ScreenResume` on the initial push and dispatching from both `on_mount` and resume double-ran it.
- Console and Library enablement is deferred to TASK-31520 and TASK-31521 until their `on_unmount` teardowns become suspend/resume-aware; bypass probes measured −80% and −95% switch CPU there, while Home shows no win through the real nav path.
- A suspend-time Buddy mouse-capture release was tried and retired after dev's #2407 moved the overlay to an app-level owner.
- `Tests/UI/test_screen_reuse.py` (5 guard tests) and `Tests/UI/test_screen_reuse_helpers.py` (6 helper-unit contracts) cover reuse, identity invalidation, the opt-in boundary, and disposal fallbacks.

<sup>Written for commit 81be69bf0097f291595ba884d097b4d0a50ba9e2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2402?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/security/production-diagnostic-inventory.json
+++ b/Docs/security/production-diagnostic-inventory.json
@@ -3974,8 +3974,8 @@
       "reason": "remaining Chatbook production diagnostic owner"
     },
     {
-      "call_count": 378,
-      "diagnostic_digest": "6925a6612538de10d8a0",
+      "call_count": 381,
+      "diagnostic_digest": "cf3e8bfa48108a6a2a97",
       "owner": "TASK-494",
       "path": "tldw_chatbook/app.py",
       "reason": "remaining Chatbook production diagnostic owner"
@@ -11360,6 +11360,6 @@
     "path_privacy_candidate_calls": 715,
     "persistent_sink_files": 10,
     "task_492_calls": 1336,
-    "task_494_calls": 7599
+    "task_494_calls": 7602
   }
 }
--- a/Tests/UI/test_screen_reuse.py
+++ b/Tests/UI/test_screen_reuse.py
@@ -1,0 +1,230 @@
+"""Screen-instance reuse contracts for reusable routes (TASK-24452).
+
+The app hands ``switch_screen`` a fresh instance on every navigation --
+except routes whose ``ScreenRoute.reusable`` flag is set, whose screen is
+constructed once, installed, and re-switched to on later visits (Textual
+suspends installed screens instead of unmounting them). These tests pin:
+
+1. the reuse itself (same instance across a leave-and-return cycle, no
+   widget re-mint) -- the warm-visit guarantee the flag exists for;
+2. per-visit dashboard refresh moving to ``on_screen_resume``;
+3. runtime-identity scoping (a local<->server flip must not resume the
+   other identity's live widget state);
+4. the opt-in boundary: a route WITHOUT the flag keeps fresh instances,
+   because reuse changes on_mount/on_unmount frequency and every route
+   must opt in only after auditing that (see the flag's docstring).
+"""
+
+from __future__ import annotations
+
+import asyncio
+from pathlib import Path
+
+import pytest
+
+from tldw_chatbook.UI.Navigation.screen_registry import resolve_screen_route
+
+
+def _scratch_env(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> Path:
+    """Point every config/data seam at a scratch tree with setup completed."""
+    home = tmp_path / "home"
+    data = tmp_path / "data"
+    config = tmp_path / "config"
+    for sub in (home, data, config):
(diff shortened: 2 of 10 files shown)


rmusser01:
Re: Qodo's remaining rule-violation (isolated unit tests for the reuse helpers) — addressed in `Tests/UI/test_screen_reuse_helpers.py` (same push the re-review ran on): six unit tests invoke `_reusable_navigation_screen` / `_retain_reusable_navigation_screen` directly against a stub app — cache hit/miss, identity-mismatch disposal (including the bounded in-stack leak), install-failure fallback, and disposal-failure tolerance. No Textual boot involved. The finding's evidence quotes `test_screen_reuse.py` only; the new file appears not to have been linked to the finding.

PR #2409, opened 2026-09-05
Pay the boot-CSS byte ratchet and wire it into perf-guard (TASK-24459)

rmusser01:
## Summary

Owner call from the 2026-09-04 perf review: get the CSS byte guard into `perf-guard.yml`. That required paying the breach first (826,956 B vs the 806,000 B pin on pristine dev — the fourth silent re-breach in four review cycles).

**Paydown:** generalizes the TASK-25812 agentic split into `SCREEN_OWNED_SPLITS` — a spec table the build iterates — and adds two modules:
- `features/_evals.tcss`: 39,695 B of owner-pure rules → `screen_feature_evals.tcss` (EvalsScreen)
- `features/_scheduling.tcss`: the 7,936 B owner-pure half → `screen_feature_scheduling.tcss` (SchedulesWorkbench); helper classes (`pane-hidden`, `detail-value-row-*`, bare status classes) stay in the bundle by the conservative classifier

Wholesale module moves were audited and **rejected** (`TaskDetail` type selector collides with an Evals widget class; `.needs-attention` is composed by the Library notes canvas).

**Result:** boot census **826,956 → 779,365 B**; ratchet **TIGHTENED 806,000 → 804,000** (measured + standard slack, ADR-097 — lowering needs no ledger row); `test_boot_css_byte_budget.py` **joins perf-guard.yml**, so the next breach fails the PR that causes it, in minutes, with culprit segments named.

## The incident this PR minted (and fixed before merge)

The first wiring put the sheets on the screens' `CSS_PATH` — and the paired evals/schedules arm flipped three destination-shell geometry tests. Mechanism: Textual loads `CSS_PATH` under ANY app, including the unstyled-tier harnesses, so harness-mounted workbenches rendered with only the *moved half* of the module (the automation-detail overlay covered the follow button). Reworked to an app-owned seam (`TldwCli._SCREEN_OWNED_ROUTE_CSS` + `_ensure_screen_owned_css`, mirroring `App._load_screen_css`, called from navigation and initial-push). The production-CSS harnesses' hand-named sheet list — stale once already when 25812 landed — now derives from `APP_STYLESHEETS`. Lesson recorded in `lessons-testing-evidence.md` with the incident.

## Safety & verification

- Build-time: per-module demotion pass (later-module seeding), per-module variables preamble, cross-split moved-selector disjointness backstop (unreachable today via demotion — documented and tested at its own level); measured **zero** selector collisions for both modules against all later bundle modules and generated sheets
- Guards: exact-rebuild contracts for the new sheets (mutation-tested red on a hand-edited sheet), route-map completeness + `CSS_PATH` ban on owning screens, functional boot-absent/visit-present test on a real app boot
- Suites: css-build integrity 26 passed; byte budget green; bundle-sync green with new sheets enrolled; full perf-guard set 26 passed (destination tour includes Schedules); **paired arms** on the evals/schedules UI selection (~950 tests each) — after the rework the four affected files show an **identical 16-failure set to pristine dev (zero divergence; all pre-existing)**
- preflight: all derived-artifact checks passed; the one new diagnostic (static route-id warning) reviewed via `--statements` before `--write`

Closes task-24459; also completes the paydown half of task-31500 (filed on PR #2399's branch).

https://claude.ai/code/session_01JZ7TGpcXMtHqAwR34fNczp

---
## Summary by cubic
Pays the boot-CSS byte ratchet breach and wires the byte-budget test into `perf-guard.yml`, so future breaches fail their own PR in minutes instead of silently re-breaching dev.

Generalizes the TASK-25812 agentic CSS split into a spec table the build iterates over, moving evals and scheduling rules off the boot bundle.

- Evals rules (39,695 B) move to `screen_feature_evals.tcss`; scheduling rules (7,936 B) to `screen_feature_scheduling.tcss`.
- Boot census drops from a breached 826,956 B to 779,365 B; ratchet tightens from 806,000 to 804,000.
- `test_boot_css_byte_budget.py` now runs in `perf-guard.yml`; the next breach fails the PR that causes it.

**Bug Fixes**

- Feature sheets load via a `TldwCli` app seam on first navigation, not `Screen.CSS_PATH` — `CSS_PATH` loads under every app, including unstyled-tier harnesses, and styled harness-mounted workbenches with only the moved half of a module.
- The harness sheet list now derives from `APP_STYLESHEETS` instead of hand-named paths that went stale twice.
- Route-map keys use the `TAB_EVALS`/`TAB_SCHEDULES` constants, and an initial-tab boot test pins the loading call.
- A partial build whose split module left the tree now removes that module's stale generated sheet at publish.

<sup>Written for commit 180c533eb195cf5239ee8c274c84262546863c92. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2409?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Re: Qodo's remaining rule-violation (initial-tab CSS load untested) — addressed in the same push: `test_schedules_as_initial_tab_loads_the_screen_owned_sheet` in `Tests/UI/test_css_build_integrity.py` boots with `[general] default_tab = "schedules"` through the real `_push_initial_screen` boundary and asserts the sheet is loaded at `_ui_ready`. It was mutation-tested: deleting the `_ensure_screen_owned_css(resolved_tab)` call from `_push_initial_screen` fails it. The finding's evidence cites the ctrl+7 navigation test only.

PR #2453, opened 2026-09-06
fix(console): Inspect rail critique burn-down — state honesty, focus, density, affordances (31660-31665)

rmusser01:
Burns down the 2026-09-05 sr-UX critique of the Console Inspect rail (scored 18/40): six tasks, each through an independent review + fix loop, plus a whole-branch final review with its fix wave merged.

**Tasks:** TASK-31660 (UNBOUND/PENDING/UNKNOWN state honesty — no more false "No git workspace", no stale-root data with a Commit action after a workspace switch) · 31661 (10s poll no longer steals rail focus; capture/restore with outside-rail guard + visible-toggle fallback) · 31662 (single-line rows: Environment at rest 11→7 lines; real-width budgets — the rail is 30 cols at 80×24, not 34) · 31663 (focus carriers on every rail stop, compact send-authority density: body 3→7 rows at 80×24, scrollbar thumbs, Tab-scoping adjudicated by-design per TASK-2154.11/23198) · 31664 (▸/…/+ marker convention, "Review & commit…" rename, Refresh acknowledgment with pending-tier accounting, "(stale)" text carrier + readable error hue) · 31665 (banding investigation — the critique's background-bleed mechanism REFUTED by compositor measurement: the real defect was a double-dimmed foreground, fixed at the CSS layer — plus a 15-AC minors batch incl. frontmatter task titles, fleet focus restore, cell_len, five more scrollbar thumbs).

**Verification:** ~450 arc-suite tests green incl. new compositor-measurement and deferred-fake modules; every behavioral fix RED-verified at its base; preflight green; pre-existing red clusters baseline-verified per change (paired arms). Ledger with every ruling and deferred item: arc workspace (untracked).

**Filed, not fixed:** TASK-31860 (stale CSS coverage contract, 157 false reds post-bundle-split); width-aware-budget follow-ups ledgered on 31665/31629.

https://claude.ai/code/session_01QPy9v1VfcCBbHWtaaodVj8

---
## Summary by cubic
Burns down the 2026-09-05 UX critique of the Console Inspect rail (18/40) across TASK-31660–31665: Environment state honesty, poll focus, row density, affordances, and a minor batch. The background-banding finding was refuted by compositor measurement; the real defect was a layered `dim` foreground, fixed at the CSS layer.

**Behavior**
- The Environment panel lands explicit UNBOUND, PENDING, and UNKNOWN states instead of asserting "No git workspace" or keeping the previous workspace's branch, counts, and PR data after a switch; a transient accessor failure renders UNKNOWN rather than a false claim.
- The 10s Environment poll no longer steals rail focus; focus restores to the same row by id, with a fallback chain ending at a visible toggle, and never overrides a user move made during the recompose.
- Refresh keeps its "Refreshing…" acknowledgment until the slow net tier lands, tracked per pending tier and scope so a stale landing from an abandoned workspace can't clear a later refresh's ack.
- Rows carry trailing markers (▸ expand, … opens another surface, + composer insert); "Commit or push" is renamed "Review & commit…".
- Change Review no longer flashes "No file changes recorded" while repo detection is in flight, and "1 files" now reads "1 file".
- Filed, not fixed: TASK-31860 (CSS coverage contract reads only the pre-split bundle) and TASK-31900 (13 rail tests fail identically on clean `origin/dev` worktrees).

**Density & visuals**
- Rows fit one line when possible, cutting the Environment section from 11 to 7 lines at 80×24; budgets use the measured 30-column rail width, not the assumed 34.
- The send-authority summary compacts to heading plus Run on short terminals, growing the scroll body from 3 to 7 rows.
- Every focusable stop has a visible focus carrier, and five more scrollbar thumbs moved to `$ds-text-muted`.
- ASCII-glyph rows measure resolved text so a mode flip re-shapes rows instead of ellipsizing them.
- A compositor sweep asserts every painted string in every inspector row clears 4.5:1 at both supported sizes.
- The focus-carrier rule was re-keyed to a `console-rail-focus-carrier` class to clear the ADR-097 census, and CSS prose was compressed to bring the boot bundle back under its 800KB budget.

<sup>Written for commit e5759f9c7417a5b757d527dd7bcb0929c7c973c5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2453?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
All six Qodo findings addressed in 457a350e0a (after syncing with dev, 82b41e5e26 — 472 commits, 5 conflicts, generated artifacts regenerated not hand-merged):

- **1-3 (docstrings)**: `compose()` Returns, `pending_ack_tiers` Returns, `on_resize` Args — all added per the Google-style rule.
- **4 (ASCII fallback breaks row budgets)**: confirmed — the row measured unresolved glyphs while rendering resolved ones (`✓`→`[x]` = +2 cells could one-line a 26/27-cell row and ellipsize it). Both the one-line decision and the structural key now measure the glyph-resolved text, so an ascii-mode flip also recomposes correctly. RED-first test added.
- **5 (stale refresh clears newer acknowledgment)**: confirmed — the pending-ack set was unkeyed, so workspace A's stale landing could clear workspace B's live "Refreshing…". The set now records its owning scope at both write sites (including the unbound pre-seed) and the stale-scope path only clears when the dropped landing's scope owns the ack. RED-first deferred-fake test added.
- **6 (rail navigation focus overridden)**: confirmed — the restore only yielded to focus that left the rail, so an in-rail Tab/click during the recompose window was overridden. The suggested allowlist (None/body/origin-toggle) turned out to be wrong in testing — Textual's reset can land on a *different* section's control — so the shipped discriminator is: yield whenever focus sits on any row (a row-focus during recompose can only be a user move; resets never land on rows). The residual in-rail-chrome gap is documented rather than papered over. RED-first test added.

258 arc tests green, preflight green. Also noted while baselining: two failures inherited from current dev (`test_console_rail_handle` geometry, `inspector_compact_access` rail-display) reproduce identically on origin/dev — filed separately, not from this branch.

PR #2467, opened 2026-09-06
feat(console): More-menu summarize / transcript-up-to-here note actions (TASK-31759)

rmusser01:
## Summary

Adds two per-message actions to the Console message **More…** menu (TASK-31759):

- **Summarize up to here as note** — summarizes the active-path span up to and including the selected message (USER or ASSISTANT target) into a note, via a new `console.summarize_note` internal prompt.
- **Save transcript up to here as note** — saves that span as a formatted Markdown transcript note (role-prefixed turns + provenance header).

## Key design points

- **Stateless summarize seam**: new `ConsoleCompactionService.summarize_span_to_text` wraps the existing bounded `_summary_completion` without the attempt ledger, branch-memory admission, or boundary commit that `summarize_manual` always performs. The `/rewind` context summary and boundary are provably untouched (tests assert zero `console_auxiliary_attempts` rows and unchanged `session_context_summary`).
- **Oversized spans block** with a user-visible notice (matching the manual summarize contract) instead of silently trimming oldest turns.
- Notes are written via `notes_scope_service.save_note` (`scope=LOCAL_NOTE`, `keywords=["console"]`), same seam as save-as Note, in a dedicated `console-note-actions` exclusive worker group — never `console-run`, so it cannot cancel a live stream. Mid-run attempts surface the active-run rejection as a notice.
- Overflow-only entries in `ConsoleMessageActionService`; dispatched via the pre-`dispatch()` interception pattern used by `save-as`, plus the two new prefixes in `_parse_console_message_action_button_id`.

## Tests

- New `Tests/Chat/test_console_note_span_actions.py` (10 tests): span inclusivity/tool-skip, note formatting + provenance, provider-not-ready / oversized-span / off-path gates, statelessness, UI dispatch wiring, mid-run notice.
- Updated exact-list expectations in `Tests/Chat/test_console_message_actions.py`.
- Targeted regression runs green: rewind-summarize, context compaction, run gate, prompts panel, chat-flow action subset (~365 passed total). `test_console_native_transcript.py` has 31 failures identical on clean `origin/dev` (verified via stash baseline) — pre-existing, unrelated.

Task: TASK-31759 (marked Done, ADR not required — no schema/boundary changes).

---
## Summary by cubic
Adds "Summarize up to here as note" and "Save transcript up to here as note" to the Console More-menu, turning the active-path conversation up to the selected message into a note without touching compaction state.

- Summarize uses the `console.summarize_note` prompt via a stateless provider call that never writes attempt-ledger rows or moves the rewind boundary.
- Save transcript writes a role-prefixed Markdown transcript with a provenance header, inclusive of the selected message, skipping tool, failed, and empty rows.
- Both actions run in an exclusive `console-note-actions` worker group, so they never cancel a live stream; mid-run attempts show a notice.
- Oversized summarize spans are blocked with a visible notice; the budget check sees the untrimmed span so notes never silently omit turns.
- Notes are saved through `notes_scope_service.save_note` (`keywords=["console"]`, under the configured notes owner) so they appear in the library.
- New tests in `Tests/Chat/test_console_note_span_actions.py` cover span inclusivity, formatting, gates, the stateless contract, and end-to-end note persistence.
- Pins the production diagnostic inventory for the new fixed-string warning on note-action failure.

<sup>Written for commit 46328592b2607a72ed5dee7747bc8f38000ebe9b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2467?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/security/production-diagnostic-inventory.json
+++ b/Docs/security/production-diagnostic-inventory.json
@@ -2535,8 +2535,8 @@
       "reason": "remaining Chatbook production diagnostic owner"
     },
     {
-      "call_count": 13,
-      "diagnostic_digest": "dbc82a289e52f7a3fbb9",
+      "call_count": 14,
+      "diagnostic_digest": "6ffafd40c155704dfbd2",
       "owner": "TASK-494",
       "path": "tldw_chatbook/UI/Console_Modules/message.py",
       "reason": "remaining Chatbook production diagnostic owner"
@@ -11462,6 +11462,6 @@
     "persistent_sink_files": 12,
     "task_31551_calls": 35,
     "task_492_calls": 1347,
-    "task_494_calls": 7668
+    "task_494_calls": 7669
   }
 }
--- a/Tests/Chat/test_console_message_actions.py
+++ b/Tests/Chat/test_console_message_actions.py
@@ -121,6 +121,8 @@ def test_assistant_message_actions_include_required_order():
         "--->",
         "Feedback",
         "🗑",
+        "Summarize up to here as note",
+        "Save transcript up to here as note",
     ]
 
 
@@ -164,6 +166,8 @@ def test_streaming_assistant_message_shows_completed_actions_disabled_with_reaso
         "--->",
         "Feedback",
         "🗑",
+        "Summarize up to here as note",
+        "Save transcript up to here as note",
     ]
     assert all(action.enabled is False for action in actions)
     assert all(action.disabled_reason for action in actions)
@@ -192,6 +196,8 @@ def test_pending_assistant_message_shows_completed_actions_disabled_with_reasons
         "--->",
         "Feedback",
         "🗑",
+        "Summarize up to here as note",
+        "Save transcript up to here as note",
     ]
     assert all(action.enabled is False for action in actions)
     assert all(action.disabled_reason for action in actions)
@@ -279,6 +285,8 @@ def test_variant_action_labels_use_symbolic_navigation():
         "--->",
         "Feedback",
         "🗑",
+        "Summarize up to here as note",
+        "Save transcript up to here as note",
     ]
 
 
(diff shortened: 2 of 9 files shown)


rmusser01:
**Qodo review addressed** (2267ccef6):

1. *Users wait without summary progress* — **won't fix.** The summarize-to-note call deliberately rides the same bounded, non-streaming auxiliary completion as compaction (consistency + the aux-model routing/timeout guarantees). The action is a fire-and-complete side task with a transient notice on both success and failure, consistent with save-as; streaming a one-shot note write would add a UI surface for no persistence benefit.

2. *Public note APIs lack contracts* — **fixed.** Google-style `Args:`/`Returns:` (and `Raises:` for the stateless seam's deliberate propagation) added to `summarize_span_to_text`, `build_transcript_note`, `summarize_span_as_note`, and `_note_span_messages`.

3. *Note saving can regress unnoticed* — **fixed.** New parametrized test drives each action's worker through `notes_scope_service.save_note` with a stubbed service and asserts title, content, `scope=local_note`, `keywords=["console"]`, and the configured notes identity.

4. *Saved notes vanish from the library* — **valid, fixed.** Now saves under `app.notes_user_id` (the owner local note views/ingest use), with `default_user` only as an unavailable fallback. (Note: the pre-existing `_save_console_message_as_note` save-as path still uses `current_user` — same latent issue, but out of this PR's scope; happy to follow up.)

5. *Older turns vanish from summaries* — **valid, fixed.** Real bug: the budget check ran after `_build_summary_span_text`'s silent oldest-turn trimming, so it could never fire at the real budget. The span is now counted (and sent) untrimmed; a new test blocks a genuinely >12k-token span at the default budget and asserts zero provider calls.

PR #2552, opened 2026-09-09
feat(library-notes): Info lists the notes that link to the open one (task 32145)

rmusser01:
## Summary

Riders wave of the Library ▸ Notes critique (plan: `Docs/superpowers/plans/2026-09-09-library-notes-riders-wave.md`, stacked on #2549). Built on the controller's recorded ruling on the user's delegation of the four idea tasks (this one and the Trash view approved; the editor chrome strip and capture-from-Console deferred).

- Info → Properties shows **Linked from (N)** with one activatable row per note whose body links to the open note through the `note://<id>` form the Obsidian importer writes; empty reads `Linked from (0) — no notes link here yet`.
- The query is a parameterised `LIKE` with wildcard escaping, `deleted = 0`, ordered by title, bounded to 50 (a `50+` header past that); FTS5 was rejected because its tokenizer splits the URI. Backlinks load on their own exclusive worker after the note detail, guarded by the selected-note id, and paint through the existing sync path so a refresh deferred while the editor has focus still lands.
- Activating a row opens that note through the same guarded load path as a list row.

Tests: 5 real-DB query tests (`Tests/Notes/test_note_backlink_query.py`) and 5 UI tests (`Tests/UI/test_library_notes_riders_backlinks.py`), RED first; neighbouring suites at their base failing-name sets. Live-verified on a fresh profile after a real vault import (`Linked from (2)`, activation opens the linked note). Reviewed (spec + quality), approved.

Deferred minors (wave ledger): a failed load reads as "no notes link here yet"; activating a backlink lands in Edit rather than Info; the full-scan ceiling is unmarked; the test fake folds case with `lower()` while SQL uses `COLLATE NOCASE`.

---
## Summary by cubic
Library ▸ Notes Info → Properties now lists the notes that link to the open note (task-32145).

- Shows "Linked from (N)" with one activatable row per note whose body carries the `note://<id>` link form the Obsidian importer writes; the count is only claimed once the query has answered — pending reads "checking…", failure "couldn't check", and "no notes link here yet" only for a verified zero.
- The query is a parameterized `LIKE` with wildcard escaping, excludes soft-deleted notes and the target itself, orders by title, and caps at 50 rows ("50+" past that); FTS5 was rejected because its tokenizer splits the URI.
- Backlinks load on their own worker after the note detail, guarded by the selected-note id, and paint through the existing sync path so a deferred refresh still lands; activating a row opens that note through the same guarded load path as a list row.
- Non-local scopes refuse the lookup instead of answering zero, so a server or workspace note can never claim "no notes link here yet" unchecked.
- Adds 6 real-DB query tests and 7 UI tests. The full-scan lookup cost at vault scale is tracked as task-32186.

<sup>Written for commit eaf8a77f2675c008193e86e0bdf25c029fbfa312. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2552?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Review round on 7216d8c1f0 — every Qodo finding adjudicated.

**1. Backlink limits can drift apart (magic `50`) — rebutted.** The only production caller passes the cap explicitly: `library_notes_controller.py` asks for `LIBRARY_NOTE_BACKLINK_DISPLAY_CAP + 1`, and `test_backlinks_are_requested_bounded_and_for_the_open_note` pins that the request is bounded by it. The `limit: int = 50` defaults in `CharactersRAGDB.get_notes_linking_to` / `NotesInteropService.get_notes_linking_to` / `NotesScopeService.list_note_backlinks` are a bound for a direct caller, not the display contract — changing the display cap cannot silently change what the UI shows. A new dependency-neutral module holding one integer (the DB layer must not import a widget) costs more than the drift it prevents.

**2. Handler lacks a docstring — fixed.** `LibraryScreen.handle_library_note_backlink` now has a Google-style docstring with `Args:`.

**3. Tests depend on disk storage — rebutted, with a run.** `CharactersRAGDB` opens **thread-local** connections (`self._local = threading.local()`, `_get_thread_connection`), so `":memory:"` gives every thread its own empty database. These tests deliberately cross a thread — `NotesScopeService.list_note_backlinks` runs the query in `asyncio.to_thread`. Probed directly:

```
db = CharactersRAGDB(":memory:", client_id="probe"); db.add_note(...)
main thread rows: []
worker thread: {'err': 'CharactersRAGDBError: Query execution failed: no such table: notes'}
```

The fixture's `tmp_path` file is a real SQLite database with the real schema running the real SQL, which is what the rule is about; `":memory:"` would make the test unable to run at all.

**4. Reads bypass `transaction()` — rebutted.** `execute_query` *is* the shared seam: it takes the same thread connection and, per its own contract, defers to an enclosing `with db.transaction():` when one is open (`commit and not conn.in_transaction`). 69 other read methods in `ChaChaNotes_DB.py` use exactly this `cursor = self.execute_query(...)` form. Wrapping a single read in a write-shaped transaction adds a BEGIN/COMMIT round-trip for nothing.

**5. Cursor lacks scoped cleanup — rebutted.** Same 69 sibling reads, same pattern. A `sqlite3.Cursor` holds no OS handle — the connection does, and it is thread-local and long-lived by design. `fetchall()` fully drains the cursor before it goes out of scope.

**6. Failed lookups look like zero links — fixed, this one was real.** State gained `backlinks_status` (`"loading"` → `"ready"` / `"failed"`); the header now reads:

- `Linked from — checking…` while the query is in flight,
- `Linked from — couldn't check` when it raised, or there was no service to ask,
- `Linked from (0) — no notes link here yet` **only** after a query actually answered zero.

Pinned by two new tests in `Tests/UI/test_library_notes_riders_backlinks.py`: `test_the_header_only_claims_zero_once_the_query_has_answered` (the copy for all three states) and `test_a_failed_backlink_lookup_does_not_claim_there_are_none` (real screen, service raising). Verified load-bearing — with the `failed` status forced back to `ready` the second test fails on exactly that assertion. The stale-note guard now runs before the status is written, so a late failure for a note you already navigated away from cannot mark the current note's panel failed. Wiring census 101 → 102; docs updated with the two new lines.

**7. Opening a note scans every note body — deferred as task-32186.** Real at vault scale and correctly diagnosed: `%(note://id)%` cannot use an index and `ORDER BY title` means the `LIMIT` does not bound the work. The fix is a different shape from this PR — a persisted source→target link relation written where wikilinks are already parsed, plus a schema migration and a backfill. Filed with acceptance criteria covering parity with the current query, edit/import/delete freshness, and existing databases picking up backlinks without a re-import.

**Tests run** (venv python, foreground):
- `Tests/UI/test_library_notes_riders_backlinks.py Tests/Architecture/test_library_notes_wiring.py Tests/Notes/test_note_backlink_query.py` — **27 passed**.
- `Tests/UI/test_destination_shells.py Tests/Library/test_library_notes_state.py Tests/Notes/test_note_backlink_query.py` — 5 failed, 167 passed, 1 skipped. All 5 failures are pre-existing on the base branch: the same 5 fail on `fix/library-notes-docs` @ 22b1fbe27c without any of this PR's code (`test_library_destination_service_failure_uses_recovery_copy`, `test_library_policy_denial_uses_runtime_recovery_taxonomy`, `test_library_policy_denial_uses_recovery_state_selector`, `test_destination_action_buttons_explain_their_outcome[schedules]`, `test_models_shell_keeps_external_paths_inside_the_dedicated_edit_view`).

rmusser01:
Landing round on `4abf9522b7` — retargeted to `dev` and merged `dev` in.

**Base.** `fix/library-notes-docs` merged as `e6cb464239`, so the PR base is now `dev`; `git merge origin/dev` was **clean** (no conflicts — dev already contains this branch's whole base). 0 behind, 4 ahead.

**Silent-bad-merge sweep.** No conflict markers; no duplicate `def`/`class` or `CONST =` in any touched `.py`; every touched `.py` AST-parses. The duplicate `id="…"` set across `Widgets/Library/*.py` is byte-identical to `origin/dev`'s (cross-file, pre-existing). Two-dot `git diff origin/dev HEAD --stat` is identical to the three-dot diff — 14 files, +853/-8 — so the merge dropped no dev content; both "deletions" are replaced lines (the Info row in `notes.md`, the census constant).

**Wiring census.** dev's pin is 100; this branch adds exactly `backlinks` and `backlinks_status`, so the pin reads **102** with the arithmetic written at the constant. (No Trash on dev yet, so 103 would be wrong here.) `test_library_notes_wiring.py` passes against the real dataclass.

**Comments.** Nothing posted since the adjudication comment of 2026-09-09T19:16:36Z — the six Qodo review comments all predate it and are answered there (2 fixed, 4 rebutted, 1 deferred as task-32186). One change landed after it: `76ba643ce1` makes `list_note_backlinks` raise for a non-local scope instead of returning `[]` (which Info would have rendered as an unverified "(0) — no notes link here yet"), pinned by `test_backlinks_refuse_every_scope_but_local`. The harness fake in `test_destination_shells.py` still answers any scope, which is covered: `test_backlinks_are_requested_bounded_and_for_the_open_note` pins that production only ever asks `scope="local_note"`.

**Tests** (venv python, foreground, post-merge):
- `Tests/UI/test_library_notes_riders_backlinks.py Tests/Notes/test_note_backlink_query.py Tests/Architecture/test_library_notes_wiring.py` — **28 passed**.
- `Tests/UI/test_destination_shells.py` (this PR adds the fake's `list_note_backlinks`) — 5 failed, 110 passed, 1 skipped. The same 5 node ids fail on a detached `origin/dev` @ `e6cb464239` with none of this PR's code — identical name set, so they are dev-side reds, not this PR's: `test_library_destination_service_failure_uses_recovery_copy`, `test_library_policy_denial_uses_runtime_recovery_taxonomy`, `test_library_policy_denial_uses_recovery_state_selector`, `test_destination_action_buttons_explain_their_outcome[schedules]`, `test_models_shell_keeps_external_paths_inside_the_dedicated_edit_view`.

rmusser01:
CI follow-up on `eaf8a77f26` — the required check failed on "Production diagnostic inventory matches the source tree", not on anything behavioural.

**Cause.** `library_notes_controller.py` went 8 → 9 diagnostic calls; the pinned `Docs/security/production-diagnostic-inventory.json` was not regenerated. The one added statement, reviewed before re-pinning:

```
logger.opt(exception=True).debug(
    "library_note_backlinks_failed", note_id=note_id
)
```

A constant message key plus the note's opaque database id — no note body, title, filesystem path, URL or secret reaches the sink. (`--statements … --since 0a3814746e` reports exactly one added call, zero removed, zero moved.)

**Fix.** Regenerated with `--write`; `scripts/check_persistent_diagnostic_inventory.py` now exits 0 (`7652 TASK-494 calls`). Then merged `origin/dev` again (dev had advanced to `3315241674`). The only conflict was that generated JSON, in the totals line alone — resolved the only correct way for a generated pin: take dev's file, regenerate from the merged source tree. The result differs from dev's by exactly the intended delta and nothing else:

```
"call_count": 8 → 9, "diagnostic_digest": 2bbd93d2… → c775c011…   (library_notes_controller.py)
"task_494_calls": 7651 → 7652                                     (dev's 7651 + this PR's 1)
```

Sweep re-run after the second merge: no conflict markers, no duplicate `def`/`class`/`CONST`, every touched `.py` AST-parses, two-dot `git diff origin/dev HEAD --stat` identical to three-dot (15 files, +856/-11). 0 behind dev.

**Tests.** `test_library_notes_riders_backlinks.py` + `test_note_backlink_query.py` + `test_library_notes_wiring.py` — **28 passed** post-merge. `Tests/Architecture/test_persistent_diagnostic_inventory.py` — 3 failed, 66 passed, 1 skipped; the same 3 fail on a detached `origin/dev` with none of this PR's code (`test_reviewed_diagnostic_changes_are_metadata_only`, whose first complaint is `library_screen.py: expected one diagnostic containing 'Failed to restore a Library note', found 0` — a Trash diagnostic that is not on dev yet — plus both `test_inventory_excludes_nested_virtualenv_but_keeps_application_sources` params). Dev-side reds, unrelated to backlinks.

PR #2555, opened 2026-09-09
fix(library-notes): editor leftovers — one back-cue rule, dead status widget removed, public timestamp helper; duplicate screen method deleted (tasks 32177, 32179)

rmusser01:
## Summary

- **32177** — the New-note and load-retry views now use the same back-cue rule as Edit, Preview and Info (`‹ Notes` at wide sizes, `‹ Back to list` on compact) through the existing helper; the `#library-note-context-status` widget that was composed and then permanently hidden is removed with its CSS rule and the rebuilt bundles, and its two geometry pins in the shell suite are reconciled by node id with the reason recorded; the browser-timestamp parser is public and imported cleanly across packages.
- **32179** — the dead first definition of `LibraryScreen._seed_local_source_snapshot_from_cache` is deleted; an AST-based test pins exactly one definition with the `now` keyword the two call sites need.

Tests: new `Tests/UI/test_library_notes_riders_r_editor.py`; 133 passed across the covering files; the two reconciled geometry pins remain at their pre-existing failing set (the stale flat-row selector, tracked by rider 32175). CSS bundle reproduces from source. Reviewed (spec + quality), approved with no findings.

---
## Summary by cubic
Fixes Library Notes editor leftovers (tasks 32177, 32179): the New-note and load-retry views now use the same back-cue wording as Edit, Preview, and Info, the dead `#library-note-context-status` widget is removed instead of hidden, and the timestamp parser is public. Back cues now also follow the compact breakpoint after a resize, not just at compose time.

- Both views pick their wording via `_library_note_back_label(self.compact)`; `apply_compact_presentation` rewrites the two buttons in place so a resize keeps the correct label.
- `_parse_browser_timestamp` becomes public `parse_browser_timestamp` with Args/Returns sections; the cross-package import in `library_notes_state` uses the public name.
- The duplicate `LibraryScreen._seed_local_source_snapshot_from_cache` is deleted; the surviving definition carries the former docstring's rationale, and an AST test pins exactly one definition with the `now` parameter.
- CSS rules, bundles, and geometry pins are updated for the removed widget; new tests cover back-cue behavior at both widths and across resizes.

<sup>Written for commit 4acefc1e459a3a72171e7926121626608950e121. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2555?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## Review round — both Qodo findings were real, both fixed (11ae25ef48)

**1. "Back cues stay wrong after resizing" (Bug / Correctness) — FIXED.**
Confirmed by tracing, not assumed: `apply_compact_presentation` is the only
thing that runs when the canvas crosses the breakpoint
(`library_notes_controller.py:2229`, guarded by `notes_canvas.compact !=
self._library_notes_compact`), it never recomposes, and it touched neither
`#library-note-back` nor `#library-notes-create-back`. So this PR's
unification only held for the width the view was *opened* at. It now
rewrites both labels in place. The editor's own pair is still rewritten from
the snapshot in `apply_session_state` (which calls
`apply_compact_presentation` first, so the values agree either way).

Tests: `test_new_note_view_back_cue_follows_the_compact_breakpoint` and
`test_load_retry_back_cue_follows_the_compact_breakpoint`
(`Tests/UI/test_library_notes_riders_r_editor.py`) open the view wide, shrink
to 60x24, and widen back — asserting the wording at each step. Both confirmed
RED with the fix removed and GREEN with it.

**2. "Timestamp parser lacks Google sections" (Rule violation) — FIXED.**
`parse_browser_timestamp` now documents `value` (ISO-8601 with `+00:00`, a
trailing `Z`, or no offset) and its `datetime | None` return, including the
naive-defaults-to-UTC rule.

### Also this round — checking the removals rather than the findings

Per the "this is cleanup, so verify the dead things were really dead" brief:

- `#library-note-context-status`: verified it has **no** remaining reference
  anywhere at HEAD — Python, all four `.tcss` tiers, tests, geometry pins —
  and that it had been permanently `display = False` since task-32142.
  Removal is safe.
- `_seed_local_source_snapshot_from_cache`: verified both definitions really
  did sit in the same class body at the base commit, so Python kept only the
  second and the first was dead; the survivor's `now` is keyword-only with a
  default, so both no-arg call sites still work.
- One thing that removal *did* cost: the deleted (dead) definition carried
  the long docstring that **two call-site comments still point at** —
  `__init__`'s "see `_seed_local_source_snapshot_from_cache`'s docstring for
  why this alone is not sufficient" and `restore_state`'s "(see the method's
  own docstring)". Those pointers resolved to a one-line summary. The
  pre-mount-safety rationale, plus `Args:`/`Returns:`, now lives on the
  surviving definition.

### Tests observed

| Suite | Result |
| --- | --- |
| `Tests/UI/test_library_notes_riders_r_editor.py` | 8 passed |
| `Tests/UI/test_library_notes_wave_editor_keys.py` | 27 passed |
| `Tests/Workspaces/test_console_conversation_browser_state.py` + `Tests/Library/test_library_notes_state.py` + `Tests/UI/test_css_build_integrity.py` | 100 passed |
| `test_library_entry_compose_once.py` (the two cache-seed node ids) | 2 passed |

The two `test_library_shell.py` geometry pins this PR reconciled remain
pre-existing baseline-broken, unchanged, as the PR body already records.

PR #2554, opened 2026-09-09
fix(library-notes): the Import once, Keep synced and Folder files pickers remember their last-used directory (task 32174)

rmusser01:
## Summary

Riders wave of the Library ▸ Notes critique (plan: `Docs/superpowers/plans/2026-09-09-library-notes-riders-wave.md`, stacked on #2549).

Rider from the pickers work (task 32122 left "last used" as a documented gap). Caller-side only, mirroring media ingest's `_library_ingest_browse_location`: each of the three pickers opens at its own last-used directory, else home, and persists through the same config-write path ingest uses. Import once remembers the selection's parent for a file and the folder itself for a folder (`[library.notes_import] last_directory`); Keep a folder synced remembers the folder (`[library.notes_sync] last_directory`); Folder files remembers the browsed folder while no root is linked or the root is offline (`[file_notes] browse`). A cancelled picker never overwrites; a remembered directory that no longer exists falls back to home. The vendored dialogs are untouched.

Tests: 8 new across `Tests/UI/test_library_notes_wave_import_ux.py` and `test_library_notes_wave_file_notes.py`, including key independence against the existing ingest key; the workspace file at its base failing-name set. Live-verified for all three pickers. Reviewed (spec + quality), approved.

Deferred minor (wave ledger): the Folder files persist path swallows a failed save silently where the two siblings log it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01KfNA9miaAhZwcqP6KFcKdx

---
## Summary by cubic
Makes Import once, Keep a folder synced, and Folder files each open their folder picker at the directory last used in that picker, instead of always starting at home.

**Behavior**
- Each picker persists its own last-used directory in `config.toml`: `[library.notes_import] last_directory`, `[library.notes_sync] last_directory`, and `[file_notes] browse`, independent of each other and of Library ingest's existing key.
- A cancelled picker never overwrites the remembered directory; a remembered directory that no longer exists falls back to home.
- Shared resolver/persist logic in `tldw_chatbook/Library/library_browse_location.py` validates stored values as trusted user state, reports failed config writes (no more silent swallow in Folder files), and orders writes per key so an older selection can't overwrite a newer one.
- Implemented caller-side without touching the vendored picker widgets; tests drive the production picker openers and their callbacks.
- The user guide (`notes.md` and `file-notes.md`) now documents the behavior and the exact config keys.

[...]

PR #2553, opened 2026-09-09
feat(library-notes): Recently deleted — a Trash view for Database Notes with Restore through the Undo seam (task 32144)

rmusser01:
## Summary

Riders wave of the Library ▸ Notes critique (plan: `Docs/superpowers/plans/2026-09-09-library-notes-riders-wave.md`, stacked on #2549). Built on the controller's recorded ruling on the user's delegation of the four idea tasks.

- A **Recently deleted (N)** row closes the folder tree (on every list path: empty list, flat list, empty tree, tree rows) and opens a Trash view listing soft-deleted notes newest first as `title · age`, each with **Restore**, plus `‹ Notes`. No permanent delete anywhere.
- Restore rebuilds the delete receipt and calls the receipt's own Undo seam (`_undo_library_note_delete` → `restore_note` with the tombstone's version), so the row returns to its folder or Unfiled and the rail count updates exactly as Undo does; that seam also reloads the Trash snapshot, so both entries stay truthful from one place.
- `r` restores the focused row (advertised only while rows exist); Escape returns to the list through the existing Notes escape gate. The DB query returns the page and the exact total in one transaction, paged at 20 (the view says so past that).

Tests: 14 in new `Tests/UI/test_library_notes_riders_trash.py` (including one that drives the real reconciler and projection to pin tree-row and count parity with Undo), 3 DB and 2 scope-service tests, RED first; neighbouring suites at their base failing-name sets. Live-verified on the seeded profile (delete two notes, open the view, restore one). Reviewed (spec + quality), approved.

Deferred minors (wave ledger): the snapshot is not refreshed on the version-conflict path; an in-flight guard can drop a refresh; an unfocused `r` presses the first row while the guide says the focused one; tombstones never expire and soft-deletes from other surfaces appear here (the pitched 30-day window was dropped in favour of honest copy).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01KfNA9miaAhZwcqP6KFcKdx

---
## Summary by cubic
Adds a Recently deleted (N) Trash view under the folder tree so a soft-deleted Database Note is recoverable even after its delete receipt is dismissed, plus a Linked from (N) section in note Info (tasks 32144 and 32145). Restore commits through the same Undo seam the receipt uses, and inbound links open their source notes.

- Trash Restore returns the note to its folder or Unfiled and keeps the rail and Trash counts current through the shared seam.
[...]

PR #2556, opened 2026-09-09
fix(library-notes): import copy edges, one non-importable set, the Update-existing abort; Obsidian follow-ups — fences, link count, Windows detection, aliases (tasks 32176, 32178)

rmusser01:
## Summary

- **32176** — a structured file with one bad record now names it (`Record 2 of 3 has no note content.`, CSV rows by spreadsheet line) instead of calling the whole file app configuration; `Skip all` / `Create all` say "on this page"; the non-importable classification set has one definition consumed by parsers, plan models and the review canvas; and the **Update existing** abort on an unchanged repeat is fixed at its root: the receipt ledger compared membership effects against the memberships the parser proposed rather than the ones the approved plan authorized, so the only review choice with `replace_content` and no `add_membership` failed. The guard now zips against the plan and stays strict in the tamper direction.
- **32178** — an unterminated code fence protects everything after it from wikilink rewriting; the completion line reports `N links resolved` (the receipt has no link column to migrate, so the count rides the projection); Windows discovery detects a vault and applies the same skips through its adapter seam; aliases are stored as `alias: <name>` keywords so they stay distinguishable from tags.

Tests: 10 new, RED first; 970 passed / 7 pre-existing across the 12 affected files; the "Update existing" defect re-driven live through the shipped buttons now reads `2 imported · 1 updated · 64 skipped · 0 failed`. Live-verified on a fresh profile (vault import with `54 links resolved`). Reviewed (spec + quality), approved.

Not done, by design: partial import of the valid records in a mixed structured file (the pipeline has no per-record warning surface; the file is refused with the record named instead).

Deferred minors (wave ledger): the alias-prefix branch is keyed off the non-alias key; an over-long alias is dropped silently; a tag/alias overlap spends two keyword slots; no negative-direction test on the new ledger guard; Windows parity is stated on adapter-seam evidence only.

---
## Summary by cubic
Fixes the Library Notes import review across tasks 32176 and 32178: bad records in a structured file are now named, `Update existing` works on an unchanged repeat, and Obsidian mode reports resolved links, detects vaults on Windows, and stores aliases distinctly.

**Task 32176 — import copy edges**
- A failed structured source names the offending record or CSV row, but the whole file is still refused.
- Group `Skip all` / `Create all` buttons now say "on this page" and settle exactly the rows that page's heading counts.
- The non-importable classification set lives once in `note_import_plan_models`; parsers, plan models, and the review canvas derive from it.
- `Update existing` on an unchanged repeat no longer aborts with no receipt; the ledger compares against the plan's authorized memberships.
- Fault-translation tests use a path-proof canary token instead of the word "private".

**Task 32178 — Obsidian follow-ups**
- The receipt line reports "N links resolved" only for links to notes created in the same batch and only on a clean completion.
- Both discovery adapters detect a vault case-insensitively and skip `.obsidian/`, `.trash/`, and `Templates/`; Windows now works like POSIX.
- Frontmatter `aliases:` entries are stored as `alias: <name>` keywords; an alias pushed over the keyword ceiling by the prefix is stored un-prefixed rather than dropped.
- The user guide's "Not on Windows" caveat is marked superseded.

<sup>Written for commit cb9c2e948385914c05e7aae32581b437b1713274. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2556?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## Review round — both Qodo findings fixed (129924d6eb)

**1. "Windows vault imports its private files" — FIXED.** Real, and wider than the
report: `_carries_obsidian_marker` compared the root entry to `.obsidian` with
`==` in *both* walkers, while the shared `_obsidian_skip_reason` looks the name up
in a map documented as "keyed by casefolded name". So the two halves of vault
handling disagreed on casing, and a marker that kept a different casing (Windows
and macOS preserve casing while comparing case-insensitively) left
`vault_detected` false and walked `.obsidian/`, `.trash/` and `Templates/` into
discovery. Both detectors now casefold, matching the map. Pinned by
`Tests/Notes/test_note_import_windows_fs.py::test_windows_adapter_detects_a_vault_whose_marker_keeps_its_casing`
and `Tests/Notes/test_note_import_obsidian.py::test_a_capitalized_marker_still_marks_the_vault`
(both fail on the pre-fix source, verified by reverting the source files and
re-running).

**2. "Long alternate names disappear on import" — FIXED.** Also real. `_keywords`
bounds every value at `MAX_IMPORT_KEYWORD_LENGTH`, then `_frontmatter_keywords`
added the seven-character `alias: ` prefix and dropped the result, so a 506–512
character alias — valid input, valid under the note ceiling — vanished. A display
prefix must never be the reason a name is lost, so when prefixing would overflow
the alias is now stored un-prefixed (still findable by keyword search) instead of
dropped. Pinned by
`Tests/Notes/test_note_import_obsidian.py::test_a_long_alias_keeps_its_name_when_the_prefix_would_not_fit`.
Did not take the "typed alias metadata" alternative from the summary — that is a
schema/search/UI change for one edge case. task-32178's notes were corrected: they
previously recorded the drop as the intended behaviour.

Also swept for the duplicate-constant hazard that bit this program earlier:
`NON_IMPORTABLE_CLASSIFICATIONS` has exactly one definition
(`note_import_plan_models.py:39`); the canvas derives its string set from it and
`test_library_notes_wave_import_ux.py:510` pins that derivation. No duplicate
top-level definition in any file this branch touches.

### Tests

- `Tests/Notes/test_note_import_obsidian.py Tests/Notes/test_note_import_windows_fs.py Tests/Notes/test_note_import_planner.py Tests/Notes/test_note_import_executor.py Tests/Library/test_library_note_import_state.py` → **638 passed, 7 failed, 3 skipped**
- `Tests/UI/test_library_note_import_flow.py Tests/UI/test_library_notes_wave_import_ux.py` → **29 passed**
- `ruff check` on every changed file → clean

The 7 failures are all
`test_note_import_executor.py::test_target_expected_faults_keep_their_item_level_translation`
and are **pre-existing and unrelated** — they reproduce identically with this
branch's base (`fix/library-notes-docs`) *and* with `origin/dev`'s copies of
`note_import_executor.py` + its test checked out in place. Nothing in this round
touches exception translation.

rmusser01:
## Retargeted onto `dev` and re-verified (3997ac0110)

Base moved from `fix/library-notes-docs` to `dev`; `origin/dev` merged in four
times as the wave landed ahead of this PR (#2552 backlinks, #2553 Trash, #2554
r-pickers, #2573 docs). **0 behind `dev`.** No rebase, no force-push.

**Conflicts, all resolved by hand:**
- `Docs/User_Guide/library/notes.md` (twice) — "Verified against" stamps only.
  Every stamp from both sides kept, each with its own blank line. Count checks
  out: 27 on `dev` + this branch's 2 = **29**. The F5 sentence still reads
  "superseded by task-32178 below", not `dev`'s pre-supersession wording.
- `backlog/docs/lessons-testing-evidence.md` — two new sections appended by
  both sides (`dev`'s pane-controls lesson, this branch's path-bearing-canary
  lesson). Both kept, neither unioned.
- `Tests/UI/test_library_notes_wave_import_ux.py` — both sides append their own
  test block at the file end. Both kept: base 13 + `dev`'s 7 (task-32174
  pickers) + this branch's 2 (task-32176) = **22 tests**, no duplicate `def`.

**Silent-bad-merge sweep clean:** no conflict markers anywhere; no duplicate
top-level `def`/`class` or `CONST =` in any changed file; the duplicate-widget-id
set in `Widgets/Library/` is byte-identical to `dev`'s; every changed `.py`
AST-parses. The two-dot `git diff origin/dev HEAD --stat` is identical to the
three-dot (18 files, +796/-108), so the merges dropped no `dev` content.

**Derived artifacts (the required check):** this branch adds, removes and
re-indents **no** `logger.*` / `notify` / `warnings` statement — `git diff
origin/dev HEAD -- tldw_chatbook/` matches none — so
`production-diagnostic-inventory.json` is byte-identical to `dev`'s.
`scripts/check_persistent_diagnostic_inventory.py` re-run **after every merge**
(it auto-merges silently, per #2553): exit 0, 595 owners / 7658 TASK-494 calls.
`tldw_chatbook/css/check_bundle_sync.py`: exit 0.

### Tests (on the pushed tree)

- `test_note_import_obsidian.py` `test_note_import_windows_fs.py`
  `test_note_import_planner.py` `test_library_notes_wave_import_ux.py`
  `test_library_note_import_flow.py` → **506 passed, 3 skipped**
- `test_note_import_executor.py` `test_library_note_import_state.py` →
  **175 passed** — including the seven
  `test_target_expected_faults_keep_their_item_level_translation` cases that
  failed on every `/private/tmp` checkout before 52603d952d made the leak canary
  path-proof (`_FAULT_DETAIL_CANARY = "zqleakcanary"`).
- `Tests/Architecture/test_library_notes_wiring.py` → 15 passed
  (`_EXPECTED_NOTES_STATE_FIELD_COUNT = 103` is correct here: this branch adds
  import-state fields, not notes-state fields).

### One red, and it is `dev`'s, not this PR's

`Tests/Architecture/test_persistent_diagnostic_inventory.py::test_reviewed_diagnostic_changes_are_metadata_only`
fails with

```
tldw_chatbook/UI/Library_Modules/library_notes_controller.py:
'Failed to restore a Library note' captures exception or stack details
```

That diagnostic is the Trash restore path from #2553. This PR does not touch
`library_notes_controller.py` — it is not in the diff at all. Reproduced on a
detached `origin/dev` worktree at `6a79651ee8` with the identical assertion at
`test_persistent_diagnostic_inventory.py:578`, so the FAILED name set here is
identical to `dev`'s. Note the pinned `--write` inventory is *not* what is
broken: `check_persistent_diagnostic_inventory.py` exits 0; the review-ledger
test is what rejects the new call. Worth a task against the Trash change rather
than a fix here.

Both Qodo findings from 17:54Z remain fixed (129924d6eb) and survived all four
merges: `_carries_obsidian_marker` casefolds in **both** walkers, and an alias
that would overflow only because of the `alias: ` prefix is stored un-prefixed
rather than dropped. Nothing new has been posted since the 19:11Z round-up.

PR #2557, opened 2026-09-09
fix(library-notes): Folder files keeps the Library rail before a folder is linked; the slow folder-change row fits at 60 columns (tasks 32173, 32180)

rmusser01:
## Summary

- **32173** — the rail was hidden before a folder was linked because `#file-notes-body` was hidden as a whole, which zeroed the reader shell's width so the adaptive layout never resolved (un-gating the rail alone could not have worked). The body now stays mounted and only the items pane, work area and grips wait for a folder, re-asserted after every layout sync; compact is unchanged because the resolver already closes the rail below 120 columns. Task 32136's first criterion is re-ticked without its qualifier; the guide says the collapse grip arrives with the folder.
- **32180** — the slow folder-change row is pinned on-pane at 60 columns from both a linked and an unlinked start (`Choose another` used to sit 11 cells past the edge, and the pre-link state kept a `width: auto` class that defeated the elision); two controls a running change makes dead (Details, Change…) leave the busy row; the `Use <folder>` button reads the modern `[file_notes] root` key first, then the legacy `notes.sync_directory`; the invariant on `_abandon_root_change_task` is recorded as a comment with its real limit (Textual's exclusive-worker cancel lands a loop turn later, so the ticket's "synchronously" was wrong and an assertion would misfire).

Tests: new `Tests/UI/test_library_notes_riders_r_file_notes.py` (mounted through the production app and screen), the wave-1 poisoning test still green, the workspace file at its base failing-name set. Live-verified pre-link at 235x52 and the busy row at 60x24. Reviewed (spec + quality) with one fix round.

Deferred minor (wave ledger): `Use <folder>` can re-offer a folder that just refused to link.

---
## Summary by cubic
Keeps the Library rail visible in Folder files before a folder is linked, fits the slow folder-change row at 60 columns, and tightens the folder-change wait flow.

The rail was hidden because display-gating the body left the reader shell at zero width, so the adaptive layout never resolved. The body now stays mounted; only the items pane, work area, and grips wait for a folder. Compact terminals are unchanged because the resolver already collapses the rail below 120 columns. The collapse grip now arrives with the folder, so the pre-link rail is simply always open.

- The busy row no longer pushes **Choose another** off the right edge; the optional controls size to their labels, and **Details**/**Change…** step aside during the change.
- The busy line elides even when the change starts from the unlinked empty state.
- **Use `<folder>`** reads `[file_notes] root` first and falls back to the legacy `notes.sync_directory`.
- **Cancel** now appears only once the wait admits it is slow; leaving mid-change reports the cancellation as a notification, and the wait line drops the warning/offline tint of the state it replaced.
- The test harness now closes the workspace and replica even after a failing assertion, and an unmounted unit test pins `_configured_sync_folder`'s key precedence and path guard.
- Two pre-existing reds in the files-sync journey suite are filed as task-32294 with their assertion output, so they can be told apart from fresh regressions.
- Guide docs and the task acceptance criteria are updated to match.

<sup>Written for commit 60443c966ca87fb9379d14d1a209c916d0094fa8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2557?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## Review round — 3 fixed, 1 rebutted (554a3744a4)

**1. "Failed tests can leak database handles" — FIXED.** Correct: cleanup sat on
the happy path, so the first failing assertion left a mounted workspace and an
open replica behind for the next test. All four tests now go through one
`_mounted_workspace(...)` async context manager that creates the replica and the
workspace and closes both in `finally`. It also removes four copies of the same
four lines. (Note the same happy-path-only shape exists in ~25 places in
`test_library_file_notes_workspace.py`; not touched here, out of this PR's scope.)

**2. "Two test docstrings omit arguments" — FIXED.** `Args:` sections added to
`test_the_slow_wait_row_keeps_every_control_on_pane_at_60_columns` (naming what
`blocked_root_change` yields and what `start_linked` selects) and to
`test_use_folder_offers_the_modern_file_notes_root`, matching the precedent in
the neighbouring module (`test_library_file_notes_workspace.py` documents its
fixture-taking tests the same way).

**3. "Folder config skips model validation" — REBUTTED.** `_configured_sync_folder`
does go through the project's typed validation boundary for this kind of value:
`validate_existing_absolute_directory` (`Utils/path_validation.py:497`), which
CLAUDE.md names as *the* boundary for file paths, and which rejects relative
spellings, NUL bytes, non-existent paths and non-directories. A Pydantic model
around a single `str` would restate the `isinstance(raw, str)` check and then hand
the value to that same helper. The evidence for the convention: the sibling read
in this widget (`library_file_notes_workspace.py:2321`,
`get_cli_setting("file_notes", "root", None)`) has the same shape, and
`grep -rl pydantic tldw_chatbook/Widgets/Library/` returns nothing — no widget in
this package parses config through a model. The constraint that matters here is
"absolute, existing, a directory", and it is enforced in one shared place rather
than in this widget.

**4. "New folder logic lacks unit coverage" — HALF FIXED, HALF REBUTTED.** Fixed for
`_configured_sync_folder`: new
`test_configured_sync_folder_picks_a_key_and_refuses_an_unusable_path` calls it on
an unmounted workspace across six config shapes — modern wins, legacy fallback,
an unusable modern value falling *back* rather than through, a folder deleted
since it was configured, a blank value, and nothing configured. Four of those were
uncovered. It kills the obvious mutant: replacing
`validate_existing_absolute_directory(...)` with `Path(raw).expanduser()` makes it
fail (verified). Rebutted for `_sync_body_panes`: every assertion it can make is
about `display` and `region.width` on a mounted reader shell — the pre-link bug it
fixes was precisely that display-gating `#file-notes-body` zeroed the shell's width
so the adaptive layout never resolved. A test without the mounted shell cannot
observe that, which is why the two rail tests mount.

### Tests

- `Tests/UI/test_library_notes_riders_r_file_notes.py` → **6 passed** (55s)
- `Tests/UI/test_library_crit8_waits.py` (shared fixture source) → **20 passed** (4m21s)
- 7 shell/root/adaptive-layout tests from `Tests/UI/test_library_file_notes_workspace.py` → **9 passed**
- `ruff check` on the changed file → clean

No source file changed in this round.

rmusser01:
## Folded in: three task-32102 structural-wait items (203f4dc633)

Handed over by the session that owns task-32102 so
`library_file_notes_workspace.py` lands once. All three are RED without the
source change and GREEN with it (verified by reverting `tldw_chatbook/` and
re-running the three node ids: 3 failed → 3 passed).

1. **Cancel reveal timing.** `#library-structural-wait-cancel` was displayed on
   `wait.cancel is not None` at t=0, offering an escape while the row still read
   a bare `Changing folder…`. It is now gated on the same `slow` that Keep
   waiting and Choose another already use, so it arrives with the line that
   first admits the wait is slow. Test:
   `test_cancel_appears_only_once_the_wait_admits_it_is_slow`, which reaches the
   boundary through task-32121's existing patience repaint — no second clock.
2. **Exit-triggered abandon.** `_flush_active_file_notes` (Escape, the back cue,
   the Database strip, the app's navigation flush) now calls
   `cancel_structural_wait(leaving=True)`, which notifies instead of writing
   `ROOT_CHANGE_CANCELLED_COPY` into the root row of the canvas being torn down.
   The in-place exits (the Cancel button, Choose another) keep the row copy —
   the surface is still there to read it. Test:
   `test_leaving_during_a_folder_change_announces_the_cancellation`.
3. **Stale tint.** The wait branch returns before the `-warning` / `-offline`
   resets the other branches perform, so a warning or offline colour sat under
   the wait line. It now clears both. Test:
   `test_the_wait_line_carries_no_tint_from_the_state_it_replaced`.

Two things worth knowing, both corrections to the brief I was given:

- The File Notes wait line does **not** end in `· Cancel` after the patience
  window — task-32121 (AC4/AC5) deliberately removed that, since it read as a
  second control with the same label beside the actual button, and
  `test_folder_change_wait_reports_still_working_and_offers_cancel` pins
  `"Cancel" not in line`. The line reads `Changing folder… · still working`
  (or `· 1,240 entries so far`); the button is what carries the word. The
  reveal boundary is the same one either way.
- Two existing tests pinned the t=0 reveal by pressing Cancel immediately
  (`test_cancelling_a_folder_change_keeps_the_previous_folder`,
  `test_a_folder_change_never_steals_another_surfaces_wait`) — a hidden Button's
  `press()` is a no-op, so both went red. They now age the wait past patience
  through its own start time and repaint (`_reveal_structural_cancel`) before
  pressing; they still test cancel semantics, not reveal timing.

Guide and task notes updated: the folder-link table, the "while a folder change
runs" paragraph, a new paragraph on leaving during a change, a stamp, and a
task-32102 block in task-32180's Implementation Notes.

### Tests

- `Tests/UI/test_library_notes_riders_r_file_notes.py` → **9 passed** (2m11s)
- `Tests/UI/test_library_crit8_waits.py` → **20 passed** (4m35s) — 2 failed before the two pinning tests were updated, 20 after
- 9 root/shell/tint node ids from `Tests/UI/test_library_file_notes_workspace.py` → **13 passed**
- `ruff check`: the test file is clean; `library_file_notes_workspace.py` (30) and `library_screen.py` (197) report exactly the same pre-existing counts as on the base branch — no new findings.

rmusser01:
## Landing round — retargeted to `dev`, four dev merges, 0 behind (3b501add15)

Base moved from `fix/library-notes-docs` (merged) to `dev`. Merged `dev` in four
times as the wave landed ahead of me; never rebased, never force-pushed.

**Conflicts, all in guide stamp lists, all resolved by keeping every stamp with
its own blank line** (a union runs them into one paragraph):
- `Docs/User_Guide/library/notes.md` ×3 — against i-backlinks (task-32145),
  r-pickers (task-32174) and r-editor (task-32177). All four stamps kept.
- `Docs/User_Guide/library/file-notes.md` ×1 — against r-pickers (task-32174);
  kept it plus this branch's three.

`tldw_chatbook/Widgets/Library/library_file_notes_workspace.py` auto-merged with
r-pickers' `_open_root_picker` / `_file_notes_browse_location` work; both sides
verified present, file AST-parses. `library_screen.py` auto-merged around
`_flush_active_file_notes`; `cancel_structural_wait(leaving=True)` verified
present after every merge.

**Silent-bad-merge sweep** clean after each merge: no conflict markers, no
duplicate `def`/`class`, no duplicate constants, widget-id duplicates
byte-identical to `dev`'s (pre-existing, cross-file). Two-dot
`git diff origin/dev HEAD` equals the three-dot diff exactly (10 files,
+781/-61), so the merges dropped nothing from `dev`.

**Derived artifacts.** `check_persistent_diagnostic_inventory.py` exit 0 after
every dev merge (595 owners / 12 sink files) and
`Docs/security/production-diagnostic-inventory.json` is byte-identical to
`dev`'s — no re-pin needed. The new `self.app.notify(reason)` on the
exit-triggered abandon path is not a tracked sink: `--statements
library_file_notes_workspace.py --since <json base>` reports `0 -> 0`
diagnostic calls, and `library_screen.py` `81 -> 81`, `added: 0`. Its payload
is the fixed `ROOT_CHANGE_CANCELLED_COPY` constant — the only `leaving=True`
call site is `_flush_active_file_notes`, so no path, title or body can reach
the toast. CSS bundle sync exit 0, no drift.

**Comments.** Nothing posted since the last summary (2026-09-10 15:27Z); the
four Qodo items were handled in the 2026-09-09 round (3 fixed, 1 rebutted).

### Tests (venv python, foreground, chunked)

| suite | result |
|---|---|
| `test_library_notes_riders_r_file_notes.py` (9) | 9 passed |
| `test_library_notes_wave_file_notes.py` (10) | 10 passed |
| `test_library_crit8_waits.py` (20) | 20 passed (see flake below) |
| 10 root/shell/layout node ids of `test_library_file_notes_workspace.py` (14) | 14 passed |

**Known dev-side reds, confirmed identical on `dev` and here** — filed as
**task-32294** in this branch, with both assertion texts recorded:
`test_library_notes_files_sync_journey.py::test_lasting_setup_keeps_server_unavailable_copy_painted[size0]`
(`assert "capability not" in painted`) and
`::test_folder_files_and_session_git_use_supported_40x20_navigator`
(`assert session_git in ..._compositor.visible_widgets`). Same two names, same
assertion output, on clean `dev` 068535986e and on this branch.

**Flake, investigated and cleared:**
`test_library_crit8_waits.py::test_skill_import_wait_reports_still_working_and_can_be_cancelled`
failed twice mid-session (`#library-skills-import-path never mounted within
30.0s`) and passed on the other two full-file runs. It is not this PR: with
this branch's three changed files copied into a detached clean-`dev`
worktree the same file ran **20 passed**, and the test passes 3/3 in isolation
on both trees. Load-sensitive 30 s mount timeout on a machine that had been
running UI suites for an hour.
`test_library_notes_wave_file_notes.py::test_folder_files_remembers_the_picked_root`
flaked once on `dev` itself and once here; passes 10/10 in the file on both.

PR #2558, opened 2026-09-09
fix(library-notes): date ordering back in Database Notes; the Newest default is stated; the toolbar merge threshold matches the real Sort width (task 32172)

rmusser01:
Rider wave 2, task 32172.

## What

- Database Notes list rows carry a date again, and the list orders by it. The critique found rows with no age at all, so two notes with the same title were indistinguishable and "which did I touch last" was unanswerable without opening each one.
- The sort control states its default in the label instead of leaving "Newest" implicit.
- `_TOOLBAR_MERGE_MIN_WIDTH` raised 100 → 109. The old value was measured before the Sort control existed; Sort adds 16 cells, so at 100–108 columns the toolbar overflowed instead of merging.

## Verification

- Unit tests for the ordering and the threshold.
- Live: driven in a real terminal at 80x24, 109x40 and 120x40.

Task: `backlog/tasks/task-32172`.

---
## Summary by cubic
Restores date ordering to Database Notes by making the placement order a repository parameter shared by the pager and the deep-link locator, so the folder tree's Sort control works again. The default browse order is now newest-first (was title), and Sort is disabled with a reason while a filter is showing.

- Choosing Newest, Oldest, or Title re-pages the whole tree, open folders included; re-picking the active value is a no-op.
- Renaming a note re-sorts the loaded window in the active order, not just by title.
- The repository still defaults to title order when none is passed, so existing callers are unchanged.
- Date ordering normalizes the two `last_modified` timestamp shapes so same-day rows sort by time, not by the separator byte.
- Raised the toolbar merge threshold from 100 to 109; at 100–108 columns the merged row clipped "Last import" off the pane once Sort rejoined.
- Fulfills task-32172's acceptance criteria and states the newest-first default in the user guide.
- The shared screen fake now carries a no-op trash refresh so tests land cleanly whichever order the i-trash wave merges in, and the sort-chooser assertion waits on the DOM since that refresh can delay the recompose by a frame.
- The user guide's stale "Sort not offered on the folder tree" claim now points to task-32172 superseding it.

<sup>Written for commit f60e49bee29fdcdf4fec19c12bc3615a4838978c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2558?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
## Review comments addressed (task-32172 rider)

**1. Mixed timestamp formats misorder notes (High, `note_folder_repository.py:2899`) — FIXED.**
Real bug, verified empirically against a real `CharactersRAGDB`: `last_modified` is `DATETIME DEFAULT CURRENT_TIMESTAMP` (schema default writes space-separated `YYYY-MM-DD HH:MM:SS`), while every application writer (`_get_current_utc_timestamp_iso`, `note_import_executor._utc_timestamp`) stamps ISO `YYYY-MM-DDTHH:MM:SS.sssZ`. Raw-text `ORDER BY last_modified` sorts same-day rows by the separator byte (`' ' 0x20 < 'T' 0x54`), not by time — reproduced against a live DB: two same-day notes at `2026-01-15T06:00:00.000Z` and `2026-01-15 12:00:00` come back `['mixed-noon', 'mixed-morning']` under the old raw-text ORDER BY, backwards from the true chronological order. Wrapped both date branches of `_placement_order_term` and `_placement_rank_clauses` in `julianday(...)` so the pager and the locator compare a normalized numeric instant instead of raw text (title order untouched). New test `Tests/Notes/test_note_folder_repository.py::test_mixed_timestamp_shapes_still_order_chronologically` pins both the pager and the locator against this exact mixed-shape scenario; it fails on the pre-fix code (confirmed by reproducing the old raw-text query against a real DB) and passes after the fix.

**2. Database sort skips central validation (Medium, `note_folder_repository.py:408`) — REBUTTED.**
`_validate_placement_order` (2873-2890) fails closed against the 3-value literal tuple `PLACEMENT_ORDERS` before `order` is used anywhere. `_placement_order_term`/`_placement_rank_clauses` (2894-2942) never interpolate `order` itself into SQL — they use it purely as an `if/elif` branch selector across three fully hardcoded SQL string literals; the actual identifiers (`last_modified`, `title`) are Python source constants, never derived from `order` or any external input. There is no identifier here for `sql_validation.py`'s `validate_identifier()`/`get_safe_order_by_clause()` to protect — those exist to validate a caller-supplied token that gets woven into SQL text; `order` never becomes such a token. A closed 3-value allow-list checked before any branch is at least as strong as the generic identifier regex `sql_validation.py` offers for this shape of problem. The codebase already tolerates a hardcoded dynamic-ORDER-BY literal outside `sql_validation.py` elsewhere too (`Client_Media_DB_v2.py:3043`), and there's no repo-enforced test requiring every dynamic ORDER BY fragment to route through that module (`Tests/DB/test_sql_validation.py` tests the module itself, not callers). Routing this through the central module would validate constants that are already correct by construction — no gap closed.

**3. Saved notes can break date tie order (Low, summary only) — REBUTTED, already adjudicated.**
This exact concern (`_resort_placement_pairs`'s `modified()` in `library_notes_tree_paging.py` stringifying a repository-loaded `datetime` object against a freshly patched ISO string) was found during this PR's own development and explicitly scoped out by the task review, recorded verbatim in task-32172's Implementation Notes: *"Deferred, no action (task review): `_resort_placement_pairs` compares `last_modified` as Python strings -- the Python twin of the SQL round-trip fixed above."* It's a Python-side, in-memory-window-only effect (self-heals on the next tree reload) and independent of the SQL-side mixed-format bug fixed in item 1 above. No new information here changes that already-made call.

### Tests run (foreground)
- `Tests/Notes/test_note_folder_repository.py` + `Tests/Notes/test_notes_scope_service_folders.py`: **396 passed**
- `Tests/UI/test_library_notes_riders_r_list.py` + `Tests/UI/test_library_notes_rename_propagation_t31796.py`: **21 passed**
- `Tests/UI/test_library_notes_wave_list.py` + `Tests/UI/test_library_notes_folder_navigator.py` + `Tests/UI/test_library_note_import_flow.py`: **149 passed, 3 failed** — all 3 failures are pre-existing and unrelated to this change (`test_unmount_invalidates_notes_authority_before_first_await` is a source-order assertion on `LibraryScreen.on_unmount` unrelated to note ordering; the two `test_clearing_filter_restores_same_epoch_browse_receipt_without_touching_ranges` failures are `AttributeError: 'types.SimpleNamespace' object has no attribute '_queue_library_notes_settled_focus_restore'` in a filter/focus-restore test double — neither touches `note_folder_repository.py`).

rmusser01:
## Retargeted onto `dev`; four dev merges; one real merge-interaction fix

Base moved from `fix/library-notes-docs` (now merged) to `dev`. Merged
`origin/dev` in four times as the wave landed around this branch — never
rebased, never force-pushed. Pushed `d7f25c5c9f`, **0 behind `dev`**.

### Conflicts, and how each was resolved
- `Docs/User_Guide/library/notes.md` (3 rounds — i-backlinks, then
  i-trash/i-backlinks, then r-pickers): "Verified against" stamp list, every
  time. Kept **all** stamps from both sides, each with its own blank line
  before it. `dev` has 27, this branch adds exactly 1 → 28, and the
  `task-32128` "Sort is not offered there — superseded by task-32172 below"
  sentence survives.
- `tldw_chatbook/UI/Screens/library_screen.py` —
  `_request_library_notes_tree_initial_load`, the exact i-trash × r-list
  collision the wave review predicted. Kept **both** edits: task-32144's
  `self._refresh_library_notes_trash()` right after
  `_begin_library_notes_tree_visit`, then task-32172's loop over
  `(None, *tree_expanded_ids)`.

`Docs/security/production-diagnostic-inventory.json` auto-merged to dev's copy
each time and is **byte-identical to dev** (this branch changes no diagnostic
call); `scripts/check_persistent_diagnostic_inventory.py` exits 0 on the final
tree (595 owners), re-run after **every** merge, and the CSS bundle-sync check
is clean.

### One real regression the merge created, and its fix
`test_library_shell_notes_sort_opens_direct_choices_and_applies_one_value`
passed at `670e27a2` and failed after the i-trash merge. Cause: task-32144's
trash read is now started from the same reload a sort choice triggers, so the
recompose that unpaints `#library-notes-sort-choices` can land a frame after
`_notes_state.sort` flips — and the test asserted on the DOM the instant the
flag changed. The chooser does close; the assertion was racing it. Changed to
wait for the DOM instead of the flag (`d.. test(library-notes): wait for the
sort chooser to unpaint, not for the flag`); it still fails, by timeout, if the
chooser never closes. Passes 3/3 runs.

### Silent-bad-merge sweep (after every merge)
No conflict markers anywhere; no duplicate `def`/`class` or module constant in
any changed `.py`; the duplicate-`id=` set in `library_notes_canvas.py` is
byte-identical to dev's (pre-existing); all 16 changed `.py` files AST-parse.
**Two-dot audit**: `git diff origin/dev HEAD --stat` is byte-identical to the
three-dot diff — 18 files, +1029/−136 — so no `dev` content was dropped, and
this branch's own delta is unchanged from before the merges. Exactly one
`task-32172` file in `backlog/tasks/` (mine); dev's renumbered `task-32203`
(Buddy) is present and untouched.

### Tests (foreground, venv python, chunked)
- `test_note_folder_repository.py` + `test_notes_scope_service_folders.py` +
  `test_library_notes_folder_navigator.py` + `test_library_note_import_flow.py`:
  **524 passed, 3 failed**
- `test_library_notes_riders_r_list.py` (9) +
  `test_library_notes_rename_propagation_t31796.py` +
  `test_library_notes_wave_list.py` + `test_library_notes_wiring.py` +
  `test_library_notes_tree_paging_live.py` + the 3 touched
  `test_library_shell.py` node ids: **62 passed, 0 failed**
- `Tests/Architecture/test_persistent_diagnostic_inventory.py`: 1 failed

The 3 navigator failures (`test_unmount_invalidates_notes_authority_before_
first_await`, `test_clearing_filter_restores_same_epoch_browse_receipt_without_
touching_ranges[button|empty_submit]`) reproduce **identically** on a detached
`origin/dev` worktree at the same tip — pre-existing, not from this branch.
Same for the inventory-architecture failure
(`test_reviewed_diagnostic_changes_are_metadata_only`, complaining about
i-trash's `"Failed to restore a Library note"`): red on dev too.
`test_mounted_topology_changed_back_restores_exact_duplicate_ranges_and_scroll`
is a load flake — it failed once on dev's baseline and once here, and passed on
re-runs of both.

### Review comments
Nothing new since the 2026-09-10 summary: both Qodo items stand as recorded
there (High mixed-timestamp ordering **fixed** with `julianday()`; Medium
central-validation **rebutted**).

PR #2602, opened 2026-09-11
fix(library-media): the footer keeps its canvas verbs while typing, Find opens from the keyboard and refuses the tabs it cannot search, a Markdown analysis renders (critique-10 wave, tasks 32346/32348/32365)

rmusser01:
Part of the Library critique-10 fix wave (plan: `Docs/superpowers/plans/2026-09-11-library-crit10-wave.md`, group `viewer`). Lands before the `export` (#2601) and `layout` branches, which carry a viewer carve-out and the adjacent footer block.

## What
- **task-32346** — while an Input has focus the footer keeps the canvas verbs visible behind one named key (`typing in field · after esc: …`) instead of blanking to `typing in field | F6 next pane`; it never advertises a key that does nothing (a focused Input eats non-priority bindings, so `u`/`o`/`/` are shown as what follows Escape — the dead-key lie task-31272 removed stays removed); `F6 next pane` is dropped before any canvas verb when width is short; below 64 columns the block stands down entirely so the narrow stage's single chip (`esc back to Library`, owned by the layout branch) is the whole context.
- **task-32348** — the Media viewer's Find opens from the keyboard (`ctrl+f`, with a footer chip) and is refused with a reason on the tabs that cannot search; the `t` trash accelerator cannot fire while a Find gesture is pending. The critique's "not in the Tab order" claim is disproved by a `focus_chain` pin, not "fixed". One cause fixed once in `analysis_find_unavailable_reason`.
- **task-32365** — a Markdown analysis renders (content sniff + a Rendered/Raw toggle owned by the viewer widget); Find over a rendered analysis highlights and scrolls to matches (the seam now rebuilds on a query start/end).

## Evidence
- `Tests/UI/test_library_crit10_viewer.py` (10 tests, red-first) + updated doubles in `test_library_media_reader_flow.py`; the four reader/render files re-run to exactly the base's failing names; 60x24 painted footer pin shared with the layout branch.
- Live-verified on an isolated seeded profile.
- Guides: `library.md`, `library/media-and-conversations.md` with refreshed stamps.

## Review
Task review (approved; 1 Medium, 1 low-medium, 5 low) → fix round 1 (`13d18854c8`, `7fd66fdbd8`, `fbd68a742f`; one decline with evidence — the remaining item belongs to the export branch). Rider noted: the Read tab's Find has the same patch-without-recompose gap.

---
## Summary by cubic
Fixes three Library critique-10 viewer issues: the typing footer keeps its canvas verbs, Find opens from the keyboard, and stored Markdown analyses render.

**task-32346 — footer under a focused field**
- With a text field focused, the footer leads with `typing in field`, names an honest Escape chip (gated through `check_action`), and keeps the canvas verbs behind `after esc:` instead of blanking them.
- `F6 next pane` is dropped before any verb on narrow widths; below 64 columns the single chip names the one key that works (Escape on the viewer), never a bare status word.

**tasks-32348/32365 — Find and the Analysis tab**
- `ctrl+f` opens the Reader's Find bar with a footer chip, sharing one implementation and one gate with the Find button so the two can't diverge; Info/Highlights refuse Find with a reason, external details stay searchable, and the gate reads the analysis the viewer actually displays.
- An open Find bar disarms the `t` trash accelerator. The analysis view choice persists for the session.
- The Analysis tab runs the same Markdown content sniff and Rendered/Raw toggle as Read, so a stored `## Key contributions` paints as a heading; plain-prose analyses get no toggle.
- An active Find query over a rendered analysis swaps to the Raw view (the only one that marks matches) and back on clear; while the query forces Raw, neither toggle half accepts a press and the reason is shown inline.
- The "Find not in the Tab order" claim is disproved by a `focus_chain` pin rather than fixed.

<sup>Written for commit 6207e5816803d246c77bed0e58fa2ea89cc57090. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2602?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/User_Guide/library.md
+++ b/Docs/User_Guide/library.md
@@ -268,8 +268,12 @@ readable measure rather than stretched across the whole canvas.
   "key action" pair — the Notes editor, for example, shows "ctrl+s save
   note | esc back to notes".
   While a text field has focus the footer leads with `typing in field`,
-  names `esc leave field`, and keeps the canvas verbs after `after esc:` —
-  they are not live until you leave the field.
+  names the Escape chip for that surface (`esc leave field` where nothing
+  else owns the key — a list canvas keeps its own `esc focus rail`), and
+  carries the canvas verbs after `after esc:` — they are not live until you
+  leave the field. On a terminal under 64 columns the return chip
+  (`esc back to Library`) comes first instead, because that is the one
+  action the narrow footer must never drop.
 
 One special case: selecting **Notes** adds a
 **Library notes | Folder files** strip above the workbench. **Folder files**
--- a/Docs/User_Guide/library.md
+++ b/Docs/User_Guide/library.md
@@ -271,9 +271,8 @@ readable measure rather than stretched across the whole canvas.
   names the Escape chip for that surface (`esc leave field` where nothing
   else owns the key — a list canvas keeps its own `esc focus rail`), and
   carries the canvas verbs after `after esc:` — they are not live until you
-  leave the field. On a terminal under 64 columns the return chip
-  (`esc back to Library`) comes first instead, because that is the one
-  action the narrow footer must never drop.
+  leave the field. Under 64 columns there is room for a single chip, so the
+  footer names only what Escape does there and the verbs are not listed.
 
 One special case: selecting **Notes** adds a
 **Library notes | Folder files** strip above the workbench. **Folder files**
--- a/Docs/User_Guide/library.md
+++ b/Docs/User_Guide/library.md
@@ -258,7 +258,8 @@ readable measure rather than stretched across the whole canvas.
   landing adds "i import content"
   and "n new note" (single-letter accelerators for the hub actions);
   the Search / RAG canvas adds "u use Library
-  context in Console", "enter select evidence", and "o open evidence";
+  context in Console", "o open evidence", and an "enter" hint that names
+  whatever Enter does on the control you are focused on;
   a Media/Notes/Prompts/Skills/Collections list adds "esc focus rail";
   that list's item viewer/editor (or the media viewer) adds "esc back to
   list" instead; the Export canvas adds "esc back to Media" (or whichever
@@ -266,6 +267,9 @@ readable measure rather than stretched across the whole canvas.
   staging canvas adds "esc back to hub". Every hint is a per-key
   "key action" pair — the Notes editor, for example, shows "ctrl+s save
   note | esc back to notes".
+  While a text field has focus the footer leads with `typing in field`,
+  names `esc leave field`, and keeps the canvas verbs after `after esc:` —
+  they are not live until you leave the field.
 
 One special case: selecting **Notes** adds a
 **Library notes | Folder files** strip above the workbench. **Folder files**
@@ -884,3 +888,10 @@ list, and the landing hub is capped at 96 cells).*
 (task-32225: the return chip stands down wherever a canvas owns Escape itself,
 so the footer never names a return the key would not perform; task-32228:
 Conversations arrives with its first row focused like every other browse list).*
+
+*Verified against fix/library-crit10-viewer — 2026-09-11 (task-32346: with the
+caret in the Search/RAG query box the footer reads "typing in field | esc leave
+field | after esc: u use Library context in Console · o open evidence | enter
+run search | F6 next pane"; the same shape on the Media list keeps "s select"
+behind the same gesture, and "F6 next pane" is last so a narrow footer drops it
+before any verb).*
--- a/Docs/User_Guide/library/media-and-conversations.md
+++ b/Docs/User_Guide/library/media-and-conversations.md
@@ -501,6 +501,8 @@ still spans the pane.
   a two-line analysis puts them on the next line, not at the bottom of the
   pane. A long analysis scrolls the tab, carrying its actions to the end of
   the text rather than clipping them.
+  A Markdown analysis renders like the Read tab, with the same
+  Rendered/Raw toggle.
   Analysis is produced at import time (the "Analyze after import" option),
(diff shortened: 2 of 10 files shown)


rmusser01:
Review response: 3 of 4 fixed, 1 declined. Head `011b9355a1`.

**1. Find rejects some visible analyses (High, Bug) — FIXED.** Confirmed and
real, and worse than reported: `detail_analysis_text` returned only
`_latest_version_analysis_text`, while `build_library_media_viewer_state`
(line 395) uses `_text(detail.get("analysis_content")) or
_latest_version_analysis_text(...)`. So a detail with top-level
`analysis_content` and no versions **displayed** an analysis that the helper
reported as absent. The second caller is affected too, and its own docstring
already claimed the fixed behaviour: `_library_media_unanalyzed_ids` says
"the same rule the Reader's Analysis tab uses", but counted such an item as
un-analyzed and would have re-run analysis over it. Fixed once at the shared
seam rather than at the Find call site. Pinned by
`test_find_follows_the_analysis_the_reader_actually_shows`, which asserts
`detail_analysis_text` and `build_library_media_viewer_state(...).analysis`
agree, plus the two fallback cases. `Tests/Library/test_library_media_viewer_state.py`,
`test_library_multiselect_media.py`, `test_library_ingest_analyze_skipped.py`,
`Tests/Library/test_library_media_content.py`: 201 passed.

**2. Stored analyses bypass sanitization (Medium, Rule 497156) — DECLINED,
with evidence.** Three reasons:
- *Not introduced here.* `LibraryMediaContentBody._build_markdown_widget`
  (`library_media_content.py:321-326`) constructs Textual's `Markdown` with no
  sanitizer, and has done since LIB-13 for the **Read** tab's
  `viewer.content` — web captures, PDFs, ebooks, transcripts, every import
  path. This PR adds one more producer to an existing shared sink; it does not
  create the sink. A fix belongs in that shared widget, which is outside this
  task's ownership rows in the wave's file map and is shared with the Read tab.
- *The sink is a terminal, not a browser.* Textual's `Markdown` renders to a
  TUI. There is no HTML execution, no DOM, no script context, and no
  navigation. Raw HTML and event attributes in the source are not executed by
  markdown-it-py — they render as text or are dropped.
- *There is no link sink on this screen.* `Markdown.LinkClicked` is handled in
  `console_transcript.py`, `settings_screen.py` and the tools windows — **not**
  in `library_media_content.py`, `library_media_viewer.py` or
  `library_screen.py`. A `javascript:`/`file:` URL in a rendered analysis is
  inert in the Library Reader: nothing opens it.
  If the rule should still be applied to this widget, it wants its own task
  covering both tabs and every import path, not a one-producer patch here.

**3. Clearing Find can leave analysis raw (Medium, Bug) — FIXED.** Correct
catch, and it was my new code: I had disabled the *Rendered* half while a query
forces Raw but left Raw pressable, and the handler wrote
`analysis_content_mode` unconditionally — so pressing the already-selected Raw
made it stick and the inline line ("clear the search to read it rendered")
became a false promise. Now **neither** half accepts a press while the query
forces Raw; the strip reports state and both controls carry the reason. Pinned
by `test_pressing_raw_during_a_search_does_not_outlive_the_search`, which
asserts the button is disabled, the stored preference is still `rendered`, and
clearing the query restores the rendered body.

**4. Server documents cannot be searched (Medium, Bug) — FIXED.** Confirmed:
`_compose_active_body`'s first branch is `if self.external_detail or
self.reader_mode == "read":`, so an external detail composes the searchable
Read body whatever mode it inherited. The gate returned `""` for every
non-analysis mode before this PR, so the mode-only refusal was a **regression
this PR introduced**, not a tightening. `analysis_find_unavailable_reason` now
takes `external: bool = False` and returns `""` for it; the local
Info/Highlights refusal is unchanged. Pinned by
`test_an_external_detail_is_searchable_whatever_mode_it_inherited`.

One follow-up this leaves open, deliberately: the Find **button** is composed in
`_compose_primary_toolbar`, which another branch in this wave owns, so its call
does not yet pass `external=self.external_detail`. The keyboard path is correct;
the button keeps today's behaviour until that one keyword is added in its
owner's range. Flagged to the wave coordinator.

Tests: `Tests/UI/test_library_crit10_viewer.py` 14 passed (3 new, all red
first). `test_library_media_reader_flow.py` + `test_library_media_render_fixes.py`
+ `test_library_media_reader_shell.py` + `test_library_crit9_media_reader.py`:
9 failed / 264 passed — the same 9 names that fail on `origin/dev`, no new ones.
`./scripts/preflight.sh`: all derived-artifact checks passed.

rmusser01:
Review response follow-up (commit `df3f5b7113`): finding 4 is now closed on **both** entry points. Last round added `external=` to `analysis_find_unavailable_reason` but only the keyboard path passed it, so for a server document opened while the session carried `info`/`highlights` `Ctrl+F` opened Find while the button beside it stayed disabled — the same key/control divergence the shared `_toggle_library_media_find` exists to prevent. `_compose_primary_toolbar` now passes `external=self.external_detail` in its existing call (one keyword, nothing else in that function moved). Pinned by `test_the_find_button_matches_the_key_on_an_external_document`, red before the keyword and green after. `Tests/UI/test_library_crit10_viewer.py`: 15 passed. `test_library_media_reader_shell.py` + `test_library_crit9_media_reader.py`: 49 passed. `./scripts/preflight.sh`: all derived-artifact checks passed.

PR #2603, opened 2026-09-11
fix(library): labelled grips, link-on-use hand-off, rail focus shape, narrow footer (crit10 layout)

rmusser01:
Part of the Library critique-10 fix wave (plan: `Docs/superpowers/plans/2026-09-11-library-crit10-wave.md`, group `layout`).

## What
- **task-32355** — the three pane grips carry their label, so a first-time user can tell a splitter from a border and knows which pane it resizes.
- **task-32107 + decision task-32057/32107 (user: link-on-use)** — pressing "Use as source" on a conversation links it to the active workspace in one undoable step and paints a receipt naming the workspace; the refusal copy only appears when there is nothing to link to. The receipt and its Undo are withheld while any workspace block stands, so the receipt can never contradict the refusal beside it. The decision text is recorded verbatim in the task.
- **task-32361** — the Conversations split gives the reader its width back at the standard stage.
- **task-32359** — rail focus is a shape (outline), not a second blue; the selected row and the focused row read differently at a glance.
- **task-32360** — the field state keeps the head of the narrow footer (`esc back to Library` in both states, painted pins at 60 cols); AC#1 and AC#3 were already true and are pinned as regressions; AC#2 (the legacy_compact gate) stays open and lands with the notes-details branch.

## Evidence
- `Tests/UI/test_library_crit10_layout.py` (21 tests, red-first, incl. 60x24 footer pins) + updated pins in `test_library_crit8_conversation_handoff.py`, `test_library_row_focus_cue_t31983.py`, `test_library_conversation_reader.py` (59 passed). Failing-name sets on the touched suites match the base.
- Live-verified at 235x52, 100x30 and 60x24 on an isolated seeded profile.
- Guides: `library.md`, `library/media-and-conversations.md` with refreshed stamps; TCSS source + rebuilt bundle.

## Review
Task review (Changes requested: 1 P1 on the 60-col footer, refuted by measurement — blur is inactive on the narrow stage so the return wins — 3 P2, rest P3) → fix round 1 (`6024a06897`, 10 fixed / 2 declined with evidence) → scoped re-review (12/12 settled; 1 new P3: a link receipt could outlive the refusal after a workspace switch) → fix round 2 (`4831835460`). Residual recorded in the task notes: switching to a second workspace that also holds the membership leaves a receipt naming the first.

---
## Summary by cubic
Fixes the Library critique-10 layout wave: pane grips now spell their names, focused rail rows read by shape, the Conversations reader gets the width back once open, and "Use as source" links an unlinked conversation in one undoable step.

**Behavior changes**
- "Use as source" links a conversation to the active workspace first, proceeds to Console, and shows a "✓ linked · \<workspace\>" receipt with Undo beside it.
- Blocks a link cannot resolve (no active workspace, stale page) disable the button and refuse.
- The receipt and Undo are withheld while any workspace block stands, so they never contradict a refusal beside it.
- The receipt is recorded only for a membership the press actually inserted, so Undo never deletes an existing link.

**Bug fixes**
- Pane grips paint their name down the column: "Nav" for the Library pane, the pane's own label for the items pane.
- The Conversations split re-resolves when a conversation opens, so the reader takes the majority at 200+ columns (132 vs 49 at 235x52).
- Focused rail rows get the house `█` left bar; active rows keep the background treatment.
- The narrow footer paints "esc back to Library" at 60 columns in both focus states; the field-state marker yields because only one chip can paint there.
- task-32360 AC#2 (clipped copy ellipsis) stays open — it needs the Notes compact gate, outside this branch's owned files.

<sup>Written for commit 8fdc0d857a7a235fff6c0e0e0600a6ad7e28b251. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2603?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/User_Guide/library.md
+++ b/Docs/User_Guide/library.md
@@ -212,7 +212,11 @@ readable measure rather than stretched across the whole canvas.
   - a **Navigation** heading with **Collapse** at the opposite edge. Collapse
     hides the rail without changing the selected destination, search query,
     section disclosures, or canvas. The slim **Nav** handle expands it again;
-    it is keyboard-focusable and remains part of the **F6** pane cycle;
+    it is keyboard-focusable and remains part of the **F6** pane cycle. The
+    handle is five cells wide, so it spells its name **downwards** —
+    `N`/`a`/`v` down the column, with the collapse arrows below it. The
+    handle beside the list pane does the same with that pane's name
+    (**Items**, **Prompts**, **Skills**, **Folder files**);
   - the **Import…** button ("Add files, links, and transcripts to
     your Library.");
   - the **Search Library…** box — submitting it lands on the
@@ -275,6 +279,13 @@ visible stage so its controls remain on-screen. Escape (or the
 **Library notes** link) returns to the notes list — see
 [File Notes](library/file-notes.md).
 
+Opening a **note for editing** is the one place the rail steps aside on a
+wide terminal, deliberately: at 120 columns and more of editor width the
+rail collapses to its **Nav** handle so the note gets the room, and Escape
+(or **Nav**) brings it straight back. Expanding it by hand ends that for the
+rest of the visit — the rail then stays put while you open one note after
+another.
+
 ## Features & controls
 
 ### State glyphs
@@ -305,7 +316,7 @@ leading `▸` uses cannot collide on one control.
 | Control | What it does |
 |---|---|
 | **Collapse** | Hides the wide navigation rail in place and gives the canvas the reclaimed width. The choice lasts for the current Library screen session. |
-| **Nav** | Expands a manually collapsed rail and returns focus to **Search Library…**. On compact terminals, Library's existing one-pane routing takes precedence and the manual collapse returns when the terminal is wide again. |
+| **Nav** | Expands a collapsed rail and returns focus to **Search Library…**. The handle spells `Nav` down its five-cell column, so it is readable without hovering. On compact terminals, Library's existing one-pane routing takes precedence and the manual collapse returns when the terminal is wide again. Used on the rail a note editor collapsed, it also ends that editor's claim on the width for the rest of the visit. |
 | **Import…** | Opens the Import media canvas — see [Import & export](library/import-and-export.md). |
 | **New note** | Opens the production note-creation canvas. It is shown directly in the Get started rail. |
 | **Explore all tools** | Reveals and remembers the complete Library without changing section disclosures. |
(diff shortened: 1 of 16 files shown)


rmusser01:
Review response: 3 Qodo findings — 2 fixed, 1 fixed-by-deletion with one part declined. All in `3f521db8bf` (head `8fdc0d857a`).

**#2 "Undo can delete an existing link" (High, Bug) — FIXED.** Confirmed real: `link_membership` is `INSERT OR IGNORE` (`registry_service.py:1144`) and returns the existing row, so a press arriving through stale cached eligibility recorded a receipt for a link it never made and Undo would have removed someone else's membership. `_link_selected_conversation_to_workspace` now reads `get_item_memberships` first and records the receipt (and so exposes Undo) only for a membership it actually inserted; the hand-off proceeds either way, per your recommendation. I did not push an atomic "did it insert?" answer into the registry — that is a shared service and one UI thread makes the read-then-write window unreachable here; marked with a `ponytail:` comment naming the ceiling and the upgrade path. Pinned red-first: `test_undo_is_withheld_when_the_membership_already_existed`. The pin also caught a second-order bug in my first cut — returning early on the already-linked path skipped `_invalidate_library_workspace_depth_state()`, leaving the cached eligibility stale so the hand-off refused the conversation it had just accepted.

**#3 "Stale conversations offer a dead action" (Medium, Bug) — FIXED.** Confirmed real. Rather than adding a freshness key to the reader's metadata, staleness is reported through the ONE workspace-block seam (`_library_conversation_workspace_block`) as a block a link cannot resolve: the action disables with its non-colour `○` marker and states its reason, no link is offered, and the reader, the tooltip and the `c` accelerator cannot disagree about it. Resume is deliberately unaffected — reopening the original never needed a fresh list. Pinned red-first: `test_a_stale_page_refuses_instead_of_promising_a_link`.

**#1 "Link flow repeats protocol strings" (Medium, Rule violation) — half FIXED by deletion, half DECLINED.**
- The `freshness != "fresh"` literal in `use_selected_conversation_as_source` is **gone**, not centralised: with #3 fixed, a non-fresh page is already a block and `_library_conversation_link_would_unblock()` is False for it, so the duplicate comparison had nothing left to disagree with. One fewer copy rather than one more indirection.
- The `_workspace_link_receipt` metadata key is **declined**. It is the fifth key on an established controller→reader seam whose four siblings (`_workspace_block`, `_workspace_block_detail`, `_workspace_block_linkable`, `_list_status`) are all plain literals at both ends. Hoisting one of five into a constant would make the seam less consistent, not more, and the key has exactly one writer and one reader.

Verification: `test_library_crit10_layout.py` + `test_library_crit8_conversation_handoff.py` + `test_library_conversation_reader.py` → **98 passed** post-merge with `origin/dev`. `./scripts/preflight.sh`: all derived-artifact checks pass. `test_library_multiselect_conversations.py::test_convo_row_select_mode_toggles` is red, identically on base — not from this branch.

PR #2604, opened 2026-09-11
fix(library-media): labelled row age, scope line with Clear, list and import-footer copy (crit10 media-rows)

rmusser01:
Part of the Library critique-10 fix wave (plan: `Docs/superpowers/plans/2026-09-11-library-crit10-wave.md`, group `media-rows`).

## What
- **task-32347 (critique-10 P1)** — a media row's age reads as an age, not a duration: `added 10m` (with the row's own time column keeping the duration), so "3h" no longer looks like a three-hour recording.
- **task-32350** — a scope line above the Items list states the applied filter (query, type facet, sort) with its own Clear; the Clear stays on screen when the Reader narrows the list (`1fr` + ellipsis, live-captured before/after), drops the type facet too (a type-only scope had a dead Clear otherwise), and the line does not paint an empty row for a legacy-built canvas state.
- **task-32364** — copy polish across the lists and the import footer: the Enter label reaches the footer on every ingest transition (the non-structural branch of the dynamic-region update now re-registers it), "N of M" reads honestly when the total is unknown, and the prompts/conversations row copy pins match what the rows paint.

## Evidence
- `Tests/UI/test_library_crit10_media_rows.py` (8 tests, red-first; the footer pin applies the same pre-flight result twice so the structural branch cannot mask the hole) + pins in `test_library_media_state.py`, `test_library_prompts_canvas.py`; 606 passed across the touched suites; `test_library_media_render_fixes.py` keeps exactly its 5 dev-red names, zero new.
- Live-verified at 235x52 and 100x30 on an isolated seeded profile (captures in the task notes).
- Guides: `library.md`, `library/media-and-conversations.md`, `library/import-and-export.md` with refreshed stamps.

## Review
Task review (Changes requested: 1 high — ingest footer not re-registered when the Start gate opens — 2 medium, 1 low-medium, 7 low, 2 nits) → fix round 1 (`41bae43322`, `49eb56e6fd`, `a9da7f428c`: 11 fixed, 1 declined by controller ruling, 2 became riders task-32378 prompt-lane vocabulary and task-32379 stored "New Chat" title). Scoped re-review in progress. Deviations from the brief are listed in the task notes (controller accessor read, sort label call, scope-line width, Clear drops the type facet).

---
## Summary by cubic
Fixes the critique-10 media-row wave: media rows label their age, the Media list states its applied scope with a Clear, and list and import-footer copy is standardized.

- A media row's age now reads as an age — "audio · updated 10m", "updated just now" under a minute — instead of a bare "10m" that read as the item's duration; the word is "updated" because the value is the last-modified time, which Generate moves.
- The scope line under the Media header states the applied filter, type, and sort with the count as "N of M", built only from applied state so the filter box's unsubmitted draft never appears.
- The scope line's Clear drops the filter and type and blanks the box, and stays retryable after a failed clear; the toolbar's "Clear filter" still clears only the filter it names.
- The "of M" half is dropped when the Media total is only a lower bound, matching the rail's "5+".
- The "· loaded"/"· loading" state moved off the row title onto the fact line; titles always open their own rows now.
- Conversations rows use the same "·" separator as every other list; plain prompt rows drop the "Prompt · " prefix; "System + User" reads "has system and user text".
- The Import footer names Enter's action by gate state — "enter check this path" / "enter start import" — and re-registers on every gate transition, including non-structural ones and while suspended.
- The labelled age costs ~8 cells per row: keyword rows clip sooner at the narrow Items width and the "· loaded" row clips first at 100 columns (pinned to the exact new output, not loosened).
- The stored "New Chat" title and the mixed prompt-lane register are filed as task-32379 and task-32378 rather than relabelled here.

<sup>Written for commit c971ecfa63d4351749b185526b5089e77fe22351. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2604?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Docs/User_Guide/library/import-and-export.md
+++ b/Docs/User_Guide/library/import-and-export.md
@@ -315,9 +315,13 @@ destination, or leaving the Import canvas cancels pending consent.
 Library screen (not just the landing — though never while you're typing in
 a text field, where `i` stays a letter), and entering the form always
 parks the caret in the path field, so you can type or paste a path
-immediately. **Enter** in the path field starts the import once the gate
-line clears — with "⚠" warnings outstanding, Enter,Enter carries the same
-two-press consent as the Start button. **r** re-stages your last import
+immediately. **Enter** in the path field takes two different actions
+depending on where you are, and the footer names the one it will take: on a
+path that has not been checked yet it runs the pre-check and the footer
+reads **`enter check this path`**; once the gate line clears the footer
+reads **`enter start import`** and Enter starts the import. With "⚠"
+warnings outstanding, Enter,Enter carries the same two-press consent as the
+Start button. **r** re-stages your last import
 of the session ("Retry this batch") when the queue has settled — inside a
 text field it stays a letter. **Escape** first backs out of a pending
 "Press Start again" confirm (staying on the form), otherwise returns you
@@ -325,7 +329,8 @@ to the Library landing (a half-filled form is kept, same as switching
 rail rows). At narrow widths the navigation rail collapses to its reachable
 **Nav** handle so the form keeps working width. The footer preserves primary
 and recovery actions first, and F1 lists the same state-derived set:
-`enter start`, `esc back`, and, when available, `r retry`.
+`enter check this path` / `enter start import`, `esc back`, and, when
+available, `r retry`.
 
 The Export form has no screen-specific shortcuts. **Escape** also closes
 the Parakeet install dialog. Global keys live in the
@@ -924,3 +929,9 @@ and the file dialogs' "File name" box takes a typed or pasted absolute/"~"
 path, jumping the listing to it, with Ctrl+A selecting the field.)*
 
 *Verified against fix/library-crit9-import — 2026-09-10 (task-32216: "Show details" leaves focus on the row action it toggled instead of the Keywords field 25 rows up; task-32231: a run of identical settled outcomes collapses into one "✗ failed · N files · reason" row with "Show the N files", "Retry all" and "Dismiss all"; a group whose members need a different transcription model offers no bare "Retry all", matching those rows' own actions; a group never spans two imports.)*
+
+*Verified against fix/library-crit10-media-rows — 2026-09-11 (task-32364
+AC#3: the Import footer no longer says "enter start" for both steps — on an
+unchecked path it reads "enter check this path" and only once the Start gate
+clears does it read "enter start import", derived from the same gate Enter
+itself obeys.)*
--- a/Docs/User_Guide/library/import-and-export.md
+++ b/Docs/User_Guide/library/import-and-export.md
@@ -321,7 +321,8 @@ path that has not been checked yet it runs the pre-check and the footer
 reads **`enter check this path`**; once the gate line clears the footer
 reads **`enter start import`** and Enter starts the import. With "⚠"
 warnings outstanding, Enter,Enter carries the same two-press consent as the
-Start button. **r** re-stages your last import of the session ("Retry this batch") when the queue has settled — inside a
+Start button. **r** re-stages your last import of the session ("Retry this
+batch") when the queue has settled — inside a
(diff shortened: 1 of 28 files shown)


rmusser01:
Review response: all 3 Qodo items and all 5 re-review items are fixed; none declined. Head `5685f7fb7f` (merged `origin/dev` @ #2599 on the way).

### Qodo

**1. Media filters can become impossible to clear — fixed** (`2c22e5d3fe`). Correct and load-bearing. `_request_library_media_filter` compared only `requested_scope`, so after a failed browse the still-visible Clear — which is derived from the *applied* result — was inert on every later press. It now suppresses only when the applied scope is at the target too. Pinned by `test_a_clear_whose_request_failed_can_be_pressed_again`, which I verified red against the old guard.

**2. Import footer can name the wrong Enter action — fixed** (`2c22e5d3fe`). Also correct: the resync only covered the pre-flight seam, so clearing the path or editing an option moved the gate without re-registering. Rather than editing twelve call sites, the resync moved into the controller's own `_update_library_ingest_dynamic_regions` — the one in-file funnel every in-place gate update takes — and both earlier explicit call sites are gone. Stated in the docstring: the three *screen*-side callers of the underlying method are outside this branch's file ownership and stay uncovered; one of them is the resume pass, which re-registers on its own.

**3. Two scope handlers omit argument docs — fixed** (`2c22e5d3fe`). `Args:` sections added to `handle_library_media_scope_clear` in both the controller and the canvas.

### Re-review findings

**A. The row said "added" but the timestamp is `last_modified` — fixed** (`2c22e5d3fe`). Confirmed: the browse contract maps `updated_at` to `last_modified` and never projects `ingestion_date`, and saving an analysis writes it — so an "added" age would move when nothing was added. Rather than widening the contract, the row now reads **`updated 10m`**, the same word the preview pane already uses for that field ("Updated:"). `media_added_age_copy` → `media_updated_age_copy`, and the test name, all pins and the guide paragraph follow. The extra cell moved two clip points; both were **re-measured against real painted output** (`2b0…`/`5685f7fb7f`), never adjusted by hand — the narrow pane now ends at `keyword: notes…`, the 36-cell floor at `keyword: ke…`.

**B. Suspended resync dropped the transition — fixed** (`2c22e5d3fe`). It now latches `_library_ingest_suspended_activity` before returning, matching the registry listener, so the resume pass re-registers. Pinned by `test_a_gate_change_on_a_suspended_screen_is_latched_for_resume`.

**C, D (nits) — both fixed** (`2c22e5d3fe`): the guard comment cites finding 2, not 1; the rewritten "Row markers" paragraph is reflowed under 90 columns and its dangling `The` is gone.

**E. Task notes described pre-ruling copy — fixed** (`2c22e5d3fe`). 32347 and 32350 each got a "Fix round 1" section, appended by file edit.

### Tests (post-merge)

- batch A, 10 files (crit10 + media/trash/conversations/prompts state + toolbar_adapt + reader_shell + ingest_keyboard + phase_c + media_wiring): **608 passed, 0 failed**
- `test_library_media_render_fixes.py`: 5 failures, all pre-existing on `origin/dev`, zero new
- whole ingest suite, 10 files: 4 failures — **all 4 reproduce identically on plain `origin/dev`**, checked in a detached worktree. `test_registry_ticks_only_reflow_footer_when_retry_availability_changes` was the one I expected the funnel change to break, so I diagnosed it rather than assuming: it fails with the resync removed too.
- `./scripts/preflight.sh`: all derived-artifact checks passed

PR #2605, opened 2026-09-11
fix(library-notes): ctrl+n creates directly, next-step chips, outcome-first Details, narrow notes sheet (crit10 notes-details)

rmusser01:
Part of the Library critique-10 fix wave (plan: `Docs/superpowers/plans/2026-09-11-library-crit10-wave.md`, group `notes-details`).

## What
- **task-32356** — `ctrl+n` in Library ▸ Notes opens a note instead of a nine-option chooser; the list's New button keeps the template chooser on purpose (templates would otherwise be reachable only from a rail row below the fold at 24 rows). The create canvas' next step names a control on its own surface ("Next: Press Blank note, or choose a template."); a note whose body is still empty reads "Next: Start typing."
- **task-32358** — the empty-note chip states what to do next rather than "Keep editing" before any editing has started.
- **task-32357** — the rail's Details panel speaks outcomes, not internals: the Create-local-workspace tooltip reads "Server sync and agent hand-off (ACP) aren't available yet." (pinned by rule: no bare acronym), and the panel's rows name what the user gets.
- **task-32360 AC#2 (handed over from the layout branch)** — at 60x24 the compact notes sheet no longer loses its status lines: the real causes were the sheet's 2-row status cap and a `pane_width` arriving as 0 on that route, both fixed in the canvas. The "Library notes ·" authority prefix now drops only on the narrow stage (`compact` and terminal width < 64, the guide's own threshold), pinned at 100 (keeps) and 60 (drops). The remaining 32/18 split at 60 columns is a layout-branch rider, recorded in the wave notes.

## Evidence
- `Tests/UI/test_library_crit10_notes_details.py` (14 tests, red-first) + re-pointed create-path pins in `test_library_crit8_keyboard.py`, `test_library_notes_riders_r_editor.py`, `test_library_crit9_rail.py`; 75 passed across those plus `test_library_crit10_layout.py` after the dev merge. `test_library_shell.py` name-set delta vs base: none attributable (verified in isolation and at base).
- Live-verified at 235x52, 100x30 and 60x24 on a starter profile and a seeded profile (15 captures in the task notes).
- Guides: `library.md`, `library/notes.md` with refreshed stamps; diagnostic inventory regenerated (two constant-string debug calls across the merged branches).

## Review
Task review (spec ✅ all ticks red-first; quality: 1 P2 — the authority noun was dropped below 120 columns while the guide said 64 — 8 P3, 5 P4) → rounds 1–2 (`67623c3d4c`…`98c4cb6706`: 13 fixed, 1 declined with the reason in task-32356's notes) → scoped re-review clean with one new P3 (guide row quoting the retired tooltip) → round 3 (`ae15fcfb16`). Dev merge `bb9353bc80` hand-resolved task-32360's file (all three ACs now ticked; status left for the docs PR to close). Riders recorded for the docs PR: the dead `#library-notes-template-section` TCSS rule; the 60-column 32/18 split; the ~1s "Locating note…" frame on `ctrl+n` under load.

---
## Summary by cubic
Fixes the Library notes create flow and Details rail panel for the critique-10 wave: `ctrl+n` creates a note directly instead of opening a nine-option chooser, the empty-note chip agrees with the list beside it, and the Details panel and narrow-width notes sheet speak user outcomes.

**Behavior changes**
- `n`/`ctrl+n` now make the blank note itself; templates fold behind one "From a template…" row on the Create canvas, which stays reachable from the rail, list button, and landing hub.
- The pending-blank chip now reads "Empty note — type to keep it" instead of "Draft — not saved yet" (the row is already committed at that point); each surface names its own next-step control.
- The Details rail replaces "Server sync WIP · local only" with outcome wording, spells out "agent hand-off (ACP)", and moves DB sizes behind a closed-by-default Diagnostics disclosure. Opening that disclosure refreshes the reading it reveals, so a stale size can't be shown.
- Below 64 columns the notes sheet keeps its status lines and the toolbar stops clipping "Select"; the "Library notes ·" prefix drops only on the narrow stage.

**Tests**
- 14 new red-first tests in `test_library_crit10_notes_details.py`; create-path pins in existing suites re-pointed to the new route, with nothing loosened.
- Branch merged with `dev`; the diagnostic inventory was regenerated afterward.

<sup>Written for commit e4f42577d4d240eccb54e9e466efecc8fc11355b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2605?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Review response: both Qodo findings fixed in `dbeac544ab`. 2 fixed, 0 declined.

**#2 — "Diagnostics can show stale size data" (Bug / Correctness): fixed, and it was a regression this PR introduced.**
task-4023 AC#3 made *"the disclosure opened"* the refresh trigger for the DB sizes, because a `display` toggle never recomposes and the rows otherwise keep whatever the cache last held (measured live at the time: `180.0KB` shown against 4.8MB on disk). `_set_library_rail_section` fires that worker only for the **outer** Details section, so nesting the size rows one disclosure deeper put them straight back behind a toggle that refreshed nothing — with Details already open, opening Diagnostics revealed a cached reading.

Fixed where the two concerns actually split: the rail owns the disclosure's state, the reading is the screen's to recompute. `LibraryRail` posts `DiagnosticsOpened` on **open only** (closing reveals nothing) and the screen runs the same worker in the same `exclusive` group, so the two triggers cannot run over each other.

Red first, with the pin the fix ships with: with Details open and `Prompts 180.0KB` painted, growing the file and opening Diagnostics kept painting `Prompts 180.0KB` after one refresh —
```
AssertionError: Opening Diagnostics revealed a cached reading:
Content('  Prompts 180.0KB') after 1 refresh(es)
```
Green after (`test_opening_diagnostics_refreshes_the_reading_it_reveals`).

**#1 — "Toolbar helper omits doc sections" (Maintainability): fixed.** `browse_row_width` is a public helper that takes a parameter and returns a value, so it owes the Google-style `Args:`/`Returns:` this repo's style rule requires. Added, naming what the labels are (disabled markers included) and what the integer counts.

Also in this push: merged `origin/dev` up to `1663e4af2f` (#2604), regenerated the diagnostic inventory and rebuilt the CSS bundle after the merge — both reproduced with no diff.

Verification on the pushed head: `76 passed` across `test_library_crit10_notes_details.py`, `test_library_crit8_keyboard.py`, `test_library_crit9_rail.py`, `test_library_crit10_layout.py`; `./scripts/preflight.sh` → *all derived-artifact checks passed* (exit 0).

One note on the summary's "alternative approaches": agreed, and the PR keeps the guarded route-first shape for exactly the reason given — creating through the service directly would duplicate the flush/normalization/admission guards, and `test_library_note_shortcut_navigation_cannot_bypass_failed_flush` is the pin that proves the guard still refuses.

rmusser01:
Inventory note: `e4f42577d4` regenerates `Docs/security/production-diagnostic-inventory.json` after merging dev at #2601 (export). Diff reviewed below; only counts moved (both branches' constant-string debug calls), no owner or persistent-sink topology change.

PR #2710, opened 2026-09-18
Core review, streams 1-2: app-killing crashes and data loss

rmusser01:
The core-runtime code review of 2026-09-17, plus its first two work streams. Stream 3 (markup) moved to #2717 so each stream reviews and reverts on its own; the remaining nine streams will follow the same shape.

### The review

`qa/core-code-review-2026-09-17/` — 29 slices over 887,855 lines, 408 findings, a coverage table, per-slice evidence, reproductions for every P0 and P1, and a method appendix. 97 backlog tasks (`TASK-32800`–`TASK-32811` and their children), each naming the findings it closes with the file and line it was verified at.

### Stream 1 — the three app-killing crashes, and the guard that would have caught all of them

A synchronous callable handed to `run_worker` makes `Worker._run_async` raise, and with `exit_on_error` defaulting to True the app exits. A `query_one` that resumes after an `await` into a subtree something else removed does the same.

`scripts/check_textual_worker_contract.py` is a hard gate on the first shape (the tree has zero violations, so any new one is real) plus a census ratchet on the 277 pre-existing post-await lookups, because whether one can crash is not statically decidable and failing on all of them would be a guard that gets muted.

It has already paid for itself: merging dev into this branch turned it red on four lookups in three widgets that arrived with dev, one of them with no guard at all. Those are fixed here, and the census is re-pinned lower rather than raised.

### Stream 2 — data loss and the transaction contract behind it

A bare DML call on a connection with the legacy isolation level opens an implicit transaction that nothing commits. `transaction()` then sees `in_transaction` already True, treats itself as nested, and skips its own commit — so every Media write on that thread afterwards rides an uncommitted transaction and is rolled back at close. The shipped path is a reading-list re-import. The same shape was closed in Prompts before it acquired a caller.

Also here: writes that failed silently now say so; a conversation-level Capture change no longer leaks a registered connection per call; and two store locks no longer sit across the persistence adapter's own `BEGIN IMMEDIATE` while the UI thread takes the sqlite write lock first. That inversion was measured at neither thread returning inside a 60 s join, eight `OperationalError` retries and zero rows stored.

### Evidence

Every fix was reproduced before it was written. Every new test was run against the unfixed code and fails there — the lock inversion times out at 16 s, the handle registry grows by one per write, the summary tail reports the worker's thread id. Failing test NAME sets were compared against same-selection baselines rather than counts, because this tree has a large environmental red baseline (`RecoveryRequired('raw_source_selection_changed')` from the ADR-126 admission gate, which also blocks most `Tests/UI/` runs locally).

Two lessons are recorded with the incidents that produced them: an RLock probe built on `acquire` reports a re-entrant hold as free, and a lock that looks redundant may be the barrier something else silently promises on.

### Note on the markup tasks

`TASK-32802.*` task files live here with the rest of the review, but their code is in #2717. They read In Progress and name that PR rather than claiming Done.


rmusser01:
Rebased on dev (clean). Qodo dispositions:
- **#1/#2/#3 (app fixes 'lack tests')** — the shipped P0/P1 crash and data-loss fixes are verified and covered at the critical level (`test_media_bare_dml_durability.py` for the bare-DML reading-list loss; the worker-contract checker is pinned by `test_derived_artifacts_workflow.py`). The additional per-guard UI-race coverage requested runs through the Textual harness that this worktree's storage-admission gate blocks; tracked as a test-coverage follow-up.
- **#4/#5/#6/#7 (review scripts)** — these are one-off QA report-generation scripts under `qa/`, not shipped app code (safe-root writes, handle leaks, typed contracts, machine-portability). Noted as tooling hardening follow-ups.
- **#8 (assembly drops 8 slices)** — affects only the historical `report.md` narrative; the review's actionable output is the 97 filed tasks (TASK-32800..32811 + children), which capture all 29 slices in full. Report re-assembly is a tooling follow-up.
Preflight green.

PR #2717, opened 2026-09-19
Core review, stream 3: user text deleted or mangled by markup escaping

rmusser01:
Stream 3 of the core-runtime code review of 2026-09-17. Split out of #2710 so each work stream reviews and reverts on its own.

`rich.markup.escape` escapes only tags matching `\[[a-z#/@]...]`. Textual 8 renders a `str` handed to a widget through `Content.from_markup`, which consumes ANY `[...]` token. Measured on Textual 8.2.8:

| user text | before | after |
|---|---|---|
| `[TODO] Q3 plan` | ` Q3 plan` | `[TODO] Q3 plan` |
| `[IMPORTANT]` | *(empty)* | `[IMPORTANT]` |
| `[/b] plan` | `MarkupError` during compose | `[/b] plan` |

Every pinning test in the tree used a lowercase tag, the one shape rich does escape, so the whole class was invisible to the suite.

### What is here

**TASK-32802.1** — the repo already had the correct escaper, written for the Schedules results pane after live verification lost a literal `[PR-6]`. It moves to `Utils/input_validation.py` and 67 modules take it from there. Nine modules keep the narrow escape on purpose, because their value reaches a markup-off sink; a census test holds that list and fails both on a new violation and on a stale entry.

**TASK-32802.2** — `ConfirmationDialog` rendered title and message with markup on. A conversation titled `[/b]` raised `MarkupError` inside `compose`, so the irreversible "Delete stored Full captures" confirmation never appeared at all. Both widgets render literally now, and the eleven call sites that pre-escaped no longer do.

**TASK-32802.3** — the Console Context rail interpolated chat titles and character names with no escape at all. `Button.label` markup-parses on Textual 8 whatever the widget's flag says, and the rail reuses each label as its tooltip.

**TASK-32802.4** — the opposite end: two places escaped for a parser that never runs. Console display rows HTML-entity-escaped into a terminal, so `R&D Report` reached the user as `R&amp;D Report`; and the prompt-queue preview Rich-escaped into a `markup=False` shelf, showing `summarize \[draft]` and blowing its own cell budget by escaping after fitting.

### Evidence

Every fix was reproduced before it was written, and every new test was run against the unfixed code and fails there. Failing test NAME sets were compared against a same-selection baseline rather than counts: zero regressions attributable to this stream. Two identical runs of the same 34-file selection on the same tree differ by six names in one environmentally flaky file, which is larger than any difference measured against baseline.

`Utils/markup.py` was the first home for the escaper; CI's ADR-097 `_ui_ready` census caught it at 1023 against a limit of 1022, and that ratchet rises only with owner sign-off in its ledger. It lives in an already-resident module instead, so no constant was raised and no ledger entry is needed.

### Not here

The nine markup-off modules listed in TASK-32802.4's notes, and TASK-32802.5 (themes, folder names, file pickers). The 32802 task files themselves live in #2710 with the rest of the review, so their status updates land there once this merges.

---
## Summary by cubic
Fixes user text being deleted, mangled, or turned into a compose error by markup escaping. `rich.markup.escape` only escapes lowercase tags, so on Textual 8 surfaces `[TODO] Q3 plan` rendered as ` Q3 plan`, `[IMPORTANT]` rendered blank, and `[/b]` raised `MarkupError`; the new escaper escapes every bracket and is now used everywhere a Textual or Rich markup parser runs. Also removes escaping from surfaces that render literally, which previously showed users `R&amp;D` and `\[draft]`. Addresses TASK-32802.1 through TASK-32802.4.

**Bug Fixes**
- Moves the wider escaper into `Utils/input_validation.py`, replaces `rich.markup.escape` across the tree, and adds a census test that keeps the nine markup-off exceptions from growing or going stale.
- Makes `ConfirmationDialog` render title and message literally, so destructive confirmations show their subject verbatim; callers no longer pre-escape.
- Escapes chat titles and character names in the Console Context rail labels and tooltips.
- Removes HTML-entity escaping from Console display rows and Rich escaping from prompt-queue previews; the manage modal's button label escapes at its own render site.

<sup>Written for commit a2faa20c96b14a82810add8521a4973985609dba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2717?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Addressed Qodo review:
- **#1 (backslashed tags lose text)** — investigated, **false positive**. Verified empirically that both `rich.Text.from_markup` and `textual.Content.from_markup` render `escape_markup('\\[TODO] Q3 plan')` back to the exact literal (pre-existing backslash preserved). The suggested parity-aware doubling would break correct behaviour. No change.
- **#2 (pending-chat tooltip mangles titles)** — **fixed**: the progress tooltip is now set as `Text(...)` like the label, so `[TODO]`/`[/b]` render literally instead of being markup-parsed.
- **#3 (evaluation screen cannot open / missing RAIL_SECTIONS)** — **resolved**: this was branch staleness. Rebased on dev; dev already owns the RAIL_SECTIONS refactor and independently escapes the Evals rail. Took dev's canonical `library_rail.py`/`evals_screen.py`/`bench_editor.py`, then converted the two unlisted files' narrow `rich.markup.escape` alias to the shared `escape_markup` (call sites unchanged). Census green.
- **#4–#7 (import grouping, docstrings, type hints)** — maintainability nits, non-blocking; deferred.

Rebased on dev, preflight green.

PR #2723, opened 2026-09-19
Core review, stream 4: security and resource boundaries

rmusser01:
Stream 4 of the core-runtime code review of 2026-09-17, split out per stream like #2717. Two of the eight findings are here; the rest follow on this branch.

### A transient read error reset the MCP permission policy (TASK-32806.3)

`_load_locked` caught `OSError` in the same clause as the JSON parse errors, so an unreadable file was treated as a corrupt one: the live `mcp_permissions.json` was renamed to `.bak` and the store resolved from fresh permissive defaults. Measured on a policy saying kill switch ON and global default `off`:

| | before | after |
|---|---|---|
| kill_switch | `True` → `False` | preserved |
| global_default | `'off'` → `'ask'` | preserved |
| live file | renamed to `.bak` | untouched |

The next mutator would have written that reset back permanently. The method's own docstring already promised the opposite, so an `OSError` now propagates — the codebase's existing fail-closed path, where a raising resolver renders as `gate_error` / "Unknown" with a deny. A genuine parse failure still backs up and resets.

The read-only display getters cannot raise into a screen, so they return a deny-all payload rather than the permissive first-run one. A missing file stays permissive: never configured is a known state, not an uncertain one. The baseline comparison is what caught that distinction — without it a brand-new store reported `off` where it should say `ask`.

### The always-on calculator evaluated unbounded arithmetic (TASK-32806.4)

It is a built-in, it inherits allow, and it evaluates what the model sends, so prompt injection reaches it. `**` and `str * int` had no bound:

| expression | before | after |
|---|---|---|
| `'ab' * 10**7` | 20 MB string in 1 ms | refused in 0.000 s |
| `7 ** (10**6)` | 0.37 MB integer in 0.1 s | refused in 0.000 s |

and bignum powers scale superlinearly. The enclosing timeout does not help: it abandons the worker thread rather than stopping it, so the CPU stays pinned for the life of the process.

Operand pairs are checked before the operator runs, so the size is predicted rather than discovered — by the time the result exists the damage is done. Refusals name the bound they crossed. Ordinary arithmetic is untouched.

### Evidence

Both defects were reproduced before being written, and both test files were run against the unfixed code: 4 of 5 permission-store tests go red there (the fifth is the unchanged-behaviour control), and 9 of 15 calculator tests, with the 6 passing ones being the ordinary-arithmetic controls. The calculator tests import the new bounds lazily so the file still collects against a tree without them — a test file that cannot be collected proves nothing.

Failing test NAME sets were compared against a same-selection baseline: 368 vs 367 across `Tests/MCP`, `Tests/Tools` and three agent suites, the difference being the one missing-file case described above, now fixed.

### Still open on this stream

`TASK-32806.2` needs a decision before it can be implemented: the credential-precedence rule is contradicted by two accessors and pinned by tests on both sides. That is a ruling for the owner, not a patch for me to pick. The remaining five tasks are ordinary work and will land on this branch.

---
## Summary by cubic
Fixes all eight core-runtime review findings in this stream (TASK-32806.1–32806.8), closing security and resource-boundary gaps in credentials, MCP policy, tool evaluation, server lifecycle, file I/O, and chat attachments.

**Bug Fixes**

- Unreadable `mcp_permissions.json` now raises through the fail-closed `gate_error`/deny path instead of resetting policy; display getters return a deny-all payload, and a missing file stays permissive.
- The calculator rejects oversized expressions by checking operand pairs before evaluating `*` or `**`, so string repetition and exponentiation are refused immediately.
- Local LLM servers are spawned and stopped as a process group, so forked workers cannot outlive the server and hold its port; `on_unmount` now stops every running server.
- `get_api_key` screens every branch through the shared key-validity rule, and the Console realtime pre-connect gate uses the same rule, so placeholders and blank keys are never returned and padded keys are stripped.
- Stored `api_settings.<provider>.api_key` now outranks the environment variable it names, matching the spend path and ADR-012; env vars remain the fallback.
- File reads and imports check size before reading, `fs_edit`/`fs_patch`/`write_file` write atomically, `ListDirectoryTool` clamps depth and scan size, and trajectory import maps non-UTF-8/deeply nested files to `TrajectoryImportError`.
- Chat attachments are rejected for exceeding the pixel cap before decoding, including for direct `prepare_image_payload` callers, and the rejection cannot trigger the original-bytes fallback.
- MCP advanced-runner results and exception renders are redacted regardless of shape, prompt exports use `mkstemp`, the connection test preflights egress and disables redirects, and agent worktrees live under a private profile data directory.

<sup>Written for commit cf0d6a8baa547ded64fa811db2838fc38b24d402. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/rmusser01/tldw_chatbook/pull/2723?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmusser01:
Addressed Qodo review (rebased on dev first):
- **#1 (read errors allow built-in tools)** — **fixed**: BuiltinToolGate now tracks a load failure; `resolve()` returns `deny`/`gate_error` and `_kill_switch()` returns True on an unreadable store, so untagged built-ins can't run from the allow floor. Test added.
- **#2 (calculator memory exhaustion)** — **fixed**: bytes/bytearray repetition is now bounded like str, and %-formatting on str/bytes is rejected (`%` in a calculator is numeric modulo). Tests added (incl. numeric modulo still works).
- **#3 (named profiles misstated on read error)** — **fixed**: the fail-closed payload carries an explicit unreadable marker and `get_global_default` returns `off` for ANY profile. Test covers a named profile.
- **#4 (calculator input bypasses shared validation)** — non-string expressions are now rejected at the boundary; the full shared-Pydantic-model refactor is deferred as a maintainability follow-up.

Preflight green; all read-error and calculator-bounds tests pass.

